PR #1, opened 2018-02-15
init commit test

bigopon:
I think we need to get eventing right from the start. It should use `handleEvent` and class methods. I leave it in there so we have a reminder.


EisenbergEffect:
Merged! Note: The framework.ts is very temporary. Just a quick example. We'll want to plumb these ideas back into the code that's in the framework folder in the future. That's the real stuff.

PR #6, opened 2018-03-07
lifecycle, <import />, behavior

bigopon:


bigopon:
@EisenbergEffect With this change, you can basically write simple app, like `app.au` requiring `name-tag.au`

bigopon:
Note that `<import from='./name-tag'/>` is used to determine if something should be scanned to be registered as custom element, custom attribute. I don't see an easy way to NOT do this, without making user config everything

PR #13, opened 2018-03-19
Dependency Injection

fkleuver:
EDIT: Most of the runtime stuff is usable at this point. There are some preliminary docs. Would like to know what you think about the API, specifically in how it's different from the "old" DI. Welcome addition or simply too much change/complexity?


fkleuver:
As discussed with @EisenbergEffect, this solution (at least the runtime part) is too large and not performant enough for Aurelia's purposes. That's okay; I was aware of that possibility and had fun & learned while building this regardless. It may show up again in some other form for some other purpose at some point :)

Before proceeding with potentially something else, I'll wait for Rob to be done with the core solution. If at that point there is still any help required with stuff like code generation or unit testing, I'll be happy to assist.

Closing this PR now; the code will stay available on my fork.

PR #17, opened 2018-06-03
chore(all): port new parser / ast, fix tests

fkleuver:
@EisenbergEffect Opening this PR to make it easier to discuss things. Work in progress.

Not entirely sure what to do with `parseCore`. `try..catch` makes a function unoptimizable by JIT. V8's latest TurboFan can do it, but the optimization itself also brings a fairly significant performance hit.. ironically. I think try/catch should only be used for wrappers around user code.
I could integrate the reporter in some other way though. Could just pass down another context flag and have the parser decide whether to throw or not based on that.


EisenbergEffect:
@fkleuver One of the main things I'm trying to do with vNext is separate very clearly what is needed at runtime from what is needed for AOT. I'm also trying to enable a middle scenario where you can run without a build as well. Basically, this amounts to things being split into:

* runtime - required pieces at runtime
* jit - add onto the runtime the pieces that would be needed without a build process
* aot - the build tooling that uses the jit to produce a build that only requires the runtime

So, the actual AST nodes would belong in runtime, but the expression parser would belong in the jit. With that in mind, can you add the parser implementation here? https://github.com/aurelia/experiment/blob/master/src/jit/binding/expression-parser.ts#L8

EisenbergEffect:
@fkleuver Can you confirm that my demo app still functions with these changes? If so, I'm happy to merge this i. I'd love to hear a bit more about your ideas for the BindingFlags.

fkleuver:
@EisenbergEffect Seems like it still works.

BindingContext would probably be a better name than BindingFlags but that name is already occupied :)

The idea is to make the whole binding / observation / change handling process (and perhaps templating as well) context-sensitive.
Extra information can be passed around without a performance overhead, so redundancy can be eliminated, multiple checks can be consolidated into single checks, recursion can be detected, change notifications can be deduplicated, shortcuts can be taken, etc.
The `Access` and `Precedence` enums in the expression parser are good examples of how I might end up applying it, but obviously it won't be the same.

But it's easier to explain when there's working code to show. I'll try to find some low-hanging fruit for a simple start to demonstrate the concept.

bigopon:
Maybe another suitable name is bindingOrigin

fkleuver:
@bigopon That would be a little too specific because there can be arbitrary information in it. For example, the flags also take the place of all `mustEvaluate` booleans:
https://github.com/aurelia/experiment/pull/17/files?utf8=%E2%9C%93&diff=unified#diff-d4e9ffa61125faebf0eecc6dd444e11fR320
https://github.com/aurelia/experiment/pull/17/files?utf8=%E2%9C%93&diff=unified#diff-d4e9ffa61125faebf0eecc6dd444e11fR338
https://github.com/aurelia/experiment/pull/17/files?utf8=%E2%9C%93&diff=unified#diff-d4e9ffa61125faebf0eecc6dd444e11fR642

bigopon:
I didn't read the code. It looks like you did incorporated the binding flags into binding ? That's huge !

fkleuver:
It's like async zombie virus, you kind of have to go all the way with it :)
Well you don't *have* to, but life gets easier if you do.

fkleuver:
Well, there has to be some kind of queue where changes can be accumulated, like a specialized ChangeQueue. The logic around the flags would allow it to be synchronous and predictable though, as well as more performant because it won't need a try/catch block.

EisenbergEffect:
The primary purpose of the task queue was to enable efficient batching of updates to the DOM. If you can come up with an alternate mechanism that is more efficient, I'm always open. The idea of the flags/context that flows through the system makes sense. The main thing I'd want to watch out for there is that we aren't creating/destroying tons of objects on each update. There was a lot of work done in binding to eliminate closures and unnecessary objects. So, I want to make sure we don't undo that. If you can get in contact with @jdanyow it would be good to run ideas by him since he did all the original optimizations.

Merging this shortly :) Looking forward to seeing how this goes.

bigopon:
For the task queue, what I really meant is the ability to update source with out it, and definitely DOM updating (target updating) will need to be batched

fkleuver:
That was my main point/goal as well. Task queue always seemed like a very suitable for the DOM thing, and the fact that it was used for source updating felt like a sort of unwanted evolution. I'm mainly looking to address that part. Perhaps by doing so the batches sent to the DOM can more easily be controlled to synchronize well with frames - that would be a nice bonus.

As for creating/destroying objects, the only thing that's created is a single primitive value (a number) per update. No need for objects - there's plenty of information you can store in 32 bits :)
No need for closures either. 

Here's some initial rough work, still need to finish the wireup and clean up a few things: https://github.com/fkleuver/experiment/commit/5f19e773301f2d510b29707e2b321bb1204c833c

There are more lines now, but this should eventually have a neutral effect on overall size.

EisenbergEffect:
All sounds great guys. I'm very excited to see where this goes :)

PR #18, opened 2018-06-05
chore(all): cleanup tests and project structure

fkleuver:
- Add `,vscode/settings` to make a few things more consistent between machines/developers
- Split up the tests between runtime/jit/debug so that they can run independently, and have independent tsconfig / setup / code coverage reports
- Add generated stuff (package lock, scripts) to .gitignore, remove unnecessary stuff, make it more explicit

@EisenbergEffect  Figured this is better than writing a long message with suggestions.

This is opinionated (in the sense that it makes stuff easier for *me*, I don't know if it also does so for you), so if you don't like/want a few things, I'd be happy to change it.

And yes, everything still seems to work :)


EisenbergEffect:
Looks good to me. The only think I'll make mention of is that eventually we'll move this into a monorepo. I'm holding out a bit longer on that. In particular, I really want to see how the new TS project system turns out. I'd like to use that if it fits well.

PR #21, opened 2018-06-22
fix(parser): left-to-right associativity for nested binary expressions

fkleuver:


fkleuver:
humm, wasn't this one merged already?

PR #22, opened 2018-07-13
chore(all): add dbmonster

fkleuver:
This adds an almost-complete dbmonster benchmark (I'm not sure how to do the class bindings or the one-time bindings with hand-written config)
It actually exposes a bug in the original repeater (or DOM?) logic that I've been wrestling with earlier. Wasn't sure if it was my test being wrong, but it seems like there's actually a bug.

@EisenbergEffect  just run `npm run dbmonster` and wait a second or so, you'll see that it starts with "element contains the parent ..." followed by the overrideContext being undefined. I'd like to have a baseline to compare my repeater rework against, do you think you can fix this?


EisenbergEffect:
I'm having some trouble running it. I get `copyfiles not found`.

fkleuver:
Did you `npm install` after pulling? It's a dev dep, it should automatically run from `node_modules`

EisenbergEffect:
I realized that was the issue as soon as I peeked in the package.json.

PR #27, opened 2018-07-26
refactor(dom): various perf tweaks

fkleuver:


src/runtime/dom.ts
```diff
@@ -96,15 +156,16 @@ export const DOM = {
...
   },
 
-  remove(node: INodeLike) {
-    if ((<any>node).remove) {
-      (<any>node).remove();
-    } else if ((<Node>node).parentNode) {
-      (<Node>node).parentNode.removeChild(<Node>node);
+  remove(node: INodeLike): void {
+    // only check the prototype once and then permanently set a polyfilled or non-polyfilled call to save a few cycles
+    if (Element.prototype.remove === undefined) {
```
bigopon: In ever green browser, which is vNext target, do we need to do this ? Feel like complicating our abstraction unnecessarily

src/runtime/dom.ts
```diff
@@ -147,36 +208,48 @@ export const DOM = {
...
     if (proxy) {
-      anchor.$proxyTarget = node;
-      anchor.hasAttribute = hasAttribute;
-      anchor.getAttribute = getAttribute;
-      anchor.setAttribute = setAttribute;
+      anchor.$proxyTarget = <Element>node;
+      // binding explicitly to the anchor instead of implicitly to ensure the correct 'this' assignment
+      anchor.hasAttribute = hasAttribute.bind(anchor);
+      anchor.getAttribute = getAttribute.bind(anchor);
+      anchor.setAttribute = setAttribute.bind(anchor);
```
bigopon: How is this better than `anchor.setAttribute = setAttribute` in the context of a perf PR. We are doing avoidable allocation. The only difference I don't know the effect on perf is bound fn vs normal fn

src/runtime/dom.ts
fkleuver: I tried to be conservative with cleaning out polyfills and wanted to leave that up to @EisenbergEffect but you're right. Pretty sure this one can go.

src/runtime/dom.ts
fkleuver: This one wasn't actually a performance thing but rather an "ensure the correct `this`" thing. I thought an arrow function invocation on this with a different closure could give it a different `this`, and cause hard to track down errors in weird places. I might be wrong about that though.

As for `bind` perf in general it outperforms normal functions (especially in hot paths) when you're binding 1 or more arguments that is not `this`. It saves the (albeit very small) overhead of the argument passing, so when an arg never changes it can be a minor perf tweak.

Yes, that's a tradeoff of one allocation versus potentially many function calls that helps a little bit in perf. Both are very small anyway.

For me it's more of a gut feeling / style preference thing and I don't feel very strongly about it. I'm 100% fine with dropping these sorts of micro tweaks if it doesn't feel right to you.

bigopon: > As for bind perf in general it outperforms normal functions (especially in hot paths) when you're binding 1 or more arguments that is not this. It saves the (albeit very small) overhead of the argument passing, so when an arg never changes it can be a minor perf tweak.

Thanks ! This is what I have been wondering about too. I think for above functions, the target is  to have the right `this`, it should not be doing allocation. If it was to always have a fixed argument then this opt would be very nice.

PR #29, opened 2018-07-26
feat(observation): add improved collection observers

fkleuver:
First step towards getting the new repeater in

### DIFF
--- a/src/runtime/binding/array-observer.ts
+++ b/src/runtime/binding/array-observer.ts
@@ -1,0 +1,330 @@
+import { collectionObserver } from './collection-observer';
+import { IObservedArray, CollectionKind, ICollectionSubscriber, IBatchedCollectionSubscriber, ICollectionObserver, IndexMap } from './observation';
+import { BindingFlags } from './binding-flags';
+
+const proto = Array.prototype;
+const nativePush = proto.push;
+const nativeUnshift = proto.unshift;
+const nativePop = proto.pop;
+const nativeShift = proto.shift;
+const nativeSplice = proto.splice;
+const nativeReverse = proto.reverse;
+const nativeSort = proto.sort;
+
+export function enableArrayObservation(): void {
+  proto.push = observePush;
+  proto.unshift = observeUnshift;
+  proto.pop = observePop;
+  proto.shift = observeShift;
+  proto.splice = observeSplice;
+  proto.reverse = observeReverse;
+  proto.sort = observeSort;
+}
+
+export function disableArrayObservation(): void {
+  proto.push = nativePush;
+  proto.unshift = nativeUnshift;
+  proto.pop = nativePop;
+  proto.shift = nativeShift;
+  proto.splice = nativeSplice;
+  proto.reverse = nativeReverse;
+  proto.sort = nativeSort;
+}
+
+// https://tc39.github.io/ecma262/#sec-array.prototype.push
+function observePush(this: IObservedArray): ReturnType<typeof nativePush> {
+  const o = this.$observer;
+  if (o === undefined) {
+    return nativePush.apply(this, arguments);
+  }
+  const len = this.length;
+  const argCount = arguments.length;
+  if (argCount === 0) {
+    return len;
+  }
+  this.length = o.indexMap.length = len + argCount
+  let i = len;
+  while (i < this.length) {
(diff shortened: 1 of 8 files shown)


src/runtime/binding/array-observer.ts
```diff
@@ -0,0 +1,330 @@
...
+  }
+  if (compareFn === undefined || typeof compareFn !== 'function'/*spec says throw a TypeError, should we do that too?*/) {
+    compareFn = sortCompare;
+  }
+  quickSort(this, o.indexMap, 0, i, compareFn);
+  o.notify('sort');
+  return this;
+}
+
+// https://tc39.github.io/ecma262/#sec-sortcompare
```
blackgwe: #sec-sortcompare does not exists any more, do you mean https://tc39.github.io/ecma262/#sec-array.prototype.sort

src/runtime/binding/array-observer.ts
```diff
@@ -0,0 +1,330 @@
...
+    this[upper] = lowerValue; o.indexMap[upper] = lowerIndex;
+    lower++;
+  }
+  o.notify('reverse');
+  return this;
+};
+
+// https://tc39.github.io/ecma262/#sec-array.prototype.sort
+// https://github.com/v8/v8/blob/master/src/js/array.js
+function observeSort(this: IObservedArray, compareFn?: (a: any, b: any) => number) {
```
blackgwe: How many elements do you expect at most in `this: IObservedArray`` ? Is it really necessary to implement a high performance quickSort  O(nlog(n)) and fallback to insertionSort if there are less then ten elements?

src/runtime/binding/array-observer.ts
fkleuver: What do you mean does not exist anymore? It is there on the page.
I included this link as a reference / justification to sort the undefineds out first.
I included a link to #sec-array.prototype.sort in the sortCompare function

src/runtime/binding/array-observer.ts
fkleuver: I'm not making any particular assumptions about the number of elements.
The insertionSort will eventually come into play for any number of elements when the quickSort has brought a section down to that number of elements. If the number of elements is less than 10 it will directly go to insertionSort, but for 1000 elements the insertionSort will also eventually come into play for specific subsections.
Apart from having made sure I understood the implementation, I am relying on the V8 team's research (and some preliminary benchmarks I run myself), the fact that Chrome is the most performant browser, and have simply "copied" their implementation (see the second url).

blackgwe: I only wonder why it is necessary to "inflate" the code with such a large/big quickSort implementation... I'm rather sure that a very simple mergeSort/quicksort does the job is good enough (if you don't want to sort millions of elements)

fkleuver: I don't know if it's "necessary". I'm sure there will never be millions of elements to be sorted, but thousands might be. The observation can also be used for non-templating stuff.

I also just tried to be as close to the spec / browser behavior as possible. And I didn't want my sort algorithm to potentially be a perf regression compared to vCurrent, hence I just took the V8 implementation as-is.

That being said, if you have an implementation that's a fair chunk smaller than this one and only gives in a few % in performance in both small and large (e.g. a few thousand) element sets, that could be worth considering.

Keep in mind this observation logic will replace the existing ArraySplices / Levhenstein distance calc from [vCurrent](https://github.com/aurelia/experiment/blob/master/src/runtime/binding/array-change-records.ts). I think you'll agree that regardless of this big and bloaty sort function, it's still an overall smaller size (also considering the new repeater that I'll soon PR)

blackgwe: Thank you very much for the detailed explanation. And of course I agree with you - you are an expert of the "aurelia" source code. And many thanks to you (and @EisenbergEffect and all the others) for the amazing code!!!

fkleuver: And thank you for having a critical look at things. It's always good to have things peer-reviewed to make sure things are well thought-through :)

src/runtime/binding/array-observer.ts
fkleuver: No problem, it's a *huge* html page that's notorious for its slow load ;)

PR #41, opened 2018-07-28
doc(runtime): clarify the purpose of several core items

EisenbergEffect:
Also, added some notes on which things are likely to go away with the removal of shadow dom emulation. In the end, I'm hopeful that a good chunk of this will be simplified. However, I think it's likely going to still be very valuable to have core abstractions like ITemplates that create IViews. The simplicity can come from the fact that you will no longer need a RenderSlot to render the IView. You can basically just have an INode (or RenderLocation, which will just be an INode) and simply append the IView at the location.

### DIFF
--- a/src/runtime/dom.ts
+++ b/src/runtime/dom.ts
@@ -1,5 +1,6 @@
 import { IContainer, IResolver, DI } from '../kernel/di';
 import { ICustomElement } from './templating/custom-element';
+import { IViewOwner } from './templating/view';
 
 export interface INodeLike {
   readonly firstChild: INode | null;
@@ -281,6 +282,11 @@ function setAttribute(this: CommentProxy, qualifiedName: string, value: string):
   this.$proxyTarget.setAttribute(qualifiedName, value);
 }
 
+// This is the most common form of IView.
+// Every custom element or template controller whose view is based on an HTML template
+// has an instance of this under the hood. Anyone who wants to create a view from
+// a string of markup would also receive an instance of this.
+// CompiledTemplates create instances of TemplateViews.
 /*@internal*/
 export class TemplateView implements IView {
   private fragment: DocumentFragment;
--- a/src/runtime/dom.ts
+++ b/src/runtime/dom.ts
@@ -1,6 +1,5 @@
 import { IContainer, IResolver, DI } from '../kernel/di';
 import { ICustomElement } from './templating/custom-element';
-import { IViewOwner } from './templating/view';
 
 export interface INodeLike {
   readonly firstChild: INode | null;
--- a/src/runtime/templating/rendering-engine.ts
+++ b/src/runtime/templating/rendering-engine.ts
@@ -40,6 +40,7 @@ export interface IRenderingEngine {
 export const IRenderingEngine = DI.createInterface<IRenderingEngine>()
   .withDefault(x => x.singleton(RenderingEngine));
 
+// This is an implementation of ITemplate that always returns a view representing "no DOM" to render.
 const noViewTemplate: ITemplate = {
   renderContext: null,
   createFor(owner: IViewOwner) {
@@ -226,6 +227,12 @@ function createDefinition(definition: Immutable<ITemplateSource>): TemplateDefin
   };
 }
 
+// This is the main implementation of ITemplate.
+// It is used to create instances of IView based on a compiled TemplateDefinition.
+// TemplateDefinitions are hand-coded today, but will ultimately be the output of the
+// TemplateCompiler either through a JIT or AOT process.
+// Essentially, CompiledTemplate wraps up the small bit of code that is needed to take a TemplateDefinition
+// and create instances of it on demand.
 class CompiledTemplate implements ITemplate {
   private createView: () => IView;
   renderContext: IRenderContext;
--- a/src/runtime/templating/template.ts
+++ b/src/runtime/templating/template.ts
@@ -3,6 +3,11 @@ import { INode, IView } from "../dom";
 import { TemplatePartDefinitions } from "./instructions";
 import { IRenderContext } from "./render-context";
 
+// The basic template abstraction that allows consumers to create
+// instances of an IView on-demand. Templates are contextual in that they are, in the very least,
+// part of a particular application, with application-level resources, but they also may have thier
+// own scoped resources or be part of another view (via a template controller) which provides a
+// context for the template.
 export interface ITemplate {
   readonly renderContext: IRenderContext;
   createFor(owner: IViewOwner, host?: INode, parts?: TemplatePartDefinitions): IView;
--- a/src/runtime/templating/view.ts
+++ b/src/runtime/templating/view.ts
@@ -24,6 +24,9 @@ export interface IViewOwner {
   $slots?: Record<string, IEmulatedShadowSlot>;
 }
 
+// This is an implementation of IView that represents "no DOM" to render.
+// It's used in various places to ensure that View is never null and to encode
+// the explicit idea of "no view".
 const noopView: IView = {
   firstChild: null,
   lastChild: null,
@@ -37,6 +40,9 @@ const noopView: IView = {
 
 export const View = {
   none: noopView,
+  // See below for an explanation of ContentView.
+  // This factory method creates the ContentView depending on whether or not
+  // shadow dom emulation is active on the host element.
   fromCompiledContent(host: INode, options: { contentOverride?: INode } = PLATFORM.emptyObject): IContentView | null {
     if (DOM.isUsingSlotEmulation(host)) {
       return new ContentView(options.contentOverride || host);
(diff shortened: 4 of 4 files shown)


src/runtime/dom.ts
```diff
@@ -281,6 +281,11 @@ function setAttribute(this: CommentProxy, qualifiedName: string, value: string):
   this.$proxyTarget.setAttribute(qualifiedName, value);
 }
 
+// This is the most common form of IView.
+// Every custom element or template controller whose view is based on an HTML template
+// has an instance of this under the hood. Anyone who wants to create a view from
+// a string of markup would also receive an instance of this.
+// CompiledTemplates create instances of TemplateViews.
 /*@internal*/
```
fkleuver: This is clear to me now, thanks.
It seems to me that this only lives in `dom.ts` because that's where the method needs access to the non-exported class.
Based on your clarification it would make more sense to me to let this class live under `view.ts`, and expose it via `ViewFactory.fromMarkup` method (similar to  `View.fromNode`, etc) and let DOM just be an abstraction of the DOM, not Aurelia-specific stuff like `TemplateView`. This way things are more logically grouped together (views with views) and you keep both the intended semantics and encapsulation of the public API. Does this make sense?

src/runtime/templating/rendering-engine.ts
```diff
@@ -226,6 +227,12 @@ function createDefinition(definition: Immutable<ITemplateSource>): TemplateDefin
   };
 }
 
+// This is the main implementation of ITemplate.
+// It is used to create instances of IView based on a compiled TemplateDefinition.
+// TemplateDefinitions are hand-coded today, but will ultimately be the output of the
+// TemplateCompiler either through a JIT or AOT process.
+// Essentially, CompiledTemplate wraps up the small bit of code that is needed to take a TemplateDefinition
+// and create instances of it on demand.
 class CompiledTemplate implements ITemplate {
```
fkleuver: So CompiledTemplate is essentially a wrapper for render context + view factory, right? Like a cached intersection point between what used to be a child container (now render context) and the view factory.

Side note, I feel there is some naming inconsistency (or I just misread) between ViewFactory and VisualFactory. FactoryProvider in RenderContext is registered as `IViewFactory` but it returns a VisualFactory. What's up with that?

Other than that, I think I get it :)

src/runtime/templating/template.ts
```diff
@@ -3,6 +3,11 @@ import { INode, IView } from "../dom";
 import { TemplatePartDefinitions } from "./instructions";
 import { IRenderContext } from "./render-context";
 
+// The basic template abstraction that allows consumers to create
+// instances of an IView on-demand. Templates are contextual in that they are, in the very least,
+// part of a particular application, with application-level resources, but they also may have thier
+// own scoped resources or be part of another view (via a template controller) which provides a
+// context for the template.
```
fkleuver: With the explanation it makes sense. The name Template is somewhat confusing to me given that it's really just a component context / component instance of sort, is it not?

src/runtime/templating/rendering-engine.ts
```diff
@@ -226,6 +227,12 @@ function createDefinition(definition: Immutable<ITemplateSource>): TemplateDefin
   };
 }
 
+// This is the main implementation of ITemplate.
+// It is used to create instances of IView based on a compiled TemplateDefinition.
+// TemplateDefinitions are hand-coded today, but will ultimately be the output of the
+// TemplateCompiler either through a JIT or AOT process.
+// Essentially, CompiledTemplate wraps up the small bit of code that is needed to take a TemplateDefinition
+// and create instances of it on demand.
```
fkleuver: So CompiledTemplate is essentially a wrapper for render context + view factory, right? Like a cached intersection point between what used to be a child container (now render context) and the view factory.

Side note, I feel there is some naming inconsistency (or I just misread) between ViewFactory and VisualFactory. FactoryProvider in RenderContext is registered as IViewFactory but it returns a VisualFactory. What's up with that?

Other than that, I think I get it :)

src/runtime/dom.ts
```diff
@@ -281,6 +281,11 @@ function setAttribute(this: CommentProxy, qualifiedName: string, value: string):
   this.$proxyTarget.setAttribute(qualifiedName, value);
 }
 
+// This is the most common form of IView.
+// Every custom element or template controller whose view is based on an HTML template
+// has an instance of this under the hood. Anyone who wants to create a view from
+// a string of markup would also receive an instance of this.
+// CompiledTemplates create instances of TemplateViews.
```
fkleuver: This is clear to me now, thanks.
It seems to me that this only lives in `dom.ts` because that's where the method needs access to the non-exported class.
Based on your clarification it would make more sense to me to let this class live under `view.ts`, and expose it via `ViewFactory.fromMarkup` method (similar to  `View.fromNode`, etc) and let DOM just be an abstraction of the DOM, not Aurelia-specific stuff like `TemplateView`. This way things are more logically grouped together (views with views) and you keep both the intended semantics and encapsulation of the public API. Does this make sense?

src/runtime/templating/view.ts
```diff
@@ -24,6 +24,9 @@ export interface IViewOwner {
   $slots?: Record<string, IEmulatedShadowSlot>;
 }
 
+// This is an implementation of IView that represents "no DOM" to render.
+// It's used in various places to ensure that View is never null and to encode
+// the explicit idea of "no view".
```
fkleuver: Not questioning the concept of a `noopView` itself, but where (and how) do we need to account for this in the core logic? It may not be null in itself, but the children are, so some checks are needed regardless right?

src/runtime/templating/view.ts
```diff
@@ -37,13 +40,31 @@ const noopView: IView = {
...
+  // This is required because there's no way to get a "loose" component into the view
+  // hierarchy without it being part of an element's ContentView or part of an IVisual.
+  // ContentViews are appended by the custom element's attach process. IVisuals can only
+  // be added via a RenderSlot. So, this form of view effectively enables a single
+  // component to be added into a RenderSlot.
+  // It's likely that after the RenderLocation abstraction is added and RenderSlot is removed
+  // that most of this functionality will go away. The rendering engine will still need the ability
+  // to create a dynamic component, but the compose element won't need a RenderSlot and an IView in
+  // order to add that dynamic component to the DOM. Instead, it should be able to take the single
+  // node and use the RenderLocation to add that node to the DOM.
```
fkleuver: Can this also be used exactly in place of `TemplateView` when you already have an existing node (rather than a template string)? And would this play a role in `enhance` kind of scenarios or am I missing the point?

src/runtime/templating/view.ts
```diff
@@ -79,6 +100,14 @@ interface IContentViewChildObserver extends IChildObserver {
...
 }
 
+// This is the default and only implementation of IContentView.
+// This type of view represents the content that is placed between the opening and closing tags
+// of a custom element. It is only used when ShadowDOM Emulation is required. It enables
+// various interactions between the content, as a logical view, the ShadowDom Emulation, 
+// and the RenderSlot.
+// Most if not all of this will go away when we remove the shadow dom emulation and render slot.
+// However, we may need to keep the IChildObserver abstraction for enabling the $children property
+// in different scenarios, depending on what optimizations we make around when/when not to use shadow dom.
```
fkleuver: Even if we remove our ShadowDOM emulation, we still need a special kind of view for interacting with the shadow dom right? So this would never completely go away in that sense, or?

EisenbergEffect: I'm not sure we'll need to worry about interacting with the shadow dom outside of a few specific times: attaching to the root, observing light dom children, and passing through the encapsulation source.

src/runtime/templating/view.ts
EisenbergEffect: It's only used inside of a custom element. That's the only case you can have a "no view" scenario. So, it has a basic check for this in the one or two places that matter.

src/runtime/dom.ts
EisenbergEffect: I ended up moving it here because I wondered whether different DOM implementations would need the ability to convert a string directly into a factory. In other words, exposing the lower-level pieces that were HTML-specific tied this more strongly to HTML than I felt might be a good idea, particularly if we want to target other things like Nidium and/or NativeScript.

src/runtime/templating/rendering-engine.ts
EisenbergEffect: I think this is all related to that visual/view naming issue we discussed in the chat. I'm hopeful that we'll arrive at the right names as we evolve these pieces.

src/runtime/templating/template.ts
EisenbergEffect: Per our discussion, this can apply to either a custom element or a "fragment" like what a repeater controls or what an if controls. Both CustomElement and Visual use templates to create views.

PR #53, opened 2018-08-03
WIP - Remove Shadow DOM Emulation

EisenbergEffect:
This PR removes all code relating to emulation of the Shadow DOM. Instead, vNext will have a collection of strategies for handling content projection. Anything that can't be statically composed together will require real shadow dom or a shadow dom polyfill. The new approach reduces the size of the runtime by removing several hundred lines of code. It also reduces framework complexity in the area of dom composition. After this is removed, we'll be able to eventually remove RenderSlot, resulting in simplification and performance gains in template controllers, esp. Repeater.


fkleuver:
Nice, love it!!
I'll fix the unit tests and build errors after merge, can't wait to do the matching cleanup in my parts :)

EisenbergEffect:
Submitting new PR shortly.

PR #57, opened 2018-08-10
feat(templating): add the low-level render location abstraction

EisenbergEffect:
## Description

This PR adds the minimal infrastructure to provide template controllers with an `IRenderLocation` instance, which is an `INode` representing the location in the DOM where the template controller can add/remove views.

## Benefits

This is a much simpler abstraction than `IRenderSlot`. The `IRenderLocation` gives the lowest-level possible power to a template controller, allowing for re-designed controllers that are more performant and use less memory.

## Next Steps

Next steps involve converting a simple template controller over to use the new abstraction, removing its usage of `IRenderSlot`. The first controllers converted will likely be `with` and `replaceable`. Once these are done, we can move on to `if` and/or `repeat`.

After all template controllers are updated to use the new abstraction, we can remove the old `IRenderSlot` and all associated code. This may also open up possibilities to re-work the `IVisual` interface and will likely lead to a re-thinking of the animation system as well. These tasks can be done incrementally.

## Testing

There's not much to test meaningfully in this PR. Futures PR which implement new template controllers will exercise this abstraction thoroughly as a side-effect.

### DIFF
--- a/packages/runtime/src/templating/render-location.ts
+++ b/packages/runtime/src/templating/render-location.ts
@@ -1,0 +1,5 @@
+import { DI } from '@aurelia/kernel';
+import { INode } from '../dom';
+
+export const IRenderLocation = DI.createInterface<IRenderLocation>().noDefault();
+export interface IRenderLocation extends INode { }
--- a/packages/runtime/src/templating/render-location.ts
+++ b/packages/runtime/src/templating/render-location.ts
@@ -1,5 +1,0 @@
-import { DI } from '@aurelia/kernel';
-import { INode } from '../dom';
-
-export const IRenderLocation = DI.createInterface<IRenderLocation>().noDefault();
-export interface IRenderLocation extends INode { }
--- a/packages/runtime/src/dom.ts
+++ b/packages/runtime/src/dom.ts
@@ -14,6 +14,11 @@ export interface INode extends INodeLike {
 
 export const INode = DI.createInterface<INode>().noDefault();
 
+export interface IRenderLocation extends INode { }
+
+
+export const IRenderLocation = DI.createInterface<IRenderLocation>().noDefault();
+
 /**
  * Represents a DocumentFragment
  */
@@ -92,10 +97,6 @@ export const DOM = {
 
   createElement(name: string): INode {
     return document.createElement(name);
-  },
-
-  createAnchor(): INode {
-    return document.createComment('anchor');
   },
 
   createNodeObserver(target: INode, callback: MutationCallback, options: MutationObserverInit) {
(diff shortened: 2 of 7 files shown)


packages/runtime/src/templating/render-location.ts
```diff
@@ -0,0 +1,5 @@
+import { DI } from '@aurelia/kernel';
+import { INode } from '../dom';
+
+export const IRenderLocation = DI.createInterface<IRenderLocation>().noDefault();
+export interface IRenderLocation extends INode { }
```
fkleuver: Can we add some methods to it that we already know we'll need?
- `appendChild(child: INode): INode`
- `cloneNode(deep: boolean): INode`
- `insertBefore(newNode: INode, refNode: INode): INode`
- `removeChild(child: INode): INode`
- `replaceChild(newChild: INode, oldChild: INode): INode`

This gives API overlap with `DOM`, but I don't think that has to be a bad thing.

EisenbergEffect: I'd prefer to hold off and see how we actually use it, esp. in the context of the repeater. I'm worried that it could turn into IRenderSlot again. For the moment, I just wanted a simple abstraction to get the anchor itself injected. Once we build some stuff out, we can change it to not be an INode, but to be something that has a simple API based on refactoring from our template controllers.

fkleuver: Just adding to this thought, overall I think we should make that `INode` a little more useful as an abstraction in itself. Everywhere in the framework we're already type-casting it to a normal `Node` just to get to those methods. If we need to do that, consumers of the framework likely also do. I think this beats the purpose of the abstraction, only adding more bloat with all the type casts. Maybe I'm missing something though

EisenbergEffect: I was trying to come up with a minimal abstraction that allowed the framework not to take a dependencies on Node. We shouldn't be casting to Node anywhere other than in DOM, otherwise we lose the ability to swap out HTML for something like NativeScript.

fkleuver: To what extent are the `DOM` and `NativeScript` APIs incompatible? I'm not sure if we should try to abstract that away in the core framework from the get-go. We can always use AOT to do some NativeScript-specific transformations later on.

For me. the problem with RenderSlot was the fact that it hid `Node` methods behind an inefficient intermediate array. There wasn't necessarily much wrong with the abstraction itself, rather, the implementation was the problem.

If we have an interface that effectively allows you to talk directly to a `Node` without casting it to that, it would solve the problem I had with `RenderSlot`, as well as the requirement of not taking a direct dependency on `Node` in the rest of the framework.

fkleuver: Comparing to `RenderSlot`, technically all I need in the repeater is `appendChild` to retain current functionality (this is what `addVisualCore` does). If we also want to support "keyed mode" (and I think we should, because the performance benefit doesn't always outweigh the benefit of physically being able to rearrange elements) then we also need `insertBefore` at least

EisenbergEffect: I hadn't thought about using AOT to transform in this way. That sounds possible, and it is probably better to make INode more meaningful by moving some things from DOM to the INode itself.

The idea behind the IRenderLocation was just to give you a "spot" in the DOM. You control adding nodes before/after that spot, keeping track of them, removing them, etc. It doesn't have any children since it's just the anchor node itself.

EisenbergEffect: @fkleuver Do you want to take a shot at re-vamping the render slot instead of merging this PR?

EisenbergEffect: The main thing is that the render slot worked in terms of visuals, not in terms of nodes.

fkleuver: How would you add nodes before/after that spot then? Via the `DOM` api?

EisenbergEffect: Via the Visual api. Repeater gets a factory that creates Visual instances, which you can insertBefore an INode.

fkleuver: If I have a clear mental picture of which kinds of DOM structures would translate to which kinds of templating class trees, I could certainly try refactoring RenderSlot. This is currently not the case - I've never fully understood the internals of templating and it's still a black box to me. I blame that on ShadowDOM :)

We can merge this PR, it doesn't introduce changes so it can't do any harm. Gives me some time to look at it and think about it. Maybe I'll better understand how this is supposed to work after letting it sink for a bit.

EisenbergEffect: I'll merge it and then submit a PR for `with` or `replaceable` so you can see how I'm using it to change those. Those are vastly simpler than `repeat` but it might be enough info to help you figure out if you want to go this route or focus on refactoring `RenderSlot`.

EisenbergEffect:
@fkleuver Please have another look at this PR. I made the terminology change throughout to remove "anchor". Full-text search on the runtime indicates no more uses of the word.

PR #58, opened 2018-08-11
feat(binding): implement ChangeSet

fkleuver:
refactor(binding): use ChangeSet instead of TaskQueue
refactor(property-observer): make reusable decorator for setter/observer


packages/runtime/src/binding/observation.ts
```diff
@@ -88,27 +131,93 @@ export interface ICollectionSubscriber { handleChange(origin: string, args?: IAr
...
- * A collection of collection subscribers
+ * Either a property or collection subscriber
  */
-export interface ICollectionSubscriberCollection {
-  subscribers: Array<ICollectionSubscriber>;
+export type Subscriber = ICollectionSubscriber | IPropertySubscriber;
+/**
+ * Either a batched property or batched collection subscriber
+ */
+export type BatchedSubscriber = IPropertySubscriber | IBatchedPropertySubscriber;
```
EisenbergEffect: Is this correct? Shouldn't this include the batched collection subscriber?

packages/runtime/src/binding/observation.ts
```diff
@@ -88,27 +131,93 @@ export interface ICollectionSubscriber { handleChange(origin: string, args?: IAr
...
+
+/**
+ * Helper type that translates from mutationKind enum to the correct subscriber interface
+ */
+export type MutationKindToSubscriber<T> =
```
EisenbergEffect: Clever.

packages/runtime/src/binding/signaler.ts
```diff
@@ -1,17 +1,18 @@
...
-  addSignalListener(name: Signal, listener: ICallable): void;
-  removeSignalListener(name: Signal, listener: ICallable): void;
+  addSignalListener(name: Signal, listener: IPropertySubscriber): void;
+  removeSignalListener(name: Signal, listener: IPropertySubscriber): void;
 }
 
 export const ISignaler = DI.createInterface<ISignaler>()
   .withDefault(x => x.singleton(class {
-    private signals: Record<string, ICallable[]>;
+    private signals: Record<string, IPropertySubscriber[]>;
```
EisenbergEffect: I think this might be a bug in my original conversion. Shouldn't this be initialized to `{}`?

packages/runtime/src/binding/subscriber-collection.ts
```diff
@@ -96,84 +98,70 @@ export abstract class SubscriberCollection {
...
       }
     }
 
-    if (context0) {
-      if (callable0) {
-        callable0.call(context0, newValue, oldValue);
-      } else {
-        context0(newValue, oldValue);
-      }
+    if (subscriber0) {
```
EisenbergEffect: Should we check explicit `!== null` in these cases for perf reasons? It's a hot code path so maybe we should make that optimization?

packages/runtime/test/unit/binding/change-set.spec.ts
```diff
@@ -0,0 +1,231 @@
+import { spy, SinonSpy } from 'sinon';
```
EisenbergEffect: We had almost no tests for the task queue, so I'm really happy to see the tests on the new change tracking mechanism from the get go. Thanks!

packages/runtime/test/unit/binding/set-observer.spec.ts
```diff
@@ -3,6 +3,7 @@ import { SetObserver, enableSetObservation, disableSetObservation, nativeSetDele
 import { expect } from 'chai';
 import { stringify, SpySubscriber } from '../util';
 import { IndexMap } from '@aurelia/runtime';
+import { ChangeSet } from '../../../src/binding/change-set';
```
EisenbergEffect: I noticed that some things are coming from `@aurelia/runtime` and some are relative. Was there a particular reason for this or was it just the tool injecting the import statements that way automatically? If it's all the same, I like the `@aurelia/runtime` import.

packages/runtime/src/binding/observation.ts
fkleuver: Yep, I missed that, fixed it. It slipped through because I'm not currently using this anywhere. This might just need to be cleaned up in the next iteration.

packages/runtime/src/binding/signaler.ts
fkleuver: I'm not entirely sure. This and a lot of other changes in binding were just to make typescript happy again. There aren't any unit tests in place for this yet, so I'd need to start with that and then see if anything needs fixing.

packages/runtime/src/binding/subscriber-collection.ts
fkleuver: We could, but I'm planning to deprecate subscriber-collection in favor of a decorator anyway. I found a slightly faster mechanism. Otherwise there are quite a few more optimizations I can think of for subscriber-collections.
As this was just an intermediary step to get it all to work first, I simply tried to keep the number of changes to a minimum.

packages/runtime/test/unit/binding/set-observer.spec.ts
fkleuver: This is actually `/*@internal*/` doing its work :)
Those things are not written out to type declarations (consumers can't import them at all), so they can only be imported relatively from within the unit tests.

PR #60, opened 2018-08-12
Observer rework

fkleuver:
@EisenbergEffect this PR contains a rework of all element observers, and accessors. 
Various tweaks are still needed (typings, naming, usage of flags), but the overall interaction between the observers and `ChangeSet` is more or less in place.

Unit tests for the element observers cover the most basic scenario (in case of `select` / `checkbox`) and they probably don't fully work yet. I'll do this iteratively as there was/is quite some ground to cover here.

I housed all element observers together under the `element-observation.ts` file and while doing that I saw more similarities emerge. Once everything works I'll probably pull a few functions out into decorators and reduce the amount of code in that file.

Let me know what you think


EisenbergEffect:
@fkleuver Looks much cleaner! I think there will be a couple of conflicts with my branch, but nothing that can't be fairly easily resolved. I'm happy to merge this if you feel it's ready. Let me know the final word.

fkleuver:
I'd just like to have another glance over it after your branch is merged in. I'll do that together with fixing and merge conflict and ping you when I'm done :)

fkleuver:
@EisenbergEffect  I cleaned things up a bit more. Extracted common logic out into a `@targetObserver` decorator, moved target accessors to their own file (and renamed the suffix Observer to Accessor), couple minor tweaks to typings and change detection and cleaned up `SelectValueObserver` a bit more.

`SelectValueObserver` and `CheckedObserver` are in need of more unit tests (and may need a few fixes as a result of testing them) but I'd prefer to take that after doing the repeater and template compiler, as well as cleaning up the unit tests a bit more. These things are tedious to test properly and I need more common helpers to be in place for doing that.

PR #61, opened 2018-08-13
fix(runtime): interface clarity

EisenbergEffect:
This PR renames the following interfaces:

* `IView` => `INodeSequence`
* `IVisual` => `IView`
* `IVisualFactory` => `IViewFactory`
* `IRenderSlot` => `IViewSlot`
* `IViewOwner` => `IRenderable`

Concrete implementations have been renamed with the same naming scheme. Properties and parameters have also been renamed where appropriate. Additionally, the `bindable` and `attachable` arrays have been pluralized.

This restores some backwards compatibility with vCurrent.


packages/runtime/src/templating/custom-element.ts
```diff
@@ -374,8 +374,8 @@ class HostProjector implements IViewProjector {
     return parentEncapsulationSource || this.host;
   }
 
-  public project(view: IView): void {
-    view.appendTo(this.host);
+  public project(nodes: INodeSequence): void {
+    nodes.appendTo(this.host);
```
fkleuver: This kind of thing is where the rename really shines. Love it!

.vscode/settings.json
```diff
@@ -22,5 +22,10 @@
     "**/coverage": true,
     "**/node_modules": true,
     "**/scripts": true
-  }
+  },
+  "cSpell.words": [
+    "Renderable",
+    "attachables",
+    "bindables"
+  ]
```
fkleuver: So they're being squiggled as spelling errors? Are these the only words for which that happens?

fkleuver:
Glad we had that discussion, these are some really good improvements! Let's get it in :)
You can go first, I'll fix the merge conflicts in my PR. I want to take another quick glance over my changes with these in.

.vscode/settings.json
EisenbergEffect: I'm just adding words as I see them with squiggles. There are probably a few more.

PR #64, opened 2018-08-14
feat(runtime): convert with attribute to use render location

EisenbergEffect:
This PR converts the `with` template controller to use `IRenderLocation` instead of `IViewSlot`. As part of this PR, various fixes needed to be made to bindable self callbacks. A `boolean` was introduced to turn on/off self callback firing, making this more explicit. Additionally, several interfaces related to lifecycles and runtime behaviors were improved.


packages/runtime/test/unit/templating/resources/template-controller-tests.ts
```diff
@@ -0,0 +1,71 @@
+import { expect } from 'chai';
+import { Constructable } from '@aurelia/kernel';
+import { BindingFlags } from "@aurelia/runtime";
+import { ViewFake } from '../fakes/view-fake';
+import { hydrateCustomAttribute } from '../attribute-assistance';
+import { createScope } from '../scope-assistance';
+
+export function ensureSingleChildTemplateControllerBehaviors(Type: Constructable) {
```
fkleuver: Nice!

packages/kernel/src/toggle.ts
```diff
@@ -0,0 +1,19 @@
+export class Toggle {
```
fkleuver: I'm a little concerned about having a whole class for this. It would make perfect sense in compiled/server-side code, but this is using a 24 bit object for storing 2 bits of data. Why not a direct property or a flag?

fkleuver:
Love what you did with the unit tests there as well as the public API.
Could you clarify what problem the toggle solves (I might be overlooking a crucial line of code somewhere) and why it needs to be in a separate class?

packages/kernel/src/toggle.ts
EisenbergEffect: Actually, we can get rid of Toggle. Originally, when I started to address some bugs, there was an issue of there being no place to store the shared boolean. After I refactored and fixed some other issues, the Toggle remained, but could be replaced by a simple boolean. I didn't see it because my head was still in the previous problem with that, which no longer existed. I'll get this switched over to a simple boolean on the instance and remove the unnecessary class.

PR #69, opened 2018-08-16
refactor(all): more flags

fkleuver:
:triangular_flag_on_post:


packages/runtime/src/aurelia.ts
```diff
@@ -32,15 +32,15 @@ export class Aurelia {
         );
       }
 
-      component.$bind(BindingFlags.none);
+      component.$bind(BindingFlags.isStarting | BindingFlags.isBinding);
```
fkleuver: @EisenbergEffect this should give you the differentiation between first bind and consecutive binds

packages/runtime/src/binding/binding-flags.ts
```diff
@@ -1,14 +1,17 @@
...
+  isBinding           = 0b0000000010000,
+  isUnbinding         = 0b0000000100000,
+  isChange            = 0b0000001000000,
+  sourceContext       = 0b0000010000000,
+  targetContext       = 0b0000100000000,
+  computedContext     = 0b0001000000000,
+  checkedArrayContext = 0b0010000000000,
+  checkedValueContext = 0b0100000000000,
+  selectArrayContext  = 0b1000000000000,
+  context             = 0b1111111000000
```
bigopon: Names like `isChange` poses a challenge for new readers without proper comments, should we change the name or add some comments ? I feel so because this `BindingFlags` is used in multiple scenarios, not a single context to understand

fkleuver: I agree. This is just me trying to keep the name short for no good reason. I'll change the name.

packages/runtime/src/aurelia.ts
bigopon: Yes! This is small but very nice improvement. Probably very few places will benefit from this clarification but it's nice nonetheless. @fkleuver 👏

fkleuver:
@EisenbergEffect @bigopon I renamed `isChange` to `isBindableCallback` and added some clarifying comments. Does this clear things up? :)

bigopon:
🏁  🥇

EisenbergEffect:
Comments are great. However, I'm not sure about the name `isBindableCallback`. Somehow `isChange` seemed better. What are some other names that might work? `isUpdate`?

fkleuver:
It's a tough one. So `isBindableCallback`, for one, is stupid because it's not proper grammar. What it would need to be is `originatesFromBindableCallback`.
`isChange` or `isUpdate` are of course cleaner but they might be too broad. Anything coming from an observer is technically an update, even if it is during `$bind`.
This flag needs to indicate *where it originates from*, rather than *what it is*.
I'm not sure on what a proper and clean name for that would be

EisenbergEffect:
We have a number of flags that use the "context" terminology. Can we use that here? Maybe something like `callbackContext` ?

fkleuver:
I was thinking about that, but I think that might water down the meaning of "context". I actually don't really like `sourceContext`, `targetContext` etc that much. They always confused me a little bit. How about changing it all to `origin`? e.g. `sourceOrigin`, `targetOrigin`, `callbackOrigin`, `$bindOrigin`

EisenbergEffect:
That works for me. Not sure if we need `$` for bindOrigin. It's a toss up.

fkleuver:
Like this? (not too sure about computed/checked/select origin, but as eluded already I need to review that anyway)

fkleuver:
Whoops looks like I broke some tests, one moment

fkleuver:
Ok fixed 🐛

EisenbergEffect:
Yeah. I agree. Merging this in the interest of moving forward, since the updates to the usage of the flags is key to get in now.

PR #70, opened 2018-08-16
refactor(all): improving a couple of method names related to resources

EisenbergEffect:
## In This PR

* I renamed the `get` method of `IResourceDescriptions` to `find`. I'm making a conscious movement away from calling things "get" in code as it has very little explanatory power.
* I renamed the `key` method of `IResourceKind` to `keyFrom` to make it clear that the key was being derived from the input. `key` seemed too much like a property name rather than a name for a pure function.

## Next Steps

I'd like to change the `IServiceLocator` interface to not use the `get` or `getAll` names but instead use `resolve` and `resolveAll`, or something similar. This would be a breaking change from vCurrent so this might be a case where we want to keep the old methods as an alias. I'm not sure at this time, so I'm delaying this change.

### DIFF
--- a/packages/jit/src/templating/binding-command.ts
+++ b/packages/jit/src/templating/binding-command.ts
@@ -17,7 +17,7 @@ export function bindingCommand(nameOrSource: string | IBindingCommandSource) {
 export const BindingCommandResource: IResourceKind<IBindingCommandSource, IBindingCommandType> = {
   name: 'binding-command',
 
-  key(name: string): string {
+  keyFrom(name: string): string {
     return `${this.name}:${name}`;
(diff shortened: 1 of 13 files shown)


fkleuver:
I think `keyFrom` and `find` are definitely improvements.
Same for `get` -> `resolve`. The breaking change issue can easily be addressed by a `vcurrent-compat` type of plugin that adds various aliases to things. I don't think we should be too concerned by that when talking purely about method names.

EisenbergEffect:
Sounds good. I'll merge this in and at some point in the future do a PR for the container API improvements.

PR #71, opened 2018-08-16
fix(runtime): dry up bindable callback behavior

EisenbergEffect:
This PR attempts to simplify the behavior around bindable callbacks.

## Background

In vCurrent, whether or not a bindable self callback is invoked during the binding process depends on whether or not you have a `bind` callback. If you have a `bind` callback, the self callbacks for each property will not be called. However, if you don't have a `bind`, they will be called. If they are called, then they are called one by one, after each property is set, rather than all together after all properties are set.

## New Behavior

For vNext I'm proposing that bindable self callbacks never fire during initialization, regardless of whether or not you have a `bound` callback. This makes it very clear how to implement behaviors. If you need to do something upon initialization, you implement `bound`. If you need to react to changes after the initial setup, you implement a callback for the properties you care about. This also makes the `bound` callback behave similarly to other frameworks `ready` callbacks. In addition to this behavior being easier to understand, it is a pure reduction in code on our part and removes odd special-case handling for bindable callbacks.

> Note: This PR also fixes an issue with `Observable` where the binding flag for `bindOrigin` was not properly preventing normal subscribers, but only self callbacks. Now all notification is skipped in this case, treating all callbacks the same and aligning Observable with the behavior of the other observers.

### DIFF
--- a/packages/runtime/src/binding/property-observation.ts
+++ b/packages/runtime/src/binding/property-observation.ts
@@ -137,7 +137,8 @@ export class Observer<T>  implements IPropertyObserver<IIndexable, string> {
   constructor(
     changeSet: IChangeSet,
     currentValue: T,
-    private selfCallback?: (newValue: T, oldValue: T) => void | T) {
+    private selfCallback?: (newValue: T, oldValue: T) => void | T
+  ) {
       this.changeSet = changeSet;
       this.currentValue = currentValue;
 
@@ -159,15 +160,15 @@ export class Observer<T>  implements IPropertyObserver<IIndexable, string> {
       this.previousValue = previousValue;
       this.currentValue = newValue;
 
-      if (this.selfCallback !== undefined) {
-        const coercedValue = this.selfCallback(newValue, previousValue);
-
-        if (coercedValue !== undefined) {
-          this.currentValue = newValue = <T>coercedValue;
-        }
-      }
-
       if (!(flags & BindingFlags.bindOrigin)) {
+        if (this.selfCallback !== undefined) {
+          const coercedValue = this.selfCallback(newValue, previousValue);
+
+          if (coercedValue !== undefined) {
+            this.currentValue = newValue = <T>coercedValue;
+          }
+        }
+
         this.notify(newValue, previousValue);
       }
     }
--- a/packages/runtime/src/templating/runtime-behavior.ts
+++ b/packages/runtime/src/templating/runtime-behavior.ts
@@ -1,4 +1,3 @@
-import { PLATFORM } from '@aurelia/kernel';
 import { BindingFlags } from '../binding/binding-flags';
 import { IChangeSet } from '../binding/change-set';
 import { IAccessor, IPropertySubscriber, ISubscribable, MutationKind } from '../binding/observation';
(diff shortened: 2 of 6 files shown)


davismj:
```javascript
export class MyCustomElement {
  @bindable value;

  bound() {
    console.log('bound', this.value);
  }

  attached() {
    setTimeout(() => this.value = 'foo', 1000);
  }

  valueChanged() {
    console.log('valueChanged', this.value);
  }
}
```

```console
bound, undefined
valueChanged, 'foo'
```

Like this?

packages/runtime/src/templating/runtime-behavior.ts
```diff
@@ -93,21 +92,17 @@ export class RuntimeBehavior implements IRuntimeBehavior {
...
-        bindableCallbacks[i] = () => instance[changeHandler](instance[name]);
-        observers[name] = new Observer(changeSet, instance[name], (n, o) => instance[changeHandler](n, o));
-      } else {
-        bindableCallbacks[i] = PLATFORM.noop;
-        observers[name] = new Observer(changeSet, instance[name]);
-      }
+      observers[name] = new Observer(
+        changeSet,
+        instance[name],
+        changeHandler in instance ? (n, o) => instance[changeHandler](n, o) : undefined
```
fkleuver: Is it possible for the prototype to change after the decorators are applied? Doesn't this just add additional overhead to the bindable self callbacks?

fkleuver:
Correct, and it would also be like this:

```
export class MyCustomElement {
  @bindable value;

  attached() {
    console.log('attached', this.value);
    this.value = 'foo';
    console.log('attached', this.value);
  }

  valueChanged() {
    console.log('valueChanged', this.value);
  }
}
```

```
attached, undefined
valueChanged, 'foo'
attached, 'foo'
```

Which in vCurrent would have been this:

```
attached, undefined
attached, 'foo'
valueChanged, 'foo'
```

packages/runtime/src/templating/runtime-behavior.ts
EisenbergEffect: In vCurrent, we had people who wanted to define some callbacks within the constructor, rather than as normal class methods. So, this attempts to support that.

EisenbergEffect:
To provide some more context around using the element.

If we have this definition:

```TypeScript
export class MyCustomElement {
  @bindable value;

  bound() {
    console.log('bound', this.value);
  }

  attached() {
    setTimeout(() => this.value = 'foo', 1000);
  }

  valueChanged() {
    console.log('valueChanged', this.value);
  }
}
```

And we use it like this:

```HTML
<my-custom-element value="bar"><my-custom-element>
```

The the event sequence will be:

* constructor
* element hydration
* `$bind` phase begins
* `bound` callback fired => console output: "bound", "bar"
* `$bind` phase ends
* `$attach` phase begins
* `attaching` callback fired
* element is projected into the DOM
* `attached` callback fired => `setTimeout` requested
* `$attach` phase ends
* 1000ms pass
* setTimeout callback invoked => `value` is set to "foo"
* valueChanged callback invoked => console output: "valueChanged", "foo"

The console output sequence is:

* "bound", "bar"
* "valueChanged", "foo"

packages/runtime/src/binding/property-observation.ts
```diff
@@ -159,15 +160,15 @@ export class Observer<T>  implements IPropertyObserver<IIndexable, string> {
...
-        if (coercedValue !== undefined) {
-          this.currentValue = newValue = <T>coercedValue;
+          if (coercedValue !== undefined) {
+            this.currentValue = newValue = <T>coercedValue;
+          }
         }
-      }
 
-      if (!(flags & BindingFlags.bindOrigin)) {
         this.notify(newValue, previousValue);
```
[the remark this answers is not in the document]
EisenbergEffect: Bindables need to prevent all callbacks during the bind phase since the element isn't stable. The initial binding should be thought of a bit as if it's part of the constructor.

packages/runtime/src/templating/runtime-behavior.ts
fkleuver: Fair enough. Side note (out of scope of this PR), wouldn't it be cleaner to just create a `BindableObserver` for this? So we can just pass instance + name and let the observer deal with this piece of logic as well as determining the name (so the `+Changed` convention lives in that class too)?
If I'm not mistaken we only have this observer defined this way so users have the ability to use it completely outside of binding/templating. But for `@bindable` and `@observable` decorators we have a known convention and a discoverable configuration, we we could save the overhead of the arrow function there.

packages/runtime/src/binding/property-observation.ts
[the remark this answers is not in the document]
fkleuver: LOL. I misread the piece of code and I thought you left `this.notify` outside of the `if (!(flags & BindingFlags.bindOrigin))` statement, and was wondering why. But you didn't, so never mind :)

packages/runtime/src/templating/runtime-behavior.ts
EisenbergEffect: 100% agree. I can do that in a future PR.

PR #72, opened 2018-08-16
(WIP) binding tests and small tweaks

fkleuver:


packages/runtime/src/binding/ast.ts
```diff
@@ -274,17 +274,11 @@ export class AccessMember implements IExpression {
 
   public evaluate(flags: BindingFlags, scope: IScope, locator: IServiceLocator) {
     const instance = this.object.evaluate(flags, scope, locator);
-    return instance === null || instance === undefined ? instance : instance[this.name];
+    return instance === null || instance === undefined ? undefined : instance[this.name];
```
bigopon: 👍

packages/runtime/src/binding/ast.ts
```diff
@@ -274,17 +274,11 @@ export class AccessMember implements IExpression {
...
 
   public assign(flags: BindingFlags, scope: IScope, locator: IServiceLocator, value: any) {
-    let instance = this.object.evaluate(flags, scope, locator);
-
-    if (instance === null || typeof instance !== 'object') {
-      instance = {};
-      this.object.assign(flags, scope, locator, instance);
-    }
-
+    const instance = this.object.evaluate(flags, scope, locator);
```
bigopon: Is this temporary behavior ?

fkleuver: Yeah, I'm trying to reduce the number of variations and code paths and make everything as strict as possible first, get it fully covered, and only then add some (configurable) fallbacks back in for backwards compat. Currently most of these features that are meant to be user-friendly are actually a source of inconsistencies and make the runtime behave differently from how normal JS would. I'd generally favor graceful failure over attempting to fix issues anyway

bigopon: I think we can separate this behavior into 2 directions: read and write. For writing, it would be fine for a graceful failure but reading it would be a real bummer

bigopon: Also failure in complex / nested access member in templating will make Aurelia feel very primitive I believe. I'm not sure how Vue and others are doing regarding this but from the look of `data` property of Vue, I think it's not developer friendly here either

EisenbergEffect: Definitely we'll want to put the code back in place that prevents failures, since that turns out to be very practical. I'm definitely fine with doing that in a future PR after certain other things are stabilized. If we can add a setting to let developers toggle that behavior, we'd fulfill an old request we've had on vCurrent as well. I think the default should be to match vCurrent if possible when you get around to that.

fkleuver:
I don't like the way my idea with the combined tests is working out here. Getting far too messy. I'm putting this PR in the fridge for now and going to try a different angle

PR #88, opened 2018-08-20
refactor(dom): use uppercase tag names

fkleuver:
HTML spec says that `Node.tagName` is always upper case when it is an element type, so we don't really need to normalize it. Slight perf improvement.

Though I'm clueless at the moment why those 4 tests are failing @bigopon you always catch these kinds of things, any idea? :bug:

### DIFF
--- a/packages/runtime/src/binding/event-manager.ts
+++ b/packages/runtime/src/binding/event-manager.ts
@@ -210,7 +210,7 @@ export class EventManager implements IEventManager {
 
   constructor() {
     this.registerElementConfiguration({
-      tagName: 'input',
+      tagName: 'INPUT',
       properties: {
         value: ['change', 'input'],
         checked: ['change', 'input'],
@@ -219,14 +219,14 @@ export class EventManager implements IEventManager {
     });
 
     this.registerElementConfiguration({
-      tagName: 'textarea',
+      tagName: 'TEXTAREA',
       properties: {
         value: ['change', 'input']
       }
     });
 
     this.registerElementConfiguration({
-      tagName: 'select',
+      tagName: 'SELECT',
       properties: {
         value: ['change']
       }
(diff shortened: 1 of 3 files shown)


bigopon:
For https://github.com/aurelia/aurelia/blob/d6a10b5fc2c30a18601e4df4d913aa466dd14a24/packages/jit/test/unit/templating/template-compiler.integration.spec.ts#L142

Does it work before ? It should always fail if initial value of `message` in component is `undefined`:

```js
let input = document.createElement('input');
input.value = undefined;
input.value === 'undefined';
```

bigopon:
For https://github.com/aurelia/aurelia/blob/c09c147981791fc526b4fa5a1257d0c45c5a9bdf/packages/jit/test/unit/templating/template-compiler.integration.spec.ts#L169-L177

It's failing because inconsistency here:

https://github.com/aurelia/aurelia/blob/c09c147981791fc526b4fa5a1257d0c45c5a9bdf/packages/runtime/src/binding/event-manager.ts#L252

We are using lowercase tag name, so your changes to upper case in config objects alone are not enough 😄 

We are doing it like this in vCurrent:
https://github.com/aurelia/binding/blob/a631a77c8f3f2676f2770c4ed5be3a801a4352cf/src/event-manager.js#L298

fkleuver:
So much for "find all" "tagName". Completely missed that one. Thanks 🤣 
I should ping you more often to help fix my bugs..

fkleuver:
@EisenbergEffect all mergable

bigopon:
@fkleuver @EisenbergEffect I think this is a tricky one. For different environment / dom engine, cases maybe preserved as is, to support XML naming

EisenbergEffect:
@fkleuver I think it's fine to remove `DOM.normalizedTagName`. Since the tagName is always uppercase in HTML and XML is supposed to be case sensitive, I don't think we'll need to do any normalization. I'd favor removing this for now. If some other platform appears where this is critical, we could put it back without too much trouble.

fkleuver:
@EisenbergEffect done

PR #90, opened 2018-08-21
fix(runtime): remove lazy bindable location and encapsulate observer

EisenbergEffect:
## Description

This PR removes the lazy location of bindable properties based on naming conventions. This change reduces the complexity of the runtime, making component hydration a little leaner. Additionally, this PR encapsulates the way that the Observer handles "self callbacks".

## Background

For vNext, I had hoped to make it possible to have completely POJO components, with bindables being automatically inferred based on property/callback naming conventions. So, this was implemented inside of the runtime. However, this approach to solving the problem didn't align with the idea that the runtime should work only off of explicit data and AOT should enable conventions that created that data. It also added needless complexity due to the fact that bindables could come from yet another location, at another time in the initialization process.

### DIFF
--- a/packages/runtime/src/binding/property-observation.ts
+++ b/packages/runtime/src/binding/property-observation.ts
@@ -139,13 +139,13 @@ export class Observer implements IPropertyObserver<IIndexable, string> {
   constructor(
     changeSet: IChangeSet,
     instance: object,
-    property: string,
-    callback: string
+    propertyName: string,
+    callbackName: string
   ) {
       this.changeSet = changeSet;
-      this.currentValue = instance[property];
-      this.callback = callback in instance
-        ? instance[callback].bind(instance)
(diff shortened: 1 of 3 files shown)


bigopon:
I remember reading somewhere `bound` function has penalty perf compared to arrow fn. Is this trade off ok ? I think if React folks are fine with `bind` everywhere, we are probably fine too

fkleuver:
@bigopon it's a close call and really depends on the exact closure you're getting. I've benched bound functions to perform better than arrows in the cases that I use them

EisenbergEffect:
Definitely agree on the naming. I'll clean that up real quick.

EisenbergEffect:
On the perf side, now that this is encapsulated, we can easily change the internals at any point in the future based on any benchmarking we do.

PR #91, opened 2018-08-21
Subscriber collection

### LINKED ISSUE #77: Rework SubscriberCollection
The disconnect/reconnect cycle is a very hot path that takes up to 50% of the CPU utilization in dbmonster-like benchmarks. There's no question that `$unbind` is where the biggest gains can be made in terms of performance.

With `BindingFlags` we don't need to store context / origin together with the subscribers anymore. We can simply pass the origin along with the flags from where the change originates, and the subscriber will have the information it needs to decide whether or not to act on a mutation.
This means `SubscriberCollection` will likely not need the relatively high-overhead array pool mechanism anymore and can become much simpler and have better performance.

Also look into making `SubscriberCollection` an injectable dependency (composition over inheritance) with the added benefit of making subscribers shareable / swappable. We might not need to fully disconnect and reconnect bindings anymore when the binding context changes.

fkleuver:
Closes #77

With the flags flowing through the system, the source/target context no longer need to be stored with the subscribers. They can simply be passed in from the source/target observers.

This allows SubscriberCollection to become an even more efficient, single-array collection rather than a dual array one.

I'm not 100% sure about the performance implications of the array pool mechanism that @jdanyow built in. There is certainly some overhead with the looping that it did, but it's also as small as it gets (due to using a boolean array). Now instead there is more overhead of the additional arrays being created by `.slice()`. On that token I will at a later stage try to benchmark this properly and see if the pooling would be worth adding back in. For now the goal was first and foremost to simplify things.

I realize there are still a few things in there that might seem vague or inconsistent. I'm not 100% sure myself yet either. I'm considering letting batched notifiers implicitly have `sourceOrigin` flags because changes coming from the target are never batched anyway.


bigopon:
I love the move, must state that first 😍 . Will have a look!

fkleuver:
:)
Seems like some e2e tests are failing (no surprises there). The checked and select value observers are still largely untested. It's a bit difficult for me to reason about what exactly they should be doing. vCurrent unit tests seem a tad strange to me. Will double check with the spec to see what it should be doing.

bigopon:
@fkleuver will this restrict the number of context types we can have ?

EisenbergEffect:
Side issue, what do think about using the naming pattern `fromSource` rather than `sourceOrigin`? Just still thinking through those names a bit.

packages/runtime/src/binding/ast.ts
```diff
@@ -264,7 +264,7 @@ export class AccessScope implements IExpression {
   public connect(flags: BindingFlags, scope: IScope, binding: IBinding) {
     const name = this.name;
     const context = BindingContext.get(scope, name, this.ancestor);
```
EisenbergEffect: Nothing to do with this PR, but this is another instance of `get`. What do you think about `BindingContext.locate()` instead?

EisenbergEffect: Or `find`?

bigopon: Its a lookup and walkup action, probably `locate` or `resolve` is better

packages/runtime/src/binding/array-observer.ts
```diff
@@ -342,18 +342,13 @@ export class ArrayObserver implements ICollectionObserver<CollectionKind.array>
...
 
-  public subscriberFlags: Array<BindingFlags>;
-  public batchedSubscriberFlags: Array<BindingFlags>;
-
   constructor(changeSet: IChangeSet, array: Array<any> & { $observer?: ICollectionObserver<CollectionKind.array> }) {
     this.changeSet = changeSet;
     array.$observer = this;
     this.collection = <IObservedArray>array;
     this.resetIndexMap();
     this.subscribers = new Array();
```
bigopon: Can we, at the very least, have some comment why this is preferred over `[]` ?

fkleuver:
> will this restrict the number of context types we can have ?

Yes and no. Let's say we'll have somewhere between 16-25 bit flags available for context. Should they run out, we can switch to a mechanism similar to what I do in the expression parser with the token types.

Instead of this:
```js
fooContext = 0b001,
barContext = 0b010,
bazContext = 0b100
```
Which can be used like so:
`if (flags & BindingFlags.fooContext)`

You'd have this:
```js
fooContext = 0b001,
barContext = 0b010,
bazContext = 0b011,
context    = 0b111
```
Which can be used like so:
`if ((flags & BindingFlags.context) === BindingFlags.fooContext)`

And then we'll have 2^16 - 2^25 contexts available. Might switch to this mechanism anyway because it would make it easier for users to add their own contexts for example by "registering" them in a context registry which then returns a number they can use, and a `isContext()` function they can use to check the flag. There are different ways this can be done.

fkleuver:
> Side issue, what do think about using the naming pattern fromSource rather than sourceOrigin? Just still thinking through those names a bit.

Thought about that too. `fromSource` is nice, we can go with that

packages/runtime/src/binding/ast.ts
fkleuver: Yeah I agree with you on the anti-get thing, though this seems like a relatively innocent one. I don't mind `locate()`. It gets me to think of `resolve()` as well, and I'm not quite sure yet which of those two (or possibly other alternatives) I'd prefer. Any particular reasoning behind `locate()` specifically?

packages/runtime/src/binding/array-observer.ts
fkleuver: Would you want that comment to be on every instance then? There are lots of places where I do this, maybe better suited for general engineering docs.
Anyway, the reason I use the array constructor is mostly for consistency with instances when I pass in the length (which is for performance reasons - eagerly allocate fixed size array and then assigning indices directly is a notch faster than using push).
I'm indifferent about `new Array()` vs `[]` otherwise.
I don't recall ever using `Array(length)`

bigopon: It may not be true any more, but i remember `Array(length)` is preferable over `new Array(length)`. And `[]` is faster than `new Array()` http://jsben.ch/rZsp6
I think this can be an exception where we don't have to enforce consitency

fkleuver: Could it be that `Array(length)` is preferable for the same reason that `[]` is faster. Maybe there's some in-memory cleverness going on with lazy (or not at all) allocating objects. I'm cool with doing it like that.

EisenbergEffect: For arrays, I think we should use `[]` if we don't know the length and `new Array(length)` if we do (accompanied by assigning indices).

fkleuver: And what about typing? `arr: Array<Foo>` or `arr: Foo[]`? We're currently mixing this.
I'm not sure about this myself. Obvious pro for `Foo[]` is that it's shorter. May be more readable in certain cases. A con is that it's not consistent with other collection types (there are no shorthands for map/set). So when you've got all 3 somewhere, it looks awkward.
A pro for  `Array<Foo>` is the consistency with other classes and not needing to use parentheses when you combine types (`Array<Foo | Bar>` vs `(Foo | Bar)[]`

StrahilKazlachev: I do mix the typings, all simple forms with `[]` - `MySpecialType[]`, and use `Array<T>` only for something like `Array<string | boolean>`, `Array<{isLoaded: string} & MyBaseClass>`. In general I don't have much `Array<>` in the types.

EisenbergEffect: I'm with @StrahilKazlachev on this one.

fkleuver: The pragmatic approach. I can go with that :)

bigopon:
I don't see any tests related to binding with `from-view` only, is it a todo ? I'm checking as I wonder will it be an error when observer relies on `setValue` to have proper `currentFlags`, which will later cause `Binding.protoype.handleChange` to have issue with this (I'm not sure how it's handled atm either)

bigopon:
@EisenbergEffect 👏 

```js
export const BindingFlags {
  fromSource,
  fromCheckedObserver
  fromValueObserver
}
```

Looks at how comprehensible they are! `context` isn't a wrong choice, but certainly I think this is better choice.

fkleuver:
Yeah it is. It could be that it doesn't work. There are a few things that don't work very well at the moment, checked observation is one of those. I am currently writing a bunch of unit tests that i'll commit to this PR when they're done

```
export const BindingFlags {
  fromSource,
  fromCheckedObserver
  fromValueObserver
}
```

I like!

packages/runtime/src/binding/binding-flags.ts
```diff
@@ -1,45 +1,40 @@
 export enum BindingFlags {
-  none                = 0b000000000,
+  none                = 0b0000000000000000000000000000000,
```
bigopon: Would be nicer if we used number separator: `0b_0000_0000_0000_0000_0000_0000_0000_0000`

packages/runtime/src/binding/binding-flags.ts
```diff
@@ -1,45 +1,40 @@
 export enum BindingFlags {
-  none                = 0b0000000000000,
+  none                = 0b0000000000000000000000000000000,
```
bigopon: Would be nicer if we used number separator: `0b_0000_0000_0000_0000_0000_0000_0000_0000`

fkleuver: Yup. How to use it with binary notation tho?

bigopon:
The new names make the code less cryptic / puzzling, which is very nice

EisenbergEffect:
I like this changes! @fkleuver Is this ready to merge then?

fkleuver:
@EisenbergEffect Yeah. I wanted to add more tests on the element side of things but that fell out of scope of this one to begin with. Turned out to be a lot of work for checked observer to do it right, so I'd rather do that in parts. Let's merge this so we can work on other stuff without 50 merge conflicts :)

PR #96, opened 2018-08-24
fix(runtime): convert if/else to use render location

EisenbergEffect:
## Description

This PR switches the if/else from using `IViewSlot` over to using `IRenderLocation`. This greatly simplifies the if/else logic and makes the entire attribute synchronous. A few basic tests have been added to ensure the right views are being rendered for true/false conditions and that their lifecycles are correctly being invoked.

## Next Steps

* Convert `compose` to use `IRenderLocation`, which will finish moving all the basic built-in resources away from `IViewSlot`
* Revisit `IView` and explore improving the API around a few scenarios:
  * Simple setting of render location.
  * Encapsulate `remove`
  * Rework animation system integration. It should be encapsulated entirely within the view implementation.
* Revisit built-ins and switch them over to new `IView` APIs


bigopon:
Lgtm. Looks very simple, so I think I wont have any further comments until we use it in real JIT compiler. 👍

PR #97, opened 2018-08-25
refactor(instructions): change number type to string

### LINKED ISSUE #84: Change TargetedInstructionType to string (or use array instead of object lookup)
As discussed with @EisenbergEffect having numeric properties on an object is a performance killer. Numeric properties / property indexers on a non-array object gives a member accessor that is between 20-50 times slower than either array indexers or normal string property accessors.
The consensus of our chat is to turn `TargetedInstructionType` into an object and prepend the numbers with a dollar `$1`, `$2`. It just needs to happen now. I'll look into this soon-ish but I'm not currently working on it, so anyone else looking for something to do feel free to pick it up

fkleuver:
Closes #84 

- Changes `instruction.type` from number to string. Also converts the enum to const enum (preserve const enums is enabled) for better performance by inlining.

- Added some unit tests to the renderer as an initial effort to improve test coverage in that area, and to have some lower level verification that the instructions work

- Made a few binding properties public on the classes themselves. Primarily for unit testing, but I believe it makes sense for them to be exposed (if you go so far as to cast to a concrete binding class, you probably want to access some of its properties too).

### DIFF
--- a/packages/jit/src/templating/template-compiler.ts
+++ b/packages/jit/src/templating/template-compiler.ts
@@ -240,8 +240,9 @@ export class OneTimeBindingInstruction implements IPropertyBindingInstruction {
   public mode: BindingMode.oneTime;
   public srcOrExpr: string | IExpression;
   public dest: string;
-  constructor(srcOrExpr: string | IExpression) {
+  constructor(srcOrExpr: string | IExpression, dest: string) {
     this.srcOrExpr = srcOrExpr;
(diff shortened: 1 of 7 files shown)


fkleuver:
@EisenbergEffect Should be good to merge if you're ok with the changes. The rest of the test coverage will follow in another PR (the current tests at least verify that nothing broke)

PR #98, opened 2018-08-25
feat(jit): move/extend attribute value parsing to the expression parser

### LINKED ISSUE #74: Move all JIT attribute value parsing to the ExpressionParser
- Interpolation parsing currently happens in the TemplateCompiler. Moving this to the ExpressionParser allows interpolation expressions to be cached just like other expressions
- Semicolon-separated custom attribute values are not being parsed yet. This should be done by the ExpressionParser.
- Pass more context information (attribute type + binding command) to the ExpressionParser so it can make the right decisions about valid/invalid syntax.
- ~~Implement loose vs strict parsing mode, where strict mode will detect and throw errors as early as possible (e.g. if a `two-way` binding command has a non-assignable expression, warn in loose mode and throw in strict mode?)~~ (not appropriate for this scope)

fkleuver:
Closes #74 

- [x] Interpolation parsing

- [ ] Semicolon-separated custom attribute values


fkleuver:
@EisenbergEffect  @bigopon template-compiler.ts should be in a more "intention-revealing" state now. I'm leaving semicolon-separated attribute parsing behind in a limbo of sorts (that is, there's a placeholder with the right shape but I'm not sure if it works). If the changes are ok with you then this should be good to merge.

bigopon:
Nice. Lgtm. This is what I wanted to do for vcurrent as well, when we already have template literal parsing in place, not really needed to have it again in the syntax intepreter.

EisenbergEffect:
Much cleaner. Very nice improvements! We can circle back around and address semicolon-separated attribute parsing at a later time. No worries there.

PR #99, opened 2018-08-26
chore(examples): initial setup for aurelia-cli

fkleuver:
@huochunpeng I tried to use `github:huochunpeng/cli#at31` for this one because the auto-tracing seems very convenient. It errors out though:

```
INFO [Bundler] Tracing files ...
commonJs.convert: COULD NOT CONVERT: (...)\jit-aurelia-cli\dist\src\app.d.ts, so skipping it. Error was: Error: Line 1: Unexpected token
ERROR [Bundler] Failed to do transforms
INFO [Bundler] Error: Line 1: Unexpected token
```

I think this is due to how the new typescript 3 project reference thing works. The references go to .d.ts files and find the originals via the mappings. I couldn't get it to work with the "normal" version of aurelia-cli either though. But that's because I didn't feel like adding 100 files mappings by hand :)

Any thoughts?

### DIFF
--- a/packages/examples/jit-aurelia-cli/src/app.ts
+++ b/packages/examples/jit-aurelia-cli/src/app.ts
@@ -1,0 +1,121 @@
+import { ArrayObserver, customElement } from '@aurelia/runtime';
+
+
+let id = 0;
+class Todo {
+  public done: boolean = false;
+  public id: number = id++;
+  constructor(public description: string, public app: App) {}
+}
+
+type ObservedTodos = Todo[] & { $observer: ArrayObserver };
+
+@customElement({
+  name: 'app',
+  templateOrNode: `
+  <template>
+    <div class="ui form">
+      <div class="fields">
+        <div class="five wide field">
+          <label>Description</label>
+          <input id="description" type="text" value.two-way="description">
+        </div>
+        <div class="two wide field">
+          <label>Count</label>
+          <input id="count" type="number" value.two-way="count">
+        </div>
+        <div class="two wide field">
+          <label>Log collection changes</label>
+          <input id="log" type="checkbox" checked.two-way="log">
+        </div>
+      </div>
+      <button id="addTodo" class="ui button" click.trigger="addTodo()">Add Todo</button>
+      <button id="clearTodos" class="ui button" click.trigger="clearTodos()">Clear Todos</button>
+      <button id="toggleTodos" class="ui button" click.trigger="toggleTodos()">Toggle Todos</button>
+    </div>
+    <div id="descriptionText">\${description}</div>
+    <div class="ui divided list todos">
+      <div repeat.for="todo of todos">
+        <div class="item todo">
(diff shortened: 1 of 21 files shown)


huochunpeng:
Why need those d.ts files for an app?

App code will be shipped in bundles. How would you use those d.ts files?

You can ask TS compiler to skip d.ts files. Or you can update transpile task to conditionally skip build.bundle() based on file name. 

Similar to this:
https://github.com/aurelia-contrib/aurelia-getting-started/blob/master/guides/how-to-support-markdown-preprocessing.md

fkleuver:
@bigopon Similar steps as jit-parcel:

From the root: `npm run init`
Then from the jit-aurelia-cli folder: `npm install && npm run watch`

@huochunpeng It's the new typescript 3 project system. Cross-project references go via the type definition mappings for some reason. Though I'm not sure the aurelia-cli is trying to resolve them? Maybe have to disable mappings?

fkleuver:
> Update: the easiest condition is to do positive condition on extname === '.js' for build.bundle

How precisely do I need to update the aurelia.json for that to work? Doesn't the `[**/*.js]` glob already take care of this?

huochunpeng:
cli is not in the position to solve it. *.d.ts is TypeScript, not JavaScript, the parser (esprima) doesn't support the format.

How would you ship those d.ts files? I don't think they work with bundle files (as bundle files are not built for someone to "import" in a TS project).

That `**/*.js` in aurelia.json only affects the decision of which bundle a file goes to. For example, if a d.ts file is bundled successfully, it will go to the last vendor-bundle as last resort, since no other bundles have wildcard to match d.ts file.

huochunpeng:
For the conditional build, it is `aurelia_project/tasks/transpile.ts` I was talking about. Not `aurelia.json`.

fkleuver:
@EisenbergEffect thanks to @huochunpeng 's amazing help on tweaking the cli to make this work, we now have a working aurelia-cli example. This one can be merged :)

packages/examples/jit-aurelia-cli/src/app.ts
```diff
@@ -0,0 +1,86 @@
...
+  public done: boolean = false;
+  public id: number = id++;
+  constructor(public description: string, public app: App) {}
+}
+
+type ObservedTodos = Todo[] & { $observer: ArrayObserver };
+
+@customElement({
+  name: 'app',
+  templateOrNode: <any>view,
```
EisenbergEffect: In the future, we'll want to do this a bit different, using a require.js plugin that actually provides the entire instruction. I've got a preliminary plugin written, so maybe a next step would be to incorporate that here. Not a reason to block merging, just mentioning it.

fkleuver: Yep I'm aware. I tried to do this with a similar mechanism to how you previous did that, but got various errors that I wasn't sure how to resolve. I'll make an issue for this

PR #101, opened 2018-08-26
feat(dbmonster): make dbmonster aot example work again

fkleuver:
This example currently gives about 70 FPS. No surprises there since RAF / connect-queue are out. We can always add this back in if we want, but I'd rather try and optimize things without it first. RAF was "masking" some perf issues in vCurrent (skipping renders doesn't really count as a fair higher repaint rate imo).
Seems a lot of the CPU goes into binding and connecting, so I'd like to try and see if we can somehow make that smarter and more efficient.


bigopon:
Some thoughts: Maybe add some number about vCurrent to make the goal obvious ? And skipping render is a valid way to have perf, the UI just needs to reflect their corresponding models good enough. I actually expected to have this when we were introducing binding flags, as we know what direction of the flush is. But like @fkleuver said, we can add this back later.

fkleuver:
I agree that skipping renders in itself is ok, but dumping everything in RAF like vCurrent does is not ideal.
We've had several discussions about just queueing the edge updates (e.g. targetObserver.setValue) via RAF but one or two people kept disagreeing with that. The internet seems very divided on this topic too.

All that said my problem with dbmonster is that I don't believe it truly measures the right thing. If you render nothing then you've got the 250 fps cap. If it took into account how many updates you're actually rendering, then that would tell something about how efficiently you're processing updates. Otherwise it's simply the more renders you skip, the better the perceived performance. That's not a helpful benchmark.

fkleuver:
Correction, it's the `connect`. Wrapping `setValue` on target observers in RAF accomplishes precisely nothing in terms of performance. Just wrapping the connect however triples FPS. Then if you also wrap the unobserve, the effect is mitigated again.

I'm concluding that the connect-queue is causing there to be small periods of the source expression not being observed. It disconnects immediately, and then re-connects only after RAF invoked.

I feel like we'd be better off by simply debouncing the observation wireup. It doesn't make sense to observe something if the instance changes again 2ms later, because the whole thing will be rebound then anyway.

PR #102, opened 2018-08-26
WIP feat(JIT): template compiler

bigopon:
@EisenbergEffect @fkleuver 

There needs to be more work for template controller attributes but I think the basic shape of instruction building is in place.

TODO:

- [ ] template controller attribute compilation
- [ ] test

### DIFF
--- a/packages/jit/src/templating/template-compiler.ts
+++ b/packages/jit/src/templating/template-compiler.ts
@@ -40,15 +40,15 @@ export class TemplateCompiler implements ITemplateCompiler {
       definition.templateOrNode = node;
     }
     node = <Element>(node.nodeName === 'TEMPLATE' ? (<HTMLTemplateElement>node).content : node);
-    while (node = this.compileNode(<Node>node, definition.instructions, resources)) { /* Do nothing */ }
+    while (node = this.compileNode(<Node>node, definition, definition.instructions, resources)) { /* Do nothing */ }
     return definition;
   }
 
   /*@internal*/
-  public compileNode(node: Node, instructions: TargetedInstruction[][], resources: IResourceDescriptions): Node {
+  public compileNode(node: Node, definition: Required<ITemplateSource>, instructions: TargetedInstruction[][], resources: IResourceDescriptions): Node {
     switch (node.nodeType) {
       case NodeType.Element:
-        this.compileElementNode(<Element>node, instructions, resources);
+        this.compileElementNode(<Element>node, definition, instructions, resources);
         return node.nextSibling;
       case NodeType.Text:
         if (!this.compileTextNode(<Text>node, instructions)) {
@@ -68,7 +68,7 @@ export class TemplateCompiler implements ITemplateCompiler {
   }
 
   /*@internal*/
-  public compileElementNode(node: Element, instructions: TargetedInstruction[][], resources: IResourceDescriptions): void {
+  public compileElementNode(node: Element, definition: Required<ITemplateSource>, instructions: TargetedInstruction[][], resources: IResourceDescriptions): void {
     let elementDefinition: Immutable<Required<ITemplateSource>>;
     let elementInstruction: HydrateElementInstruction;
     const tagName = node.tagName;
@@ -89,13 +89,21 @@ export class TemplateCompiler implements ITemplateCompiler {
     const attributeInstructions: TargetedInstruction[] = [];
     let liftInstruction: HydrateTemplateController;
     for (let i = 0, ii = attributes.length; i < ii; ++i) {
-      const instruction = this.compileAttribute(attributes.item(i), node, resources, elementInstruction, false);
+      const instruction = this.compileAttribute(
+        attributes.item(i),
+        node,
+        resources,
+        elementInstruction,
+        definition,
+        false
+      );
       if (instruction !== null) {
         // if it's a template controller instruction
         // it could be nth template controller instruction: <div if.bind="" repeat.for="" with.bind=""></div>
         // just keep refreshing the reference to the last template controller instruction
         if (instruction.type === TargetedInstructionType.hydrateTemplateController) {
-          i--; --ii;
+          --i;
+          --ii;
           if (liftInstruction) {
             liftInstruction.instructions.push(instruction);
           }
@@ -117,18 +125,23 @@ export class TemplateCompiler implements ITemplateCompiler {
       }
     }
     if (elementInstruction || liftInstruction || attributeInstructions.length) {
-      node.classList.add('au');
-      instructions.push(liftInstruction
-        ? [liftInstruction]
-        : elementInstruction
+      if (liftInstruction) {
+        const marker = createMarker();
+        node.parentNode.replaceChild(marker, node);
+        node = marker;
+        instructions.push([liftInstruction]);
+      } else {
+        node.classList.add('au');
+        instructions.push(elementInstruction
           ? [elementInstruction]
           : attributeInstructions
-      );
+        );
+      }
     }
     node = <Element>(node.nodeName === 'TEMPLATE' ? (<HTMLTemplateElement>node).content : node);
     let currentChild = node.firstChild;
     while (currentChild) {
-      currentChild = this.compileNode(currentChild, instructions, resources);
+      currentChild = this.compileNode(currentChild, definition, instructions, resources);
     }
   }
 
@@ -153,6 +166,7 @@ export class TemplateCompiler implements ITemplateCompiler {
     node: Element,
     resources: IResourceDescriptions,
     elementInstruction: HydrateElementInstruction,
+    definition: Required<ITemplateSource>,
     isForSurrogateElement: boolean
   ): TargetedInstruction {
     const { name, value } = attr;
@@ -231,7 +245,14 @@ export class TemplateCompiler implements ITemplateCompiler {
           const parts = binding.split(':');
           const attr = document.createAttribute(parts[0]);
           attr.value = parts[1];
-          instructions.push(this.compileAttribute(attr, node, resources, elementInstruction, false));
+          instructions.push(this.compileAttribute(
+            attr,
+            node,
+            resources,
+            elementInstruction,
+            definition,
+            false
+          ));
         }
         return new HydrateAttributeInstruction(targetName, instructions);
       }
@@ -259,6 +280,7 @@ export class TemplateCompiler implements ITemplateCompiler {
         src = <ITemplateSource>this.compile(<Required<ITemplateSource>>src, resources);
       }
       const expression = this.expressionParser.parse(value, bindingCommand);
+      (definition.dependencies || (definition.dependencies = [])).push(CustomAttributeResource)
       return new HydrateTemplateController(src, targetName, [
         new ToViewBindingInstruction(expression, 'items'),
         new SetPropertyInstruction('item', 'local')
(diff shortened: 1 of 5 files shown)


packages/jit/src/templating/template-compiler.ts
```diff
@@ -68,17 +69,60 @@ export class TemplateCompiler implements ITemplateCompiler {
...
       if (instruction !== null) {
-        attributeInstructions.push(instruction);
+        // if it's a template controller instruction
+        // it could be nth template controller instruction: <div if.bind="" repeat.for="" with.bind=""></div>
+        // just keep refreshing the reference to the last template controller instruction
+        if (instruction.type === TargetedInstructionType.hydrateTemplateController) {
+          if (liftInstruction) {
+            liftInstruction.instructions.push(instruction);
+          }
+          liftInstruction = instruction;
```
fkleuver: Does this mean that multiple template controllers on the same element turn into a tree structure of instructions?

fkleuver:
Glad you're helping out with this. So excited!! :)

EisenbergEffect:
@bigopon Is this ready for merge?

bigopon:
@EisenbergEffect It's not ready yet, but soon, ... hopefully 😄

packages/jit/src/templating/template-compiler.ts
bigopon: Yes, if @EisenbergEffect can confirm the following is right:

```html
<div
  if.bind=""
  repeat.for=""
  with.bind=""></div>
```
equals:
```js
[
  {
    type: 'template controller',
    src: ...,
    res: 'if',
    instructions: [
      {
        type: 'template controller',
        src: ....
        res: 'repeat',
        instructions: [
          {
            type: 'template controller',
            src: ...,
            res: 'with',
            instructions: [... the rest of the attributes]
          }
        ]
      }
    ]
  }
]
```

EisenbergEffect: I'm not sure that we should allow more than one template controller per element. If I recall, the reason we didn't support this in vCurrent is because browsers weren't consistent in the order that they provided the attribute list.

bigopon: that's new to me, I've always thought the opposite. I'm not doing it but I can imagine someone would, as it looks valid to me. Nice to have the confirmation though.

StrahilKazlachev: @EisenbergEffect it's concerning to have different behaviour between AOT and JIT.

fkleuver: @StrahilKazlachev @EisenbergEffect We can easily alleviate the problem of attribute ordering by reading them all out before processing them, sorting them (maybe having our own internal ordering based on template controller priority?) and then process them in our own deterministic order..

fkleuver: I'd say let's go for multiple template controllers per element. People will want it anyway. And I'm looking forward to unit testing that :P

fkleuver:
@bigopon See if that test runs? https://github.com/aurelia/aurelia/blob/master/packages/jit/test/unit/templating/template-compiler.integration.spec.ts#L478
If it does (e.g. template controllers other than the repeater work) I'd say we can merge this in and look at template-controller-on-custom-elements in a next iteration.

packages/jit/src/templating/template-compiler.ts
StrahilKazlachev: @fkleuver if I remember correctly we couldn't get past the ordering issue because we first created a `HTMLTemplateElement` before using it. At this point it was already messed up.
So are you suggesting we read it before creating the `HTMLTemplateElement`?

fkleuver: I'm not sure I understand. How can we create a `HTMLTemplateElement` before reading any attributes on an element?

bigopon:
@fkleuver thanks for the pointer. I'm handing some more details of template controller, will update soon

packages/jit/src/templating/template-compiler.ts
StrahilKazlachev: I may be missing some change in vNext, but currently the parsing of the raw view template strings is *not* done by us, but by creating a `HTMLTemplateElement`. If for vNext this has changed then my concern is invalid.

fkleuver: Well we're doing it with `div.innerHTML` in the JIT, but still, why does it matter? We can reorder and process the attributes however we want, regardless of the order in which they appear on the element.
This did get me to realize we probably need a concept of TemplateController `Precedence`. E.g. repeat always comes first, then `if`, then other stuff. Something like that. Having standard rules for that would make everything consistent and predictable regardless of browser. I would certainly not expect the order of attributes to matter.

StrahilKazlachev: I don't agree this is doable:
- `if.bind="showItems" repeat.for="item of items"` - now if you write it in this order you probably have the mind set that if the `if` condition is *truthy* it will repeat over the items.
- what about custom `TemplateController`s, how will they be able to compete with the *out-of-the-box* ones?

If we are targeting some kind of order dependent behaviour, anything different from what the author has written, for me is is wrong.

bigopon:
@EisenbergEffect @fkleuver 

I'm a bit confused by this block https://github.com/aurelia/aurelia/blob/9f633c3657f90494850e253f7b841b22a7508228/packages/runtime/src/templating/renderer.ts#L51-L59

It seems like every target should have a corresponding instruction, but for the test template 

```html
<template><div if.bind="foo"></div></template>
```

The `src` of template controller `if` will have only 1 target and no instruction, thus should be an error.

packages/jit/src/templating/template-compiler.ts
fkleuver: We could make it possible to specify the precedence on the templateController decorator itself. The order in which attributes are declared doesn't matter anywhere in the HTML spec, so if it matters in Aurelia then (imo) we're being inconsistent.
You could still work around the precedence order by using nested template tags, much like you would use parentheses in JavaScript to override the natural precedence of operators.

StrahilKazlachev: I would really like *not* to go in a scenario where implementations specify their precedence. I'm writing html, I already have to know which attr is a `TemplateController` *and* now to have take into consideration their precedence/ranking? Please not.

A `TemplateController`s looks just as any other attr, but they do something *quite* different from the spec defined ones, at least I can't think of one that acts remotely similar to a `TemplateController`, so order being irrelevant by spec isn't much of a surprise. I suppose this is why a lot of people put more than 1 `TemplateController` on the same element and expect the ordering to give them the desired outcome, instead of nesting elements as @fkleuver said.

EisenbergEffect:
@bigopon Every target has an array of instructions that should be applied to that target. There should be at least one instruction per target.

packages/jit/test/unit/templating/template-compiler.spec.ts
```diff
@@ -64,19 +64,19 @@ const attrNameArr = [
...
-  { $type: BindingType.CallCommand,     attrName: 'foo.call' },
-  { $type: BindingType.Interpolation,   attrName: 'foo.options' },
-  { $type: BindingType.ForCommand,      attrName: 'foo.for' },
-  { $type: BindingType.Interpolation,   attrName: 'foo.foo' },
-  { $type: BindingType.BindCommand,     attrName: 'bind.bind' },
+  { $type: BindingType.CallCommand, attrName: 'foo.call' },
+  { $type: BindingType.Interpolation, attrName: 'foo.options' },
+  { $type: BindingType.ForCommand, attrName: 'foo.for' },
+  { $type: BindingType.Interpolation, attrName: 'foo.foo' },
+  { $type: BindingType.BindCommand, attrName: 'bind.bind' },
```
fkleuver: :(

bigopon: I know, will turn off the auto format :(

packages/jit/src/templating/template-compiler.ts
```diff
@@ -117,18 +125,23 @@ export class TemplateCompiler implements ITemplateCompiler {
       }
     }
     if (elementInstruction || liftInstruction || attributeInstructions.length) {
-      node.classList.add('au');
-      instructions.push(liftInstruction
-        ? [liftInstruction]
-        : elementInstruction
+      if (liftInstruction) {
+        const marker = createMarker();
```
bigopon: I think because of this, a bunch of unit tests in `TemplateCompiler` starts to fail, I need to investigate further. If you had any pointer, it would be great.

packages/jit/src/templating/template-compiler.ts
```diff
@@ -40,15 +40,15 @@ export class TemplateCompiler implements ITemplateCompiler {
...
     }
     node = <Element>(node.nodeName === 'TEMPLATE' ? (<HTMLTemplateElement>node).content : node);
-    while (node = this.compileNode(<Node>node, definition.instructions, resources)) { /* Do nothing */ }
+    while (node = this.compileNode(<Node>node, definition, definition.instructions, resources)) { /* Do nothing */ }
     return definition;
   }
 
   /*@internal*/
-  public compileNode(node: Node, instructions: TargetedInstruction[][], resources: IResourceDescriptions): Node {
+  public compileNode(node: Node, definition: Required<ITemplateSource>, instructions: TargetedInstruction[][], resources: IResourceDescriptions): Node {
```
bigopon: I thought `dependencies` list of `definition` was supposed to be updated whenever we hit a custom behavior, but it doesn't seem like it. Probably I'll remove it. @fkleuver @EisenbergEffect

packages/jit/test/unit/templating/template-compiler.integration.spec.ts
```diff
@@ -476,13 +476,13 @@ describe('TemplateCompiler (integration)', () => {
   });
 
   // TODO: implement this in template compiler
-  it.only(`if - shows and hides`, () => {
+  it(`if - shows and hides`, () => {
     component = createCustomElement(`<template><div if.bind="foo">bar</div></template>`);
     component.foo = true;
     au.app({ host, component: component }).start();
+    cs.flushChanges();
```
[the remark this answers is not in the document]
fkleuver: Only the target accessors are async, so updates going towards the dom still need to be flushed. You don't need the changeSet per se, you can also just make the method async and do `await Promise.resolve()`, to await the tick.

packages/jit/src/templating/template-compiler.ts
```diff
@@ -259,6 +280,7 @@ export class TemplateCompiler implements ITemplateCompiler {
         src = <ITemplateSource>this.compile(<Required<ITemplateSource>>src, resources);
       }
       const expression = this.expressionParser.parse(value, bindingCommand);
+      (definition.dependencies || (definition.dependencies = [])).push(CustomAttributeResource)
```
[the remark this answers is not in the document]
fkleuver: Yeah I don't think that actually does anything..

packages/jit/src/templating/template-compiler.ts
fkleuver: Those are just DI registrations scoped to that particular resource. I think we (can) use it to register an `Element` instance so the user could inject that in their component, among other things. Might also be used for AOT

packages/jit/src/templating/template-compiler.ts
fkleuver: Most of these are just fixed tests that I put in place for finding edge cases. I sort of hacked them into a fixed state where they passed specifically for the code I wrote. Make it refactor proof etc. Now you are changing (adding) behavior, so you will break a lot of these unit tests. You can safely ignore most of them (just change `it` to `xit` or something).

The tests that do need to work are the integration tests and those under `describe('compile()')` and of course the e2e tests. Maybe @EisenbergEffect can merge your PR into a separate branch so you can work off of this repo and let the CI server verify your commits as well.

bigopon:
@EisenbergEffect @fkleuver 

The following block is how I guessed out of our AOT, please have a look

https://github.com/aurelia/aurelia/blob/82500d37995c4781aa4809afb3c16e2250c87d8a/packages/jit/test/unit/templating/template-compiler.spec.ts#L395-L430

EisenbergEffect:
@bigopon Looks good to me :)

packages/jit/test/unit/templating/template-compiler.spec.ts
```diff
@@ -272,102 +284,149 @@ describe('TemplateCompiler', () => {
...
 
+  // });
+
+  describe.only('compileElement()', () => {
+
+    it('throws on <let/> + <slot/>', () => {
+      const markup = '<template><slot></slot></template>';
+      expect(() => {
+        sut.compile(<any>{ templateOrNode: markup, instructions: [] }, resources);
+      }).to.throw();
```
fkleuver: Doesn't this need to be `.to.throw`? Not sure if this works

fkleuver:
Yep, looks like the right thing :)

bigopon:
Closed in favor of #121

PR #106, opened 2018-08-27
fix(di): convert invokers to an array

EisenbergEffect:
This PR converters the invoker lookup to an array, ordered by dependency count. Previously, it was an object with the dependency count as key and the invoker as value.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -25,6 +25,7 @@
   },
   "cSpell.words": [
     "Constructable",
+    "Injectable",
     "Renderable",
     "attachables",
     "bindables"
--- a/packages/kernel/src/di.ts
+++ b/packages/kernel/src/di.ts
@@ -126,19 +126,19 @@ export const DI = {
 
       Key.register = function(container: IContainer, key?: any) {
         return configure({
-          instance(value: any) {
+          instance(value: any): IResolver {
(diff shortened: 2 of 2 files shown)


EisenbergEffect:
@fkleuver I *think* you had indicated that using integers index into an object in JavaScript was not as efficient as just using an array. I switched the invoker lookup over to use an array in response to that. Can you confirm? If not, I can revert that part and we can keep the rest of this PR, as it includes a number of lint fixups.

fkleuver:
Numeric indexers are faster on arrays then on objects yes. String properties on objects are always fastest however.
I just double checked, turns out the *huge* difference actually only applies to classes: http://jsben.ch/FCU4s

Drilling down on that, you can see that arrays are still a little faster than objects for numeric properties but not a lot (remember they're both about 100x faster than numeric properties on classes):
http://jsben.ch/7Fd92

I'm guessing this is because chrome internally uses Array.prototype methods on normal objects to handle numeric properties, so there is only the overhead of explicitly adding/removing properties as dictated by the spec.

So your change is a minor perf improvement at least. Let it boil down to a preference :)

EisenbergEffect:
The `resolve` method should already return a typed result if you use it wither with one of the DI.createInterface return types or with an actual constructor function. That's handled through the `IServiceProvider` interface.

PR #107, opened 2018-08-28
refactor(runtime): lift resource methods

EisenbergEffect:
## Description

This PR refactors the internals of the `define` methods of each of the resources types so that they use shared functions rather than creating new function instances on every method invocation.

While working on the `CustomElementResource` a helper factory method for basic scopes was also factored out and placed on the `BindingContext`.


bigopon:
Nice. There are few other places in, such as the renderer context, that I think can benefit from the same refactoring

fkleuver:
LGTM :)

PR #108, opened 2018-08-28
chore(examples): add e2e-benchmark

fkleuver:
First step towards having some automated performance tests to detect regressions etc.
Next step is to add a similar benchmark but with vCurrent, so we have a baseline to compare against.
Also need to write out the results (to json?) and post them back to github


fkleuver:
I'm trying to get a vCurrent one in as well. I don't expect that to be much work. Should be in a few minutes

PR #109, opened 2018-08-28
feat(runtime): let binding

bigopon:
@EisenbergEffect @fkleuver 

I original copy-pasted the whole `Binding` class to create `LetBinding` class, but then changed `private` fields of `Binding` to `protected` and made `LetBinding` extends it.

### DIFF
--- a/packages/jit/src/templating/template-compiler.ts
+++ b/packages/jit/src/templating/template-compiler.ts
@@ -2,6 +2,7 @@ import { inject } from '@aurelia/kernel';
 import { BindingType, CustomAttributeResource, DOM, IExpression, IExpressionParser, Interpolation, IResourceDescriptions, ITargetedInstruction, ITemplateCompiler, TemplateDefinition, TargetedInstructionType, TargetedInstruction, DelegationStrategy, ITextBindingInstruction, IPropertyBindingInstruction, IListenerBindingInstruction, ICallBindingInstruction, IRefBindingInstruction, IStylePropertyBindingInstruction, ISetPropertyInstruction, ISetAttributeInstruction, IHydrateElementInstruction, IHydrateAttributeInstruction, IHydrateTemplateController, BindingMode, ITemplateSource, INode } from '@aurelia/runtime';
 import { Char } from '../binding/expression-parser';
 import { BindingCommandResource, IBindingCommandSource } from './binding-command';
+import { ILetBindingInstruction } from '@aurelia/runtime/src';
 
 const domParser = <HTMLDivElement>DOM.createElement('div');
 const marker = document.createElement('au-marker');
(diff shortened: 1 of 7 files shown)


EisenbergEffect:
Since most of this is in the binding space, let's have @fkleuver do the primary review and then after he's ok'd it, I'll do a second quick pass. Very excited about this!

fkleuver:
@bigopon Hm it won't build on CI or locally.. did you rebase the latest master?

bigopon:
I did before I started it, must have been some more commits. Let me do it

fkleuver:
Does it build locally for you? Still fails here :(

packages/jit/src/templating/template-compiler.ts
```diff
@@ -2,6 +2,7 @@ import { inject } from '@aurelia/kernel';
 import { BindingType, CustomAttributeResource, DOM, IExpression, IExpressionParser, Interpolation, IResourceDescriptions, ITargetedInstruction, ITemplateCompiler, TemplateDefinition, TargetedInstructionType, TargetedInstruction, DelegationStrategy, ITextBindingInstruction, IPropertyBindingInstruction, IListenerBindingInstruction, ICallBindingInstruction, IRefBindingInstruction, IStylePropertyBindingInstruction, ISetPropertyInstruction, ISetAttributeInstruction, IHydrateElementInstruction, IHydrateAttributeInstruction, IHydrateTemplateController, BindingMode, ITemplateSource, INode } from '@aurelia/runtime';
 import { Char } from '../binding/expression-parser';
 import { BindingCommandResource, IBindingCommandSource } from './binding-command';
+import { ILetBindingInstruction } from '@aurelia/runtime/src';
```
bigopon: @fkleuver This was the cause

bigopon:
@fkleuver Tests run fine locally, is there anything particular I should be aware of ?

bigopon:
Ignore my previous comment, somehow CircleCI showed tests were failing 😕

bigopon:
https://github.com/aurelia/aurelia/blob/83298fcd3c5ba1a290c215de18f11442a7807b7a/packages/runtime/src/binding/binding.ts#L92

I think there's something wrong with this `context` ? It seems to be out of nowhere. Intellisense suggests it's from Chai, but I guess it's not correct ?

fkleuver:
Lel. That's supposed to be source / target context. Must have been some screw-up in copy pasting and then intellisense letting it through because it's a valid global requirejs/mocha variable. Changing that to `flags` (and then modifying the reporter to correctly interpret and report on those flags) should do. If it causes anything to fail, just pass in the flags.

bigopon:
closed in favor of #132

PR #110, opened 2018-08-28
e2e benchmark fixes + perf improvements

fkleuver:
fix(e2e-benchmark): simplify markers and use setTimeout to include render time
perf(binding): remove unnecessary work / cleanup
refactor(binding): convert subscriber-collection to decorator for sync and async

I'm a little hesitant on whether I prefer the approach I took on property observer vs the approach on collection observers.
- On the property observers there is a partial interface implementations with as few properties and methods declared as possible. This gives a much cleaner view of the implementation, but lacks a degree of transparency perhaps. It also means observers need to be cast explicitly.
- On the collection observers the full interfaces are implemented. This means everything that decorators would set on the prototype is explicitly declared. I think it's ugly and messy, but it may also be more practical

Thoughts? @bigopon  @EisenbergEffect

### DIFF
--- a/packages/examples/e2e-benchmark/vcurrent/app.ts
+++ b/packages/examples/e2e-benchmark/vcurrent/app.ts
@@ -52,13 +52,17 @@ export class App {
       }
       this.todos.push(...newTodos);
     }
-    this.tq.queueTask(instrumenter);
+    setTimeout(() => {
+      instrumenter.markEnd();
+    });
   }
 
   public clearTodos(): void {
     instrumenter.markActionStart(`app-clearTodos-${this.todos.length}`, true);
     this.todos = [];
-    this.tq.queueTask(instrumenter);
+    setTimeout(() => {
+      instrumenter.markEnd();
+    });
   }
 
   public toggleTodos(): void {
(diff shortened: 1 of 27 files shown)


EisenbergEffect:
@fkleuver Check out what I've done for some of the template controllers like `If`. I'm using TypeScript's interface merging. So, I have the `templateController` decorator, which is going to add a bunch of properties and methods at runtime. But I also have the an interface with the same name as the class that inherits from an interface that declares all the properties, for compile-time.

```
export interface If extends ICustomAttribute {} // adds the props at compile time
@templateController('if') // adds the props at runtime
export class If { }
```

You could do the same thing with the observers. Note that I hope this problem will go away once TS implements the new decorators spec and allows decorators to return intersection types that affect the `this` of the decorated class. Probably a ways out, but I think the interface merging works well enough as a work around for now.

fkleuver:
@EisenbergEffect Awesome tip, didn't know it worked like that! Much cleaner now I think :)

Ready to merge as far as I'm concerned

EisenbergEffect:
Agreed. Looks great!

PR #111, opened 2018-08-29
refactor(view): increase cohesion of view code

EisenbergEffect:
## Description

This PR does two things:

1. Moves the View and ViewFactory code into the `view.ts` file with the related interfaces. There was a reason to have it where it was, once upon a time, I think. That reason has gone and it's better to keep things together.
2. Reworks View and ViewFactory so that a custom class doesn't need to be created for every View that a factory creates. This should be much more efficient.

## Next Steps

After the refactoring to remove the custom class generation for the primary scenarios, the component view scenario is left in an odd state. I intend to fix this up, but since it's strongly coupled to the `compose` element scenarios, I chose not to do it as part of this PR. I'd like to fix this up at the same time as the `compose` element fixups.

### DIFF
--- a/packages/runtime/src/templating/rendering-engine.ts
+++ b/packages/runtime/src/templating/rendering-engine.ts
@@ -1,26 +1,21 @@
-import { all, Constructable, DI, IContainer, Immutable, inject, PLATFORM, Reporter } from '@aurelia/kernel';
-import { IScope } from '../binding/binding-context';
-import { BindingFlags } from '../binding/binding-flags';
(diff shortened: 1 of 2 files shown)


packages/runtime/src/templating/rendering-engine.ts
```diff
@@ -127,11 +122,11 @@ export class RenderingEngine implements IRenderingEngine {
       public component: ICustomElement;
 
       constructor() {
-        super(null, animator);
+        super(null, null, animator);
         this.$context = context;
       }
 
-      public render(): INodeSequence {
+      public createNodes(): INodeSequence {
```
fkleuver: nice :)

PR #113, opened 2018-08-30
build outputs / package.json fields

fkleuver:
chore(all): add tsconfig and npm scripts for other build outputs
chore(all): add additional package.json fields

In preparation for publishing

@EisenbergEffect


fkleuver:
re-created PR under different branch name to bust the circleci cache

PR #115, opened 2018-08-30
Prepare publish

fkleuver:
chore(all): add tsconfig and npm scripts for other build outputs
chore(all): add additional package.json fields

In preparation for publishing

@EisenbergEffect 

(re-created PR under different branch name to bust the circleci cache)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1466,12 +1466,6 @@
         }
       }
     },
-    "alphanum-sort": {
-      "version": "1.0.2",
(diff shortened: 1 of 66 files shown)


fkleuver:
@EisenbergEffect All's good from a technical point of view. Do you agree with the package.json changes and structure for tsconfig? I filed a feature request related to this https://github.com/Microsoft/TypeScript/issues/26763 but no idea if/when that would be picked up. If it does get implemented, we can get rid of the configs again

PR #121, opened 2018-09-01
Template compiler

bigopon:
@EisenbergEffect @fkleuver 

The template compiler now understands custom elements and template controllers, but I'm unable to create an integration test for custom element. Any pointers would be nice.

https://github.com/aurelia/aurelia/blob/8ce5f90b51a650c34e8a48dcefb795f252ac1f7e/packages/jit/test/unit/templating/template-compiler.integration.spec.ts#L505-L524

@fkleuver I've commented `parseAttribute()` tests. I think I will need to rework them a bit as now compiling attribute has been changed a bit.


EisenbergEffect:
@bigopon Do you need advice on setting up the test or with the implementation of the compiler itself?

packages/jit/src/templating/template-compiler.ts
```diff
@@ -154,60 +338,115 @@ export class TemplateCompiler implements ITemplateCompiler {
...
+        instruction
+      ]);
+    }
+  }
 
-    // TODO: other template controllers / binding commands
-
-    return null;
+  private swapWithMarker(node: Element, parentNode: Element): Element {
+    const marker = createMarker();
```
fkleuver: Small nit: this could be a loose function instead of a private method

packages/jit/src/templating/template-compiler.ts
```diff
@@ -127,22 +237,96 @@ export class TemplateCompiler implements ITemplateCompiler {
...
+        if (elementProperty) {
+          // IPropertyBindingInstruction = with binding command / interpolation
+          // ISetPropertyInstruction = plain html attribute
+          let attrInstruction: IPropertyBindingInstruction | ISetPropertyInstruction;
+          if (bindingCommand) {
+            // When it's `.bind`, needs to select the right instruction based on
+            // default binding mode of the element property
+            if (bindingCommand & BindingType.BindCommand) {
+              switch (elementProperty.mode) {
+                case BindingMode.fromView:
```
fkleuver: `new BindingInstruction[bindingCommand & BindingType.Command]` :P

bigopon:
@EisenbergEffect I think I'm having issue with dependencies not picked up, as in this block https://github.com/aurelia/aurelia/blob/8ce5f90b51a650c34e8a48dcefb795f252ac1f7e/packages/jit/test/unit/templating/template-compiler.integration.spec.ts#L63-L72

Is that the correct way to declare dependencies for a custom element ? I'm having empty instructions list inside `app` custom element description

packages/jit/src/templating/template-compiler.ts
```diff
@@ -38,52 +67,126 @@ export class TemplateCompiler implements ITemplateCompiler {
...
+  ): void {
+    let elementDefinition: Immutable<Required<ITemplateSource>>;
+    let elementInstruction: HydrateElementInstruction;
+    const tagName = node.tagName;
+    if (tagName === 'SLOT' || tagName === 'LET') {
+      throw new Error('<slot/> or <let/> not implemented');
+    }
+    elementDefinition = resources.find(
+      CustomElementResource,
+      hyphenate(node.tagName.toLowerCase())
```
EisenbergEffect: I understand toLowerCase here but I'm not sure I understand why you need to hyphenate the tagName.

bigopon: Yes, it should be gone

packages/jit/src/templating/template-compiler.ts
```diff
@@ -38,52 +67,126 @@ export class TemplateCompiler implements ITemplateCompiler {
...
+    );
+    if (elementDefinition) {
+      elementInstruction = new HydrateElementInstruction(
+        resources,
+        [],
```
fkleuver: `res` needs to be the registered name of the resources (in the case of your failing integration test "name-tag"). Then it works :)

packages/jit/test/unit/templating/template-compiler.integration.spec.ts
```diff
@@ -382,120 +383,141 @@ describe('TemplateCompiler (integration)', () => {
...
+  //   expect(host.innerText).to.equal('baz');
   //   component.foo = false;
+  //   cs.flushChanges();
   //   expect(host.innerText).to.equal('bar');
   //   cs.flushChanges();
   //   expect(host.innerText).to.equal('baz');
   // });
 
+  it(`custom elements`, () => {
+    @customElement({
```
fkleuver: This needs to be:

```
@customElement({
  name: 'name-tag',
  templateOrNode: '<template>asdf</template>',
  build: { required: true, compiler: 'default' },
  dependencies: [],
  instructions: [],
  surrogates: []
})
```

packages/jit/src/templating/template-compiler.ts
bigopon: Wow, doh.!

packages/jit/test/unit/templating/template-compiler.integration.spec.ts
fkleuver: Actually just put this whole thing at the top of the file:

```
@customElement({
  name: 'name-tag',
  templateOrNode: '<template>asdf</template>',
  build: { required: true, compiler: 'default' },
  dependencies: [],
  instructions: [],
  surrogates: []
})
export class NameTag {
  @bindable()
  name: string;
}
```

And add it to globalresources:

```
const globalResources: any[] = [
  If,
  Else,
  Repeat,
  SortValueConverter,
  JsonValueConverter,
  OneTimeBindingBehavior,
  ToViewBindingBehavior,
  FromViewBindingBehavior,
  TwoWayBindingBehavior,
  DebounceBindingBehavior,
  ThrottleBindingBehavior,
  NameTag
];
```

And the test to this:

```
  it.only(`custom elements`, () => {
    component = createCustomElement(`<template><name-tag name="bigopon"></name-tag></template>`);
    au.app({ host, component: component })
    au.start();
    cs.flushChanges();
    expect(host.innerText).to.equal('bigopon');
  });
```

That will give a bindable error, but if you remove the bindable and instead put plain text in the custom element it will work. Not sure yet what that error is about

packages/jit/test/unit/templating/template-compiler.integration.spec.ts
```diff
@@ -382,120 +383,141 @@ describe('TemplateCompiler (integration)', () => {
...
+      templateOrNode: '<template>${name}</template>'
+    })
+    class NameTag {
+
+      @bindable()
+      name: string;
+    }
+    component = createCustomElement(
+      `<template><name-tag name="bigopon"></name-tag></template>`,
+      NameTag
```
fkleuver: Maybe you just need to change this to `Registration.transient('name-tag', NameTag)`. The dependency would get autoregistered by the function instead of the custom element name otherwise

bigopon:
@fkleuver @EisenbergEffect With the help from Fred, it's now working fine. There are the following issues:

- [ ] local dependencies not working
- [ ] link-able template controllers not linked
- [ ] process content / process attribute
- [ ] surrogate instructions

I think it's safe to merge in and then we can continue the above work later.

fkleuver:
@EisenbergEffect let's get this merged in, I want to get my hands on this stuff :D

bigopon:
```diff

   private templateFromSource(context: IRenderContext, definition: TemplateDefinition): ITemplate {
     if (definition && definition.templateOrNode) {
       if (definition.build.required) {
         const compilerName = definition.build.compiler || defaultCompilerName;
         const compiler = this.compilers[compilerName];
 
         if (!compiler) {
          throw Reporter.error(20, compilerName);
        }
-       definition = compiler.compile(definition, new RuntimeCompilationResources(<ExposedContext>context));
+       definition = compiler.compile(definition, new RuntimeCompilationResources(<ExposedContext>context), ViewCompileFlags.surrogate);
      }
```

@EisenbergEffect the above changes in `rendering-engine` may not be correct as I'm not sure how the entire flow is yet.

EisenbergEffect:
@bigopon I haven't read through all the code yet, but the ViewCompileFlags are new to me. What is the significance of that in this case?

fkleuver:
Merge conflict was my bad, it's fixed up now :)

fkleuver:
@EisenbergEffect good to merge

bigopon:
@EisenbergEffect what used to be our `ViewCompileInstruction` is now `ViewCompileFlags`

@fkleuver  looks like some changes in my latest commit were lost. Can you add them again. 😁

fkleuver:
Hold on, we got a major perf regression for some reason `addTodo (x10000) took 2390ms` i'm just gonna check what that's all about

packages/jit/src/templating/template-compiler.ts
```diff
@@ -31,59 +67,179 @@ export class TemplateCompiler implements ITemplateCompiler {
...
+        }
       }
     }
-    if (attributeInstructions.length) {
-      node.classList.add('au');
-      instructions.push(attributeInstructions);
+    if (liftInstruction) {
+      node = swapWithMarker(node, parentNode as Element);
+      const template = DOM.createTemplate() as HTMLTemplateElement;
+      template.content.appendChild(liftInstruction.src.templateOrNode as Node);
```
EisenbergEffect: Is this safe? Can't `templateOrNode` be a string?

packages/jit/src/templating/template-compiler.ts
```diff
@@ -31,59 +67,179 @@ export class TemplateCompiler implements ITemplateCompiler {
...
+          attributeInstructions.push(instruction);
+        }
       }
     }
-    if (attributeInstructions.length) {
-      node.classList.add('au');
-      instructions.push(attributeInstructions);
+    if (liftInstruction) {
+      node = swapWithMarker(node, parentNode as Element);
+      const template = DOM.createTemplate() as HTMLTemplateElement;
```
EisenbergEffect: Should we improve the `createTemplate` signature so that it can take the template content as input, optionally?

packages/runtime/src/templating/view-compile-flags.ts
```diff
@@ -0,0 +1,5 @@
+export enum ViewCompileFlags {
+  none        = 0b0_001,
+  surrogate   = 0b0_010,
+  shadowDOM   = 0b0_100,
```
EisenbergEffect: I'm not sure what shadowDOM would do. For vNext, we only need to know if there are slots or not, at least for now.

fkleuver:
Sorry, go ahead and merge. I'll check the perf later. Might be a circleci hiccup as well

packages/jit/src/templating/template-compiler.ts
bigopon: The template of this instruction is from currently processed DOM tree, so it won't be a string

packages/jit/src/templating/template-compiler.ts
bigopon: @EisenbergEffect Could be nice improvement, it can also take care of adopting node / etc, less code everywhere else.

PR #123, opened 2018-09-01
refactor(all): remove viewslot

fkleuver:
It's a little brash perhaps, but considering only the (currently unused/untested) `compose` was referencing it (and only a small portion of the API), I think it's ripe for removal. At least we shouldn't need it to be in the component context for every element and attribute..

Note: this will of course break `compose` (if that was working)


EisenbergEffect:
I'm fine with this. I was about to remove the compose dependency in my next PR anyway. So, this should be fine.

PR #124, opened 2018-09-01
Dynamic Composition - Phase 1

EisenbergEffect:
## Description

This PR adds a `createElement` function with the same signature that compilers generate JSX for. It is used for the following things:

1. Enabling custom elements to support a `render` method that returns JSX for the element's template.
2. Enabling developers to use a low-level API for creating element templates, view factories and views, for extremely dynamic template generation without the need for the view compiler.
3. Providing the low-level API used by the `compose` element in order to support declarative dynamic composition.

The old infrastructure code for the previous `compose` implementation has been removed and some core APIs have been fixed up to better represent parameter purposes. There is now a new implementation of the `compose` element, which uses the `createElement` element API internally when composing elements. `compose` is now able to compose `IView` or anything that can produce an `IView`.

### Subject to Change

Currently, the property of the `compose` element that designates what should be composed is called `subject`. We may wish to change this property name in the future. Currently, it's the best I could think of, since it describes the subject of the composition, which could be any of the following: `IView`, `IViewFactory`, `TemplateDescription`, `PotentialRenderable`, or a custom element constructor.

## Tests

A basic set of unit tests for the `compose` element have been provided. The tests exercise both raw and Promised values for `IView` and all possible `IView` producers.

## Next Steps

This PR includes the `createElement` implementation, but no direct tests for that. A follow-up PR will provide a set of tests specifically for `createElement` and `PotentialRenderable`. Once that is in place, additional tests of the `compose` element can be added to exercise specifically how the `compose` element interacts with `createElement`. e.g. content, pass-through bindings


packages/runtime/src/templating/resources/compose.ts
```diff
@@ -14,14 +13,14 @@ const composeSource: ITemplateSource = {
...
-const composeProps = ['component', 'composing'];
+const composeProps = ['subject', 'composing'];
 
 export interface Compose extends ICustomElement {}
 @customElement(composeSource)
 @inject(IRenderable, ITargetedInstruction, IRenderingEngine)
 export class Compose {
-  @bindable public component: any;
-  @bindable public composing: boolean;
+  @bindable public subject: any = null;
```
fkleuver: Subject? :P is this the viewmodel?

EisenbergEffect: I originally called this `component` but that turned out to be inaccurate since it could be a component or a view. I'm also going to be able to open this up to view factories, what I'm calling "potential renderables" and probably raw template strings. So, not sure what the best name for this is just yet. `subject` as the first thing that came to mind. "It's the subject of the composition." Definitely open to ideas...

fkleuver: Why not make it explicit and have a `component` as well as a `view` property, and document the invariant that only either of them can be defined at a time? Makes things clearer in terms of typing as well

AureliaEffect:
This pull request **introduces 3 alerts** and **fixes 1** when merging 7eba1824149c00e0fef1b190fc1c0ecc93735d77 into a2be6185b805de6e5a61bdad88250e3d41cd2ef7 - [view on LGTM.com](https://lgtm.com/projects/g/aurelia/aurelia/rev/pr-3029788623a62d83fe620df3a9b8d45b36e2fc93)

**new alerts:**

**fixed alerts:**

---

AureliaEffect:
Allure Report: https://2576-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/e2e/allure-report/index.html

AureliaEffect:
Allure Report: https://2571-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/e2e/allure-report/index.html

AureliaEffect:
This pull request **fixes 1 alert** when merging 8b8de7ea303c25d1fc844184c36ccebae8019bfe into a2be6185b805de6e5a61bdad88250e3d41cd2ef7 - [view on LGTM.com](https://lgtm.com/projects/g/aurelia/aurelia/rev/pr-5a34ebf300e84df3dd70148ae428f25e9e16fc61)

**fixed alerts:**

---

AureliaEffect:
Allure Report: https://2581-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/e2e/allure-report/index.html

packages/runtime/test/unit/templating/resources/compose.spec.ts
```diff
@@ -0,0 +1,265 @@
...
+    },
+    {
+      description: 'via a Promise',
+      create(subject) { return Promise.resolve(subject); }
+    }
+  ];
+
+  for (let subjectPossibility of subjectPossibilities) {
+    for(let producerPossibility of producerPossibilities) {
+      it(`can compose a ${subjectPossibility.description} ${producerPossibility.description}`, done => {
```
fkleuver: Nice loops! :)

packages/runtime/test/unit/templating/resources/compose.spec.ts
```diff
@@ -0,0 +1,265 @@
...
+          element.subject = producer.create(value);
+        });
+
+        element.subject = view1;
+      });
+    }
+  }
+
+  function waitForCompositionEnd(element: Compose, callback: () => void, done?: () => void) {
+    const endComposition = element.endComposition;
```
fkleuver: Why is this callback needed? Doesn't changeSet.flushChanges (or just making the tests async and `await Promise.resolve()`) work?

packages/runtime/src/templating/create-element.ts
```diff
@@ -0,0 +1,172 @@
...
+
+function createElementForTag(tagName: string, props?: any, children?: ArrayLike<ChildType>): PotentialRenderable {
+  const instructions: TargetedInstruction[] = [];
+  const allInstructions = [];
+  const dependencies = [];
+  const element = DOM.createElement(tagName);
+  let hasInstructions = false;
+
+  if (props) {
+    Object.keys(props)
```
fkleuver: I was going to say why not make this a `for` loop for performance, but then I benched it and it turns out this `Object.keys(props).forEach` is actually faster than storing the keys and for-looping over them. I'm baffled.. http://jsben.ch/Ln3nE
@bigopon

packages/runtime/src/templating/custom-element.ts
```diff
@@ -31,7 +32,7 @@ export type ElementDefinition = Immutable<Required<ITemplateSource>> | null;
 /*@internal*/
 export interface IInternalCustomElementImplementation extends Writable<ICustomElement> {
   $behavior: IRuntimeBehavior;
-  $child: IAttach;
+  $encapsulationSource: INode;
```
fkleuver: Nice that we got rid of `$child`. Now there's more consistency :)

packages/runtime/src/templating/custom-element.ts
```diff
@@ -127,22 +130,34 @@ function register(this: ICustomElementType, container: IContainer): void {
...
+    if (result.getElementTemplate) {
+      template = result.getElementTemplate(renderingEngine, Type);
+    } else {
+      this.$nodes = result;
+    }
+  } else {
+    template = renderingEngine.getElementTemplate(description, Type);
+  }
+
+  if (template) {
```
fkleuver: Not necessarily for this PR but I think we should try to achieve some consistency with how we check for something's existence. I like the explicit null checks we have in various places. They expose bugs - sometimes an undefined comes through where it shouldn't and stuff breaks so you learn about it.

packages/runtime/src/templating/custom-element.ts
```diff
@@ -177,11 +192,11 @@ function unbind(this: IInternalCustomElementImplementation, flags: BindingFlags)
       bindables[i].$unbind(flags | BindingFlags.fromUnbind);
     }
 
+    this.$isBound = false;
```
fkleuver: I think this makes more sense. Only when it's already bound, you call `bound`. Did this affect any existing behavior (e.g. is it a fix) or did you just change it for that reason?

packages/runtime/src/templating/custom-element.ts
```diff
@@ -203,17 +220,14 @@ function attach(this: IInternalCustomElementImplementation, encapsulationSource:
     attachables[i].$attach(encapsulationSource, lifecycle);
   }
 
-  if (this.$child !== null) {
-    this.$child.$attach(encapsulationSource, lifecycle);
-  }
-
   this.$projector.project(this.$nodes);
 
+  this.$isAttached = true;
```
fkleuver: Same questions as for the `$isBound`

packages/runtime/src/templating/custom-element.ts
```diff
@@ -257,7 +268,7 @@ export function createCustomElementDescription(templateSource: ITemplateSource,
       required: false,
       compiler: 'default'
     },
-    bindables: Object.assign({}, (Type as any).bindables, templateSource.bindables),
+    bindables: {...(Type as any).bindables, ...templateSource.bindables},
```
fkleuver: Nice :)

packages/runtime/src/templating/create-element.ts
bigopon: If we are to embrace perf in this, probably `Object.values` is best http://jsben.ch/RHlzi (on Chrome / Chromium based)
image

I remember reading v8 member saying something related to this but I don't remember any more.

<hr>

Firefox is different though 
image

<hr>

Edge
image

packages/runtime/src/templating/resources/compose.ts
```diff
@@ -1,163 +1,166 @@
...
 export interface Compose extends ICustomElement {}
 @customElement(composeSource)
-@inject(IRenderable, INode, ITargetedInstruction, IRenderingEngine)
+@inject(IRenderable, ITargetedInstruction, IRenderingEngine)
 export class Compose {
-  public component: any;
-  //public swapOrder: SwapOrder;
-  public isComposing: boolean;
+  @bindable public subject: any = null;
+  @bindable public composing: boolean = false;
```
bigopon: Why does this need to be bindable ? Isn't this an internal state of this element ? You anticipated something useful with this ?

bigopon:
@EisenbergEffect I'm very excited for this rework. It helped reshape the APIs very nicely. Though I'm not sure about interconnection composition state between compose elements, did you think of that ?

packages/runtime/src/templating/resources/compose.ts
StrahilKazlachev: @bigopon I can see it used with `.fromView` for progress. It can also allow for hierarchies roots to await their contents.

StrahilKazlachev:
@EisenbergEffect one thing missing in vCurrent composition infrastructure(`CompositionEngine`) is cancellation. More of the race conditions in `<compose>` originated from the inability to cancel ongoing composition when new one had to be made. Is there something the new APIs can do in that department?

fkleuver:
@StrahilKazlachev We can probably work that in via lifecycles somehow, but I think it's a bit too early for it now

EisenbergEffect:
@StrahilKazlachev The new `compose` element automatically cancels any ongoing composition task when a new one is made :)

packages/runtime/src/templating/resources/compose.ts
EisenbergEffect: Exactly. If there's a long-running composition due to a Promise that is making network requests, for example. This will allow the developer to bind to this property and display status.

EisenbergEffect:
@bigopon I'm not trying to tackle interconnected compose or having async compose stop composition at this point. We may be able to enable that through the Lifecycle objects that we now have, but I think we can do that at a later time, after a few more things settle down.

packages/runtime/src/templating/custom-element.ts
EisenbergEffect: VS Code did it for me :) I do like how the lint rules can sometimes trigger refactor helpers in the UI.

packages/runtime/src/templating/custom-element.ts
EisenbergEffect: And now you can dynamically add any number of children :) So, it's much more flexible as well.

packages/runtime/src/templating/custom-element.ts
EisenbergEffect: While working on the new `$addChild`/`$removeChild` methdods for renerables, I realized that it was possible and likely for something (such as compose) to try to add a renderable child *during* the `bound` callback. Since the `$addChild`/`$removeChild` methods only bind if the renderable is already bound, this would mean that adding a child during bound would never get bound because the renderable would have already processed its bindables but not yet set the flag indicating it had done so. When I was looking for a solution to that, it just seemed to make sense that these properties should be set before the past-tense callbacks were called. So, it solved the specific issue with compose and other things that might add children and I think also made it more consistent in general.

packages/runtime/src/templating/custom-element.ts
EisenbergEffect: Same as above, related to `$addChild`/`$removeChild` invocation timing, but also updated for consistency as well, as discussed above.

packages/runtime/src/templating/custom-element.ts
EisenbergEffect: Agreed!

packages/runtime/test/unit/templating/resources/compose.spec.ts
EisenbergEffect: See, you can teach an old 🐶 new tricks 🤣

packages/runtime/test/unit/templating/resources/compose.spec.ts
EisenbergEffect: The `endComposition` method always gets called at the end of a composition, regardless of whether it's sync or async, a valid value or null. So, this just a simple way to hack into that and get a guaranteed safe time to make the asserts. There isn't necessarily a promise involved.

I think a future improvement could be to add a `composition-complete` DOM event that fires. These tests could be reworked to tie into that. I don't want to add too much more to this PR since it's been hanging out a bit (mostly due to me dragging my feet).

EisenbergEffect: Actually, maybe `composition-start` and `composition-end` events could be added as an incremental improvement.

packages/runtime/src/templating/create-element.ts
```diff
@@ -0,0 +1,172 @@
...
+
+export function createElement(tagOrType: string | Constructable, props?: any, children?: ArrayLike<ChildType>): PotentialRenderable {
+  if (typeof tagOrType === 'string') {
+    return createElementForTag(tagOrType, props, children);
+  } else {
+    return createElementForType(tagOrType as ICustomElementType, props, children);
+  }
+}
+
+export class PotentialRenderable {
```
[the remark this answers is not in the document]
EisenbergEffect: I wasn't really sure what to name this class yet. I would consider this a placeholder name actually. This basically represents the return value of the `createElement` call, which wraps up all the data from that operation and then allows the end user several APIs to turn it into something usable, depending on their use case e.g. view, factory, element template, etc. It also has an internal API that's used when it's provided as a child to a `createElement` call, so it can properly build up the structure. No tests for this API in this PR. I'll send a separate PR for that. Along with that, we can rename this class. What are some other names we could use?

* CreateElementResult
* RenderableBuilder
* RenderableSpecification
* ProtoRenderable
* RenderableFactory
* ???

packages/runtime/src/templating/create-element.ts
EisenbergEffect: I bet it has something to due with the fact that they can shove more of it down into native code somehow.

packages/runtime/src/templating/create-element.ts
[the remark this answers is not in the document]
fkleuver: Isn't this just an `ElementDefinition`?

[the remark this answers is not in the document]
EisenbergEffect: It doesn't define an element, no, if that's what you mean ;) It defines a renderable, which can be used to render an element or can be used to create a view factory/view which has nothing to do with an element directly. It's something that holds all the data needed to create various types of renderables, either views for elements or raw views.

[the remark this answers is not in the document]
EisenbergEffect: I'm favoring `RenderableSpecification` as I think that's the closest to what this actually is. It holds all the data needed to create renderables of various types.

AureliaEffect:
This pull request **fixes 1 alert** when merging 9badcfb172a91daad339347411070e0c79c5b59e into ac133febcb65e939d923848fd92e2ca24622623c - [view on LGTM.com](https://lgtm.com/projects/g/aurelia/aurelia/rev/pr-1075dab013d004b2dbd7e62c9eecb3b3fea5782a)

**fixed alerts:**

---

AureliaEffect:
Allure Report: https://2642-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/e2e/allure-report/index.html

EisenbergEffect:
Merged this since there didn't seem to be any major issues. I can follow up with additional PRs. I didn't want this to get stale (had to merge master in twice already and resolve conflicts).

PR #129, opened 2018-09-02
fix(template-compiler): template controller instrs order, linking

bigopon:
Besides some tests, I tried to add if / else linking. But it seems rather difficult to get access to resource constructor during compilation, which makes it hard to determine if a template controller requires linking. Test for if / else
https://github.com/aurelia/aurelia/blob/6bfc9e1914ec4314bb5c3b5256a2034817198d32/packages/jit/test/unit/templating/template-compiler.integration.spec.ts#L512-L524

@EisenbergEffect @fkleuver

### DIFF
--- a/packages/jit/src/templating/template-compiler.ts
+++ b/packages/jit/src/templating/template-compiler.ts
@@ -33,7 +33,7 @@ import {
 import { Char } from '../binding/expression-parser';
 
 const domParser = <HTMLDivElement>DOM.createElement('div');
-const marker = document.createElement('au-marker');
+const marker = DOM.createElement('au-marker') as Element;
 marker.classList.add('au');
 const createMarker: () => HTMLElement = marker.cloneNode.bind(marker, false);
 const swapWithMarker = (node: Element, parentNode: Element) => {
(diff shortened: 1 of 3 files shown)


packages/jit/src/templating/template-compiler.ts
```diff
@@ -33,7 +33,7 @@ import {
 import { Char } from '../binding/expression-parser';
 
 const domParser = <HTMLDivElement>DOM.createElement('div');
-const marker = document.createElement('au-marker');
+const marker = DOM.createElement('au-marker') as Element;
 marker.classList.add('au');
```
fkleuver: I respectfully disagree with @EisenbergEffect on that we should use the DOM api here. I don't think we should be trying to abstract away the specificity of browser runtime in the JIT anytime soon, especially considering we (inevitably) directly talk to `document` in various other places here anyway..

fkleuver:
@bigopon I think at least for template controllers we can create a global registry of sorts where you can get the type based on the name. I'm working on something like that right now for custom binding commands. I think for the time being it's ok to hard-code the `link` thing based on the name of the controller being `else`

packages/jit/src/templating/template-compiler.ts
EisenbergEffect: I really, really do not want references to the gobal `document` object anywhere.

EisenbergEffect:
@bigopon We can remove the `link` part of the instruction. That can actually be determined fairly easily by the runtime, much like any lifecycle method. It really is a sort of lifecycle method anyway.

bigopon:
@EisenbergEffect I think in a future PR where `link` is adjust, everything can be adjusted together. At the moment it's a nice way to have things tested and write the example apps

EisenbergEffect:
@bigopon It looks like the parcel e2e tests are failing.

bigopon:
It doesn't seem like an error on our side @EisenbergEffect 
```
Error: Timeout of 10000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/home/circleci/repo/packages/examples/jit-parcel/e2e/test.spec.ts)
```

EisenbergEffect:
I reran the workflow and everything passed :)

PR #131, opened 2018-09-02
chore(all): add dev script and enable parallelism in some scripts

fkleuver:
`npm run dev` will tell tsc to build in watch mode. Running it from the top level will do so for all packages in parallel, useful for developing/changing stuff across packages for a single PR
@EisenbergEffect  @bigopon


EisenbergEffect:
Looks like the browserify e2e tests are failing. I saw that only parcel was failing in @bigopon 's PR. Are there intermittent issues with CircleCI?

EisenbergEffect:
Rerunning the workflow now...

PR #132, opened 2018-09-02
feat(Let): let binding

bigopon:
With the template compiler in, I was able to verify the implementation. It's pretty ready, except I need to handle the scenario where `<let/>` is defined not at the start of a template:

```html
<template>
  <div>${fullName}</div>
  <let full-name.bind="firstName + ` ` + lastName"></let>
</template>
```

@EisenbergEffect @fkleuver Is there already a way to add a instruction without a target ? If no, probably I'll need to splice let instructions into the start of a definition instructions, beside adding some marker based on it (yak)

Note: Or we can also only support `<let/>` as is, which means if it's defined after a binding already created a property on binding context, it will just use it.

### DIFF
--- a/packages/jit/src/templating/template-compiler.ts
+++ b/packages/jit/src/templating/template-compiler.ts
@@ -126,7 +126,6 @@ export function inspectAttribute(name: string, resources: IResourceDescriptions)
   return attributeInspectionBuffer;
 }
 
-
 @inject(IExpressionParser)
 export class TemplateCompiler implements ITemplateCompiler {
   public get name(): string {
@@ -147,7 +146,7 @@ export class TemplateCompiler implements ITemplateCompiler {
     // Parent node is required for remove / replace operation incase node is the direct child of document fragment
     const parentNode = node = isTemplate ? (<HTMLTemplateElement>node).content : node;
 
-    while (node = this.compileNode(node, parentNode, definition.instructions, resources)) { /* Do nothing */ }
+    while (node = this.compileNode(node, parentNode, definition, definition.instructions, resources)) { /* Do nothing */ }
 
     // ideally the flag should be passed correctly from rendering engine
     if (isTemplate && (flags & ViewCompileFlags.surrogate)) {
@@ -160,13 +159,14 @@ export class TemplateCompiler implements ITemplateCompiler {
   public compileNode(
     node: Node,
     parentNode: Node,
+    definition: Required<ITemplateSource>,
     instructions: TargetedInstruction[][],
     resources: IResourceDescriptions
   ): Node {
     const nextSibling = node.nextSibling;
     switch (node.nodeType) {
       case NodeType.Element:
-        this.compileElementNode(<Element>node, parentNode, instructions, resources);
+        this.compileElementNode(<Element>node, parentNode, definition, instructions, resources);
         return nextSibling;
       case NodeType.Text:
         if (!this.compileTextNode(<Text>node, instructions)) {
@@ -226,13 +226,15 @@ export class TemplateCompiler implements ITemplateCompiler {
   public compileElementNode(
     node: Element,
     parentNode: Node,
+    definition: Required<ITemplateSource>,
     instructions: TargetedInstruction[][],
     resources: IResourceDescriptions
   ): void {
     const tagName = node.tagName;
 
     if (tagName === 'SLOT') {
-      throw new Error('<slot/> not implemented.');
+      definition.hasSlots = true;
+      return;
     } else if (tagName === 'LET') {
       const letElementInstruction = this.compileLetElement(node, resources);
       instructions.push([letElementInstruction]);
@@ -310,7 +312,7 @@ export class TemplateCompiler implements ITemplateCompiler {
     const current = node;
     let currentChild = node.firstChild;
     while (currentChild) {
-      currentChild = this.compileNode(currentChild, current, instructions, resources);
+      currentChild = this.compileNode(currentChild, current, definition, instructions, resources);
     }
   }
 
--- a/packages/jit/src/templating/template-compiler.ts
+++ b/packages/jit/src/templating/template-compiler.ts
@@ -12,7 +12,10 @@ import {
   IHydrateAttributeInstruction,
   IHydrateElementInstruction,
   IHydrateTemplateController,
+
   ILetBindingInstruction,
+  ILetElementInstruction,
+
   IListenerBindingInstruction,
   INode,
   IPropertyBindingInstruction,
@@ -43,16 +46,6 @@ const domParser = <HTMLDivElement>DOM.createElement('div');
 const marker = DOM.createElement('au-marker') as Element;
 marker.classList.add('au');
 const createMarker: () => HTMLElement = marker.cloneNode.bind(marker, false);
-<<<<<<< HEAD
-=======
-const swapWithMarker = (node: Element, parentNode: Element) => {
-  const marker = createMarker();
-  const template = DOM.createElement('template') as HTMLTemplateElement;
-  (node.parentNode || parentNode).replaceChild(marker, node);
-  template.content.appendChild(node);
-  return marker;
-};
->>>>>>> feat(runtime): Let binding, priorityInstructions
 
 const enum NodeType {
   Element = 1,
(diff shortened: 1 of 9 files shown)


EisenbergEffect:
All instructions have assumed a target up to this point.

packages/jit/src/templating/template-compiler.ts
```diff
@@ -180,9 +181,21 @@ export class TemplateCompiler implements ITemplateCompiler {
...
+    if (tagName === 'SLOT') {
+      throw new Error('<slot/> not implemented.');
+    } else if (tagName === 'LET') {
+      const letInstructions = this.compileLetElement(node);
+      if (letInstructions.length > 0) {
+        instructions.push(letInstructions);
+        // theoretically there's no need to replace, but to keep it consistent
+        DOM.replaceNode(createMarker(), node);
+      } else {
+        node.remove();
```
bigopon: @EisenbergEffect Do you think we should embrace DOM level 4 with methods like `remove()`, `replaceWith()`, etc ?

EisenbergEffect: I haven't recently checked what browsers support what. I think we still need to support IE11. So, if that has them, we are probably ok. Still, I prefer to be a bit conservative. I hadn't seen `replaceWith` before. `remove` is one we're already using.

fkleuver: Maybe it's a good idea to add a new `@aurelia/polyfills` package. We definitely don't need the vCurrent ones anymore, but we could have a new set.
Here's the "official" `replaceWith` polyfill for IE10+: https://developer.mozilla.org/en-US/docs/Web/API/ChildNode/replaceWith#Polyfill

EisenbergEffect: I really do not want to maintain our own polyfills going forward. It's been a constant pain with vCurrent...

I'd rather refer people to some other resources if we can. Otherwise, we need to engineer a proper polyfill build that is independent of Aurelia's load, and lets people pick which polyfills they want. I'd prefer to not do this right now...

bigopon:
Alright, changes:

1. `<let/>` will be turned into a `ILetElementInstruction`
2. attributes on let will go into `ILetElementInstructions.instructions`
3. The renderrer will just remove the marker when rendering (not sure if swapping the let element with marker is necessary

@fkleuver @EisenbergEffect

EisenbergEffect:
Removing the marker should be fine. Ready for review?

bigopon:
Yes. Unless @fkleuver has something to add

packages/runtime/src/binding/let-binding.ts
```diff
@@ -0,0 +1,132 @@
...
+import { IObserverLocator } from './observer-locator';
+
+const slotNames: string[] = new Array(100);
+const versionSlotNames: string[] = new Array(100);
+
+for (let i = 0; i < 100; i++) {
+  slotNames[i] = `_observer${i}`;
+  versionSlotNames[i] = `_observerVersion${i}`;
+}
+
```
fkleuver: Not that important but these slotNames aren't used here and can be removed

packages/runtime/src/binding/let-binding.ts
```diff
@@ -0,0 +1,132 @@
...
+    if (this.$isBound) {
+      if (this.$scope === scope) {
+        return;
+      }
+      this.$unbind(flags);
+    }
+
+    this.$isBound = true;
+    this.$scope = scope;
+    this.target = this.toViewModel ? scope.bindingContext : scope.overrideContext;
```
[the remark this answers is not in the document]
bigopon: It sounds very nice though I don't see much usage of it beside let. the logic of `getContextFor` conflicts with the flag I think.

bigopon:
@EisenbergEffect @fkleuver It's ready for a final review. I'll start updating the vCurrent PRs based on this soon

packages/runtime/src/binding/let-binding.ts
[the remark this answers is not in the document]
fkleuver: For example targeting extra meta properties such as `$odd`, `$even` on the repeater, maybe something like `$shown` on the `if` binding. They're all conceptually equivalent to `let` in that they're assigning computed properties to the override context. So then instead of hard-coding those computed properties into the template controllers, we could also allow users to hook into them and add extra info (or remove what they dont need), and decide whether to put it on the binding or override context.

packages/jit/test/unit/templating/template-compiler.integration.spec.ts
```diff
@@ -518,4 +518,28 @@ describe('TemplateCompiler (integration)', () => {
...
+      '<template><let full-name.bind="firstName + ` ` + lastName"></let><div>\${fullName}</div></template>',
+    );
+    au.app({ host, component: component }).start();
+    expect(host.textContent).to.equal(' ');
+    component.firstName = 'bi';
+    component.lastName = 'go';
+    expect(host.textContent).to.equal(' ');
+    cs.flushChanges();
+    expect(host.textContent).to.equal('bi go');
+  });
```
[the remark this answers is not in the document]
bigopon: 😎 I'm thinking of writing a loop to test every possible combination of instruction with let together with some usage bindings for integration test.

fkleuver:
Yep love it. It's probably fine from technical POV, just a liiitle bit lackluster on the testing side of things. I would verify stuff like:
- If you change one of the properties, is the computed value updated? (I can tell from the code that it should, but adding tests for this helps defend against regressions later on with refactoring) 

I haven't been setting the right example on this yet (tests on Binding class are also quite poor), but OTOH that class has been battle-tested quite extensively through other means.

- Are the change handlers / evaluates not called more than needed?

- What happens if you use the result of one `let` in the computed input of another `let`?

- What's the actual breakage when you declare the `let` after its usage? (verify that it's either undefined or throws or whatever, just so that that's established and "documented")

I could come up with more but I also need to do this myself for some of the other stuff. I can (will) help you with it

packages/jit/test/unit/templating/template-compiler.integration.spec.ts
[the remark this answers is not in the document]
fkleuver: Yep, I'm going to work towards a boatload of combinatory tests in a few places too. Just a head up though that generating loops for that kind of tests can easily take a day if you want to do it properly

packages/jit/src/templating/template-compiler.ts
```diff
@@ -224,8 +230,15 @@ export class TemplateCompiler implements ITemplateCompiler {
     resources: IResourceDescriptions
   ): void {
     const tagName = node.tagName;
-    if (tagName === 'SLOT' || tagName === 'LET') {
-      throw new Error('<slot/> or <let/> not implemented');
+
+    if (tagName === 'SLOT') {
+      throw new Error('<slot/> not implemented.');
```
EisenbergEffect: Note that there's nothing to do for slots other than set `hasSlots: true` at least for now.

packages/jit/test/unit/templating/template-compiler.integration.spec.ts
```diff
@@ -518,4 +518,28 @@ describe('TemplateCompiler (integration)', () => {
...
+
+  it('<let [to-view-model] />', () => {
+    component = createCustomElement(
+      '<template><let to-view-model full-name.bind="firstName + ` ` + lastName"></let><div>\${fullName}</div></template>',
+    );
+    au.app({ host, component: component }).start();
+    component.firstName = 'bi';
+    component.lastName = 'go';
+    cs.flushChanges();
+    expect(host.textContent).to.equal('bi go');
```
EisenbergEffect: Should this test check to see if the property was added to the view model as well?

packages/runtime/src/binding/let-binding.ts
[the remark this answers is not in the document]
EisenbergEffect: That's an interesting idea. Maybe something for a future pass. I think we could do that after we get a few other things squared.

bigopon:
@EisenbergEffect Done.

packages/jit/src/templating/template-compiler.ts
```diff
@@ -220,12 +226,21 @@ export class TemplateCompiler implements ITemplateCompiler {
...
     instructions: TargetedInstruction[][],
     resources: IResourceDescriptions
   ): void {
     const tagName = node.tagName;
-    if (tagName === 'SLOT' || tagName === 'LET') {
-      throw new Error('<slot/> or <let/> not implemented');
+
+    if (tagName === 'SLOT') {
+      definition.hasSlots = true;
+      return;
```
EisenbergEffect: You will have to traverse the children though as there could be fallback slot content that we need to search for elements, bindings, etc. Sorry for not making that clear. Feel free to do it in a separate PR tough.

EisenbergEffect: I'll merge this but we may need to revisit that. It looks like it's returning rather than traversing the children of the slot.

PR #133, opened 2018-09-02
implement instructionCompiler / renderStrategy (formerly bindingCommand)

fkleuver:
In short:

`@instructionCompiler` (jit) is a slightly more generic name for binding command (it could do more when the other extensibility points I had in mind are in place). You put it on top of a class that implements the `compile` method. This method takes a heap of information from the template compiler's current state and returns an `IRenderStrategyInstruction`

The `IRenderStrategyInstruction` is handled separately by the renderer. It has a name that corresponds to the `@renderStrategy` (runtime) decorator. You put this on top of a class that basically behaves like a single-method mini renderer.

As an example I implemented the `.keydown` binding command that was discussed a few times, with the instructionCompiler / renderStrategy. You can find it in `instruction-compiler.spec.ts`. (note: I was planning to add an extensibility point to the parser as well to allow other binding operators, but I didn't get to that yet)

There was also an issue with the unit tests that caused them to give false positives. Cleaned things up a bit and fixed the tests.

@bigopon can you get behind these changes or did I mess something up (sorry in that case!)


fkleuver:
@EisenbergEffect @bigopon 
Ok so for the sake of experiment I tried @jods suggestion of moving (most of) the command compilation out of the template compiler and into decorators. I renamed it to `bindingCommand` because it just makes more sense that way imo. I think it's actually pretty clean like this, feels more extensible too. What do you guys think?

bigopon:
I think the following https://github.com/aurelia/aurelia/blob/cd36ceb56b9c2154c60a25aeaf65f3e4ed620ea5/packages/jit/src/templating/template-compiler.ts#L193-L195

Will be invalid when it's an attribute with binding, on a custom element, but not one of the element bindable properties. @fkleuver

fkleuver:
Actually.. @EisenbergEffect isn't every attribute instruction supposed to be a child of the element instruction, regardless of whether the attribute is a bindable property of the element (except for template controllers ofc)?

fkleuver:
`hydrateElementInstance`:

```js
      if (currentType === TargetedInstructionType.stylePropertyBinding || currentType === TargetedInstructionType.listenerBinding) {
        realTarget = target;
      } else {
        realTarget = component;
      }
```

Here's what I think: if attribute instructions are siblings of an element instruction, then those instructions will be hydrating the wrong thing because targets are found by the `"au"` class (and you can only have one per element). So whether or not instructions are targeting bindable properties of a custom element, they should always be child instructions of the custom element or they would end up on the wrong target.

Now the renderer only picks the element as the real target if it's a style/class binding or event listener. Does this mean you cannot set "raw" properties on the attributes of a custom element? Don't we perhaps need an additional property in an instruction saying whether it's targeting the node or the component of the element that it's placed on? @EisenbergEffect

bigopon:
@fkleuver 
> Here's what I think: if attribute instructions are siblings of an element instruction, then those instructions will be hydrating the wrong thing because targets are found by the "au" class (and you can only have one per element). So whether or not instructions are targeting bindable properties of a custom element, they should always be child instructions of the custom element or they would end up on the wrong target.

I think you are right for the above, but I'm not sure if the rendering engine should be defensive about this. Ideally I think there is nothing but bindable instruction in child instructions of a hydrate element instruction. It seems a bit hard to understand, so lets have some example:

```ts
export class El {
  @bindable() a: any;
  @bindable() b: any;
}
```

```html
<template>
  <el a.bind="a" b.bind="b" c.bind="d" e.trigger="e" style.bind="s"></el>
</template>
```

the instructions should be

```js
[
  { hydrateElement, instructions: [ bindableA, bindableB ] },
  { elementPropertyC },
  { elementListenerE },
  { elementStyle }
]
```
`style` and `normal` element properties shouldn't be in

fkleuver:
@bigopon I need to rethink this, probably we may need to move the bindable check further up and split out the attribute compilation into 2 methods or something.

EisenbergEffect:
@fkleuver I haven't read the code yet. About to do that. But, regarding your question above, yes you are correct. Those need to be child instructions. I do think we'll need some further metedata to clarify whether it's the class or the element that needs to be targeted. That code that's in there is left over from the very earliest version of the instruction application code and it's definitely not suitable for production.

EisenbergEffect:
Ok, finally got the chance to look over the code...

I LOVE THIS! 😍 

I know there are still some details and refinements to make, but this is a fantastic direction to take things. Is this ready enough to merge in now? or does more work need to be done on this branch first?

fkleuver:
Nice :)
I was trying to further reorganize and fine tune the compiler but it's too much work for one sitting, so best probably merge this in so it doesn't block @bigopon 
I'll continue the rest later. (no need to wait for me to finish btw @bigopon, is any merge conflicts show up I'll resolve them)

EisenbergEffect:
Ok, merged in!

PR #134, opened 2018-09-04
refactor(template-compiler): move stuff around / fix various edge cases

fkleuver:
@bigopon @EisenbergEffect

### DIFF
--- a/packages/jit/src/templating/template-compiler.ts
+++ b/packages/jit/src/templating/template-compiler.ts
@@ -41,7 +41,6 @@ const marker = DOM.createElement('au-marker') as Element;
 marker.classList.add('au');
 const createMarker: () => HTMLElement = marker.cloneNode.bind(marker, false);
 
-
 const enum NodeType {
   Element = 1,
   Attr = 2,
@@ -91,6 +90,8 @@ export function resolveTarget(target: string, element: ElementDefinition, attrib
   return target;
 }
 
+const attributeInspectionBuffer: [string, AttributeDefinition, IBindingCommand | null] = <any>Array(3);
+
 /*@internal*/
 export function inspectAttribute(name: string, resources: IResourceDescriptions):
   [string, AttributeDefinition, IBindingCommand | null] {
@@ -113,7 +114,10 @@ export function inspectAttribute(name: string, resources: IResourceDescriptions)
   }
   const attributeDefinition = resources.find(CustomAttributeResource, targetName) || null;
 
-  return [targetName, attributeDefinition, bindingCommand];
+  attributeInspectionBuffer[0] = targetName;
+  attributeInspectionBuffer[1] = attributeDefinition;
+  attributeInspectionBuffer[2] = bindingCommand;
+  return attributeInspectionBuffer;
 }
 
 
--- a/packages/jit/src/templating/template-compiler.ts
+++ b/packages/jit/src/templating/template-compiler.ts
@@ -1,4 +1,4 @@
-import { Immutable, inject, PLATFORM } from '@aurelia/kernel';
+import { inject, PLATFORM } from '@aurelia/kernel';
 import {
   BindingMode,
   BindingType,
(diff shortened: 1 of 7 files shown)


packages/jit/src/templating/template-compiler.ts
```diff
@@ -60,37 +57,91 @@ const enum NodeType {
   Notation = 12
 }
 
+/*@internal*/
+export function resolveTarget(target: string, element: ElementDefinition, attribute: AttributeDefinition): string {
+  // give custom attributes priority over custom element properties (is this correct? should we throw if there's a conflict?)
```
bigopon: Yes. It is how we current do it with vCurrent. Unless it's well documented and configurable or a very good reason to change it, I think we keep the behavior.

packages/jit/src/templating/template-compiler.ts
```diff
@@ -60,37 +57,91 @@ const enum NodeType {
...
+      bindingCommand = resources.create(BindingCommandResource, commandName) || null;
+      if (bindingCommand !== null) {
+        // keep looping until the part after any dot (doesn't have to be the first/last) is a bindingCommand
+        break;
+      }
+    }
+  }
+  const attributeDefinition = resources.find(CustomAttributeResource, targetName) || null;
+
+  return [targetName, attributeDefinition, bindingCommand];
```
bigopon: I think for JIT, with the big amount of attributes, this could cause unnecessary objects. Maybe hoist the array and dynamically change the element values

bigopon: I think for JIT, with the big amount of attributes, this could cause unnecessary objects. Maybe hoist the array and dynamically change the element values

fkleuver: Good idea, I changed it. Would this work?

packages/jit/src/templating/template-compiler.ts
```diff
@@ -168,74 +223,87 @@ export class TemplateCompiler implements ITemplateCompiler {
...
-        if (instruction.type === TargetedInstructionType.hydrateTemplateController) {
-          liftInstruction = instruction;
-          // lift instructions will take over the compilation
-          // no need to process further if hit 1
-          break;
+      const attr = attributes.item(i);
+      const { name, value } = attr;
+      const [target, attribute, command] = inspectAttribute(name, resources);
+
+      if (attribute !== null) {
```
bigopon: In vCurrent, it's the binding command that decides the instruction of an attribute on template. With this check, it's no longer the behavior. I think we shouldn't make this change as command should be in control. By doing so, we can also have away to specify dynamic attributes / instructions. @fkleuver  @EisenbergEffect

fkleuver: Is there any situation in which the binding command would tell a template controller to not be a template controller..?

bigopon: Please ignore. Just checked again, I actually misstated how the vCurrent compiler behaves. Sorry for this. I was under an impression that binding command will decide everything, so `if.fun` will be completely different with `if.bind`. (though it could make sense)

packages/jit/src/templating/template-compiler.ts
bigopon: 👍

fkleuver:
@EisenbergEffect I think this is good to merge?

packages/jit/test/unit/templating/template-compiler.spec.ts
```diff
@@ -67,7 +68,7 @@ const attrNameArr = [
 ];
 
 
-describe('TemplateCompiler', () => {
+describe.only('TemplateCompiler', () => {
```
EisenbergEffect: Should we remove this `.only` before we merge?

fkleuver: I think it's already removed?

PR #139, opened 2018-09-05
fix(observer-locator): add collection length back in, fix precedence, add tests

fkleuver:

### DIFF
--- a/packages/runtime/src/binding/collection-observer.ts
+++ b/packages/runtime/src/binding/collection-observer.ts
@@ -1,5 +1,7 @@
-import { CollectionKind, CollectionObserver, ICollectionObserver, IndexMap, MutationKind } from './observation';
+import { BindingFlags } from './binding-flags';
+import { Collection, CollectionKind, CollectionObserver, IBindingTargetAccessor, ICollectionObserver, IndexMap, MutationKind } from './observation';
 import { batchedSubscriberCollection, subscriberCollection } from './subscriber-collection';
+import { targetObserver } from './target-observer';
 
 function flushChanges(this: CollectionObserver): void {
   this.callBatchedSubscribers(this.indexMap);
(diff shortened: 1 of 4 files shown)


bigopon:
Will to string work in the same way witg instanceof for sub classes ? @fkleuver

fkleuver:
As far as I know, yes

```
class Arr1 extends Array{}
class Arr2 extends Arr1 {}
var a2 = new Arr2();
a2 instanceof Array // true
Object.prototype.toString.call(a2) // [object Array]
```

```
class Arr1 extends Array{}
class Arr2 extends Arr1 {
  constructor() {
    super();
    Object.setPrototypeOf(this, Arr2);
  }
}
var a2 = new Arr2();
a2 instanceof Array // false
Object.prototype.toString.call(a2) // [object Object]
```

bigopon:
I think this approach has its own issue too. This is what I meant https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/toStringTag

In short, it will fail for:

```js
class Collection extends Array {
  [Symbol.toStringTag]() {
    return 'Collection';
  }
}

Ojbect.prototype.toString.call(new Collection()) === '[object Collection]'
```

@fkleuver @EisenbergEffect 

Instance isn't 100% reliable and slower but I think it works fine for this.

fkleuver:
```
class Collection extends Array {
  get [Symbol.toStringTag]() {
    return 'Collection';
  }
}
```

Do we *want* that to succeed? If you try hard enough you can always break the framework..

bigopon:
It's a normal & arbitrary usage of language feature, there is nothing wrong with the above, and we may even see more rare stuff in many applications. I wouldn't expect it to cause issue to my usage of a framework.

EisenbergEffect:
Since the new implementation is faster, could we keep that and then only revert to instance checks if an actual developer hits the issue?

fkleuver:
I have no problems reverting it, the perf difference is negligible. I just thought it was a bit cleaner. We do need to define some of these scenarios and edge cases and have extensive tests for those too. Maybe we need to gather them up in a separate issue

PR #140, opened 2018-09-06
docs

fkleuver:
@EisenbergEffect @bigopon 

In this PR:
- Add licenses and basic readme's to each package so that each standalone install has all the things an npm package needs
- Remove the progress checklist from the frontpage (this belongs in issues, milestones, etc -> we should create those asap)
- Add up-to-date and complete information on what the various npm scripts do, how to set up the repo, testing/debugging
- Add --bail mode for debug testing, blackbox framework scripts, and clarify how to more easily test new stuff without having to use `.only`
- Add overview readme at the root of packages folder
- Add instructions on setting up and running the examples


docs/engineering/README.md
```diff
@@ -1,3 +1,165 @@
...
+Install dev dependencies:
+
+```bash
+npm ci
+```
+
+Install/symlink package dependencies:
+
+```bash
+npm run bootstrap # just like npm ci, this command only needs to be run once
```
bigopon: thanks for this. I don't know how many times it will be run without it. I ran it ... not less than 10 👍

bigopon:
looks great to me

PR #149, opened 2018-09-09
test(binding): increase tests coverage

bigopon:


packages/runtime/src/binding/computed-observer.ts
```diff
@@ -98,14 +98,16 @@ export class CustomSetterObserver implements CustomSetterObserver {
...
 
     this.observing = true;
     this.currentValue = this.instance[this.propertyName];
 
     Reflect.defineProperty(this.instance, this.propertyName, {
-      set: function(newValue) {
-        setter(newValue);
+      // tslint:disable-next-line:no-any
+      set: function(newValue: any): void {
+        setter.call(that.instance, newValue);
```
bigopon: Should it be this way ? @EisenbergEffect

bigopon:
At the moment, the logic to decide when to augment a property on an object doesn't cater for DOM elements thus volatile is `false` for them, which I think is inappropriate. I think we need to add an extra check to make sure it always go to dirty checking when trying to observe a property that is in the prototype of the element. @EisenbergEffect @fkleuver

packages/runtime/src/binding/computed-observer.ts
fkleuver: The `this` context of the setter needs to be the observer, so that would be `setter.call(that, newValue);`
But `that`, `this`, `there`, `thus`, I think this is a legacy thing and we should use an arrow function or a `.bind` (truth be told I'm still not completely certain which of the two is better).

bigopon: It should be `that` if `this` of the `setter` is `that`, but in this case, it the instance.

fkleuver:
Just so you know, I completely broke the computed observer when I rewrote the collection observers. I haven't properly tested it since - the only thing that's currently covered is its initialization.

But don't the dependencies go through the observerLocator anyway? That always accounts for DOM elements first, so I'm not sure if DOM dependencies can ever make it through. DOM elements themselves should never make it through to begin with for the same reason

bigopon:
It accounts for a certain list of properties. It used to fall to dirty checking for the rest. And yes, agree that anything that is in prototype should not go through.

EisenbergEffect:
@bigopon Ready for review or still working on this?

packages/runtime/src/binding/computed-observer.ts
fkleuver: Oops you're right. But in that case I think you can just make it `this` instead of `that.instance`? I believe the `this` inside a descriptor with a normal function expression will always refer to the instance

bigopon:
@EisenbergEffect Fred will take over this. My part here is done

packages/runtime/src/binding/computed-observer.ts
bigopon: Yes. I was focusing on `that` and didnt pay attention what `this` is.

AureliaEffect:
Allure Report: https://2792-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/e2e/allure-report/index.html

AureliaEffect:
Allure Report: https://2852-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/e2e/allure-report/index.html

AureliaEffect:
Allure Report: https://2857-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/e2e/allure-report/index.html

AureliaEffect:
Allure Report: https://2892-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/e2e/allure-report/index.html

AureliaEffect:
Allure Report: https://2898-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/e2e/allure-report/index.html

AureliaEffect:
Allure Report: https://2903-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/e2e/allure-report/index.html

AureliaEffect:
Allure Report: https://2941-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/e2e/allure-report/index.html

AureliaEffect:
Allure Report: https://3066-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/e2e/allure-report/index.html

PR #150, opened 2018-09-09
DI tests

fkleuver:
@EisenbergEffect @bigopon 
I'm creating this PR ahead of time because I want to show you what's going on here.
Please look at `the case at [line 133](https://github.com/aurelia/aurelia/pull/150/files#diff-88bcb8378393687b8b5af91397a1298eR133) and compare this with the case at [line 249](https://github.com/aurelia/aurelia/pull/150/files#diff-88bcb8378393687b8b5af91397a1298eR249). This could explain some of the wonkyness we experienced with resolving types.

TLDR:
```ts
function decorator(): ClassDecorator { return (target: any) => target; }

const Foo1 = class{};
@decorator() // designParamTypes = [Foo1] = OK
class Bar1 { constructor(foo: Foo1) }

interface Foo2 {}
const Foo2 = class{};
@decorator() // designParamTypes = [Foo2] = OK
class Bar2 { constructor(foo: Foo2) }

interface Foo3 {}
const Foo3: Foo3 = class{};
@decorator() // designParamTypes = [Object] = WRONG
class Bar3 { constructor(foo: Foo3) }

const Foo4 = function(){};
@decorator() // designParamTypes = [Foo4 ] = OK
class Bar4 { constructor(foo: Foo4) }

interface Foo5 {}
const Foo5 = function(){};
@decorator() // designParamTypes = [Object] = WRONG
class Bar5 { constructor(foo: Foo5 ) }
```

There are a few more strange cases which honestly surprised me. I'm not sure how this affects our `InterfaceSymbol` implementation but I'll find that out soon enough I guess.


EisenbergEffect:
Where is it that we're actually using the TS param metadata in the framework or apps?

EisenbergEffect:
Maybe there's an issue inside the inject decorator and we're allowing TS to override the explicit dependencies? and it's messing them up sometimes?

fkleuver:
From the runtime coverage report:

image

From the jit coverage report:

image

Not sure (yet) where exactly these calls come from, but plenty it seems.

To my knowledge this will happen anytime we're relying on a decorator to get the type info (probably non-DI decorators)

fkleuver:
> Maybe there's an issue inside the inject decorator and we're allowing TS to override the explicit dependencies? and it's messing them up sometimes?

The comment I posted at the top had nothing to do with the inject decorator. It's applying *any* decorator to a class that gets these `design:paramTypes` generated, and the point I'm making is that the returned value will be the wrong thing in some cases. Explicitly casting certain symbols to an interface and then using the interface as the constructor argument type would break it.

EisenbergEffect:
The design time types should be irrelevant, meaning it doesn't matter if they are wrong, if the end user has explicitly provided an inject decorator that lists the dependencies. That `getDesignParamTypes` should never be called if there is an explicit dependency list...unless there is an error in the DI logic somewhere.

fkleuver:
> The design time types should be irrelevant, meaning it doesn't matter if they are wrong

It matters if you're using another decorator. If you apply `@customElement` to a class, that would traditionally always work the same way as if you applied `@inject` to it. I've always considered that a big elegance of TypeScript, that you don't need to explicitly declare the inject everywhere. `design:paramTypes` are very important to be correct or a lot of people (including me) will have a lot of code that breaks and needs to change just to get DI working again.

fkleuver:
But I'll add some more tests first and see where exactly things break. It's not completely broken, we just need to be mindful of what we cast those symbols to.

EisenbergEffect:
It's never worked the way you are asking for it though. The TS team knows it's broken and has no plan to fix it until after they land the new decorators implementation. It's out of our control.

EisenbergEffect:
You can't use design params with interfaces, even if you have a same-named constant. It's not supported by the compiler, basically.

EisenbergEffect:
You have to use the symbols in an explicit dependency list OR you have to use concrete classes with design params. You can't mix and match if you want the design params to be inferred. It's a bug in the TS compiler. We should definitely call this out in the DI docs.

EisenbergEffect:
I have a small worry that doing that will open up a different set of problems.

fkleuver:
I wouldn't even attempt it before we have 200% test coverage on DI ;)
Once that is the case I'd say we could try a couple things to see if we can accomplish the best of both world.

EisenbergEffect:
The TS team is going to fix this, but not until after they update to the new decorators version. That's my understanding. This is a known bug reported by a number of people, but they aren't making any fixes since design params is an experimental feature built on an experimental feature. They are waiting for the decorators proposal to mature so they can get that working and then address the typing and design params issues (if they keep that feature) after that.

EisenbergEffect:
Note that I don't think we should by any means release vNext to 1.0 until after the official decorators proposal is implemented in TS. Otherwise, we have to turn around and break things. I'm hoping that these issues will get fixed automatically.

fkleuver:
@EisenbergEffect it's ready for review. A few fixes:

fix(di): invoke correct method on array strategy resolver
feat(di): recurse through static registrations to find register methods in more edge cases
fix(di): invalidate Object keys to help diagnose invalid design:paramTypes
refactor(di): append new resolvers on existing keys to a single array strategy resolver instead of nesting them
fix(di): add a non-any type alternative to the InterfaceSymbol<T> so that container.get() returns correctly typed instances

I'm not 100% sure if all of them make sense in your vision, perhaps I made some wrong assumptions. Especially the array strategy is a bit unclear to me as it seems that only the first resolver is ever invoked.

Apart from that, it's mostly tests tests tests.

Jit registration, child container behavior, `getAll` and certain complex interactions would still need to be tested but I do believe DI test coverage is about 70-80% there now.

EisenbergEffect:
Thanks @fkleuver I'm going to review this via the new VS Code plugin and see how that works ;)

AureliaEffect:
This pull request **fixes 1 alert** when merging 4a9d93101e6fe2fc89f0ccd570d92627ca253d0d into a87b09bb6bc20c6f3694fefe1850d59249001edb - [view on LGTM.com](https://lgtm.com/projects/g/aurelia/aurelia/rev/pr-352e901e71b91c12d91239b29961b7a9571268f9)

**fixed alerts:**

---

AureliaEffect:
Allure Report: https://2494-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/e2e/allure-report/index.html

packages/kernel/src/di.ts
```diff
@@ -251,7 +255,7 @@ export class Resolver implements IResolver, IRegistration {
       case ResolverStrategy.callback:
         return (this.state as ResolveCallback)(handler, requestor, this);
       case ResolverStrategy.array:
-        return this.state[0].get(handler, requestor);
+        return this.state[0].resolve(handler, requestor);
```
EisenbergEffect: Good catch! This was a refactoring that was made not too long ago and got missed due to lack of tests (and wide typing).

packages/kernel/src/di.ts
```diff
@@ -376,10 +382,10 @@ export class Container implements IContainer {
 
     if (result === undefined) {
       resolvers.set(key, resolver);
-    } else if (resolver instanceof Resolver && (resolver as Resolver).strategy === 4) {
+    } else if (result instanceof Resolver && (result as Resolver).strategy === ResolverStrategy.array) {
```
EisenbergEffect: Nice catch again!

packages/kernel/src/di.ts
```diff
@@ -376,10 +382,10 @@ export class Container implements IContainer {
...
       (result as Resolver).state.push(resolver);
     } else {
-      resolvers.set(key, new Resolver(key, 4, [result, resolver]));
+      resolvers.set(key, new Resolver(key, ResolverStrategy.array, [result, resolver]));
```
EisenbergEffect: Thanks for moving this over to use the enum! I must have missed these literals when I made the refactor to add the enum.

packages/kernel/src/di.ts
```diff
@@ -510,7 +516,7 @@ export class Container implements IContainer {
       return keyAsValue.register(handler, keyAsValue);
     }
 
-    const resolver = new Resolver(keyAsValue, 1, keyAsValue);
+    const resolver = new Resolver(keyAsValue, ResolverStrategy.singleton, keyAsValue);
```
EisenbergEffect: Thanks for updating this!

packages/kernel/src/di.ts
```diff
@@ -566,13 +572,15 @@ export const Registration = {
 
 /*@internal*/
 export function validateKey(key: any): void {
-  if (key === null || key === undefined) {
+  // note: design:paramTypes which will default to Object if the param types cannot be statically analyzed by tsc
+  // this check is intended to properly report on that problem - under no circumstance should Object be a valid key anyway
```
EisenbergEffect: I love this! Very nice developer-happiness improvement :)

packages/kernel/src/di.ts
```diff
@@ -566,13 +572,15 @@ export const Registration = {
...
+  if (key === null || key === undefined || key === Object) {
     throw Reporter.error(5);
   }
 }
 
 function buildAllResponse(resolver: IResolver, handler: IContainer, requestor: IContainer): any[] {
-  if (resolver instanceof Resolver && resolver.strategy === 4) {
+  if (resolver instanceof Resolver && resolver.strategy === ResolverStrategy.array) {
```
EisenbergEffect: Looks like I missed quite a few of these hdardcoded values when I made the refactor. Thanks for sorting it all out!

packages/kernel/src/di.ts
```diff
@@ -348,22 +352,24 @@ export class Container implements IContainer {
     this.resolvers.set(IContainer, containerResolver);
   }
 
+  public register(registry: (IRegistry | Record<string, Partial<IRegistry>>)): void;
```
EisenbergEffect: This threw me for a second. I actually didn't know you could have multiple signatures on a concrete class. I thought this was only allowed on an interface. Good to know :)

packages/kernel/src/di.ts
```diff
@@ -348,22 +352,24 @@ export class Container implements IContainer {
...
   public register(...params: (IRegistry | Record<string, Partial<IRegistry>>)[]): void {
-    const resolvers = this.resolvers;
-
     for (let i = 0, ii = params.length; i < ii; ++i) {
       const current = params[i];
-
       if (isRegistry(current)) {
-        current.register(this);
+        (<any>current)['register'](this);
```
EisenbergEffect: Can we instead do `current.register(this);`?

packages/kernel/src/di.ts
```diff
@@ -348,22 +352,24 @@ export class Container implements IContainer {
...
-
-          if (value.register) {
-            value.register(this);
+        const keys = Object.keys(current);
+        for (let j = 0, jj = keys.length; j < jj; ++j) {
+          const value = current[keys[j]];
+          // note: we could remove this if-branch and call this.register directly
+          // - the extra check is just a perf tweak to create fewer unnecessary arrays by the spread operator
+          if (isRegistry(value)) {
+            value['register'](this);
```
EisenbergEffect: Can we instead do `value.register(this);`?

AureliaEffect:
This pull request **fixes 1 alert** when merging 8245742a89d660abefcaad144ebe402b3f44d7ad into a87b09bb6bc20c6f3694fefe1850d59249001edb - [view on LGTM.com](https://lgtm.com/projects/g/aurelia/aurelia/rev/pr-50c926bc20bd954331d7860492ef493252c1644f)

**fixed alerts:**

---

packages/kernel/src/di.ts
fkleuver: Easy peasy :)

AureliaEffect:
Allure Report: https://2501-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/e2e/allure-report/index.html

AureliaEffect:
This pull request **fixes 1 alert** when merging b46d36ce73300378a6b8219bee4831f75be3cbcf into a87b09bb6bc20c6f3694fefe1850d59249001edb - [view on LGTM.com](https://lgtm.com/projects/g/aurelia/aurelia/rev/pr-d0ed5dc19b88dd15bd462eca62167fe15ce259df)

**fixed alerts:**

---

AureliaEffect:
Allure Report: https://2505-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/e2e/allure-report/index.html

AureliaEffect:
This pull request **fixes 1 alert** when merging 5b35d61d9781ff52bec11603908892d4d54e7f19 into a87b09bb6bc20c6f3694fefe1850d59249001edb - [view on LGTM.com](https://lgtm.com/projects/g/aurelia/aurelia/rev/pr-25ef16aba850069dba6173a390b6da640c364706)

**fixed alerts:**

---

bigopon:
@EisenbergEffect the decorator proposal doesnt look like it will be moving forward anytime soon. Iy got tied with private field, and that proposal isnt going to be resolved soon either. Maybe we need to have some pre official decorator milestone

PR #155, opened 2018-09-11
chore(all): cleanup build scripts, fix nightly dependency versions

fkleuver:
This PR is intended as a small step in the direction of a more coherent build/test/release workflow. Adds a `project.ts` file that exposes frequently used information such as paths and package names. Other scripts use this instead of resolving the same paths over and over.

The publish script will now replace the dependency version names in the packages package.json files with the nightly version as well. This will not in itself solve the "outdated dependency" warning we have in the readme, but will ensure version integrity when you install the dev lane. This mechanism will work the same way later on when we start publishing to `@latest`


AureliaEffect:
Allure Report: https://2637-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/e2e/allure-report/index.html

PR #157, opened 2018-09-12
chore(all): add validate-commit-msg hook and commitizen script

### LINKED ISSUE #127: Add commitizen for enforcing commit message conventions

fkleuver:
fix #127

Commit messages are automatically validated via husky hook.
Use `npm run commit` to get an interactive CLI for building a properly formatted commit message

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,6 +3,7 @@
 [![TypeScript](https://img.shields.io/badge/%3C%2F%3E-TypeScript-%230074c1.svg)](http://www.typescriptlang.org/)
 [![CircleCI](https://circleci.com/gh/aurelia/aurelia.svg?style=shield)](https://circleci.com/gh/aurelia/aurelia)
 [![Test Coverage](https://api.codeclimate.com/v1/badges/5ac0e13689735698073a/test_coverage)](https://codeclimate.com/github/aurelia/aurelia/test_coverage)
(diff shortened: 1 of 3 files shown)


AureliaEffect:
Allure Report: https://2671-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/e2e/allure-report/index.html

bigopon:
@fkleuver For a PR, does this apply to only first commit (surrogate commit) or every commit needs to conform this ?

fkleuver:
The validation will happen whenever you do `git commit` locally. You can bypass the validation with `--no-verify` for WIP commits. We could adjust the validation rules to allow those sort of special cases as well, this is just a first setup

EisenbergEffect:
I like the idea of this :) Based on the various issues I've seen on vCurrent over the last few years, I think this will definitely help. I do worry that will be a bit much on each commit but we can play around with it a bit.

Shall we merge this?

fkleuver:
> I do worry that will be a bit much on each commit but we can play around with it a bit

I think the `--no-verify` flag will be sufficient for WIP stuff, but in the end we do want each commit to follow some kind of convention and get ourselves in the habit of making those descriptive. The better the commit messages are, the less manual work there will be left when it comes to the change logs and writing blog post etc.

Yeah it should be good to merge.

PR #161, opened 2018-09-12
build(all): first attempt at rolling up

fkleuver:
@bigopon Ok, by the looks of it rollup definitely won't work for dev/test purposes. It just doesn't have the proper ts3 integrations etc. Maybe in a year it will.

For publish outputs it might work (just missing something for the type declarations), but we'd need to replace the `module`, `main` and `jsnext:main` fields in package.jsons with the rollup outputs right before publishing.

I like the idea of having just 2 or so module outputs though. `umd` accounts for `amd`, `commonjs` and `iife`. Then `esm` accounts for native, and all that's missing is systemjs (we could add it back in when we're closer to alpha, and save ourselves a few hours of building time in the meantime..)

I don't know if I want to burn my fingers on this right now though. Getting a few warnings/errors I don't fully understand. Stuff about names and `this` and whatnot. It's a brand new native es2018 target / esnext module output. I don't understand what rollup could possibly be complaining about..


bigopon:
@fkleuver I tried running build, it produced nice output and we got circular dependencies resolution for free 👍 We should specify only the bundled output with  typings in `files`. Beside some circular dependencies and guessing work for external module for classic script loading way, all look very good to me. Nice work

scripts/bundle.ts
```diff
@@ -0,0 +1,44 @@
...
+      name: PLATFORM.camelCase(pkg.name),
+      format: 'esm',
+      sourcemap: 'inline'
+    });
+
+    await bundle.write({
+      file: join(pkg.path, 'dist', 'index.js'),
+      name: PLATFORM.camelCase(pkg.name),
+      format: 'umd',
+      sourcemap: 'inline'
```
bigopon: Maybe just `true` ? We wouldn't want to distribute the source map next to the distribution code

scripts/bundle.ts
```diff
@@ -0,0 +1,44 @@
...
+        }),
+        typescript2({
+          tsconfig: join(pkg.path, 'tsconfig.json'),
+          typescript: ts
+        })
+      ],
+      external: project.packages.filter(p => p.name !== pkg.name).map(p => p.scopedName)
+    });
+
+    await bundle.write({
```
bigopon: @fkleuver @EisenbergEffect This is absolutely what I wanted / planned to do: using rollup to roll a bundle to enable usage of vNext in a script tag. With this in place, maybe we should start distributing that too

scripts/bundle.ts
bigopon: @EisenbergEffect Should we also specify `tslib` to be required, or just let rollup inline necessary helper functions. It's one time import per package, so relatively small, but could still be an annoyance for folks

fkleuver: I think we already settled on including tslib. It "just needs to happen". Regarding the sourcemaps, I understood somewhere somehow that inline was the only thing that'd work. Not entirely sure. But to be fair, ES2018 isn't the worst code to debug. It's practically typescript anyway. So we may as well just lose the sourcemaps

scripts/bundle.ts
fkleuver: I'm not 100% sure how that works for `UMD` (it's possible, I just don't know how to invoke it). We could explicitly specify an `iife` output. That really makes it easy. You just get a global `Aurelia` object on the window. We tried to move away from that idea but over these last few years I've started to grow rather fond of simple globals..

scripts/bundle.ts
bigopon: What I meant is this format
image

Instead of the sourcemap inlined into end of the output file

fkleuver:
> Beside some circular dependencies warnings

Is this something we should strive to get rid of? I don't like warnings. In my experience today's warnings are tomorrow's errors.

> warnings for external module for classic script loading

What do you mean by this? There are no external modules and there is no classic script loading afaik? How would we get rid of those?

> all look very good to me.

Hmm, I think we're looking at this from different angles. I guess you're happy that it's one file - I'm quite unhappy that all the typescript goodness (type declarations and source mappings) and a warning-free build process are out the window :P

bigopon:
> Is this something we should strive to get rid of? I don't like warnings. In my experience today's warnings are tomorrow's errors.

Yes, except that maybe it could make more sense to be circular. I anticipate some restructuring / refactoring down the line so it doesn't seem to matter to me.

---

> What do you mean by this? There are no external modules and there is no classic script loading afaik? How would we get rid of those?

For amd, commonjs, `a module name can be mapped 1:1 as they are all strings. For global format of umd (umd supports 3 format: amd, commonjs and global), module name will need to be transformed into a property on global object. example:

```js
import { PLATFORM } from '@aurelia/kernel';
// amd
define(['exports', '@aurelia/kernel']...)
// commonjs
require('@aurelia/kernel')
// global
// cannot have the equivalent of the loaders, so just assume the module `@aurelia/kernel` comes from a global name
window.kernel
```
Because of that characteristic, with some more configuration, we can bundle all core packages together in a single script file for distribution.

---
> ... all the typescript goodness (type declarations and source mappings) and a warning-free build process are out the window :P

They are not. For development purpose, we can still use the normal output from tsc, and I think it's better that way. The bundled packages are meant for npm distribution only, and for application, I don't know if having separate files is more or less desired than a packed version of them

fkleuver:
>  for application, I don't know if having separate files is more or less desired than a packed version of them
Tbh I've never cared about it because in the end, an end-user that uses rollup or webpack or whatever will get exactly the same end result after bundling all their stuff together.
That said, I've also never used Aurelia in `iife` format because it wasn't possible. Now it seems like it is, and rollup makes it rather clean.

I tweaked a couple more things and there are only 2 circular dep warnings left. I forgot that the build/test process was already wiring up the packages in a way independently of package.json, so I was able to just change those fields without breaking stuff. I think it's actually all good now?

scripts/bundle.ts
```diff
@@ -32,7 +32,15 @@ async function bundle() {
 
     await bundle.write({
       file: join(pkg.path, 'dist', 'index.js'),
+      exports: 'named',
       name: PLATFORM.camelCase(pkg.name),
+      globals: {
+        ...project.packages.reduce((g, pkg) => {
+          g[pkg.scopedName] = pkg.name;
```
bigopon: For `umd`, maybe we only need to point everything to a global namespace like `au` or `aurelia`. It's a bit weird when you do
```html
<script src='aurelia.bundle.js'></script>
<script>
  new aurelia.Aurelia({...})
</script>
```

But the benefit is we only have 1 namespace to remember, and it reflects the barrel export like we have in `aurelia-framework'

scripts/bundle.ts
```diff
@@ -32,7 +32,15 @@ async function bundle() {
...
+          return g;
+        }, {}),
+        'tslib': 'tslib'
```
bigopon: Also, for `umd`, maybe we may want to just inline `tslib` helpers ... in case it is used in script tag. What you think ?

EisenbergEffect: If we can make it work, for the single file version only, it would be cool to inline those helpers and make it truly a single file.

scripts/bundle.ts
EisenbergEffect: Would it be possible to have something that was namespaced like this?

```typescript
new au.jit.Aurelia();
au.kernel.DI.createContainer();
```

Basically, one namespace for each package. That might help with people who want to correlate the single file build with the packages and documentation.

Thoughts?

bigopon: @EisenbergEffect Brilliant!

For the following config:

```js
    await bundle.write({
      file: join(pkg.path, 'dist', 'index.js'),
      name: `au.${PLATFORM.camelCase(pkg.name)}`,
      format: 'umd',
      globals: {
        '@aurelia/kernel': 'au.kernel',
        '@aurelia/runtime': 'au.runtime',
        '@aurelia/jit': 'au.jit'
      },
      sourcemap: true
    });
```

we have this umd:

```js
(function (global, factory) {
    typeof exports === 'object' && typeof module !== 'undefined' ? factory(exports, require('@aurelia/kernel')) :
    typeof define === 'function' && define.amd ? define(['exports', '@aurelia/kernel'], factory) :
    (factory((global.au = global.au || {}, global.au.runtime = {}),global.au.kernel));
}(this, (function (exports,kernel) { 'use strict';
```

scripts/bundle.ts
fkleuver: For single file we'll need to create separate bundles that include the Aurelia deps themselves. Probably jit+runtime+kernel+tslib?

scripts/bundle.ts
fkleuver: Yep love it :)

scripts/bundle.ts
bigopon: that'd be ideal 👍 😄

AureliaEffect:
Allure Report: https://2807-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/e2e/allure-report/index.html

AureliaEffect:
Allure Report: https://2813-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/e2e/allure-report/index.html

AureliaEffect:
Allure Report: https://2818-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/e2e/allure-report/index.html

AureliaEffect:
Allure Report: https://2822-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/e2e/allure-report/index.html

bigopon:
This is how it looks like when built:
image


So yes, fully compiled to es6.

fkleuver:
@obedm503 @bigopon It's compiled to es2018 actually, a few versions higher than es6 (es2015)

obedm503:
@fkleuver then does it really make sense to name it `index.es6.js`?
it also might make sense to rename the `umd` build to just `index.js`

so `index.es6.js` -> `index.mjs` and `index.umd.js` -> `index.js`?

fkleuver:
@obedm503 I don't feel strongly about it at all. Whatever makes the most sense to the community as a whole is what we should go with in this case.

If you look at `scripts/bundle.ts` you can see that there are a few more build outputs that can be toggled on/off via configuration. I had it `index.js` before but I saw some elegance/clarity in keeping them consistent with how rollup names them.
I think if/when we settle on a definite set of build outputs we may want to follow a naming scheme such as the one you're suggesting, if that's considered to be more standard.

obedm503:
no particular use case. if there are more outputs, the way it currently is makes more sense

PR #170, opened 2018-09-15
Upload iife bundles to azure

fkleuver:
This is just a simple azure blob container I created in my private Azure account. It can be moved to an official Aurelia account simply by changing the `AZURE_STORAGE_CONNECTION_STRING` env variable in CircleCI. Container will automatically be created if it doesn't exist inside the storage account.

Automatically runs after the nightly `dev` publish and the triggered `latest` publish. Root `dist` folder is uploaded twice: once to the root of the target lane (that's overwritten with each upload) and once to a versioned subfolder (which never changes). Example output:

(note: the version will be the actual dev version if it runs on the CI environment. it just reads out the lerna.json which contained 0.1.1 when I ran it locally)
image

Cleaned up the iife example to target the latest dev file directly. Now it's nothing but an index.html in there, using `https://aurelia.blob.core.windows.net/releases/dev/au.bundle.js` :)


EisenbergEffect:
Very nice! Is there a way to transfer ownership of the aurelia.blob... in the future?

Issue #163, opened 2018-09-13
View Lifecycle Improvements


EisenbergEffect:
## Background

At present, when a renderable (custom element or view) is told to detach, it always removes its DOM nodes, even if its parent has already been removed from the DOM. This is highly inefficient and serves no purpose.

## Plan

We should evolve the lifecycle code to ensure that only the owner of the lifecycle removes its nodes. This will result in the minimal amount of DOM modification on a detach. Following up from that, if the same renderable is attached again later, only its top-level nodes should be placed back in the DOM, since all the children are still connected into their parents. One final scenario needs to be explored: custom element view projection. A custom element should only project its view once, regardless of how many times the element is added or removed from the DOM. An exception to this (read complication) is containerless, which would need to remove their nodes when the element is the lifecycle owner, since they don't have a single container (host or shadowDOM root) to remain attached to. This should be handled by the IElementProjector abstraction, so it's hopefully not too complicated.

PR #171, opened 2018-09-15
feat(lifecycle): new callbacks, async tasks, and updated template controllers

EisenbergEffect:
## Background

Previously, when a renderable (custom element or view) was told to detach, it always removed its DOM nodes, even if its parent had already been removed from the DOM. This behavior is a carryover from vCurrent and is highly inefficient. This PR attempts to address the issue.

## Solution

There are several pieces to this:

1. Update the DetachLifecycle so that when a renderable owns the lifecycle, only its nodes are removed.
2. Update the DetachLifecycle so that if a non-renderable owns the lifecycle (e.g. Repeat), only top-level renderables (the views managed by the Repeat, not children of those views) have their nodes removed.
3. Abstract node removal for renderables so that renderables can individually track whether or not their nodes are addeded.
4. Improve View's abstraction for assigning a location for the view by adding a `mount` API to replace the previous `onRender` callback.
5. Update renderables to only add their nodes when they are not already mounted.
6. Add special behavior for containerless custom elements, since they are the only custom element scenario where the actual nodes need removal. (This is because all other scenarios are handled by the fact that the container is removed.)

> Note: As part of this PR, some vestigial code left over from the previous Compose refactor was removed. A few minor refactors/renames were performed as well.

## Related Issues

* [View Lifecycle Improvements](https://github.com/aurelia/aurelia/issues/163)

## TODO

All existing tests pass but new tests that describe lifecycle behavior need to be added. Once those are added, this PR will be ready.


AureliaEffect:
This pull request **introduces 1 alert** when merging 1c288e0330864631dc0c76698b43007da306492d into 978c2f01a4d3c27feb1f48454ee9990f193e928e - [view on LGTM.com](https://lgtm.com/projects/g/aurelia/aurelia/rev/pr-7319f82652efe4a13c5f254e27c0086cedc4cf55)

**new alerts:**

---

AureliaEffect:
Allure Report: https://2917-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/e2e/allure-report/index.html

AureliaEffect:
Allure Report: https://2947-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/e2e/allure-report/index.html

AureliaEffect:
Allure Report: https://2979-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/e2e/allure-report/index.html

bigopon:
Does this have any implication on granularity of each View that doesn't own the `LifeCycle` ?

EisenbergEffect:
Can you clarify the question or provide an example? I'm not quite sure.

AureliaEffect:
Allure Report: https://2998-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/e2e/allure-report/index.html

bigopon:
For views with behaviors, callback life cycle `detached` means the view is detached and put back in its own place, without the parent location that it was attached to. With this change, from what I understand, it's no longer true. It could affect caching?

EisenbergEffect:
Oh, man, that's a problem probably. I'll have to think about that. It might be worth it to just remove caching. This new behavior is probably more important. I'm not sure.

EisenbergEffect:
Well, for started, when we return a view to cache, we would need to:

* null out the location
* remove the nodes

However, there's a trick to this, since we don't want to remove the nodes or return to cache at the wrong time WRT the lifecycle itself. So, we may need to tweak these APIs more.

EisenbergEffect:
This stinks...massive complication I think. Ugh...

EisenbergEffect:
Maybe we can make it work with a requirement that you must own the lifecycle object to return things to the cache? Thoughts?

bigopon:
> Oh, man, that's a problem probably. I'll have to think about that. It might be worth it to just remove caching. This new behavior is probably more important. I'm not sure.

Caching requires lowest amount of LOC and effort for the same amount of perf gain, I think it will be very hard to justify removing caching. I remember repeat used to have it own caching mechanism and then later merged into part of templating. Maybe you can give some rationale behind that move so we can have stronger base for discussion?

> Maybe we can make it work with a requirement that you must own the lifecycle object to return things to the cache? Thoughts?

From what it looks like, at any point, there is only one owner. This seems limited, and I think it's inappropriate to put that restriction on.

EisenbergEffect:
I'll probably need to rethink the entire detach lifecycle again then 😭

bigopon:
For your inspiration, there is this plugin from @huochunpeng https://github.com/buttonwoodcx/bcx-aurelia-reorderable-repeat , it may / may not employ caching, but it show the strength of granularity control: ability to have any arbitrary way to implement the behavior that fits your need.

fkleuver:
Problem is that efficient caching means components often need to know something about their ancestors and descendants. This is something that I managed to solve partly in binding with the flags, because an ancestor can tell a descendant what it's doing and the descendant can, based on that, decide whether or not to actually remove the views during `$unbind`. I think something similar can be done with attach/detach.
I got a few ideas on how to improve this but I need to sit down for a bit on that. For now, it's important that we get a few meaningful benchmarks first so we can play around with different things and compare the impact rather than theorize about it. Perfect caching is impossible anyway as some assumptions need to be made.

bigopon:
@EisenbergEffect Yes, that makes perfect sense. It seems to me only view from template controller will behave differently comparing to normal views, which means it could very much be the responsibility of the template controller implementation do decide what to do with those views. I think it's safe to assume that there shouldn't be any issue. My previous concern was probably unnecessary.

Beside that, I think another potential issue could be memory leak where view nodes has reference to view model, either directly or through a model object. But maybe I'm just over-worrying.

fkleuver:
I think caching should be done first and foremost on template string level, preferably via some globally accessible view factory, much like the expression parser cache. Key difference being of course that you can have multiple cached elements per template string.

A repeater creates 10.000 of some view, returns it to cache, then there will be 10.000 cached views for that particular string available for any other component to grab. This becomes a memory problem when you have dynamically generated templates that differ and are never reused, you'll have an ever-expanding cache. This is another reason to have caching done globally - a cache size limit will be able to keep dynamic templates in check, not just large amounts of the same single template

fkleuver:
> It might be worth it to just remove caching.


> Maybe you can give some rationale behind that move so we can have stronger base for discussion?

I can give a rationale for that. "premature optimization is the root of all evil" (and yes, I'm very guilty myself). I think if we remove caching *for the time being* (and take the animator along with it please) we'll significantly simplify the templating code base, allowing us to focus on getting everything to work *correctly* first. When everything is 100%, tested, proven, and beautiful, we can build caching back in (and have the tests tell us exactly when we're doing something wrong)

EisenbergEffect:
I think it's possible that you are missing out on some of the conversations (and painful experiences) we had with vCurrent on the core team in the past @fkleuver One of the biggest issues with vCurrent is that we waited too long to integrate caching...and as a result, created a lot of problems. Animation was also integrated in a poor way and caused complexity and oddity throughout. The vNext template system, while lacking many tests, has been rebuilt incrementally and is technically almost feature-complete, with the exception of animation. For me, the next step is working with fixing a core issue of vCurrent around how nodes are added/remove in hierarchies. This is an issue that has bothered me for over 3 years and is one I've literally waited on for 9mo since the vNext effort was started. I believe the time is right to revisit it. After that, things will be in place to handle animation. Some of the work to set that up is part of this. Some of the work was part of the template controller re-writes I did, that established the patterns that would enable this. So, this isn't out of nowhere for me, it's been on my radar for a while and I've been strategically doing other incremental work to prepare the way for this.

fkleuver:
I understand how you've been working towards this, and it makes perfect sense for the most part. I'm also trying to understand and help ensure that all design decisions are coherent with some of the fundamental changes that have been made in vNext.

All that said, whenever you work on something it always comes out a lot better so I'll wait this one out :)

EisenbergEffect:
I think I need to probably write a little spec for this, to think through some of the scenarios. The PR here (once it has tests) is not a bad first crack. It's definitely an improvement, but not quite what it needs to be.

AureliaEffect:
Allure Report: https://3181-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/e2e/allure-report/index.html

AureliaEffect:
Allure Report: https://3186-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/e2e/allure-report/index.html

AureliaEffect:
Allure Report: https://3292-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/e2e/allure-report/index.html

AureliaEffect:
This pull request **fixes 1 alert** when merging 63959e7df680c05aaf8d869a67930fdd287e5d52 into 57c07e8ffac77dfd6cc55529d972c55016f316e8 - [view on LGTM.com](https://lgtm.com/projects/g/aurelia/aurelia/rev/pr-1b38df0314a326f1efaec2569b4eb77616818ff2)

**fixed alerts:**

---

AureliaEffect:
Allure Report: https://3298-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/e2e/allure-report/index.html

AureliaEffect:
This pull request **fixes 1 alert** when merging 039309a2f6015455ecd07ad6654d6623e6b1ff43 into 57c07e8ffac77dfd6cc55529d972c55016f316e8 - [view on LGTM.com](https://lgtm.com/projects/g/aurelia/aurelia/rev/pr-1541905ae079d51575e0a77cd20cddae22c67bbd)

**fixed alerts:**

---

bigopon:
@EisenbergEffect From the commit, I understand that if we want to implement a list with staggering enter animation, we need to scope lifecycle to list and create a child for each list item for both `detached` and `attached` ?

EisenbergEffect:
For staggering enter animation, you don't need to do anything with the lifecycle. The lifecycle is now generalized to simply allow a task to block node add and attached callback or block node remove and detached callback. It's a general hook that an animation system could use to prevent nodes from being removed before an animation finishes (also enables synchronous compose and other scenarios). I don't think an animation system would use that for attach since you don't want to block adding nodes. This removes the need for the animation system to have to exist in the core and allows it to exist entirely as a plugin. It's possible we'll still have a standard mechanism like before, but that doesn't need to live inside the runtime now.

AureliaEffect:
This pull request **introduces 4 alerts** and **fixes 1** when merging 14bbed60a11a5954cf067d08d679977bc4a0cb98 into 57c07e8ffac77dfd6cc55529d972c55016f316e8 - [view on LGTM.com](https://lgtm.com/projects/g/aurelia/aurelia/rev/pr-c2d2c07ad0691a0c8152d610a7344297999eefef)

**new alerts:**

**fixed alerts:**

---

AureliaEffect:
This pull request **introduces 4 alerts** and **fixes 1** when merging d540ea0408e56aee25e001e8ad74d08829af7816 into 57c07e8ffac77dfd6cc55529d972c55016f316e8 - [view on LGTM.com](https://lgtm.com/projects/g/aurelia/aurelia/rev/pr-e469c4bb33e16c4b313f80b6cc08fb19ede02519)

**new alerts:**

**fixed alerts:**

---

AureliaEffect:
This pull request **introduces 6 alerts** and **fixes 1** when merging b2fef296502208a6c2eed3aa1fa4b8b9a86701c6 into 57c07e8ffac77dfd6cc55529d972c55016f316e8 - [view on LGTM.com](https://lgtm.com/projects/g/aurelia/aurelia/rev/pr-fcb6712458326da0d78ada3959f422bf79fc37ec)

**new alerts:**

* 6 for Unused variable, import, function or class

**fixed alerts:**

---

AureliaEffect:
This pull request **introduces 4 alerts** and **fixes 1** when merging f120c74985472d84c9c37a1975c347ae634c0a0a into 57c07e8ffac77dfd6cc55529d972c55016f316e8 - [view on LGTM.com](https://lgtm.com/projects/g/aurelia/aurelia/rev/pr-e4c6c3b70c8391de87309a1731a290ba14607e0c)

**new alerts:**

**fixed alerts:**

---

AureliaEffect:
This pull request **introduces 1 alert** and **fixes 1** when merging fb8fbd2409d634d0d5e68f749464ed8cc420280f into 9eb0764711f930d4b5ff930bffac2501508b8d5e - [view on LGTM.com](https://lgtm.com/projects/g/aurelia/aurelia/rev/pr-3f06d466c202933a007817809d819ec62c39142d)

**new alerts:**

**fixed alerts:**

---

PR #175, opened 2018-09-16
ci: temporarily disable flaky and time-consuming jobs

fkleuver:
With the new circleci 2.1 this will be much easier to manage. Parameterized commands will also allow us to configure individual jobs more granularly based on the workflow they're running in (e.g. only run the bundle script if it's inside the nightly workflow). For the time being, I'm disabling the benchmarks and the allure reports to make the jobs run faster and eliminate some intermittent failures

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -30,6 +30,12 @@ aliases:
       command: |
         cd ./test/e2e
         npm run start
+  - &step_bootstrap_build
+    run:
+      name: Bootstrap and build the packages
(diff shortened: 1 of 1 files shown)


fkleuver:
Another small change: e2e job will itself run the minimum needed bootstrap+build scripts. The build job then only verifies that the whole build works, but the e2e job will not wait for it to finish. All 3 jobs (test, build, e2e) running in parallel means about 2 minute less wait per push

Issue #114, opened 2018-08-30
Dynamic Behavior APIs


EisenbergEffect:
I'm using this issue to put down some design thoughts on our dynamic behavior APIs. There are APIs that would be used by the `compose` element internally, but that also would be used directly by developers who need to create and apply attributes or elements on the fly.

> Note: No interface or method names are final here. I'm just trying to get some ideas down...

## Idea 1

For convenience, I'm thinking we'll hang these APIs off of the `Aurelia` object. We'll have APIs for elements, attributes and template controllers. Something like this:

```typescript
const au = new Aurelia();
const a = au.createElement(MyCustomElement);
const b = au.createAttribute(MyCustomAttribute);
const c = au.createTemplateController(MyTemplateController, myTemplate);
```

Rather than having these APIs return the element, attribute and template controller directly, I'm thinking they will return something like an `IBehaviorManipulator`. The reason for this is that once you've got an instance of one of these things, what do you do with it? Well, there are a couple of different things I think developers might want to do. The most common is that I think they'd want to graft it into an existing view hierarchy within their app. With vCurrent, developers had some clunky APIs to create things on the fly and then once they created it, they had to manage the entire lifecycle directly. For vNext, I think you should be able to create one of these behaviors and then graft it into the existing view hierarchy so that it can manage the lifecycle for you from that point forward. Here's what that might look like:

```typescript
const ele = au.createElement(MyCustomElement);
const handle = ele.insertBeforeViewNode(document.getElementById('foo'));
```

At some later time, you could remove the behavior, if you want as well:

```typescript
handle.removeFromView();
```

The interfaces should be defined so that you can't call the wrong methods at the wrong time. So, you might have something like this:

```typescript
class Aurelia {
  createElement(elementType: Constructable): IElementManipulator;
}

interface IElementManipulator {
  insertBeforeViewNode(node: INode): IConnectedElement;
  appendToViewNode(node: INode): IConnectedElement;
}

interface IConnectedElement {
  removeFromView(): IElementManipulator;
}
```

I think the manipulator will look a bit different for attributes and template controllers:

```typescript
interface IAttributeManipulator {
  applyToViewNode(node: INode): IConnectedAttribute;
}

interface IConnectedAttribute {
  removeFromView(): IAttributeManipulator;
}
```

Beyond this, I think people will want to add behaviors to the dom and maintain manual control over the lifecycle. So, I think there would be additional methods on the behavior manipulator, perhaps something like this:

```typescript
[...]

PR #179, opened 2018-09-16
examples: jit-fuse-box-ts

PraveenGandhi:
Hi @EisenbergEffect @fkleuver 

I have added fuse-box example, keeping jit-webpack-ts as reference.

Thanks,
Praveen.

### DIFF
--- a/examples/bundlers/jit-aurelia-cli-ts/README.md
+++ b/examples/bundlers/jit-aurelia-cli-ts/README.md
@@ -14,5 +14,5 @@ To install and run the app (in "prod" mode):
 ```bash
 npm install
 npm run build
-npm run start
+npm run serve
 ```
--- a/examples/bundlers/jit-aurelia-cli-ts/package.json
+++ b/examples/bundlers/jit-aurelia-cli-ts/package.json
@@ -4,7 +4,7 @@
   "scripts": {
     "watch": "au run --watch",
     "build": "rimraf scripts && au build",
-    "start": "http-server -c-1 -p 9000 ."
+    "serve": "http-server -c-1 -p 9000 ."
   },
   "dependencies": {
     "@aurelia/jit": "dev",
--- a/examples/bundlers/jit-browserify-ts/README.md
+++ b/examples/bundlers/jit-browserify-ts/README.md
@@ -14,5 +14,5 @@ To install and run the app (in "prod" mode):
 ```bash
 npm install
 npm run build
-npm run start
+npm run serve
 ```
--- a/examples/bundlers/jit-fuse-box-ts/.gitignore
+++ b/examples/bundlers/jit-fuse-box-ts/.gitignore
@@ -1,0 +1,3 @@
+.fusebox
+dist
+node_modules
--- a/examples/bundlers/jit-fuse-box-ts/.npmrc
+++ b/examples/bundlers/jit-fuse-box-ts/.npmrc
@@ -1,0 +1,1 @@
+package-lock=false
--- a/examples/bundlers/jit-fuse-box-ts/README.md
+++ b/examples/bundlers/jit-fuse-box-ts/README.md
@@ -14,5 +14,5 @@ To install and run the app (in "prod" mode):
 ```bash
 npm install
 npm run build
-npm run start
+npm run serve
 ```
--- a/examples/bundlers/jit-fuse-box-ts/README.md
+++ b/examples/bundlers/jit-fuse-box-ts/README.md
@@ -1,0 +1,18 @@
+# jit-fuse-box-ts
+
+## Getting started
+
+To install and run the app (in watch mode):
+
+```bash
+npm install
+npm run watch
+```
+
+To install and run the app (in "prod" mode):
+
+```bash
+npm install
+npm run build
+npm run start
+```
--- a/examples/bundlers/jit-fuse-box-ts/fuse.js
+++ b/examples/bundlers/jit-fuse-box-ts/fuse.js
@@ -1,0 +1,39 @@
+const { FuseBox, HTMLPlugin, EnvPlugin, UglifyJSPlugin, QuantumPlugin, WebIndexPlugin } = require("fuse-box");
+const { src, task } = require("fuse-box/sparky");
+
+let run = (production) => {
+    const fuse = FuseBox.init({
+        target: "browser@es6",
+        homeDir: 'src',
+        output: 'dist/$name.js',
+        runAllMatchedPlugins: true,
+        plugins: [
+            production && UglifyJSPlugin(),
+            production && QuantumPlugin(),
+            EnvPlugin({
+                devMode: !production
+            }),
+            HTMLPlugin(),
+            WebIndexPlugin({
+                template: './index.html'
+            })
+        ]
+    });
+    fuse.bundle("vendor")
+        .cache(true)
+        .instructions('~ startup.ts');
+    if (!production) {
+        fuse.bundle("app")
+            .instructions(" > [startup.ts]")
+            .hmr()
+            .watch();
+        fuse.dev();
+    } else {
+        fuse.bundle('app').instructions(" > [startup.ts]")
+    }
+    fuse.run();
+};
+
+task('clean', async () => await src('dist/*').clean('dist').exec());
+task("dev",     ['clean'], () => run(false));
+task("prod",    ['clean'], () => run(true));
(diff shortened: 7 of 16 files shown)


fkleuver:
Never seen fuse-box before..
Looks a bit like a mix breed between gulp and webpack but cleaner. I kinda like it! Thanks :)

@EisenbergEffect works like a charm, let's get this in

PraveenGandhi:
@fkleuver 

Fuse-box is simple and performant. 

By the way in all examples, the serve npm script is not in sync with README (npm run start). I can submit a pull request if required.

fkleuver:
@PraveenGandhi Oops, I must have missed that through all the renaming and moving stuff around :) The readme's across this repo are indeed lagging behind a bit. By all means feel free to PR readme fixes if you want to. It needs to happen at some point :)

PraveenGandhi:
OK. I will submit another pull request in this regard

HamedFathi:
@fkleuver 

You should see [fuse-box-aurelia-loader](https://github.com/fuse-box/fuse-box-aurelia-loader) by @vegarringdal

[Aurelia-Toolbelt](https://github.com/aurelia-toolbelt/aurelia-toolbelt) is based on it.

PraveenGandhi:
@HamedFathi 

You are right, however I think now with vNext we no longer need any kind of loader specific to "Aurelia". Now the vNext of Aurelia is bundler free if I an not wrong.

fkleuver:
@PraveenGandhi I don't feel strongly about it. I've been switching back and forth between the two a few times. It **start**s a **serve**r so what should it be called.. I think `npm run start` is a bit of a tautology so I'm more on the side of serve, but it's stretchy.

fkleuver:
@HamedFathi @PraveenGandhi There will be no more runtime loader for vNext. So the `aurelia-loader` package and its implementations won't be needed anymore in that form. We will however be having various bundler plugins for compile time processing

PraveenGandhi:
@EisenbergEffect 

Changes are done, can be merged now.

Thanks,
Praveen.

vegarringdal:
> You are right, however I think now with vNext we no longer need any kind of loader specific to "Aurelia". Now the vNext of Aurelia is bundler free if I an not wrong.

@PraveenGandhi I really hope this is true. Loaders was important when Aurelia was made but in the end it ended up being a pain/weak point if you did not like webpack.
Thx for making this PR.

Loader free Aurelia will be great. 
Might get me motivated fixing my [vgrid](https://github.com/aurelia-v-grid/vGrid) where I know I can do better. (get scared sometimes looking at the code:joy:)
Hopefully there isnt any breaking changes to how binding works that cant be fixed by editing code.


Is there any early migration docs made for how/what need to be changed ?
Is there any eary docs for how to work on aurelia ?
Only see link to a starter kit/app.
Never used lerna, guess I just haveto to start with npm install.. maybe  need to use yarn ?

vegarringdal:
@PraveenGandhi 

Maybe add 
`document.querySelector('app').innerHTML = '';`

To startup.ts, so we clear app element on saves/reloads ?

This way new users dont end up with something like this if they edit the `message`

image

fkleuver:
@vegarringdal I wouldn't recommend spending particularly much time on it just yet as a fair amount of core functionality isn't sufficiently tested yet. I'm still working on binding, looking to get every edge case covered from the get-go. We're already 11k unit tests in and still got quite a way to go.

> Is there any early migration docs made for how/what need to be changed ?
Not yet. Only low-level APIs will mostly change. Binding syntax stays the same, as do most decorators.

> Is there any eary docs for how to work on aurelia ?
docs/engineering contains some preliminary info. It's not 100% up to date but the basic setup (npm ci -> npm run bootstrap -> npm run build -> npm run dev etc) will work fine.

Do feel free to create issues and ping us if you have specific questions. I have been looking at your v-grid as one of the first things I'd like to try and see working in vNext. It's a complex component that will be a very good migration test for us. I'll be more than happy to help you out with it. We might even want a proper grid component in this repository as a plugin package.

PraveenGandhi:
> This is a HMR integration problem that should be solved in an aurelia bundler plugin package imo. If any of you have some ideas on that feel free to make a PR. We'd like to eventually have `@aurelia/plugin-webpack`, `@aurelia/plugin-parcel`, `@aurelia/plugin-fuse-box`, etc, and no one is working on that right now

@fkleuver 

I am interested, and requesting guidance on the starting this up and the path to progress.
I can start building one of the loader plugins with guidance.

fkleuver:
@PraveenGandhi The idea for the bundler plugins is to translate plain imports to the appropriate component configuration.

Roughly speaking, this:

```ts
import * as view from './app.html';

@customElement(view)
export class App {}
```

Should be translated to this:

```ts
import * as view from './app.html';

@customElement({
  name: 'app',
  templateOrNode: view,
  build: {
    required: true,
    compiler: 'default'
  },
  instructions: []
})
export class App {}
```

This tells the runtime that this custom element needs to be compiled by the template compiler, and contains the necessary information for the template compiler to do its work.

@EisenbergEffect has already implemented a working variant of this for requirejs (although not yet in the examples - this is a todo for is) which you can see in `packages/plugin-requirejs`.

From a bundler plugin perspective that's really all there is to it: non-ts/js imports need metadata added to them for the runtime to process them correctly.

vegarringdal:
> @PraveenGandhi The idea for the bundler plugins is to translate plain imports to the appropriate component configuration.
Roughly speaking, this:
```ts
import * as view from './app.html';

@customElement(view)

export class App {}
```
>Should be translated to this:
```ts
import * as view from './app.html';

@customElement({
  name: 'app',
  templateOrNode: view,
  build: {
    required: true,
    compiler: 'default'
  },
  instructions: []
})
export class App {}
```

@fkleuver 
This sounds like you are bring back issues that loader gave us.
Making it very hard for all bundlers to just work out the box. Places like stackblitz might get hard to use too if we need to use plugins with bundlers.

Why not just force minimum like this if you are making a custom element

```ts
@customElement({
  name: 'app',
  templateOrNode: view 
})
export class App {}
```

And remove stuff like `require` tag we had in html, I feel that is a bad way to do it.
If you need to load resources it should be a method on the class/decorator that accept a promise.


Please dont take any of this in a bad way :-)
There is a lot I dont know and you guys have been doing this way longer then me. So maybe I just dont understand it.

EisenbergEffect:
Allowing a promise at that point would introduce a ton of complexity back in and hurt performance. In fact, we'd have to change the entire component model. I don't think that's the direction we want to go, esp. since most apps will be bundled and can have their dependencies statically linked.

Loaders/packagers already do this work, so there's no reason not to let them do it. The problem is that the current version of Aurelia abstracted the loader then built a bunch of stuff on top of that abstraction, which ended up being a problem in the real world, esp. with Webpack. The "correct" abstract is the ES2015 import system itself. By leveraging that, and the fact that each loader has a way to plug into that, we can actually create simpler, easier to maintain plugins. It was hardly any work to create the require.js plugin, for example. I did it in an afternoon and most of the time it took came from the fact that I'd never built an r.js plugin before and didn't even know how to do it.

Most other frameworks have plugins of this type. They are common use with React and Vue, for example. I don't think we'll have the same type of problems as in the past.

Furthermore, we'll need to write plugins like this to enable AOT probably. 

@vegarringdal Please note, there is a plan to enable writing almost exactly what you show above. We need to tweak our apis and samples to not require the `build` and `instructions` properties. With that handled, you can write any custom element exactly as you show above (assuming it's using global resources). If you need resources scoped only to that element, then you have to add them in a dependency array. The loader plugins only layer on top to do that automatically, and AOT only layers on top of that yet again, to do more during the build and get you out of writing the decorator at all.

I hope that makes sense.

vegarringdal:
@EisenbergEffect 

> @vegarringdal Please note, there is a plan to enable writing almost exactly what you show above. We need to tweak our apis and samples to not require the build and instructions properties.

great. 👍 
Ill just add the build/instructions while I try and convert my grid then
Any changes to customElement on how to use, build/instructions here too ?

EisenbergEffect:
CustomAttributes are pretty similar. With TemplateControllers, you don't need both the attribute decorator and the template controller decorator anymore; you can just use the template controller decorator. There may be a minor difference or two but I think it will be pretty self explanatory via the TS definitions.

I should also note that we have some ideas around async components for vNext as well. I'll add those notes to the Dynamic Behavior API issue #114 shortly.

PR #184, opened 2018-09-18
add src files and sourcemaps to published outputs

fkleuver:


fkleuver:
@EisenbergEffect including the `src` files will cause the `index.d.ts.map` to actually work for consumers. Ought to clean it up in some way later on, but for the time being I think it's more valuable to have the best possible debugging experience rather than a smaller tarball

PR #187, opened 2018-09-20
fix(@bindable): respect alias and transform to kebab case

bigopon:


packages/runtime/src/templating/bindable.ts
```diff
@@ -34,7 +34,8 @@ export function bindable(configOrTarget?: BindableSource | Object, key?, descrip
     }
 
     config.property = key2;
-    bindables[key2] = config;
+    config.attribute = validateAttributeName(config.attribute);
```
fkleuver: Hmm why do we need to validate that it has an uppercase? If it doesn't and you send it to the kebabcaser, it won't get transformed anyway and will get cached..?

bigopon: the name is `validate...` but it actually does `validate, report if invalid + transform` 😓

packages/runtime/test/unit/templating/bindable.spec.ts
```diff
@@ -1,10 +1,29 @@
...
+      @bindable({ attribute: 'notprop' }) isProp: any
+    }
+
+    expect((Test as any).bindables['notprop']).not.to.be.undefined;
+  });
+
+  it('kebabs alias', () => {
+    const errorSpy = spy(Reporter, 'error');
+    class Test {
+      @bindable({ attribute: 'notProp' }) isProp: any
```
fkleuver: Don't you mean either `attribute: 'not-prop'` or `property: 'notProp'`? I don't believe attribute names are supposed to get kebabcased

bigopon: I did it that way to demonstrate 2 cases: it's always kebab-cased and it should report error.

packages/runtime/src/templating/bindable.ts
fkleuver: Yeh this makes sense too, it'll always be a bad idea to pass uppercased stuff into the decorator

packages/runtime/src/templating/bindable.ts
```diff
@@ -45,3 +46,11 @@ export function bindable(configOrTarget?: BindableSource | Object, key?, descrip
 
   return deco;
 }
+
+function validateAttributeName(attrName: string): string {
```
fkleuver: Would it make sense to you to do something like this insteadL
```ts
const kebab = PLATFORM.kebabCase(attrName);
if (attrName !== kebab) {
    Reporter.error(100, attrName);
}
```

Since it's always going through the kebabcaser in the template compiler anyway, and hides the implementation detail

packages/runtime/src/templating/bindable.ts
```diff
@@ -34,7 +34,8 @@ export function bindable(configOrTarget?: BindableSource | Object, key?, descrip
     }
 
     config.property = key2;
-    bindables[key2] = config;
+    config.attribute = validateAttributeName(config.attribute);
+    bindables[config.attribute] = config;
```
fkleuver: Why does this need to be the attribute name instead of the property name that it's declared on?

packages/runtime/src/templating/bindable.ts
bigopon: Yes, I was thinking of it too. It should be that way.

packages/runtime/src/templating/bindable.ts
bigopon: Because attribute is public API of a bindable property, while property is private one. We see an example with vGrid:
```js
@bindable({ attribute: 'name' }) AttributeName = 'Im missing!!'
```

fkleuver: Yes but that's a template compiler problem. What you're doing here is all correct except for the change in the bindable property imo. Template compiler should just loop through the bindables and find a bindable with a matching `attribute` property. If that becomes a performance issue (which I doubt it would), that's easy to optimize.
In the mean time various parts of the core framework rely on the bindable lookup properties mapping to the normal property names.

bigopon: Then that's a design issue: do we want bindable to be lookup of attributes, or of properties. We can also have 2 separate lookups, `bindables` for property and `$bindable` for attributes

EisenbergEffect: The actual metadata should contain both the property name and the attribute name already, I think. For now, maybe just create a map that reverses the lookup and use that. if we need, we can later find a way to do that more efficiently.

bigopon: @EisenbergEffect Then we can have `bindables` for property, can `bindableAttrs` for attributes ?

fkleuver: That is pretty much what I implied with:

> Template compiler should just loop through the bindables and find a bindable with a matching attribute property. If that becomes a performance issue (which I doubt it would), that's easy to optimize.

bigopon:
Resolved by #189

PR #188, opened 2018-09-20
binding / call / observer fixes and coverage

fkleuver:
In this PR, aside from test coverage:
- `PropertyAccessor` no longer batches any changes. This was an oversight on my part that would cause bindings in nested custom elements to be updated *after* the attach lifecycle. (note to self: urgently need unit tests for custom elements)
- Lazy initialize `slotNames` and `versionSlotNames` per 5, based on demand. This lifts the implicit restriction of 100 observers per binding without additional cost. It's very possible no one will ever end up benefiting from this, but it's a cheap change to ensure that the silent and impossible to track down bug this would case, can never happen.
- `$event` is removed from `Call`. Even though the 'price' to keep it is negligible in terms of size, it can be a significant source for confusion. Nothing wrong with the object itself but we could at least rename it to `$args`. Perhaps a packwards compat package thing, or perhaps will add it back in later if people turn out to rely heavily on it.


AureliaEffect:
This pull request **introduces 2 alerts** when merging 9a3dd6ccfe6729c853cfa1a9571a533e7b720a27 into fb417d4f4cd46a5f543949f3013101de2b420e6c - [view on LGTM.com](https://lgtm.com/projects/g/aurelia/aurelia/rev/pr-54d6d8ddc49592a1869464c6dfc376b0ca02ea9c)

**new alerts:**

---

AureliaEffect:
This pull request **introduces 2 alerts** when merging 5c91f2eafa3070d90edc12ad733c3449623affaa into fb417d4f4cd46a5f543949f3013101de2b420e6c - [view on LGTM.com](https://lgtm.com/projects/g/aurelia/aurelia/rev/pr-d8795abaca5c452d17f71b5791a6eaafda1f0cbe)

**new alerts:**

---

bigopon:
> PropertyAccessor no longer batches any changes. This was an oversight on my part that would cause bindings in nested custom elements to be updated after the attach lifecycle. (note to self: urgently need unit tests for custom elements)

This feels nice to me, in an synchronous binding system, I wouldn't expect the before-fix behavior. Though in a framework like Aurelia that already does too many of the heavy lifting, the chance that you have to change value across view models / view / bindings and rely on their sequence / readiness are rare, but this will certainly help make that more robust.

> Lazy initialize slotNames and versionSlotNames per 5, based on demand. This lifts the implicit restriction of 100 observers per binding without additional cost. It's very possible no one will ever end up benefiting from this, but it's a cheap change to ensure that the silent and impossible to track down bug this would case, can never happen.

This is true, a binding would highly unlikely have to subscribe more than 10 observers

For `$event`, I believe sooner or later, for migration purpose, we will need to have it back, but for now, it lgtm 👍

PR #189, opened 2018-09-21
template compiler rework

### LINKED ISSUE #185: Differently named attributes/properties on custom elements
Custom element unit tests are still on my todo-list for the template compiler, but @vegarringdal reported a [bug](https://github.com/aurelia-v-grid/vGrid/issues/99) that's specifically worth mentioning as it relates to the intricate combinations of possible edge-case configurations that we may get. 

@bigopon This may also be related to what you mentioned before. Due to the lack of tests in this particular use case I'm not sure yet if this is a template compiler issue or a renderer issue (it might be both), but at this point I agree with you it's probably better to keep this sort of complexity out of the renderer.

fkleuver:
fixes #185


fkleuver:
@EisenbergEffect @bigopon Ready for review.
Summary:
- Attribute instructions on custom elements are now only made into child instructions if they reference an existing binding description; otherwise, they become siblings. (this was a pretty important fix too, because it was blocking any normal attributes from ever being placed on the html element itself)
- Semicolon-separated attribute bindings are handled with similar rules and fallbacks to normal attributes in other contexts, making them equally flexible and robust
- Property name, attribute name and binding mode are correctly resolved even when the name of the descriptor, attribute and property are all different.
- Added about 2k tests to validate all of the above

Todos:
- shuffle some code around to reduce duplication in target resolution
- add caching here and there for better performance
- implement an intermediary compiler AST to serve as a semantic model

packages/jit/src/templating/template-compiler.ts
```diff
@@ -69,23 +69,25 @@ const resolved = {
...
 
 /*@internal*/
-export function resolveTarget(target: string, element: ElementDefinition, attribute: AttributeDefinition): typeof resolved {
-  let bindable: IBindableDescription;
-  const propertyName = PLATFORM.camelCase(target);
+export function resolveTarget(target: string, element: ElementDefinition, attribute: AttributeDefinition, isMultiAttrBinding: boolean): typeof resolved {
   // give custom attributes priority over custom element properties (is this correct? should we throw if there's a conflict?)
   if (attribute) {
     // only consider semicolon-separated bindings for normal custom attributes, not template controllers
     // users must not have a semicolon-separated binding with the same name as the attribute; behavior would be unpredictable
```
bigopon: Why is this, can you inline some example in the comment ?

packages/jit/src/templating/template-compiler.ts
```diff
@@ -69,23 +69,25 @@ const resolved = {
...
+        if (bindable.attribute === target) {
+          resolved.target = bindable.property;
+          resolved.mode = (bindable.mode && bindable.mode !== BindingMode.default) ? bindable.mode : BindingMode.toView;
+          resolved.bindable = bindable;
+          return resolved;
+        }
       }
+      resolved.target = target;
+      resolved.mode = BindingMode.toView;
+      resolved.bindable = null;
```
bigopon: It's better if you always reset `bindable` to `null` at the start of this function

fkleuver: Why?

bigopon: The following is what I meant in code:
```js
function resolveTarget() {
  resolved.bindable = null;
  ...
  // only need to set resolved.bindable to something else when we need to
  // no need to reset it based on if / else branch
}
```
Can save few bytes 🅱️

packages/jit/src/templating/template-compiler.ts
fkleuver:
```html
<div foo="foo.bind: 'bar';"></div>
```
That inner foo would also be resolved as a custom attribute rather than a bindable property. I don't know what exactly will happen.
But this comment was from before I wrote all this code. I could disable custom attribute resolution when it's doing semicolon separated bindings. Maybe that's better?

fkleuver: I'm a bit unsure about the precise resolution rules on this anyway. But it makes sense to disable custom attribute resolution inside a custom attribute, whether or not vCurrent does that, we should probably do it.
We don't even have `primaryProperty` atm though. Just picking the first one by default..

For your example, this would work the same:
```ts
@customAttribute('visible')
export class Visible {
  @bindable()
  visible: boolean
}
```

If there's no semicolon in the html template it will always just grab the first property it finds, and fallback to `value` if there are no bindables.

bigopon: It's not the same when you expand it to this:
```js
@customAttribute('visible')
export class Visible {
  @bindable({ primary: true })
  visible: boolean

  @bindable()
  visibility: number
}
```
```html
  <img visible="visible.bind: imgIsVisible; visiblity.bind: imgVisibility" />
```
Inside a custom attribute, all attributes are expected to be its properties. I think for vNext instruction style, it's a lot simpler and we don't have to define `@bindable` on properties of attribute at all

fkleuver: > Inside a custom attribute, all attributes are expected to be its properties

Argh I was overthinking it. Good point.

EisenbergEffect: Note that there's no `primary` property of a bindable in vNext currently. Instead, my idea is that `value` should always be the primary property. This was the initial idea for vCurrent even, but somehow got lost along the way. If an attribute has multiple bindables, and they don't want to use the name 'value', then they would specify an attribute name.

For example, here's how `route-href` would be configured:

```
@customAttribute('route-href')
export class RouteHref {
  @bindable({ attribute: 'route' }) value;
  @bindable params;
  @bindable attribute = 'href';
}
```

```xml
<a route-href="foo">Go to Route Foo</a>
<a route-href="route: foo; params.bind: bar;">Go to Parameterized Route</a>
```

AureliaEffect:
This pull request **introduces 15 alerts** and **fixes 4** when merging eb3889d3ce896de54dfa1c1d47167d6aa028499d into 57c07e8ffac77dfd6cc55529d972c55016f316e8 - [view on LGTM.com](https://lgtm.com/projects/g/aurelia/aurelia/rev/pr-b4ddf1dd054be30ec98a4a744ca0a089cbb30822)

**new alerts:**

* 14 for Unused variable, import, function or class
* 1 for Useless assignment to local variable

**fixed alerts:**

* 1 for Comparison between inconvertible types

---

AureliaEffect:
This pull request **introduces 14 alerts** and **fixes 4** when merging 66a81522b1fcc2681a87e4ce867d2d775f389118 into 57c07e8ffac77dfd6cc55529d972c55016f316e8 - [view on LGTM.com](https://lgtm.com/projects/g/aurelia/aurelia/rev/pr-1068d5e053042029088d6a848ef815ab4ea0e031)

**new alerts:**

* 13 for Unused variable, import, function or class
* 1 for Useless assignment to local variable

**fixed alerts:**

* 1 for Comparison between inconvertible types

---

AureliaEffect:
This pull request **introduces 2 alerts** and **fixes 4** when merging de58b952e103d6edbf2deb28f068dfc4251faf76 into 57c07e8ffac77dfd6cc55529d972c55016f316e8 - [view on LGTM.com](https://lgtm.com/projects/g/aurelia/aurelia/rev/pr-59d8faed03ecfd3401f862720a5450de9c6973af)

**new alerts:**

**fixed alerts:**

* 1 for Comparison between inconvertible types

---

AureliaEffect:
This pull request **introduces 1 alert** and **fixes 5** when merging f3da8300a540c7d1426360cf459c812c21031951 into 57c07e8ffac77dfd6cc55529d972c55016f316e8 - [view on LGTM.com](https://lgtm.com/projects/g/aurelia/aurelia/rev/pr-6261b0fd5276aa5c52d7401cdc3e49fa56cf377c)

**new alerts:**

**fixed alerts:**

* 1 for Comparison between inconvertible types

---

fkleuver:
@EisenbergEffect @bigopon The initial implementation for semantic model is done, so this is ready for review.

packages/jit/src/templating/template-compiler.ts
```diff
@@ -37,196 +32,87 @@ import {
...
-    const isTemplate = node.nodeName === 'TEMPLATE';
-
-    // Parent node is required for remove / replace operation incase node is the direct child of document fragment
-    const parentNode = node = isTemplate ? (<HTMLTemplateElement>node).content : node;
-
-    while (node = this.compileNode(node, parentNode, definition, definition.instructions, resources)) { /* Do nothing */ }
+  public compile(definition: ITemplateSource, resources: IResourceDescriptions, flags?: ViewCompileFlags): TemplateDefinition {
+    let $el = new SemanticModel(this.attrParser, this.elParser.parse(definition.templateOrNode), resources).getElementSymbolRoot(definition);
+    const $root = $el;
+    while ($el = this.compileNode($el.$content || $el));
```
bigopon: Will this make the template compiler ignore the `<template/>` element itself? If it's already handled elsewhere, there may need to have some comment describing the flow in high level overview.

packages/jit/src/templating/template-compiler.ts
```diff
@@ -375,27 +242,21 @@ export class TemplateCompiler implements ITemplateCompiler {
...
   /*@internal*/
-  public compileAttribute(
-    target: string,
-    value: string,
-    node: Node,
-    attribute: AttributeDefinition,
-    element: ElementDefinition,
-    command: IBindingCommand | null): TargetedInstruction {
+  public compileAttribute($attr: IAttributeSymbol): TargetedInstruction {
       // binding commands get priority over all; they may override default behaviors
```
bigopon: @EisenbergEffect @fkleuver Comment like this is so nice. Maybe we should start making a small document outlining the template compiler behavior

packages/jit/src/templating/template-compiler.ts
```diff
@@ -406,41 +267,48 @@ export class TemplateCompiler implements ITemplateCompiler {
       //   return new StylePropertyBindingInstruction(expression, target);
       // }
       // plain custom attribute on any kind of element
-      if (attribute) {
-        const resolved = resolveTarget(target, element, attribute);
-        value = value && value.length ? value : '""';
-        const expression = parser.parse(value, BindingType.Interpolation) || parser.parse(value, BindingType.ToViewCommand);
-        switch (resolved.mode) {
+      if ($attr.isCustomAttribute) {
+        let expression = parser.parse($attr.rawValue, BindingType.Interpolation);
```
bigopon: For later, we may want to invalidate this usage: `attr.bind="inter${polation}"`

packages/jit/src/templating/element-parser.ts
```diff
@@ -0,0 +1,85 @@
...
+      const nodeLen = nodeChildNodes.length;
+      children = Array(nodeLen);
+      for (let i = 0, ii = nodeLen; i < ii; ++i) {
+        children[i] = this.parse(nodeChildNodes[i]);
+      }
+    }
+
+    let attributes: AttrSyntax[];
+    const nodeAttributes = node.attributes;
+    if (nodeAttributes) {
```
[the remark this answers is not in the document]
fkleuver: No, this check is necessary because it could also be a `Text` or `DocumentFragment` instance. `attributes` will be undefined then. That said, the length check would be a good addition to help save some memory

packages/jit/src/templating/template-compiler.ts
fkleuver: Yep good catch. I mixed this up with the parent logic. Will fix

packages/jit/src/templating/template-compiler.ts
fkleuver: Yeah but any documentation we write now will have to be maintained as we go forward as well. This already bit me in the behind with the CI/CD workflow. There is still too big a feature gap imo for this to be worthwhile right now

packages/jit/src/templating/template-compiler.ts
fkleuver: This is already invalid. A plain attribute with `.bind` will be caught by a binding command at line 253, which parses the expression with `BindingType.BindCommand`. An interpolation will throw then.

bigopon: I see, it seems that you already cover the `.isCustomAttribute` inside `compileElementNode`, maybe no need to have that check here again? The logic doesn't seem to be straightforward to me as it seems to be checking `is custom attribute` twice, and the check in `compileAttribute` is redundant?

AureliaEffect:
This pull request **introduces 1 alert** and **fixes 5** when merging 7b8c77949dd34d177456e0b87d912852d6dc16ab into b296ed359f8c9c2f84c41e5e9e0a6b2c73927c9c - [view on LGTM.com](https://lgtm.com/projects/g/aurelia/aurelia/rev/pr-9d9d127d2e0f789abe652599dd1f86e7a6266358)

**new alerts:**

**fixed alerts:**

* 1 for Comparison between inconvertible types

---

packages/jit/src/templating/template-compiler.ts
fkleuver: The logic wrapped inside `isCustomAttribute` must not be executed when `compileElementNode` passes it a plain attribute. It's called from multiple code paths, not all of them do this check

packages/jit/src/templating/template-compiler.ts
fkleuver: Hm, shouldn't this be handled by `compileSurrogate?`

bigopon: The compile surrogate wont be appropriate for nested template? Normally `<template>` is only used together with one or many template controllers

fkleuver: I don't believe this worked in the pre-refactored version of the template compiler either (they weren't ignored like they are now, but they threw regardless). This is why I've intentionally been avoiding putting template controllers on template elements in the unit tests up to this point. Simply changing it to be similar to pre-refactored version gives those same errors so I'm going to have to look for this special "x" that's missing somewhere. Not sure if I'll be able to solve that today :(

packages/jit/src/templating/template-compiler.ts
bigopon: It's not as simple as I thought. Nice flow

EisenbergEffect:
@fkleuver I didn't go over every detail, but I like the general direction this is all going in. I'm fine to merge this work. Let me know if that's ok or if you wanted to make any other updates before that.

fkleuver:
@EisenbergEffect It started bothering me, I'm trying to fix it. It's quite tricky, constantly fixing one edge case breaks another. I'm going to cat-and-mouse this a little more :)

fkleuver:
@EisenbergEffect Okay, go ahead and merge this. I won't be able to fix it before the nightly probably. I'll continue in a new branch, probably better anyway

PR #190, opened 2018-09-25
Cleanup

fkleuver:
Removes the TaskQueue class and TaskQueue-specific utilities, and tidy up a few things along the way
@EisenbergEffect


AureliaEffect:
This pull request **fixes 1 alert** when merging 7897860e9d5e0a4e55176b471df22b73d53104c1 into 57c07e8ffac77dfd6cc55529d972c55016f316e8 - [view on LGTM.com](https://lgtm.com/projects/g/aurelia/aurelia/rev/pr-a3a6453ebcf7893d23c3a2e85689e3d309700df6)

**fixed alerts:**

---

packages/debug/src/reporter.ts
```diff
@@ -102,10 +102,6 @@ const codeLookup: Record<string, IMessageInfo> = {
     type: MessageType.error,
     message: 'Signal name is required.'
   },
-  13: {
```
EisenbergEffect: What should we do about error code gaps? Just make a practice of looking for them to fill them in before adding new codes?

fkleuver: I personally think we should have a convention of error code ranges tied to specific types or errors / areas of the code base, similar to http status codes. That way we'll naturally have some gaps, and  similar errors will be grouped together, I think that would make things easier to manage and maintain

EisenbergEffect: That's a fantastic idea! I love it.

Issue #164, opened 2018-09-13
[WIP] Route Configuration API Proposal [RFC] 


davismj:
- [x] `IRouteConfig` interface
- [x] Basic routes
- [x] Canonical urls
- [x] Aliased routes
- [x] Custom viewports
- [x] Child routes
- [x] Absolute paths
- [ ] Case sensitive routes
- [x] Parameterized routes
- [x] Custom parameterizers
- [ ] Navigation model
- [x] Route metadata

Please let me know if there are any *use cases* that are missing from this specification.

# The Route Configuration API

Route configuration can be done one of three ways: Up-front (no definition yet), using the `@route` decorator, or by defining a `static routes` property. For clarity, I've included the most terse definition with the decorator approach and the most verbose definition with the static routes approach for each example. This will demonstrate what they will look like in a typical development case as well as an explicit definition what they imply. I've also commented out implicit or conventional values that will be set by Aurelia under the hood.

## `IRouteConfig`

The names of properties are up for grabs. I have chosen names that I believe are most illuminating and most standard.

```javascript
interface IRouteConfig {

  // A string or array of strings of matching route paths.
  // (Renamed from route in vCurrent)
  path: string | string[],

  // A constructor function for the view-model to attach for this route. 
  // For route decorator and static routes approaches, Aurelia will set this
  // value under the hood. 
  // (Repurposed from moduleId in vCurrent)
  component?: IComponent,

  // A uniquely identifiable name for the route, for canonical navigation.
  // For route decorator and static routes approaches, Aurelia will try to
  // set this value by convention if not specified explicitly.
  name?: string,

  // Optional, the name of the viewport to attach the controller to. If not
  // specified, the default viewport will be used.
  viewport?: string,

  // Optional, the name of the parent route or routes, matched by the `name` property.
  parent?: string | string[],

  // Optional, flag to opt the route out of the navigation model. Defaults 
  // to true.
  nav?: boolean,

  // Optional, an object with additional information available to the 
  // view-model throughout the activation lifecycle. 
  // (Renamed from settings in vCurrent)
  meta?: any
}
```

## A basic route

Defining a basic route is dead simple.

```javascript
@route('home')
export class MyViewModel {
  static routes = [
    { 
      path: 'home',
      //, name: 'my'
      //, component: this
    }
  ];
}
```

## Canonical urls

Multiple routes can redirect to the same route, creating a canonical url for that page.

```javascript
@route('home')
@route({ path: '', redirect: 'home' })
[...]

PR #193, opened 2018-09-25
WIP: router

fkleuver:
@EisenbergEffect @bigopon @davismj

This is just about the most quick-and-dirty implementation I could come up with to make some simple button navigation work.

### IRoute
I based the `IRoute` interface on @davismj's description in #164, so I did give some extra attention to the validation of that config. I think a more defensive approach to configuration would help users catch errors earlier and prevent them from surfacing in convoluted ways.

### RouterView
The `RouterView` class is loosely based on the `Compose` element and by no means representative of what I have in mind for it in the end. For one, it should probably be a custom attribute or a template controller, rather than a custom element. This is just step 1 to get something working. "move along people" :)

### Lifecycle hooks
I'm thinking the lifecycle/pipeline could go in the same direction as attach/detach lifecycle (with a linked list etc). The `$behavior` concept would be an interesting extensibility point for additional lifecycle hooks. Combined with some potentially custom ordering/precedence and the ability to specify which hooks to invoke via flags, this could give us more flexibility than vCurrent with at the same time a much simpler underlying mechanism.

Custom lifecycle hooks could also be registered via decorators. Multiple ways to do the same thing, really.

The way I currently implemented these is definitely not how I have it in mind for how it should be. Again, just step 1, "move along people" :)

### Resource definition
We don't want routes to conflict with the definition of custom elements. Should there be a separate resources API for routes or should they just not be resources at all?

### IRouter 
The `IRouter` interface is probably a bit more interesting. As as starting point I thought to make it a combination of the `History` and `Location` interfaces (staying as close to native api's as possible) with the addition of an API to register routes and router views. The `History` and `Location` API's are currently not much more than pass-through methods but these would get the appropriate logic eventually.

### DIFF
--- a/packages/router/src/history.ts
+++ b/packages/router/src/history.ts
@@ -1,0 +1,31 @@
+import { DI, IIndexable } from '@aurelia/kernel';
+
+// https://html.spec.whatwg.org/multipage/history.html#the-history-interface
+export interface IHistory {
+  back(): void;
+  forward(): void;
+  go(delta?: number): void;
+  pushState(data: IIndexable, title?: string, url?: string | null): void;
+  replaceState(data: IIndexable, title?: string, url?: string | null): void;
+}
+
+export const IHistory = DI.createInterface<IHistory>().withDefault(x => x.singleton(HtmlHistory));
+
+/*@internal*/
+export class HtmlHistory implements IHistory {
+  public back(): void {
+    history.back();
+  }
+  public forward(): void {
+    history.forward();
+  }
+  public go(delta?: number): void {
+    history.go(delta);
+  }
+  public pushState(data: IIndexable, title?: string, url?: string | null): void {
+    history.pushState(data, title, url);
+  }
+  public replaceState(data: IIndexable, title?: string, url?: string | null): void {
+    history.replaceState(data, title, url);
+  }
+}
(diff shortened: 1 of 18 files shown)


davismj:
I'm going to prioritize the other specs then. I didn't think we were close to an implementation.

AureliaEffect:
This pull request **introduces 4 alerts** when merging 481a468c07596c500fff4b73e8855e0da81813c6 into b296ed359f8c9c2f84c41e5e9e0a6b2c73927c9c - [view on LGTM.com](https://lgtm.com/projects/g/aurelia/aurelia/rev/pr-31d64492a5a252b0855addfec994c771011079dc)

**new alerts:**

---

bigopon:
✅ 

> We don't want routes to conflict with the definition of custom elements. Should there be a separate resources API for routes or should they just not be resources at all?

Maybe the we can employ static convention and make it even simpler, since we only need to care about route definition after all?

StrahilKazlachev:
Is there a way to determine a *route* or the *view model* of a route dynamically? Like with [`navigationStrategy`](https://aurelia.io/docs/routing/configuration#dynamically-specify-route-components) in the current router.

fkleuver:
There will be. Just let me be absolutely clear here: this isn't even 5% of the implementation. Right now I'm personally just looking to be able to navigate between view models even if everything must be hard coded, so we can do a little more meaningful testing.

fkleuver:
It was too early for me to open this PR, i'm closing this for now so it doesn't keep drawing attention

PR #196, opened 2018-09-27
template compiler tests

fkleuver:
@bigopon @EisenbergEffect 

I'm not sure in which scenario the flush is actually needed (is there any demo or tests to prove the situation?) because in this integration test you can see the value propagating all the way from the app to the 5th level nested custom element on startup, without a single flush..

### DIFF
--- a/packages/jit/src/templating/semantic-model.ts
+++ b/packages/jit/src/templating/semantic-model.ts
@@ -397,7 +397,7 @@ export class ElementSymbol {
     switch (this.name) {
       case 'TEMPLATE':
         this._isTemplate = true;
-        this._$content = this.semanticModel.getTemplateElementSymbol(syntax.$content, this, definition, this);
+        this._$content = this.semanticModel.getElementSymbol(syntax.$content, this);
         break;
       case 'SLOT':
         this._isSlot = true;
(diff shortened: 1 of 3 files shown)


fkleuver:
@bigopon now it does :p see latest commit

bigopon:
Nice, though the actual error was from `Observer.callback` being noop instead of identity :D

fkleuver:
`identity` instead of `noop` makes sense.. sort of. I'm just really curious how exactly you managed to get an error from that. That callback should only do anything if you actually have a `xxxChanged` method in your component that needs to listen to those changes. Can you demonstrate your error?

bigopon:
Here https://github.com/aurelia/aurelia/blob/beec33ffed1a75258360046ef88c55db8bdc7cb9/packages/runtime/src/binding/property-observation.ts#L122-L126 is where i got that issue: With `noop`, it's always undefined and the value was never set properly during startup

fkleuver:
> is where i got that issue: With noop, it's always undefined and the value was never set properly during startup

The part `if (!(flags & BindingFlags.fromBind)) {` makes sure the callback never gets invoked during bind.
Furthermore, the check: `if (coercedValue !== undefined) {` will prevent the value from ever being *adjusted* when the callback is either `noop`, or you don't return anything from your own `handleChange` callback. The only time you would get an issue is if you intentionally try to set the value to `undefined` from your callback by returning undefined. Then that would be ignored..

bigopon:
Yes, that's what I meant by commenting that. Only realized that after looking at it again. I'm pretty sure I always got undefined during bound, now I can no longer reproduce it

EisenbergEffect:
Definitely a great kind of test to have. Ready to merge?

fkleuver:
Yeah it's good to go. There's a tiny fix in there too. Will clean up this test and add more similar ones later :)

PR #199, opened 2018-09-27
Templating tests

fkleuver:
In this PR:
- Some basic "happy path" unit tests for `createElement`.
- Automatically mark a definition with `build: { required: true,.. }` if `templateOrNode` is a string
- A utility method for getting all property descriptors in a prototype chain (not utilized right now, but I've been meaning to add this a couple times and keep finding myself rewriting it every time I want to try something with bulk testing element attributes)


fkleuver:
Merging this into the other PR

PR #202, opened 2018-09-28
fix(DI): alias registration param order and tests

### LINKED ISSUE #191: Fix alias registrations
@EisenbergEffect Just so it doesn't get forgotten, there are still a few disabled DI integration tests related to aliases: https://github.com/aurelia/aurelia/blob/master/packages/kernel/test/unit/di.integration.spec.ts#L101
The test will hang due to an infinite loop. I looked into it briefly but couldn't determine the root cause. I can look into it later again, unless you can see at a glance what's wrong

EisenbergEffect:
There was a small bug in the alias registration fluent API for InterfaceSymbol and a few other minor oversights. Additionally, there were a number of errors in the tests themselves. This PR corrects all these issues and gets the full test set working with no more ignored tests.

Fixes #191

### DIFF
--- a/packages/kernel/src/di.ts
+++ b/packages/kernel/src/di.ts
@@ -129,21 +129,22 @@ export const DI = {
       };
 
       Key.register = function(container: IContainer, key?: Key<T>): IResolver<T> {
+        const trueKey = key || Key;
         return configure({
           instance(value: T): IResolver {
-            return container.registerResolver(Key, new Resolver(key || Key, ResolverStrategy.instance, value));
+            return container.registerResolver(trueKey, new Resolver(trueKey, ResolverStrategy.instance, value));
           },
           singleton(value: Function): IResolver {
-            return container.registerResolver(Key, new Resolver(key || Key, ResolverStrategy.singleton, value));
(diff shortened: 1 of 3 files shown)


packages/kernel/src/di.ts
```diff
@@ -129,21 +129,22 @@ export const DI = {
...
-            return container.registerResolver(Key, new Resolver(key || Key, ResolverStrategy.transient, value));
+            return container.registerResolver(trueKey, new Resolver(trueKey, ResolverStrategy.transient, value));
           },
           callback(value: ResolveCallback): IResolver {
-            return container.registerResolver(Key, new Resolver(key || Key, ResolverStrategy.callback, value));
+            return container.registerResolver(trueKey, new Resolver(trueKey, ResolverStrategy.callback, value));
           },
           aliasTo(destinationKey: T): IResolver {
-            return container.registerResolver(destinationKey, new Resolver(key || Key, ResolverStrategy.alias, Key));
+            return container.registerResolver(trueKey, new Resolver(trueKey, ResolverStrategy.alias, destinationKey));
```
fkleuver: Wow, can't believe I missed *that* xd

PR #204, opened 2018-10-02
expression parser tests and fixes

### LINKED ISSUE #81: Review and add full test coverage for expression parsing
I need to add some notes about which precise pieces of the spec we are implementing and where we deviate. We need to ensure full test coverage for every possible edge case. This can be done by simply cross-checking with the spec: https://tc39.github.io/ecma262.
We're almost there already, but I know some things are missing. A good example is the various restrictions on member expressions (currently object and array literals are not allowed on the left-hand side, which is already fixed in vCurrent).
Also need to think about (and document) the various rules that would apply to semicolon-separated custom attribute expressions, interpolations, restrictions on `repeat.for`, and so forth.

fkleuver:
Closes #81 

- Big increase in test density - almost every thinkable combination is tried.
- The tests revealed 3 or 4 rather sneaky bugs which are now fixed
- Updated the Unparser in Debug
- Added `IVisitor` interface and implemented `accept` methods on the AST (going to try and move some logic from the AST into visitors where it makes sense, and make the AST a bit more like what an AST is meant to be: data)

@bigopon @EisenbergEffect


fkleuver:
@EisenbergEffect @bigopon As a POC I've now moved the `connect` and `evaluate` logic into visitors. The means to instantiate/get a reference is not really the point atm.

I think this visitor approach has certain benefits, e.g. you can have different AST implementations at runtime (or just different settings) without having to touch or monkey patch the AST. The AST stays small. One visitor contains all logic for a certain process, rather than it being scattered around multiple classes. You could even define entirely new processes, layer them, have an async version in there, etc

What do you think about this idea of decoupling the logic from the AST?

bigopon:
The idea is beautiful, but I'm concerned by the fact that the 2 extra layers on top of the hottest heart of Aurelia like this will have big perf implication. I think Visitor could be decoupled in a one time way, at the parsing time to teach AST how to connect, evaluate and assign etc

fkleuver:
Yeah that's an option. I'm of course also thinking about AOT here.
Keep in mind though that while it's an additional method call, it's also 2-3 fewer arguments per call. And there's less for the browser JIT to inline this way, so better optimizations may kick in. We'd have to bench it to know for sure though.

EisenbergEffect:
Same thoughts from me. Design-wise, its nice. I'd want to feel that we're good on perf though.

bigopon:
For the AST, I think only connect makes sense to have it done this way, the rest need to stay as strictly close as possible to normal JS expression.

fkleuver:
Yeah I'm going to mark this little experiment as failed for now. There was a ~10% perf drop in dbmonster. I dumped the visitors in a separate branch for potentially revisiting at a later point in time (no pun intended).
I'll keep this pr scoped to the tests and parser fixes :)
Ready to merge if all is good

fkleuver:
> I'll keep this pr scoped to the tests and parser fixes
I lied. There's always more scope :p

PR #205, opened 2018-10-03
typings and api docs

fkleuver:
This PR contains for the most part improvements to the precision of types, and some preliminary API documentation for what tends to be the most misunderstood part of the framework: the lifecycle hooks.

### DIFF
--- a/packages/runtime/src/dom.ts
+++ b/packages/runtime/src/dom.ts
@@ -1,4 +1,5 @@
 import { DI, IContainer, IResolver, PLATFORM } from '@aurelia/kernel';
+import { ICustomElement } from '.';
 
 export interface INodeLike {
   readonly firstChild: INode | null;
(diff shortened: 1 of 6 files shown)


packages/runtime/src/dom.ts
```diff
@@ -1,4 +1,5 @@
 import { DI, IContainer, IResolver, PLATFORM } from '@aurelia/kernel';
+import { ICustomElement } from '.';
```
bigopon: Will importing like this have potential issue?

packages/runtime/src/templating/lifecycle.ts
```diff
@@ -15,6 +16,211 @@ export interface IAttach {
...
 }
 
+export interface IElementTemplateProvider {
+  getElementTemplate(renderingEngine: IRenderingEngine, customElementType: ICustomElementType): ITemplate;
+}
+
+/**
+ * Defines optional lifecycle hooks that will be called only when they are implemented.
+ */
+export interface ILifecycleHooks extends Partial<Omit<IRenderable, '$addNodes' | '$removeNodes'>> {
```
bigopon: Nice comments 👍

bigopon:
lgtm. I'd love to see more like this. Thanks @fkleuver

PR #207, opened 2018-10-06
expression parser #2

### LINKED ISSUE #200: Runtime ExpressionParser caching issues
@bigopon discovered a problem today where trying to parse an interpolation expression will actually return the cached value of a previously parsed non-interpolation expression.

I didn't think the caching through very well at the time I changed it, nor tested it very properly, so after looking at it a bit more closely I think this mechanism is due for a rework. I'm just putting my thoughts out here to give you a heads up on what I'll be working on soonish (I may or may not do some more templating tests first)

I'm considering tackling this together with improving the coverage on the renderer and making targeted instruction types numeric again (using an array as the backing store) so they can more easily be inserted (addressing #182), swapped, and have some cleverness with bitwise ops applied to them when appropriate*.

There would be 3 or so lookups in the expression parser, one for each set of parsing rules (interpolation, iterator statement and normal expressions)

The instruction types would be split up again per binding type / command, each having a unique number. The numbers of the instruction types can then, via bitwise operators, be translated to one out of these cache indices.

With all this in mind there are some const enums that need to be cleaned up. The `BindingType` and `ExpressionKind` enums have become an incoherent mess and I'd like to try to just get rid of them. Same goes for some of those direct prototype assignments on instruction classes and AST nodes.

Naturally this would also touch the AST and expression parser and probably break a few thousand tests, so those need to be tidied up as well. So that's a fairly big chunk of work upcoming, but I think things will be much cleaner and easier to understand when it's done.

fkleuver:
Fixes #200

### DIFF
--- a/packages/runtime/src/binding/ast.ts
+++ b/packages/runtime/src/binding/ast.ts
@@ -926,27 +926,31 @@ export class ForOfStatement implements IExpression {
 export class Interpolation implements IExpression {
   public $kind: ExpressionKind;
   public assign: IExpression['assign'];
+  public readonly isMulti: boolean;
+  public readonly firstExpression: IExpression;
   constructor(
     public readonly parts: ReadonlyArray<string>,
-    public readonly expressions: ReadonlyArray<IExpression>) { }
+    public readonly expressions: ReadonlyArray<IExpression>) {
+      this.isMulti = expressions.length > 1;
+      this.firstExpression = expressions[0];
+    }
 
   public evaluate(flags: BindingFlags, scope: IScope, locator: IServiceLocator): string {
-    const expressions = this.expressions;
-    const parts = this.parts;
-    let result = parts[0];
-    for (let i = 0, ii = expressions.length; i < ii; ++i) {
-      result += expressions[i].evaluate(flags, scope, locator);
-      result += parts[i + 1];
-    }
-    return result;
-  }
-
-  public connect(flags: BindingFlags, scope: IScope, binding: IConnectableBinding): void {
-    const expressions = this.expressions;
-    for (let i = 0, ii = expressions.length; i < ii; ++i) {
-      expressions[i].connect(flags, scope, binding);
+    if (this.isMulti) {
+      const expressions = this.expressions;
+      const parts = this.parts;
+      let result = parts[0];
+      for (let i = 0, ii = expressions.length; i < ii; ++i) {
+        result += expressions[i].evaluate(flags, scope, locator);
+        result += parts[i + 1];
+      }
+      return result;
+    } else {
+      const parts = this.parts;
+      return parts[0] + this.firstExpression.evaluate(flags, scope, locator) + parts[1];
     }
   }
+  public connect(flags: BindingFlags, scope: IScope, binding: IConnectableBinding): void { }
 
   public accept<T>(visitor: IVisitor<T>): T {
     return visitor.visitInterpolation(this);
--- a/packages/runtime/src/binding/ast.ts
+++ b/packages/runtime/src/binding/ast.ts
@@ -6,6 +6,7 @@ import { IBinding } from './binding';
 import { BindingBehaviorResource } from './binding-behavior';
 import { BindingContext, IScope } from './binding-context';
 import { BindingFlags } from './binding-flags';
+import { IConnectableBinding } from './connectable';
 import { Collection } from './observation';
 import { ISignaler } from './signaler';
 import { ValueConverterResource } from './value-converter';
@@ -52,7 +53,7 @@ export interface IVisitor<T = any> {
 export interface IExpression {
   readonly $kind: ExpressionKind;
   evaluate(flags: BindingFlags, scope: IScope, locator: IServiceLocator | null): any;
-  connect(flags: BindingFlags, scope: IScope, binding: IBinding): any;
+  connect(flags: BindingFlags, scope: IScope, binding: IConnectableBinding): any;
   accept<T>(visitor: IVisitor<T>): T;
   assign?(flags: BindingFlags, scope: IScope, locator: IServiceLocator | null, value: any): any;
   bind?(flags: BindingFlags, scope: IScope, binding: IBinding): void;
@@ -122,7 +123,7 @@ export class BindingBehavior implements IExpression {
     return (<any>this.expression).assign(flags, scope, locator, value);
   }
 
-  public connect(flags: BindingFlags, scope: IScope, binding: IBinding): void {
+  public connect(flags: BindingFlags, scope: IScope, binding: IConnectableBinding): void {
     this.expression.connect(flags, scope, binding);
   }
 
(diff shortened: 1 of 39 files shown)


AureliaEffect:
This pull request **introduces 1 alert** and **fixes 2** when merging b0fd2c32d14f7e09ca9013468e1aa9f2ac850854 into a0c9b0ecd41c556f9ea20f409f94810c8eb5a3de - [view on LGTM.com](https://lgtm.com/projects/g/aurelia/aurelia/rev/pr-f04d9fd8ffcadadc7ec4d3981e0df2147ed074f1)

**new alerts:**

**fixed alerts:**

---

AureliaEffect:
This pull request **introduces 1 alert** and **fixes 2** when merging 77aabc77d70e86b3645b949a18ae96259b0dc88a into a0c9b0ecd41c556f9ea20f409f94810c8eb5a3de - [view on LGTM.com](https://lgtm.com/projects/g/aurelia/aurelia/rev/pr-4e5555bbbb3de3b95bac2275c018eafe3637d8dc)

**new alerts:**

**fixed alerts:**

---

AureliaEffect:
This pull request **introduces 1 alert** and **fixes 2** when merging 38760bbeef8c21739134c0bed045663373a947db into a0c9b0ecd41c556f9ea20f409f94810c8eb5a3de - [view on LGTM.com](https://lgtm.com/projects/g/aurelia/aurelia/rev/pr-b6dd0aae120bc716f77be3059226958fd5a8235e)

**new alerts:**

**fixed alerts:**

---

AureliaEffect:
This pull request **introduces 2 alerts** and **fixes 4** when merging 3c6627c38662641c4210ef553ff34888d473efcb into a0c9b0ecd41c556f9ea20f409f94810c8eb5a3de - [view on LGTM.com](https://lgtm.com/projects/g/aurelia/aurelia/rev/pr-d0dd6180744b0f598a8b38d3444f76c919714c65)

**new alerts:**

**fixed alerts:**

---

AureliaEffect:
This pull request **introduces 2 alerts** and **fixes 4** when merging 4565244b6d181154b5ad481c8a2c57a5c9f1b320 into a0c9b0ecd41c556f9ea20f409f94810c8eb5a3de - [view on LGTM.com](https://lgtm.com/projects/g/aurelia/aurelia/rev/pr-6569010298ac6c13611e794752ccaaab710e5b00)

**new alerts:**

**fixed alerts:**

---

AureliaEffect:
This pull request **introduces 2 alerts** and **fixes 4** when merging 4b0e45f2d9f5e1d5a87aadddcec49846297816ce into a0c9b0ecd41c556f9ea20f409f94810c8eb5a3de - [view on LGTM.com](https://lgtm.com/projects/g/aurelia/aurelia/rev/pr-6047a1c14c35af70d9d78eff859fa08b4b86ac1f)

**new alerts:**

**fixed alerts:**

---

fkleuver:
@EisenbergEffect @bigopon This piece is ready for review/merge. In summary:

### ExpressionParser
- ForOfStatement and Interpolation are now also fully tested, bringing the parser to 100% coverage (with a test density that gives confidence)
- The caching issue with the runtime expression parser is fixed. There are now 3 separate lookups corresponding to the 3 main BindingTypes: Interpolation, ForOfCommand, and the rest. These caches can no longer return parsed expressions for the wrong type.
- parse/parseCore return types are strongly typed to return the correct set of possible expressions depending on bindingType passed in

### @decoratable
- Re-introduced the `@connectable` decorator now that there is more than one binding type that can use it. Refactored `Binding` and `LetBinding` to use the decorator.
- Introducated `Decoratable` and `Decorated` interfaces in the kernel that make it easier to write strongly typed decorators. `@connectable` uses these and provides two overloads (invocation and non-invocation), both of which will type-check the class it is applied to and give a compile time type error if the class doesn't implement the correct interface. Should apply this to other decorators as well.

### Interpolation
- Implemented the `InterpolationBinding` based on the one in `templating-binding` from vCurrent, also uses `@connectable`. A few differences in implementation (imo it's cleaner) that makes especially the parsing part more performant (uses the proper parser and caches the results). The runtime aspect should also be slightly faster (not verified) because it has less work to do.

(This is a pretty important piece that was missing.)

packages/runtime/src/binding/ast.ts
```diff
@@ -923,29 +924,33 @@ export class ForOfStatement implements IExpression {
...
+        result += expressions[i].evaluate(flags, scope, locator);
+        result += parts[i + 1];
+      }
+      return result;
+    } else {
+      const parts = this.parts;
+      return parts[0] + this.firstExpression.evaluate(flags, scope, locator) + parts[1];
     }
   }
+  public connect(flags: BindingFlags, scope: IScope, binding: IConnectableBinding): void { }
```
bigopon: Is this intentionally left out?

fkleuver: Yeah, it's just a noop like some other AST classes have that cant be / dont make sense to connect. It's just there to satisfy the interface. I also don't really see a point in throwing. It's just kind of there. Maybe the interface needs to change or something

bigopon: I was confused at first, but it seems connecting job is delegated to the expression part, so this is fine. Probably in a polishing final touch, we may want to comment the flow how a interpolation is translated and observed

bigopon:
@fkleuver what is the benefit of having two dedicated classes to handle the interpolation vs the existing way? They look similar to me.

fkleuver:
@bigopon They only look similar to the existing way because I've simplified/normalized some of the logic from vCurrent and offloaded some responsibility to the `Interpolation` expression type. There is still a key difference in that this mechanism allows both a single and multiple interpolations to effectively be treated as a single binding from the renderer/view's perspective.

The `MultiInterpolationBinding` corresponds to vCurrent's [InterpolationBinding](https://github.com/aurelia/templating-binding/blob/master/src/interpolation-binding-expression.js#L49) and the `InterpolationBinding` corresponds to [ChildInterpolationBinding](https://github.com/aurelia/templating-binding/blob/master/src/interpolation-binding-expression.js#L116)

The existing mechanism would break (not necessarily error, but unpredictable / wrong behavior) with more than one interpolation per binding with value converts and/or binding behaviors.
The other thing (albeit small) is performance. Interpolation binding can never be two-way so it needs a bit less stuff on it.

bigopon:
I was under the impression that `Binding` + `Template` ast could replace the need for dedicated interpolation bindings. Maybe I was wrong. I'll have a look and do a follow up PR with more examples around this. Other than that, lgtm.

fkleuver:
Perhaps it can, but the subtle differences would create a need for a few extra checks. The expressions of a template (by their parsing rules) cannot contain value converters or binding behaviors. Theoretically we could introduce this and make template behave the same way as interpolation, but that wouldn't remove the need for InterpolationBinding (in fact it would create the need for a recursive one)

There are more things we could cram into single classes but that wouldn't necessarily benefit library size more than it would negatively impact performance/maintainability.

PR #208, opened 2018-10-06
refactor(SelectObserver): adjust SelectValueObserver synchronization

bigopon:
Added some "spec" for select value observer as a base for discussion, it's based on vCurrent behavior.
Moving some tests around since the element-observation spec file got busy, and is expected to be busier.


packages/runtime/src/binding/element-observation.ts
```diff
@@ -373,37 +377,87 @@ export class SelectValueObserver implements SelectValueObserver {
...
+    //      b. If is an array:
+    //        i. gather all current selected <option/>, in to array called values
+    //        ii. loop through the currentValues array and remove items that are nolonger selected based on matcher
+    //        iii. loop through the values array and add items that are selected based on matcher
+    //        iv. Return false to signal value hasn't changed
+    // B. If the select is single
+    //    1. Let value equal the first selected option, if no option selected, then value is null
+    //    2. assign currentValue to oldValue
+    //    3. assign value to currentValue
+    //    4. return true to signal value has changed
```
fkleuver: This is great! 🥇

bigopon:
@EisenbergEffect @fkleuver 
I think this fix / refactor could be called a day. There are many more unit & integration tests needed but I think it's good for review / merge for now.

About spec in observer behavior, I quite like it but found it hard to apply anywhere else. Codes are quite often self explanatory. The parser would be a nice target but it probably not too much different from JS grammar spec so maybe not need for it either. Thoughts?

PR #215, opened 2018-10-09
chore(all): use custom script for generating changelog instead of lerna

fkleuver:
It looks like the convention only allows one feature, fix or perf improvement per commit if it is to be included in the changelog, even if it's a squash merge commit.

What this means is that the repo admin @EisenbergEffect would need to tweak the merge commit messages by hand before merging them in, to ensure some meaningful information ends up in the change log. Then still PRs can't be too big since all the extra details will be crammed under one single feature or fix header.

I'm a bit underwhelmed by the flexibility of this and I don't any benefit in strictly adhering to conventional-commits, at least at this stage of development.

I wrote my own little script that accounts for our non-compliance to conventional commits and includes all information found in commit messages, not just the headers. It also includes refactorings because I believe that's nice information to have in the changelogs as well. 
It doesn't have all the extra floof in it such as automatically removing reverts, including specially formatted sections/links for issues closed and such, but that's partly on us for not consistently including all the necessary information in commits.

We should try a bit harder to start complying to [conventional commits](https://www.conventionalcommits.org/) so that we can perhaps eventually remove this custom script again. But until we've figured out the precise workflow to make that happen nicely with our squash workflow, this custom script will at least ensure the changelogs include everything.

@EisenbergEffect with this script in place you should now also be able to release v0.3.0 without any fuss


AureliaEffect:
This pull request **introduces 4 alerts** when merging f9c6760d8bfd799fb18a2eeb0bb20b4a3d8f3fd0 into 1e804a022226965fdbb99a2c3962c05576956e69 - [view on LGTM.com](https://lgtm.com/projects/g/aurelia/aurelia/rev/pr-c7494b538bd3c1a2b936008fb57bac744f080c0d)

**new alerts:**

---

huochunpeng:
I don't think the limitation of conventional-commits is necessarily bad thing. It enforces you to write atomic commit, also means you cannot squash and merge.
I would suggest to have a conventional-changelog (maybe the underneath standard-changelog) fork to handle this task, instead of adding logic to this repo itself. The fork may be usable to other people too. Call it non-conventional-changelog 😄

fkleuver:
> I don't think the limitation of conventional-commits is necessarily bad thing.

That depends.

> It enforces you to write atomic commit

A good thing indeed

> also means you cannot squash and merge.

I believe squash merge is a legitimate approach that should be compatible with best practices in software development, but I could be wrong.

Do you think we should perhaps stop doing squash merge and do normal merge instead?

> I would suggest to have a conventional-changelog (maybe the underneath standard-changelog) fork to handle this task, instead of adding logic to this repo itself. The fork may be usable to other people too. Call it non-conventional-changelog

For the couple of lines of code that it ultimately boils down to, deliberately bypassing all the fancy templating and tracing that they have built in, I don't think my little snippet would really fit suit that repo. I've hard-coded it to a subset of a convention that I think would work for us, for now.

Furthermore I cloned their repo locally and tried a few things with it, but it's layers upon layers of (heavily backwards compatible) vanilla JS. I'm incapable of working productively with that kind of code base.

For now, this snippet will allow us to release 0.3.0 with a properly discovered changelog and leak a bit of non-convention into 0.4.0 as well, giving us some time to figure out how to best solve this for the long term. That's my take on it at least

huochunpeng:
> Do you think we should perhaps stop doing squash merge and do normal merge instead?
No matter what's the opinion on merge, I would rather not fight the tooling (changelog) 😄

fkleuver:
@EisenbergEffect do you have anything against doing normal (non-squash) merges to retain the full history in the master branch?

EisenbergEffect:
No issues on my part. I think there's a GitHub setting that even will make the UI default to the preferred merge. Shall we shift our approach?

fkleuver:
Yeah. I agree with @huochunpeng's stance on this. Rather not fight the tooling. Keep things as simple as possible. We need this little workaround for the changes up to this point, but we can try to do it properly from now on.

fkleuver:
@EisenbergEffect what shall we do with this? my suggestion would be to use it as a one-off for v0.3.0 and just leave it afterwards, both scripts exist and it could be useful for fixing some changelog oopses later on

EisenbergEffect:
That sounds fine. I'll go ahead and merge it in.

Issue #78, opened 2018-08-19
Rework ObserverLocator


fkleuver:
ObserverLocator has some room for improvement in clarity and correctness as well as in performance.

My idea is to use `Object.prototype.toString.call(obj)` on the passed-in object and have an object lookup with things like this:
```
{
    ['[object Array]']: createArrayObserver,
    ['[object HTMLInputElement]']: createElementObserver,
}
```

We'd need to make a practical tradeoff between the size of this lookup (including all possible element constructors is probably too much) and the performance overhead of having a more generalized solution (for example stripping off everything between "HTML" and "Element" in a charAt loop)

In any case we can probably get rid of the regexes without adding too much code.

fkleuver:
@EisenbergEffect I'm closing this as it was already sort-of addressed with the runtime separations, and things have partly taken a different direction since, so this issue doesn't make a whole lot of sense right now

Issue #75, opened 2018-08-19
Make binding and templating more intelligent with information from to the w3/whatwg specs


fkleuver:
Just some notes:

- Defaults/conventions for input element attributes: 
    - https://www.w3.org/TR/html52/sec-forms.html#common-input-element-attributes
- Look into `document.createRange` as an alternative for / supplement to using document fragments
- Look into edge cases for content editable:
    - https://www.w3.org/TR/html52/editing.html#making-document-regions-editable-the-contenteditable-content-attribute
- Look into making HTML commands first-class citizens:
    - https://www.w3.org/TR/html52/interactive-elements.html#commands
- Look into making HTML element activation as a concept a first-class citizen (activationEvents are `change`, `click`, `contextmenu`, `dblclick`, `mouseup`, `pointerup`, `reset`, `submit`, `touchend`): 
    - https://www.w3.org/TR/html52/editing.html#activation-behavior
    - https://html.spec.whatwg.org/#activation
    - https://dom.spec.whatwg.org/#eventtarget-activation-behavior
- Review event handling:
    - https://dom.spec.whatwg.org/#concept-event-dispatch
    - https://developer.mozilla.org/en-US/docs/Web/API/EventTarget
    - https://w3c.github.io/input-events/index.html#event-definitions
- Make templating / two-way bindings aware of (and more intelligent about) input modes (`none`, `text`, `tel`, `url`, `email`, `numeric`, `decimal`, `search`):
    - https://html.spec.whatwg.org/#input-modalities:-the-inputmode-attribute

fkleuver:
In response to https://github.com/aurelia/binding/issues/347

@EisenbergEffect 
I'm thinking about a specialized `InputObserver` instead of the generic `ValueAttributeObserver` to handle `input` elements, since they have so many special behaviors and edge cases.
- There are about 25 different `type` attributes (making up 15 different types of values that an input may have).
- In addition there are about 25 content attributes that affect the constraints of these values
- And a total of 9 attributes whose value can be changed from the UI (meaning they can be two-way bound)

It's not just the `input` element (although that is by far the biggest one). Several other elements also have behaviors that vCurrent isn't explicitly dealing with.
Building all of the behavior in is sure to take up quite some library space so we need to make sure it can be tree-shaken, which means a loosely-coupled configuration class that can be injected into observers at startup might be a good way to go.

[...]

PR #216, opened 2018-10-10
refactor(binding-context): explicitly model scope, context and observer lookup

fkleuver:
## 📖 Description

This PR should lay the groundwork for more intelligent features, optimizations, and increase the precision of typings by explicitly defining classes for `Scope`, `BindingContext`, `OverrideContext` and `ObserversLookup`.

The key thing is the ability to tell the difference between a "synthetic" binding context (as indicated by the `$synthetic` property) and a regular binding context (e.g. a user view model). By checking this property, we can take certain shortcuts or do additional things that would otherwise be unsafe to do. Especially template controllers such as the repeater will benefit from this, once utilized properly.

All that aside, classes perform slightly better in terms of memory than regular objects, so memory consumption will go down with this.

### 🎫 Issues

Loosely related to anything that has to do with observation and lifecycles, and specifically:
- https://github.com/aurelia/aurelia/issues/78

This is the non-DOM equivalent of: https://github.com/aurelia/aurelia/issues/75

## 👩‍💻 Reviewer Notes

The APIs to create scopes and override contexts have moved to static methods on `Scope` resp. `OverrideContext`:

- `BindingContext.createOverride` -> `OverrideContext.create`
- `BindingContext.createScope` -> `Scope.create`
- `BindingContext.createScopeFromOverride` -> `Scope.fromOverride`
- `BindingContext.createScopeFromParent` -> `Scope.fromParent`

BindingContext can be created via `BindingContext.create`. You can pass in a single property + value (as used by the repeater), an existing object (which will clone it) or nothing (which just creates the equivalent of an empty object)

One important change is that the `parentOverrideContext` parameter is no longer optional: you should explicitly pass in null. I believe this helps with preventing accidentally not including the parent context where it should be (I think this is the case with `@customElement`, marked as a todo there)

Furthermore, the `ObserversLookup` provides a shortcut for the observer locator to just directly return a `SetterObserver` since that is the only thing you'll ever want when observing a synthetic binding context.

## 📑 Test Plan

This affects areas that were already heavily covered by tests. Fixing these tests consisted mostly of a few simple find/replaces, and explicitly passing in nulls. All tests then passed again, suggesting that this should not cause any regressions.

## ⏭ Next Steps

This leaves a few potential loose ends wrt observer-locator, `SetterObserver`, the AST and some other pieces that should now be easier to optimize. I may or may not visit those pieces on the short term. I don't want to make any too drastic changes to binding anymore in a short amount of time. No concrete plans at the moment but a few rough ideas.

### DIFF
--- a/packages/runtime/src/binding/binding-context.ts
+++ b/packages/runtime/src/binding/binding-context.ts
@@ -1,14 +1,6 @@
-import { Reporter } from "@aurelia/kernel";
-
-export interface IOverrideContext {
-  parentOverrideContext: IOverrideContext;
-  bindingContext: any;
-}
-
-export interface IScope {
-  bindingContext: any;
-  overrideContext: IOverrideContext;
-}
+import { IIndexable, Reporter } from '@aurelia/kernel';
+import { StrictPrimitive } from './ast';
+import { BindingSourceObserver, IBindScope, ObservedCollection } from './observation';
 
 const enum RuntimeError {
   UndefinedScope = 250, // trying to evaluate on something that's not a valid binding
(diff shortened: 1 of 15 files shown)


AureliaEffect:
This pull request **introduces 2 alerts** when merging 00cb17f405e6c45c0b3b76821cfee8dc18c13525 into 72671654f43951788cfc9252c70861e7c874dce8 - [view on LGTM.com](https://lgtm.com/projects/g/aurelia/aurelia/rev/pr-0a48fddc7725fc617daef52b9601defadc207bc1)

**new alerts:**

---

packages/runtime/src/binding/binding-context.ts
```diff
@@ -92,4 +119,60 @@ export const BindingContext = {
...
+  ) { }
+
+  public static create(bc: IBindingContext | IBindScope, poc: IOverrideContext | null): OverrideContext {
+    return new OverrideContext(bc, poc === undefined ? null : poc);
+  }
+
+  public getObservers(): ObserversLookup<IOverrideContext> {
+    let observers = this.$observers;
+    if (observers === undefined) {
+      this.$observers = observers = new InternalObserversLookup();
```
bigopon: This may be fine, but I'd prefer to have all `$observers` non enumerable.

packages/runtime/src/binding/binding-context.ts
```diff
@@ -17,46 +10,81 @@ const enum RuntimeError {
...
-    if (overrideContext === null || overrideContext === undefined) {
-      throw Reporter.error(RuntimeError.NilOverrideContext);
+}
+
+export type ObserversLookup<TObj extends IIndexable = IIndexable, TKey extends keyof TObj =
+  Exclude<keyof TObj, '$synthetic' | '$observers' | 'bindingContext' | 'overrideContext' | 'parentOverrideContext'>> =
+  { [P in TKey]: PropertyObserver; } & { getOrCreate(obj: IBindingContext | IOverrideContext, key: string): PropertyObserver };
+
+/*@internal*/
+export class InternalObserversLookup {
```
bigopon: I think this is nice. With instance of class normally is smaller than POJO (maybe completely different between JS engines), this could be nice in some extreme scenarios. Though we probably need a link to benchmark to properly mention that. Or in our engineering doc maybe?

packages/runtime/src/binding/observer-locator.ts
```diff
@@ -56,7 +57,10 @@ export class ObserverLocator implements IObserverLocator {
     private svgAnalyzer: ISVGAnalyzer
   ) {}
 
-  public getObserver(obj: IObservable, propertyName: string): AccessorOrObserver {
+  public getObserver(obj: IObservable | IBindingContext | IOverrideContext, propertyName: string): AccessorOrObserver {
+    if (obj.$synthetic === true) {
```
bigopon: Not against this but the more we have random property names prefixed by `$`, the more uncomfortable I feel. Doc should have very intensive section for these properties at least. @EisenbergEffect @fkleuver

packages/runtime/src/binding/subscriber-collection.ts
```diff
@@ -46,7 +45,7 @@ function addSubscriber<T extends MutationKind>(this: ISubscriberCollection<T>, s
     this._subscriberFlags |= SubscriberFlags.SubscribersRest;
     return true;
   }
-  nativePush.call(this._subscribersRest, subscriber);
+  this._subscribersRest.push(subscriber);
```
bigopon: Is this revert because of perf or readability. If it's readability, I think we have already sacrificed many parts for perf (need clear benchmarks though), this particular hot path should be a good one to keep.

packages/runtime/test/unit/mock.ts
```diff
@@ -357,8 +358,8 @@ export class MockTemplate implements ITemplate {
 
 export class MockTextNodeTemplate {
   constructor(
-    public sourceExpression: IExpression,
-    public observerLocator: IObserverLocator
+    public sourceExpression: any,
```
bigopon: 😢

fkleuver: It's a mock, I think mocks shouldn't have really any restrictions per se on what you pass in. Wouldnt you agree?

packages/runtime/src/binding/subscriber-collection.ts
fkleuver: It's actually to remove a circular dependency that the JIT tests started failing because of. And I just figured we could/should do these kinds of optimizations (like decorator inlining) from AOT when it's there. Maybe the prototype observables should be in a separate file as an alternative means to remove the circular dep. For now this was the easy way to do it

packages/runtime/src/binding/observer-locator.ts
fkleuver: Yeah it's just there to reduce possibility of clash. I also thought about prefixing things with `au` or `au$` instead, which could be a bit safer and make it even clearer that its an internal property

packages/runtime/src/binding/binding-context.ts
```diff
@@ -92,4 +119,60 @@ export const BindingContext = {
     // the name wasn't found.  return the root binding context.
     return scope.bindingContext || scope.overrideContext;
   }
-};
+
+  public getObservers(): ObserversLookup<IOverrideContext> {
+    let observers = this.$observers;
+    if (observers === undefined) {
+      this.$observers = observers = new InternalObserversLookup() as ObserversLookup<this>;
```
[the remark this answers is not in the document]
fkleuver: Users generally wouldn't enumerate over a synthetic context. Normal properties perform a little better than non enumerable + non writable ones, so I figured it could be okay here. Are there special reasons to have it otherwise? I think if this is non enumerable then the `$synthetic` properly should be too.. and i'm not a huge fan of making both non enumerable for perf reasons..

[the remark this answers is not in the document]
bigopon: I see. I missed this. Nice 👍

PR #223, opened 2018-10-11
tests and templating fixes

fkleuver:
# Pull Request

### test-suite
The main piece of this PR is [`test-suite.ts`](https://github.com/aurelia/aurelia/pull/223/files#diff-b8d644c2dd4f1e290e3be38e23dea9b6). It's motivated by some of the ways I've been applying `eachCartesianJoinFactory`, namely to combine multiple data points as well as multiple test actions to test different logical flows in a single test.

Example: 

```ts
class Something {
  constructor(public a, public b) {}
  public returnA() { return this.a += 'a'; }
  public returnB() { return this.returnA() + (this.b += 'b'); }
}

const suite = new TestSuite<string, string, Something>('something');

// data slots a-o are strongly typed, order of adding them does not matter
// the data is assigned to the ctx by property name
// a value is a single static value shared across all contexts
suite.addDataSlot('a')
  .addData('1').setValue('foo')
  .addData('2').setValue('bar');

// a factory will be invoked for each combination and thus guarantees a new value per test
// for each slot the values from previous slots will be available
suite.addDataSlot('b')
  .addData('1').setFactory(ctx => ctx.a + 'baz')
  .addData('2').setFactory(ctx => ctx.a + 'qux');

// technically you can put an action in a data factory and vice versa
// but the results of actions are not assigned to any property
suite.addActionSlot('setup')
  .addAction(null, ctx => {
    expect(ctx.a.length).to.equal(3);
    expect(ctx.b.length).to.equal(6);
    expect(ctx.a).to.equal(ctx.b.slice(0, 3));
    ctx.c = new Something(ctx.a, ctx.b);
  });

// actions are executed in the order that they're added, the strings are just cosmetic for generating
// the titles of the tests
suite.addActionSlot('act')
  .addAction('returnA', ctx => {
    const { c: sut } = ctx;
    expect(sut.a).to.equal(ctx.a);
    expect(sut.b).to.equal(ctx.b);
    expect(sut.returnA()).to.equal(ctx.a + 'a');
    expect(sut.a).to.equal(ctx.a + 'a');
    expect(sut.b).to.equal(ctx.b);
  })
  .addAction('returnB', ctx => {
    const { c: sut } = ctx;
    expect(sut.a).to.equal(ctx.a);
    expect(sut.b).to.equal(ctx.b);
    expect(sut.returnB()).to.equal(ctx.a + 'a' + ctx.b + 'b');
    expect(sut.a).to.equal(ctx.a + 'a');
    expect(sut.b).to.equal(ctx.b + 'b');
  });

// calling load() loops through all combinations and generates the contexts
suite.load();
// run() runs them, runOnly() calls describe.only, etc
suite.runOnly();
```

Output:
image

You can also build a base suite, clone it, and then reuse it for multiple derived suites. I was experimenting with that [here](https://github.com/aurelia/aurelia/blob/8c98cd0ce6e14deeaf7e6c52219781ce6763fdd5/packages/jit/test/integration/template-compiler.repeater-if-else.spec.ts)

The other place I applied it is [here](https://github.com/aurelia/aurelia/blob/8c98cd0ce6e14deeaf7e6c52219781ce6763fdd5/packages/jit/test/integration/template-compiler.repeater.spec.ts)

A few additional tests are added and the jit tests are reorganized a bit.

### if/else and lifecycles

The ChangeSet is now passed into lifecycle controllers and they will call `flushChanges()` first thing inside `processAll()`, to ensure any queued state changes in binding are processed before the final attach/detach lifecycles are. This may need some tweaks. Need to test more advanced scenarios to see how this plays out when flushChanges itself causes stuff to be added to the lifecycle. Generally it shouldn't be needed often anyway.

The `If` template controller queues itself to the ChangeSet when it receives an update via `valueChanged`. This makes it consistent with how the repeater works and should cause changes to be processed in the correct order in most cases, as well as prevent some redundant work. Again, this will need more tests to make sure.

## 👩‍💻 Reviewer Notes

My intent was to make the tests themselves (as well as writing them) more approachable for people who haven't been spending as much time on it as I have. I'd like to see if @EisenbergEffect, @bigopon would actually use the test-suite implementation.

## ⏭ Next Steps

One notable optimization is disabled as discussed with @EisenbergEffect, which is the behavior to only remove root nodes. The plan is to somehow get that behavior back in again while keeping things working.

Furthermore, more work could be done to make the test-suite api cleaner but I'd first like to try and clean up some tests first and see how things go. Probably going to focus a bit more on runtime tests next.


EisenbergEffect:
Haven't had a chance to review the actual code yet, but looking at the example in the description, I think this is a nice improvement on readability.

bigopon:
The setup is very nice by itself, though I'd argue it introduces noise / obscurity on reading the real test / test intention. I think this should be used together with traditional way to increase test density and expose edge cases, not a main way to communicate our assertion.

AureliaEffect:
This pull request **introduces 1 alert** when merging 4cb15a909988e0f3b5b609ff93dc199da8f772e1 into 6ba681c5c0255796f55855f2e64b3adcc4c494c3 - [view on LGTM.com](https://lgtm.com/projects/g/aurelia/aurelia/rev/pr-a30657ec182a0025e6d23751669b26934b29600b)

**new alerts:**

---

AureliaEffect:
This pull request **introduces 1 alert** when merging 0222b629fc8084ace390c05239d2084a5781455f into 6ba681c5c0255796f55855f2e64b3adcc4c494c3 - [view on LGTM.com](https://lgtm.com/projects/g/aurelia/aurelia/rev/pr-d7f96a35753dcc8732d93d445910de2b1e3873d7)

**new alerts:**

---

AureliaEffect:
This pull request **introduces 1 alert** and **fixes 2** when merging 0e66eb3e6576527256890b535cea662044d74d72 into fccbeb679530defb409fc7750ee814e1811765a8 - [view on LGTM.com](https://lgtm.com/projects/g/aurelia/aurelia/rev/pr-04b8c5ab1643fa9d1d71169349913de8822dccd9)

**new alerts:**

**fixed alerts:**

---

fkleuver:
Latest update is perhaps a little [unorthodox](https://github.com/aurelia/aurelia/pull/223/commits/0e66eb3e6576527256890b535cea662044d74d72#diff-0bbea5c3a3b006ddb3de803b5136bd53R262), but it addresses a problem that's caused by the renderer and doesn't have a straightforward and clean solution imo (at least not one that would perform well)
Problem is the `au-marker` node that gets removed by the renderer stays in the `FragmentNodeSequence` as `firstChild`. Since it's removed, its `nextSibling` is null and thus the following text node doesn't get removed.
I could have gone for a different approach of looking a little further through the childNodes inside the fragment and pick up from the next item, or simply replace that firstNode in the fragment directly. I don't think any of it is particularly clean.

So I created a specialized `TextNodeSequence` specifically for this (pretty common) scenario of an `au-marker` followed by an empty textnode. The marker is discarded and the nodeSequence itself is simply an object with the `text` node assigned to it as `nextSibling`. As long as the renderer doesn't try to do things with the marker other than `nextSibling` and `remove`, and consumers of the nodeSequence don't violate the contract, it should just work.

I wanted to do something like this already for textnodes for performance reasons, but it turns out that it can also solve a problem while we're at it :)

bigopon:
This sounds like a nice approach to me. But i'm unsure why `nextSibling` would be null, since, theoretically, `<au-marker/>` should always be followed by an empty textnode ?

fkleuver:
Line 108 in the renderer: `DOM.remove(target);`
At that point the textNode will sit under the DocumentFragment and the target (`<au-marker>`) has a null parent and nextSibling.  So you could say just don't remove the marker, but why keep a marker around when it serves no purpose anymore. That's just a perf hit for no benefit

AureliaEffect:
This pull request **introduces 1 alert** and **fixes 2** when merging 7d87ad1fa0aff253f2e7917354a383fa4a67ed87 into fccbeb679530defb409fc7750ee814e1811765a8 - [view on LGTM.com](https://lgtm.com/projects/g/aurelia/aurelia/rev/pr-7f084401ae3256aaf536dd1e7f12fc833f6f5703)

**new alerts:**

**fixed alerts:**

---

AureliaEffect:
This pull request **introduces 1 alert** and **fixes 2** when merging e91be57197926dde47e79e8fa455ec43e0084538 into fccbeb679530defb409fc7750ee814e1811765a8 - [view on LGTM.com](https://lgtm.com/projects/g/aurelia/aurelia/rev/pr-acc8f618681cb4d279949c9008777991a3563a43)

**new alerts:**

**fixed alerts:**

---

fkleuver:
@EisenbergEffect It's ready for review. A few fixes allowed me to uncomment some tests, but there are still a couple of things left to do.

Note going forward:
We should probably have a chat about the general testing approach, so far I've neglected usability in favor of coverage/density (it's just a trade-off), but I'm getting worried that I might be locking everyone in to something that's too hard to work with. I need to hear that if it's the case because I can't always judge it that well.

AureliaEffect:
This pull request **introduces 1 alert** and **fixes 2** when merging 72b8905d54a2aa79ca39765d1a1945f36b05bf72 into fccbeb679530defb409fc7750ee814e1811765a8 - [view on LGTM.com](https://lgtm.com/projects/g/aurelia/aurelia/rev/pr-1b3b952d77128442ae19e30e3909c9dfdd32e55f)

**new alerts:**

**fixed alerts:**

---

PR #231, opened 2018-10-15
Renames

fkleuver:
@EisenbergEffect

# Pull Request

### Rename `src` to `def`
I think the term `Source` (and, likewise, `src`) is a little ambiguous here. Source as in a URI? an import statement? source code? some other kind of reference? `Definition` simply says: it *defines* a template. That definition may very well be (or contain) some source, if you want to put it that way, but it is unambiguously the thing to consult to be able to construct the template it represents. 

Since we already have the type `TemplateDefinition`, this change would make that relation clearer as well. A `TemplateDefinition` is simply a completed and immutable instance of `ITemplateDefinition`

### Rename interface `ICustomAttributeSource` to `IAttributeDefinition`
For similar reasons as `ITemplateSource`. Leaving out the "custom" part because there is no model of a non-custom attribute in Aurelia. It is implied. And changing from Source to Definition makes it consistent with `ITemplateDefinition` as well as with `AttributeDefinition`

### Rename `templateOrNode` to `template` (reverts a change I introduced)
A `Node` may just as well be considered a template (for a view/resource etc). The fact that it can be a raw string or a parsed html node is an implementation detail.

### Rename `srcOrExpr` to `from` (modifies a change I introduced)
I first renamed this from `src` to `srcOrExpr` to convey that it could be a raw string or a parsed AST node. Like with the `templateOrNode`, this is really just an implementation detail. The rename from `src` to `from` and from `dest` to `to` is to put in place a more explicit correlation between the source and target of a binding while keeping the names short.

## 👩‍💻 Reviewer Notes

I don't feel strongly about any of these and have no issues reverting/changing any of it if preferable. These are just some suggestions to get the ball rolling on this

## ⏭ Next Steps

Let me know if you merge either this PR or #223 before merging the other one. Even if there are no merge conflicts, I'd still like to go over it by hand to make sure loose ends are tied up


fkleuver:
@EisenbergEffect This is ready for review. I have some more renames in mind on some APIs but that would turn into a merging hell, so it's better to get the current PRs in first before I do any other work on this topic.

EisenbergEffect:
@fkleuver Looks like this needs to be updated now that I merged your other PR.

fkleuver:
@EisenbergEffect it's fixed

EisenbergEffect:
I felt that the name changes helped the readability, particularly to/from. The change to template is good for dev ergonomics. So, great changes across the board here.

bigopon:
Everything feels quite natural correction, except for `src` to `def`, I think align with vnext terminologies better.

fkleuver:
> Everything feels quite natural correction, except for `src` to `def`

I feel the same about `src` to `def` honestly. But it felt like it made more sense since the thing that's in there is a definition now, not a source.

>  I think it aligns with vnext terminologies better.

You mean `def` right?

I wouldn't mind making some adjustments even if most if it is good, maybe a small part isn't. Feel free to nit.

bigopon:
Yes. `def` goes along with the rest definitely better than src

PR #234, opened 2018-10-17
Templating refactor

fkleuver:
# Pull Request

## 📖 Description

- Refactor out all (or at least most) of the direct dependencies on `lib.dom.d.ts` in prep for non-browser / nodejs runtime readiness
- Tweaks and typing improvements to most templating decorators
- Small increase to decorator (lifecycle hooks) and compose test coverage
- Rename projector `onElementRemoved` to `take` ("take" as in the opposite of "project".. sort of)
- Rename `$addNodes` to `$mount`, `$removeNodes` to `$unmount`
- Rename (previous) `mount` to `hold` ("hold" as in the opposite of "release" w.r.t RenderLocation)
- Refactor `ChangeSet` into `LinkedChangeList` (better performance and working towards better consistency and co-ordination between lifecycles)
- Replace the various state properties `$isBound`, `$isAttached` etc with a single `LifecycleState` enum flags property
- Replace the `RuntimeBehavior` properties with a single `LifecycleHooks` enum flags property
- Centralize `TemplateDefinition` creation into `buildTemplateDefinition` with more thorough checks and (hopefully) clearer preconditions
- Split up `IViewFactory` into `IViewFactory` and `IViewCache` interfaces (where the former now only has `create()`), to reduce leakiness of the abstraction. `View` now takes `IViewCache` instead
- Move `createNodeSequence` into a factory class to (hopefully) make the node creation flow easier to digest
- Rename `PotentialRenderable` to `RenderPlan`
- Fix customAttribute alias key registration
- Fix customElement cache value (use provided value)
- Add missing exports (`LetBinding` and some others)
- Remove all (except the top-level) `index.ts` files to reduce library size and lower the chance of accidentally introducing circular references
- Add missing `BindingFlags` bitwise ORs to various `$bind` and `$unbind` methods to ensure the flags are always correct
- Make the `instructions` and `build` properties optional on `@customElement`
- Implement `BindLifecycle` to enforce proper call order for `bound` and `unbound` in the same manner as `attached` and `detached`, but as a single const object that uses a depth counter to determine when to end the lifecycle. I went with this to lower the "buy-in" (not much else had to change) and I don't consider it in its final state yet. Still requires integration with attach/detach and change-set
- **major** restructuring of the interfaces and enums. This significantly reduces the number of files you need to import from when trying to accomplish certain tasks, and overall gives you less places to look at, and hopefully makes the structure more logical. This pertains to the following files, where I've found that they're typically imported together anyway:
  - `definitions.ts` - Consolidated from `templating/instructions.ts` and the descriptions/definitions from `templating/bindable.ts`, `templating/custom-attribute.ts` and `templating/custom-element.ts`. 
  - `lifecycle.ts` - Consolidated from `templating/lifecycle.ts` and lifecycle-describing interfaces from various files in binding and templating
  - `observation.ts` - Consolidated from `binding/observation.ts`, `binding/change-set.ts` and the `binding-context.ts` interfaces
- And a medium-sized restructuring of the runtime tests (mostly separated out into unit/integration tests, first step clean up)

### 🎫 Issues

Surprisingly, this closes no open issues.. :)
These were mostly things discussed on Discord, as well as found "en route" towards a rework of the lifecycles (which is not completed yet)

## 👩‍💻 Reviewer Notes

Pay special attention to the keyword "rename" (in the above notes as well as commit messages). Some of these were ad hoc decisions I made in the flow of refactoring, without discussing them, and they may not be to everyone's liking. They are easy enough to find/replace (this is why I did it without discussing first), so do speak up if they're not OK.

In particular I'm referring to: `mount` -> `hold` and `onElementChanged` -> `take`. It was tricky to find good names for these.

## 📑 Test Plan

I have for the most part been relying on existing test coverage, but I did add a number of customElement/customAttribute tests and duplicated the tests for `LinkedChangeList`. Almost not a single change in logic has gone by without breaking a bunch of tests, so they seemed to have done their job.
Other changes did not cause any tests to fail which surprised me a little:
- a bug in LinkedChangeList went unnoticed (apparently the code base doesnt cause any changes to be queued recursively, which is how I designed it with `BindingFlags.fromFlushChanges`, but we do need some integration tests to simulate users causing recursion)
- The implementation of `BindLifecycle` did not cause any tests to fail, which isn't necessarily a reason for worry but it does mean that more complex scenarios need to be tested where this callback order actually matters, and makes tests fail if it is *not* in place

## ⏭ Next Steps

- Figure out a way to properly wire up the `flushChanges`, `attach`/`detach` and `bind/unbind` queues, re-introduce a `connect-queue` for binding and ensure it is properly timed with relation to `mount`/`unmount`. In other words, the real lifecycle overhaul has yet to begin.
- Find the cleanest way to ensure stability of RenderLocations when only the top-level node is removed on `$unmount`, and enable this optimization
- Tests, tests, tests


AureliaEffect:
This pull request **introduces 66 alerts** and **fixes 3** when merging 92068eea5b8e3411d68c0a72dbc22a167310ac77 into 0ef3e8ec4508eb2f6855d0610da471ccf91d5413 - [view on LGTM.com](https://lgtm.com/projects/g/aurelia/aurelia/rev/pr-7af16840c85bb0c46c0f6d6457bc90643e16df91)

**new alerts:**

* 33 for Superfluous trailing arguments
* 33 for Illegal invocation

**fixed alerts:**

---

EisenbergEffect:
The main thing I'm not clear on is why there's a need to change `DOM` in this way. It seems as if this would cause use to lose the ability to replace the DOM methods with alternate implementations, which might be needed by other platforms or environments.

packages/runtime/src/binding/event-manager.ts
```diff
@@ -81,14 +81,14 @@ export class ListenerTracker {
   public increment(): void {
     this.count++;
     if (this.count === 1) {
-      DOM.addEventListener(this.eventName, this.listener, null, this.capture);
+      addEventListener(this.eventName, this.listener, null, this.capture);
```
bigopon: I quite like it. Kind of good way to opt for bundle size 😬

packages/runtime/src/dom.ts
```diff
@@ -364,3 +410,30 @@ export class FragmentNodeSequence implements INodeSequence {
...
+  public readonly nodeName: 'AU-MARKER';
+  public readonly nodeType: typeof ELEMENT_NODE;
+  public textContent: string = '';
+
+  constructor(next: INode) {
+    this.nextSibling = next;
+  }
+  public remove(): void { /* do nothing */ }
+}
+(<Writable<AuMarker>>AuMarker.prototype).previousSibling = null;
```
bigopon:
```suggestion
(auMarkerProto => {
  ...
})(AuMarker.prototype);
```

bigopon:
I like the interfaces refactoring work. It makes things very nice. For Rob's concern, maybe there probably no way to avoid `pal` - `pal-browser` ... `pal-....` abstraction.

fkleuver:
I think he's referring to the fact that I'm exporting the methods on `DOM` it. I did this primarily for performance reasons and bundle size. Granted, it's not going to work in its current shape. As @EisenbergEffect said this won't allow for methods to be replaced.
OTOH I think we shouldn't replace methods anyway. Monkey patching is still (and will be for some time to come) a performance killer.

fkleuver:
@bigopon @EisenbergEffect what do you think about the readability and contributer-friendliness of these tests? https://github.com/aurelia/aurelia/pull/234/commits/b499495d9634136b89ca9d1c3cebb8cf3e4299c7#diff-d06056c1e0c28176be428f7703c02472R1
I'd like an honest and nit-picky opinion because I'm looking for a formula for generated tests that is still easy to understand and follow (if at all possible) that I'll be applying going forward. So it's not a question of whether it's good enough for this PR, but whether it's good enough for just about all the remainder of the tests (and changing some existing ones into).
I would say "not really" myself, but I'm also not even sure whether this is better than fully inlined loops. There is simply way too much text. But I'd like to know what you think about this.

AureliaEffect:
This pull request **introduces 66 alerts** and **fixes 3** when merging b499495d9634136b89ca9d1c3cebb8cf3e4299c7 into 092834b3919bbcf6e74e02fa50cfa39cb1eb379b - [view on LGTM.com](https://lgtm.com/projects/g/aurelia/aurelia/rev/pr-7727dfaca197221c76698735ffe8763cfa717ec6)

**new alerts:**

* 33 for Superfluous trailing arguments
* 33 for Illegal invocation

**fixed alerts:**

---

packages/runtime/src/dom.ts
fkleuver: good idea, i adjusted it

AureliaEffect:
This pull request **introduces 66 alerts** and **fixes 3** when merging cfacba4837569dceff2058d6e1f0aef386bb2c80 into 092834b3919bbcf6e74e02fa50cfa39cb1eb379b - [view on LGTM.com](https://lgtm.com/projects/g/aurelia/aurelia/rev/pr-9a68723f00fd8065506cd8702b1175e9fc49dcc1)

**new alerts:**

* 33 for Superfluous trailing arguments
* 33 for Illegal invocation

**fixed alerts:**

---

AureliaEffect:
This pull request **fixes 3 alerts** when merging b608223f8c3c28099a9841ad403a55edf1ba53eb into 092834b3919bbcf6e74e02fa50cfa39cb1eb379b - [view on LGTM.com](https://lgtm.com/projects/g/aurelia/aurelia/rev/pr-0baab3561c90bc83c2591428915dc74cfa0f8ab2)

**fixed alerts:**

---

AureliaEffect:
This pull request **fixes 3 alerts** when merging 7156d18589c69ad1a609e0b51bd5569d47d8136c into 092834b3919bbcf6e74e02fa50cfa39cb1eb379b - [view on LGTM.com](https://lgtm.com/projects/g/aurelia/aurelia/rev/pr-6435fcfa329efd69defab90151f4ff26fd291b49)

**fixed alerts:**

---

packages/runtime/src/templating/definition-builder.ts
```diff
@@ -0,0 +1,152 @@
...
+  instructions?: ReadonlyArray<ReadonlyArray<TargetedInstruction>> | null,
+  dependencies?: ReadonlyArray<unknown> | null,
+  surrogates?: ReadonlyArray<TargetedInstruction> | null,
+  containerless?: boolean | null,
+  shadowOptions?: { mode: 'open' | 'closed' } | null,
+  hasSlots?: boolean | null): TemplateDefinition {
+
+  const def = new DefaultTemplateDefinition();
+
+  // all cases fall through intentionally
```
bigopon: Perf in this block will never matter, so I highly doubt we want switch / fall instead of multiple if blocks.

bigopon: It's clever use of switch behavior, though I must say, but I hope we don't have to employ this? Plus I though maximum number of parameters supported is 10?

fkleuver: My main goal here was actually to minimize the amount of code / duplication while having the full proper defaults and fallbacks on everything. Organizing the fallthrough like this automatically ensures the correct definition precedence. It even comes at a slight performance loss compared to existing decorators because some properties are checked / written redundantly. But as you said, the perf difference on that won't ever be measurable let alone noticeable.

EisenbergEffect:
@fkleuver This looks to be coming along well. How are you feeling about things?

fkleuver:
@EisenbergEffect Alright, so two parts:

### Lifecycles
Regarding lifecycles, I'm trying to make everything  as natural and deliberate as possible, not resorting to things like "flush before attach". Things should just be in the correct state at the end of a particular lifecycle. So I'm thinking about a single lifecycle manager to handle binding as well as templating. Either `change-set` would get a different name and take on the responsibilities of the lifecycle controllers, or vice versa, giving us more granular control over when to batch changes, in which order to do things and e,g, when to delay and/or skip property descriptor creation.

Having `$attach`/`$detach`, `$mount`/`$unmount` and `hold`/`release` feels like one lifecycle too much. I can't really put my finger on what `$attach` is meant to be when it's `$mount` that really adds the nodes. mount and attach have slightly different meanings, but it doesn't feel to me like we're letting them behave that way. And the `needsMount`, while I understand its purpose, feels like a redundancy. It should mean the same thing as "not attached", but it doesn't for some reason. It's like we're really close to something that make sense, but not quite there yet.

### HTML
 I was planning to pull out HTML-specific stuff but I'm having difficulties deciding where to draw the line:
- `binding/element-observation.ts`: no-brainer, this is HTML
- `binding/target-accessors.ts`: all classes except property-accessor are HTML
- `binding/svg-analyzer.ts`: kind of HTML, but the implementation already lives elsewhere, so not really. Maybe put that implementation on the same pile with the rest of HTML and get rid of svg-analyzer package?
- `binding/event-manager.ts`: seems mostly html-specific, but should the event-manager as a whole move or should just the event path handling and event configuration go? What kind of events do pixi / webgl have?
- `binding/resources/attr-binding-behavior.ts`: definitely HTML
- `binding/resources/update-trigger-binding-behavior.ts`: quacks like HTML due to the dependency on event-manager, but is it really HTML or is it just a tight coupling we should get rid of?
- `binding/resources/sanitize.ts`: seems like HTML but has no real DOM dependencies, so could go either way
- `dom.ts`: make an injectable HTML dependency, and only leave the interface? Will introduce DI into components that currently go without DI, may lead to over-engineering. Perhaps use the `DOM.instance` singleton approach similar to vCurrent `Container.instance`? Or put it on the `Aurelia` instance on `start`, taking the document from the passed-in `host` (seems to make most sense to me, but then `Aurelia` becomes a global app dependency)?

fkleuver:
Seems like starting to think of mount/unmount as a lifecycle is where I went wrong then. It's not a lifecycle, but rather as an action in the attach/detach lifecycle (much like observeProperty is an action in the bind lifecycle). And it can go in queue, much like connect used to. Perhaps thinking of mounting and connecting as the binding/templating analogues would add some coherence to the whole. Does that make sense?

EisenbergEffect:
Yes. Absolutely. I think this is a good parallel.

fkleuver:
I'm thinking about a `mount-queue` and a `connect-queue`, where `connect` is forced to happen after `mount`. So that would mean separating mounting out into a little process of its own that's not directly embedded in the lifecycle controller. The attach process mainly just figures out what needs to be mounted, and queues it up (so the views and elements don't have to do this themselves). Does that make sense? Maybe I just need to play a little with that idea

EisenbergEffect:
There might need to be a bit more nuance. I'm not sure. Is "connect queue" the same as in vCurrent? I just want to make sure we consistently bind before attach. Connecting up listeners could happen later as along as the initial pull of the values happens at the appropriate time.

fkleuver:
There is no connect queue right now, so it's open for being shaped however it needs to be shaped. I'm aware of the nuances, I was just summarizing it. It will take some experimenting to see what works best. We can `RAF` or `setTimeout` the connect queue and pull updates right before connect happens, but then the DOM is already attached so there needs to be another pull before attach. That theoretically two redundant pulls, so a few flag and state checks may help with reducing that overhead.

fkleuver:
Yeah, it's not a bind queue :)

fkleuver:
@EisenbergEffect @bigopon it's actually ready for review now. I have summarized the changes as best I could in the opening comment.

bigopon:
From the summary, all sound fine. I will have a look in details soon.
> For which belongs in core vs. HTML-specific package, it seems like some of the biggest questions relate to events.

About the above, I think I will at least draw some diagram for us to analyze the most loose joints inside runtime, to do it properly.

fkleuver:
> for us to analyze the most loose joints inside runtime, to do it properly.

For me it's not about the current runtime but rather about future runtime. When there are alternative DOM implementations, which pieces can actually be reused and which pieces are only usable in browser DOM? If we can answer that question, the problem is solved imo.

fkleuver:
Don't feel rushed by the way. The lifecycles is something we need to work together on so I'd rather have some delay with the merge and be sure that you understand what I'm trying to work towards. I'm worried that if I completely redo the lifecycles by myself that I'll miss use cases you had in mind.

PR #237, opened 2018-10-19
chore(karma-webpack): fix double error logging and stack trace sourcemaps

fkleuver:
# Pull Request

## 📖 Description

- Failed test assertions were logged to the console twice. Fixed this by switching around coverage and progress reporters.
- Failed tests were showing useless stack traces in both normal and `ChromeDebugging` mode where everything is on line 1. Fixed this by adding `source-map-support` and switching to devtool `inline-source-map`.
- Failed tests in `ChromeDebugging` mode were linking to nonexistent files after the above change, fixed this by switching to `eval-source-map` only in debugging mode

This makes debugging and troubleshooting tests a little nicer again @EisenbergEffect @bigopon

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8106,6 +8106,15 @@
         }
       }
     },
+    "karma-source-map-support": {
+      "version": "1.3.0",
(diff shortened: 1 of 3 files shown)


bigopon:
Nice little tweak!

PR #241, opened 2018-10-21
chore(jit-aurelia-cli): update package.json

fkleuver:
# Pull Request

## 📖 Description

Updates the package.json to target latest version. aurelia-cli seems to run into a parser error of which I'm not sure what's causing it, but I'll try a few things in the framework. For example getting rid of the star imports/exports.


fkleuver:
Will open a new PR with all examples

PR #242, opened 2018-10-21
bundler e2e tests

fkleuver:
# Pull Request

## 📖 Description

This adds locally running webdriverio e2e tests for all bundler variants to run with:
- Locally bundled packages (after bundling) on each commit
- published `dev` packages (after publishing to dev) on each nightly
- locally bundled packages (after pushing to release) on a release, and after publishing to release on the published `release` packages (note: the approval job will run before the local tests are done, giving you the option of letting a release go through even if not all bundler tests pass - for example if the failure is known to be caused by some external factor we cannot control)

This will likely cause builds to fail much more frequently due to the instability of some bundlers and their dependencies, but will allow us to catch any problems before they ever make it to the `@latest` channel at least.

## ⏭ Next Steps

- Fix the currently failing bundler examples
- Make a script to generate (part of) the test apps, to reduce code duplication in the repo


fkleuver:
@EisenbergEffect It's ready. The ones that are currently failing will cause the regular `build_test` builds to fail, but that's kind of the point. We should try to get those to work again. Having these tests running on each commit will make fixing them easier too.

The nightly dev release will still go through because it only tests the dev channel after publishing (this is intentional), otherwise the dev release could stale and prevent it from being tested if just 1 thing fails.

PR #244, opened 2018-10-25
chore(di): fix get() and getAll() return types on IServiceLocator interface

fkleuver:
# Pull Request

## 📖 Description

This PR fixes the type inference of the return type for `get()` and `getAll()` against the `IContainer` or `IServiceLocator` interfaces.

Before:
```ts
const cs = container.get(IChangeSet):
// cs is of type "{}"
const cs = container.get<IChangeSet>(IChangeSet):
// cs is of type "IChangeSet" (casting or explicitly type arguments are necessary)
```

After:
```ts
const cs = container.get(IChangeSet):
// cs is of type "IChangeSet"
const cs = container.get<IChangeSet>(IChangeSet):
// cs is of type "IChangeSet"
```

### DIFF
--- a/packages/kernel/src/di.ts
+++ b/packages/kernel/src/di.ts
@@ -32,14 +32,16 @@ export interface IFactory<T = any> {
 export interface IServiceLocator {
   has(key: any, searchAncestors: boolean): boolean;
 
-  get<K>(key: Constructable<unknown> | Key<unknown> | K):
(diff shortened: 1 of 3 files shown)


bigopon:
Nice. If folks want to do it explicitly, i guess they can do type casting at the end?

EisenbergEffect:
I think that `any` is ok in this case particularly since everything uses the interface. We've got some pretty fancy conditional types going on now...might make a good blog post at some point. I'm not sure what we'd do without those now.

Issue #89, opened 2018-08-21
Cleanup / separate the tests


fkleuver:
Some sets of unit tests have gotten a bit too messy and take a long time to run, in particular the observer and repeater tests.
I would propose getting rid of the `unit` subfolder and appending the "suite" name to the file:
- `.unit.spec.ts`
  - Very fast (< 5 seconds total). Stick to the definition of a unit test to the extent that it's practical.
  - Every module that has code, should have a corresponding `.unit.spec.ts`
- `.integration.spec.ts`
  - Relatively slow (up to a minute or more) and thorough
  - Mostly non-kernel/runtime packages, and high-level api's like `aurelia.ts`, `renderer.ts`
- `.smoke.spec.ts`
  - Similar to integration tests, but higher level, fewer in numbers and run fast, making them a good supplement for TDD / watch mode. Typically use jit/aot to cover lots of ground per test. Not sensitive to low-level api changes.
  - Don't need to map to specific files - one file per package (e.g. `runtime.smoke.spec.ts`) could be just fine
- `.regression.spec.ts`
  - Can fall under any of the 3 above categories and more. Target parts of the code base that probably won't change much (if ever) and thus don't need to be verified with every run. Examples: extra thorough variants of the parser and observer tests, and compiler tests that try many combinations of elements/attributes/events/data types.
  - May map to specific files or packages. May also map to issue numbers or pieces of the ES / HTML specs, etc.

Test commands:
- `npm run tdd` - runs the unit and smoke tests in watch mode, for during development. Only a few seconds to finish
- `npm run test` - runs the unit, integration and smoke tests in non-watch mode (single run), for before pushing. CI will also run this on each commit. Up to 2-3 minutes to finish for all packages
- `npm run test:full` - runs all tests. CI will run this on a nightly basis and will only publish a new `dev` version when the tests passed. Up to 30 minutes to finish
- `npm run e2e` - runs the e2e tests (can only be run from within examples that have them). CI will run these on each commit. Might later be split up as well, but no plans for that yet.

@EisenbergEffect @bigopon This is just an initial idea, I don't feel particularly strongly about this. Thoughts?

EisenbergEffect:
This is an interesting idea. I'm wondering if anyone else is doing something like this? Is "regression" really needed or could we get away with just three categories?

fkleuver:
Regression would be home to 3 kinds of tests:
- Tests that are reactive to issues/bug reports (and the corresponding fix)
[...]

PR #250, opened 2018-11-02
Lifecycles

fkleuver:
# Pull Request

## 📖 Description

This is a rework of templating/lifecycles in an attempt to make things a bit more coherent in terms of how the code is structured (bring related code paths together etc), and to fully integrate bind/attach lifecycles and batching of changes to improve the timing and co-ordination between different lifecycles.

~In order to prepare the lifecycles for this surgery I had to take out the async / task-related logic @EisenbergEffect put in place for the compose/if/else resources, as well as a few other important pieces that are not yet utilized.~

~I will add those back in once I've managed to put the various moving parts in the correct places. Taking them out was the only way to keep this exercise somewhat manageable.~

## 👩‍💻 Reviewer Notes

~I would advise against looking at this PR at the moment. Many things are in a (messy) intermediate state, only serving as a stepping stone for the next refactor phase.~
~Most of the remaining work is around restoring removed functionality (aggregate lifecycle tasks, encapsulation source) and fixing unit / integration tests.~
There is a lot of currently untapped potential in doing the lifecycles this way, but those will have to surface gradually as the need arises. It's hard to predict exactly how it's best utilized.

While many files are touched, on a fundamental level the changes are relatively scoped:
- The LifecycleControllers and ChangeSet / LinkedChangeList, as well as a variant of the former connect-queue are all merged into a single `Lifecycle` class. `lifecycle.ts` is the primary place to look.
- Templating has undergone a large reorganization mostly revolving around getting the lifecycle-related code together in the same place:
  - All rendering related logic is in `lifecycle-render.ts`
  - The bind and attach lifecycles are in `lifecycle-bind.ts` and `lifecycle-attach.ts`, respectively. The view lifecycles are also extracted out of view.ts and into these files

## 📑 Test Plan

~I am refactoring with exclusively the template-compiler integration tests enabled. I believe they're reliable and thorough enough to behave like an adequate indicator for whether things are working as they should. Fixing the unit tests will be the very last step.~
All existing tests (except for lifecycle unit tests) have been reworked to pass again with the minimum amount of changes. A small number of tests were added to specifically address some edge cases that this PR solves, but more are needed for thorough verification. This (along with potential minor fixes to address failing tests) is better tackled in an overall test cleanup PR that addresses #89 since so many areas are affected.

## ⏭ Next Steps

- Add a more fine-grained `registerTask` API for placing tasks before specific lifecycle hooks
- Add `encapsulationSource` back in, potentially via DI (depending on precise needs for consumers)
- Add more lifecycle tests that deeply verify the various potential interactions between lifecycles
- Add more integration tests that verify real world usages of lifecycles and tasks


packages/runtime/src/lifecycle.ts
```diff
@@ -418,55 +418,84 @@ export class AggregateLifecycleTask implements ILifecycleTask {
...
   }
 
   public queueMount(requestor: ILifecycleMount): void {
-    this.mountTail.$nextMount = requestor;
+    if (this.mountHead === null) {
+      this.mountHead = requestor;
+    } else {
+      this.mountTail.$nextMount = requestor;
+    }
     this.mountTail = requestor;
```
akircher: Know this is a WIP, but noticed that this seems to overwrite the assignment in the else statement

packages/runtime/src/lifecycle.ts
```diff
@@ -418,55 +418,84 @@ export class AggregateLifecycleTask implements ILifecycleTask {
...
   }
 
   public queueAttachedCallback(requestor: ILifecycleAttached): void {
-    this.attachedTail.$nextAttached = requestor;
+    if (this.attachedHead === null) {
+      this.attachedHead = requestor;
+    } else {
+      this.attachedTail.$nextAttached = requestor;
+    }
     this.attachedTail = requestor;
```
akircher: Know this is a WIP, but noticed that this seems to overwrite the assignment in the else statement

packages/runtime/src/lifecycle.ts
```diff
@@ -517,14 +532,22 @@ export class LifecycleController implements ILifecycleController {
...
   public queueUnmount(requestor: ILifecycleUnmount): void {
     if (this.allowUnmount) {
-      this.unmountTail.$nextUnmount = requestor;
+      if (this.unmountHead === null) {
+        this.unmountHead = requestor;
+      } else {
+        this.unmountTail.$nextUnmount = requestor;
+      }
       this.unmountTail = requestor;
       this.allowUnmount = false; // only remove roots
```
akircher: Know this is a WIP, but noticed that this seems to overwrite the assignment in the else statement

packages/runtime/src/lifecycle.ts
```diff
@@ -517,14 +532,22 @@ export class LifecycleController implements ILifecycleController {
...
   }
 
   public queueDetachedCallback(requestor: ILifecycleDetached): void {
-    this.detachedTail.$nextDetached = requestor;
+    if (this.detachedHead === null) {
+      this.detachedHead = requestor;
+    } else {
+      this.detachedTail.$nextDetached = requestor;
+    }
     this.detachedTail = requestor;
```
akircher: Know this is a WIP, but noticed that this seems to overwrite the assignment in the else statement

packages/runtime/src/lifecycle.ts
fkleuver: In the else statement:
```ts
this.attachedTail.$nextAttached = requestor;
```
Assigns the requestor to the $next property of the current tail.
Then:
```ts
this.attachedTail = requestor;
```
Replaces the current tail with the requestor. It's a standard LinkedList implementation. Is this what you're referring to?

akircher: Yes.  And I definitely might be missing something.  But the order of operations seems wrong here.

this order would work and create a recursive object
```
this.attachedTail = requestor;
this.attachedTail.$nextAttached = requestor;
```

this order would overwrite and nothing would be linked (unless this.attachedTail already == requestor in which the 2nd assignment is redundant.
```
this.attachedTail.$nextAttached = requestor;
this.attachedTail = requestor;
```
For example pseudocode, 
```
this.attachedTail = {tail:true}
this.requestor = {requestor:true}

this.attachedTail.$nextAttached = requestor; 
// attachedTail  == {tail:true, $nextAttached:{requestor:true}}

this.attachedTail = requestor; 
//attachedTail == {requestor:true}
```

akircher: My bad I missed the linked list comment!  I was thinking recursive object which is why I got this wrong.   I understand now.  Sorry for the trouble :-)

fkleuver: The point is that the `head` always points to the very first item so it is only set once. The tail must always point to the last item so it's replaced every operation. Before replacing the tail with the incoming node, the incoming node is first assigned to the `$next` property of the current tail node.

If you set the tail node to the incoming node directly, the `$next` property of the existing tail node will stay null and the incoming node's `$next property will just point to itself`

To clarify, here would be the flow from starting state up to two linked objects:

```ts
// Initial state
{
  head: null,
  tail: null
}

// Initial state -> first linking operation, incoming item.name = "first"
if (this.head === null) { // this executes
  this.head = item;
} else { // does not execute
  this.tail.$next = item;
}
this.tail = item;

// State after first linking operation
{
  head: {
    name: "first",
    $next: null
  },
  tail: {
    name: "first",
    $next: null
  }
}

// After first operation -> second linking operation, incoming item.name = "second"
if (this.head === null) { // does not execute
  this.head = item;
} else { // this executes
  this.tail.$next = item;
}
this.tail = item;

// State after second linking operation
{
  head: {
    name: "first",
    $next: {
      name: "second",
      $next: null
    }
  },
  tail: {
    name: "second",
    $next: null
  }
}

// After second operation -> third linking operation, incoming item.name = "third"
if (this.head === null) { // does not execute
  this.head = item;
} else { // this executes
  this.tail.$next = item;
}
this.tail = item;

// State after third linking operation
{
  head: {
    name: "first",
    $next: {
      name: "second",
      $next: {
        name: "third",
        $next: null
      }
    }
  },
  tail: {
    name: "third",
    $next: null
  }
}
```

Swapping it around would result in this:

```ts
// Initial state
{
  head: null,
  tail: null
}

// Initial state -> first linking operation, incoming item.name = "first"
this.tail = item;
if (this.head === null) { // this executes
  this.head = item;
} else { // does not execute
  this.tail.$next = item;
}

// State after first linking operation (so far, so good)
{
  head: {
    name: "first",
    $next: null
  },
  tail: {
    name: "first",
    $next: null
  }
}

// After first operation -> second linking operation, incoming item.name = "second"
this.tail = item;
if (this.head === null) { // does not execute
  this.head = item;
} else { // this executes
  this.tail.$next = item;
}

// State after second linking operation
{
  head: {
    name: "first",
    $next: null
  },
  tail: {
    name: "second",
    $next: {
      name: "second",
      $next: // recursively points to itself
    }
  }
}

// After second operation -> third linking operation, incoming item.name = "third"
this.tail = item;
if (this.head === null) { // does not execute
  this.head = item;
} else { // this executes
  this.tail.$next = item;
}

// State after third linking operation
{
  head: {
    name: "first",
    $next: null
  },
  tail: {
    name: "third",
    $next: {
      name: "third",
      $next: // recursively points to itself
    }
  }
}
// Additionally, the second item is now essentially "lost"
```

I should also point out there are currently roughly 7000 integration tests pounding this code. I can promise you it works ;)

fkleuver: I still think your comments were helpful in the sense that it made me realize the lifecycles probably need some clarification as the intent won't be obvious to everyone. So I added a little explanation to it for people who run into it in the future :)

fkleuver:
@EisenbergEffect @bigopon There is still some mess to clean up, some things to move around and some features to restore, but I'm reasonably satisfied with the current state of the lifecycles itself. It feels like there is much more control (and overview) in how the various flows connect and interact. What's left in this area is mostly utilizing it, small tweaks and adding features.
Feel free to start taking a look at `lifecycle.ts` if you have the time. There's no rush though.

EisenbergEffect:
@fkleuver I see you added back the root removal code. Should I review now or are you planning to add back the task stuff as well or are you reserving that for a future PR?

fkleuver:
I was planning to do it all in one go but it became apparent to me in the last few commits that it might be possible to add async functionality in a more integrated manner as well. A bit like simply allowing async bind/attach (not necessarily in that form, but at that level of simplicity).
In any case, far more tests are needed first to throw different kinds of complex scenarios at it and validate the correctness (or even necessity) of for example queueing the bound callbacks.

Since the lifecycles are at the center of it all, I'm estimating somewhere between 10k-20k integration and unit tests will be needed to cover this sufficiently. At that point the runtime could also be considered stable I think.

If you are okay with an intermediary state where the async lifecycle is temporarily gone, then it's probably better to review this now and take the async stuff later. Let everything sink in a bit, stabilize the effects of this refactor, get new inspiration, etc.

EisenbergEffect:
That's fine with me as long as we can add it back since we'll need it for compose, router and animation scenarios.

fkleuver:
Yeah that shouldn't be a problem. In the interim I could probably hook it in a similar manner to what you had it before, I just need to think about where precisely to hook it in. I'm mostly worried about the `unbindAfterDetached` mechanism which in itself is very convenient, but it also changes the lifecycle interaction of multiple views being processed in parallel.

fkleuver:
@EisenbergEffect @bigopon it's getting pretty close to finished and it's pretty ready for review. Most unit tests are back in action again (what remains is largely just the lifecycle-specific tests) and the tasks are back in (still needs as much testing as it did before, but the async compose integration tests are working at least).
Reminder, main files to look at are:
- `packages/runtime/src/lifecycle.ts`
- `packages/runtime/src/templating/lifecycle-attach.ts`
- `packages/runtime/src/templating/lifecycle-bind.ts`
- `packages/runtime/src/templating/lifecycle-render.ts`

Lifecycle-render is still due for cleanup / split-up though I'm not entirely sure what should or shouldn't be in there. The idea is to keep render-related stuff together (same with attach and bind) but to actually keep the renderer and rendering-engine in there may be a bit too much Any thoughts on that?

EisenbergEffect:
Apologies for being the bottleneck on this! I've had to work on other tasks for a bit. This is my next thing to do.

packages/runtime/src/templating/lifecycle-attach.ts
```diff
@@ -0,0 +1,245 @@
...
+    this.$state |= State.isDetaching;
+    flags |= LifecycleFlags.fromDetach;
+
+    const hooks = this.$hooks;
+    if (hooks & Hooks.hasDetaching) {
+      this.detaching(flags);
+    }
+
+    if (this.$state & State.isMounted) {
+      if ((flags ^ LifecycleFlags.parentUnmountQueued) | (flags & LifecycleFlags.allowUnmount)) {
```
EisenbergEffect: I like this better than my previous hack :)

packages/runtime/src/lifecycle.ts
```diff
@@ -164,10 +245,10 @@ export interface ILifecycleAttaching {
    * This is the time to add any (sync or async) tasks (e.g. animations) to the lifecycle that need to happen before
    * the nodes are added to the DOM.
    */
-  attaching?(encapsulationSource: INode, lifecycle: IAttachLifecycle): void;
+  attaching?(flags: LifecycleFlags): void;
```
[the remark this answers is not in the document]
fkleuver: I had no intent to remove it permanently (no need to convince me :)), I just wanted to start with a uniform interface. What do you think about using DI for this (similar to RenderLocation)?

fkleuver:
@EisenbergEffect latest changes merged, it's good to go then

Issue #249, opened 2018-11-01
WIP Fix linting warnings


BBosman:
# 💬 RFC

I started pocking around the codebase of vNext today and noticed that running `npm run lint` gives a huge amount of linting warnings. Do these need fixing? If so, I can start making PR's for them. Added benefit for me is that I get to learn the codebase a bit while doing it.

## 🔦 Context

I thought I'd ask the question here first, before I started doing the work and it turns out that it's not (yet) needed/wanted. And also to get some feedback on how to approach this, if it is wanted.

My idea would be to create a PR per rule to keep the PR's easier to review and less prone to merge conflicts, or is a split along package (for Changelog purposes) also wanted? 

## 💻 Progress

- [X] Many linting fixes (#251, #252, #253, #255, #256, #257, #262, #274, #277, #290)
- [X] Use official Reflect metadata API type definitions (#261)
- [X] Disable promise-function-async linting rule (#280)
- [X] Remove StrictAny in favour of unknown (#294)
- [X] Turn off `no-reserved-keywords` rule. (#297)
- [X] Rename `fromIndex`, `toIndex` and `messageType` back to `from`, `to` and `type`. Partially reverts #290. (#301)
- [X] Fix `no-floating-promises` violations. (#304)
- [X] Fix `no-shadowed-variable` violations. (#338)
- [X] Fix `no-unnecessary-type-assertion` violations (in `expression-parser.ts`) (#337)
- [X] Fix `no-any` violations in `di.ts` (#338)
- [ ] ~~Fix other `no-any` violations~~
- [ ] Set `completed-docs` to `warning` and all other linting violations to `error` (#418)

EisenbergEffect:
I do think this would be helpful but we'll want to synchronize where you are making those changes so it doesn't create conflicts with other work. I'm thinking that you could look through the code in the `kernel` package and make some improvements there. @fkleuver You do you expect you'll be in there? Or are you mostly in runtime right now?

fkleuver:
> Do these need fixing?

Yep, eventually we would like everything to be linting error-free.

It's probably safest (in terms of not risking wasted effort and/or merge conflicts) to start in the more stable areas. That would be `packages/kernel`, `packages/jit`, and `packages/runtime/binding`. I have some refactorings planned for the observers but those will not affect the typings.

`packages/runtime` (top-level) and `packages/runtime/templating` is currently undergoing a huge refactor so better not spend too much time there right now :)

[...]

PR #252, opened 2018-11-02
style(jit): fix some linting warnings

BBosman:
# Pull Request

## 📖 Description

Initial round of fixing lining warnings in `@aurelia/jit`.

There are still warnings left, but they require a more intimate knowledge of the code that I don't yet have, so I'll pick those up once I have the feeling I'm ready for that.

### 🎫 Issues

Related to #249

### DIFF
--- a/packages/jit/src/element-parser.ts
+++ b/packages/jit/src/element-parser.ts
@@ -1,5 +1,5 @@
 import { DI, inject, PLATFORM } from '@aurelia/kernel';
-import { DOM, INode } from '@aurelia/runtime';
+import { DOM, IAttr, INode } from '@aurelia/runtime';
 import { AttrSyntax, IAttributeParser } from './attribute-parser';
 
 const domParser = <HTMLDivElement>DOM.createElement('div');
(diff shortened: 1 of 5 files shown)


packages/jit/src/semantic-model.ts
```diff
@@ -357,13 +357,13 @@ export class ElementSymbol {
     const siblings = this.$parent.$children;
     for (let i = 0, ii = siblings.length; i < ii; ++i) {
       if (siblings[i] === this) {
-        return siblings[i + 1] || null;
+        return siblings[i + 1] ? siblings[i + 1] : null;
```
fkleuver: Referring to what I said about accessing member variables, this holds even more true for array indexers. They're miniscule nitpicks, granted, but I'd prefer a variable declaration for the conditional in this case.

packages/jit/src/element-parser.ts
```diff
@@ -49,14 +49,14 @@ export const IElementParser = DI.createInterface<IElementParser>()
 export class ElementParser implements IElementParser {
   constructor(public attrParser: IAttributeParser) {}
 
-  public parse(markupOrNode: string | INode): ElementSyntax {
-    let node: Element;
+  public parse(markupOrNode: string | Node): ElementSyntax {
+    let node: Node;
```
fkleuver: This is a bit of a tricky one. So we're trying to avoid any references to `lib.dom.d.ts` wherever possible because eventually the framework needs to be platform-independent (e.g. it may also run in NodeJS, render WebGL, NativeScript and such). Internally these casts can be fine on occasion (though we try to eliminate that too), but especially in the public API we try to avoid it.
Just curious, what linting error does this fix?

BBosman:
Updated the PR with the requested changes and I think I got all the spots needing intermediate variables.

packages/jit/src/element-parser.ts
fkleuver: If you have any ideas for this by introducing an additional type, that's fine by me too. We'll need to extend the interfaces accordingly at some point anyway. But for this PR this is fine for now. Removing `lib.dom.d.ts` dependency is a larger undertaking that may require some additional refactoring which doesn't really fall under the scope of linting.

PR #255, opened 2018-11-04
style(plugin-svg): fix linting warnings

BBosman:
# Pull Request

## 📖 Description

Initial round of fixing lining warnings in `@aurelia/plugin-svg`.

### 🎫 Issues

Related to #249

## 📑 Test Plan

Manually imported the fixes from #257 (not included in the final PR).

Executed without errors:
- lerna run lint --scope=@aurelia/plugin-svg
- lerna run build
- lerna run test

### DIFF
--- a/packages/plugin-svg/src/svg-analyzer.ts
+++ b/packages/plugin-svg/src/svg-analyzer.ts
@@ -2,86 +2,86 @@ import { IContainer } from '@aurelia/kernel';
 import { DOM, INode, ISVGAnalyzer } from '@aurelia/runtime';
 
 const svgElements = {
-  a: ['class','externalResourcesRequired','id','onactivate','onclick','onfocusin','onfocusout','onload','onmousedown','onmousemove','onmouseout','onmouseover','onmouseup','requiredExtensions','requiredFeatures','style','systemLanguage','target','transform','xlink:actuate','xlink:arcrole','xlink:href','xlink:role','xlink:show','xlink:title','xlink:type','xml:base','xml:lang','xml:space'],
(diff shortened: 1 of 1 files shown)


fkleuver:
I'm not sure about this one. The problem is that *not* having a typedef here causes the object to get the correct generated types. Intellisense will tell you exactly what's in the object if you type `svgElements.`.

Manually specifying all of this would create a typing that is even bigger than that object itself which is also a bit silly and a potential maintenance burden.

Personally my preference here would be to somehow utilize `lib.dom.d.ts` for this (keep in mind that, as opposed to the runtime package, it's perfectly fine for this plugin to have hard dependencies on the DOM since it's 100% DOM specific code anyway) or simply have this file generated based on some specification data.

In either case, this piece of code is rather bulky, raw, and unlikely to change very often. In that light, I think the appropriate middle ground for now would be to simply make sure the public API has the correct typings and simply disable the `typedef` tslint rule for the big object itself. I think the other fixes are fine.

fkleuver:
In addition to the above, can you please include in the "Test Plan" which steps you've undertaken to ensure that it's all still working.
For example: "I've ran `npm run build` from the project root to ensure all dependent packages still build in addition to this one. I also ran `npm run lint` and confirmed that no new linting errors were introduced"

Note: this is a bit of a trick question because I happen to know that you broke the build in one of your previous PR's ;) - which is a little bit on me as well because I assumed that you built the project before submitting.

fkleuver:
Conditional types would be an option here if `lib.dom.d.ts` was complete on the SVG side of things which it seems like it's not, unfortunately. But since this list isn't part of the public API I think `tslint:disable-next-line:type-def` would also be acceptable

BBosman:
I updated the PR.

I dropped those `type-def`'s as it turns out they weren't needed anyway. At one time the linter was complaining about it, but apparently one of the other linting fixes I did made them irrelevant.

I also updated the PR description on what testing I did.

fkleuver:
Yup I like :)

PR #256, opened 2018-11-04
style(plugin-requirejs): fix linting warnings

BBosman:
# Pull Request

## 📖 Description

Initial round of fixing lining warnings in `@aurelia/plugin-requirejs`.

### 🎫 Issues

Related to #249

## 👩‍💻 Reviewer Notes

Adding @types/requirejs introduces more issues than it solves. It's incomplete (i.e. isBuild is missing on the RequireConfig interface ) and it's creating conflicts:

```
node_modules/@types/node/index.d.ts(5984,5): error TS2300: Duplicate identifier 'mod'.
node_modules/@types/requirejs/index.d.ts(38,11): error TS2300: Duplicate identifier 'mod'.
node_modules/@types/requirejs/index.d.ts(422,13): error TS2403: Subsequent variable declarations must have the same type.  Variable 'require' must be of type 'NodeRequire', but here has type 'Require'.
```

So I went for a (simplified) version of duck-typing. Just enough to match the parts of the API being used.

## 📑 Test Plan

Executed without issues:
- lerna run lint --scope=@aurelia/plugin-requirejs
- npm run build
- npm run test

### DIFF
--- a/packages/plugin-requirejs/src/processing.ts
+++ b/packages/plugin-requirejs/src/processing.ts
@@ -7,16 +7,16 @@ export interface ITemplateImport {
 
 export interface ITemplateDescription {
   template: string;
-  imports: ITemplateImport[]
+  imports: ITemplateImport[];
 }
 
-export function processImports(toProcess: ITemplateImport[], relativeTo): string[] {
+export function processImports(toProcess: ITemplateImport[], relativeTo: string): string[] {
   return toProcess.map(x => {
     if (x.extension === '.html' && !x.plugin) {
-      return 'component!' + relativeToFile(x.path, relativeTo) + x.extension;
+      return `component!${relativeToFile(x.path, relativeTo) + x.extension}`;
     }
 
-    let relativePath = relativeToFile(x.path, relativeTo);
+    const relativePath = relativeToFile(x.path, relativeTo);
     return x.plugin ? `${x.plugin}!${relativePath}` : relativePath;
   });
 }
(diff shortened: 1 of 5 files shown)


fkleuver:
If I'm not mistaken these originate from the requirejs API, the typings for which can be found here: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/requirejs/index.d.ts

We could install `@types/requirejs` as a dev dependency into this plugin package, but we do need to double check whether those types are actually correct and up-to-date. I've learned the hard way to distrust type definitions that were not actually generated by the tsc compiler.

BBosman:
Adding `@types/requirejs` introduces more issues than it solves. It's incomplete (i.e. `isBuild` is missing on the `RequireConfig` interface ) and it's creating conflicts:
```
node_modules/@types/node/index.d.ts(5984,5): error TS2300: Duplicate identifier 'mod'.
node_modules/@types/requirejs/index.d.ts(38,11): error TS2300: Duplicate identifier 'mod'.
node_modules/@types/requirejs/index.d.ts(422,13): error TS2403: Subsequent variable declarations must have the same type.  Variable 'require' must be of type 'NodeRequire', but here has type 'Require'.
```

So I went for a (simplified) version of duck-typing. Just enough to match the parts of the API being used.

packages/plugin-requirejs/src/processing.ts
```diff
@@ -107,13 +107,14 @@ export function relativeToFile(name: string, file: string): string {
...
+  nodeRequire(name: string): unknown;
 }
 
+// tslint:disable-next-line:no-reserved-keywords
 declare const require: Require;
 
-export function loadFromFile(url: string, callback: Function, errback: Function) {
-  const fs = require.nodeRequire('fs');
+export function loadFromFile(url: string, callback: (content: string) => void, errback: (error: Error) => void): void {
+  const fs = require.nodeRequire('fs') as {readFileSync(url: string, encodig: string): string};
```
fkleuver: To fix a small typo and make it just a little bit more accurate in the subset of `readFileSync` that it represents
```suggestion
  const fs = require.nodeRequire('fs') as {readFileSync(path: string, options?: { encoding?: string } | string): string};
```

BBosman: Done

BBosman:
Updated the PR with the requested changes. :)

fkleuver:
Excellent thank you :)
LGTM @EisenbergEffect

PR #257, opened 2018-11-04
Typing fixes

fkleuver:
# Pull Request

## 📖 Description

This PR fixes two small regressions introduced by @BBosman 's otherwise excellent work on reducing the number of linting errors.

One is related to decorators which I believe is either impossible to fix at this point (see the clarifying comments), or would be so complicated that it's simply not worth the trouble right now. So I changed the `unknown` back to `any` and disabled the tslint rule for that one.

The other is related to the `IIndexable` type which can probably be changed back to `unknown` again at some point, but doing so will require a number of changes in the runtime for its various consumers. Since this one *can* be fixed, I did not disable the "no-any" rule for this line.

## 👩‍💻 Reviewer Notes

@BBosman please check the comments I put in place :)

## 📑 Test Plan

I've simply made sure that the whole project builds again. One linting error is reintroduced and the other is suppressed.

## ⏭ Next Steps

See if `IIndexable`'s index signature return type can be changed back to `unknown` without needing to overhaul the entire runtime's typings (if it is needed however, then it should happen at some point anyway)

### DIFF
--- a/packages/kernel/src/interfaces.ts
+++ b/packages/kernel/src/interfaces.ts
@@ -29,7 +29,13 @@ export type Decorated<TOptional, TRequired> = Function & {
 
 export type Injectable<T = {}> = Constructable<T> & { inject?: unknown[] };
 
-export type IIndexable<T extends object = object> = T & { [key: string]: unknown };
+// Note: use of "any" here can perfectly well be replaced by "unknown" but that would also involve fixing consumers of this
+// interface since their indexed properties are now all returning "unknown" which is not assignable to anything else.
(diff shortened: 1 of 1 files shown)


fkleuver:
@BBosman No problem and completely understandable :)
You don't need to stay away from the runtime entirely by the way. It's completely fine to fix typings in binding, even if I'm working on the same file merge conflicts on typings are very easy to resolve (usually they can be auto merged). I only mentioned templating and top-level runtime files as a bad idea to work on because I'm also changing and adding/removing types there :)

PR #260, opened 2018-11-05
chore(examples): fix jit-parcel-ts

fkleuver:
# Pull Request

## 📖 Description

So much for "zero config" huh :)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -16,6 +16,7 @@ map-1: &package_globs
     - dist
     - packages/*/dist
     - packages/*/node_modules
+    - packages/*/*.tgz
 
 map-2: &filter_ignore_develop_release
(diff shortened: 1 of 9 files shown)


EisenbergEffect:
The CircleCI tests still seem broken. Any ideas @fkleuver ?

fkleuver:
Yeah it's failing for a different reason now. Parcel doesn't seem to be smart enough to figure out the symlinks thing so the kernel is included in the bundle 4 times, and the template compiler is registered in a different DI than the one the runtime uses -.-
Guess i'll have to dust off the local publish script

fkleuver:
@EisenbergEffect it's working now with a little bit of special treatment for parcel. This is only needed for the local variant (the rest worked already)

PR #261, opened 2018-11-05
style(kernel): use official Reflect metadata API type definitions

BBosman:
# Pull Request

## 📖 Description

Use the official Reflect metadata API type definitions instead of rolling our own.

API proposal:
https://www.typescriptlang.org/docs/handbook/decorators.html#metadata
https://rbuckton.github.io/reflect-metadata/

Type definitions:
https://github.com/rbuckton/reflect-metadata

### 🎫 Issues

Related to #249.

## 👩‍💻 Reviewer Notes

As `@types/reflect-metadata` is deprecated we use the definitions bundled with `reflect-metadata`. Because this is outside the `@types` scope we need to manually reference the type definitions.

This also reintroduces some `any` types, but as they're part of the official spec proposal it seems a valid use-case, so I suppressed those warnings.

## 📑 Test Plan

- `lerna run lint --scope=@aurelia/kernel` -> No new linting warnings
- `npm build` -> succeeds without errors.
- `npm test` -> succeeds without test failures.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10657,6 +10657,12 @@
         "strip-indent": "^1.0.1"
       }
     },
+    "reflect-metadata": {
+      "version": "0.1.12",
+      "resolved": "https://registry.npmjs.org/reflect-metadata/-/reflect-metadata-0.1.12.tgz",
+      "integrity": "sha512-n+IyV+nGz3+0q3/Yf1ra12KpCyi001bi4XFxSjbiWWjfqb52iTTtpGXmCCAOWWIAn9KEuFZKGqBERHmrtScZ3A==",
+      "dev": true
+    },
     "regenerator-runtime": {
       "version": "0.11.1",
       "resolved": "https://registry.npmjs.org/regenerator-runtime/-/regenerator-runtime-0.11.1.tgz",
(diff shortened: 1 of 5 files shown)


fkleuver:
I think this is a nice detail. Better to use official typings so we don't risk missing important changes if we roll our own.
@BBosman does the project still correctly install from a clean clone? Since lerna manages the dependencies on a package level by hoisting them

BBosman:
As I used lerna itself to add the package using `lerna add reflect-metadata --dev --scope=@aurelia/kernel` I would have expected it would.

Just tested by nuking all `node_modules`, including the root one and ran `npm run init` and it failed. 😞

So I just added it to the root `package.json` and `package-lock.json` and now it works. I would have thought `lerna` would "just work" here.

fkleuver:
Just a tip on the nuking, this should get everything cleaned up (from the root, assuming you're on windows, otherwise use the other slash and rm -rf instead of rmdir):
```
.\node_modules\.bin\lerna clean -y && git clean -xfd && git checkout -- .
npm ci
```

BBosman:
Or the shorthand: `npm run refresh` 😉

fkleuver:
Yeah I've had that git stash trick bite me once or twice so I stopped using it even though I wrote it ;P
(it really just breaks when there are e2e tests and/or examples installed though - the symlinks mess up the git clean command)

PR #262, opened 2018-11-05
style(runtime): fix linting warnings in binding

BBosman:
# Pull Request

## 📖 Description

Linting fixes in the `binding` part of `@aurelia/runtime`

### 🎫 Issues

Related to #249.

## 👩‍💻 Reviewer Notes

Mostly syntactical linting fixes, but also a couple of strict type checks which could use some scrutiny.

Hit a few files also touched in #250, but I think these changes are compatible.

## 📑 Test Plan

- `lerna run lint --scope=@aurelia/runtime` -> Significant drop in the number of reported warnings
- `npm run build` -> Succeeded without errors.
- `npm run test` -> Succeeded without test failures.

## ⏭ Next Steps

Once #250 lands the linting warnings that have a bit more impact in `binding` can be tackled, together with the warnings in the root and templating files.

Another follow up would be implementing the outcome of https://github.com/aurelia/aurelia/pull/262#discussion_r230880271 outside `binding` and work on https://github.com/aurelia/aurelia/pull/262#discussion_r230879707

### DIFF
--- a/packages/runtime/src/binding/array-observer.ts
+++ b/packages/runtime/src/binding/array-observer.ts
@@ -1,7 +1,6 @@
 import { IIndexable, Primitive } from '@aurelia/kernel';
 import { BindingFlags, CollectionKind, IChangeSet, ICollectionObserver, IndexMap, IObservedArray } from '../observation';
 import { collectionObserver } from './collection-observer';
-// tslint:disable:no-reserved-keywords
 const proto = Array.prototype;
 export const nativePush = proto.push; // TODO: probably want to make these internal again
 export const nativeUnshift = proto.unshift;
@@ -93,7 +92,7 @@ function observeSplice(this: IObservedArray, start: number, deleteCount?: number
   }
   const indexMap = o.indexMap;
   if (deleteCount > 0) {
-    let i = start || 0;
+    let i = isNaN(start) ? 0 : start;
     const to = i + deleteCount;
     while (i < to) {
       if (indexMap[i] > -1) {
@@ -191,12 +190,12 @@ function preSortCompare(x: IIndexable | Primitive, y: IIndexable | Primitive): n
   return 0;
 }
 
-function insertionSort(arr: IObservedArray, indexMap: IndexMap, from: number, to: number, compareFn: (a: IIndexable | Primitive, b: IIndexable | Primitive) => number): void {
+function insertionSort(arr: IObservedArray, indexMap: IndexMap, fromIndex: number, toIndex: number, compareFn: (a: IIndexable | Primitive, b: IIndexable | Primitive) => number): void {
   let velement, ielement, vtmp, itmp, order;
   let i, j;
-  for (i = from + 1; i < to; i++) {
+  for (i = fromIndex + 1; i < toIndex; i++) {
     velement = arr[i]; ielement = indexMap[i];
-    for (j = i - 1; j >= from; j--) {
+    for (j = i - 1; j >= fromIndex; j--) {
       vtmp = arr[j]; itmp = indexMap[j];
       order = compareFn(vtmp, velement);
       if (order > 0) {
@@ -209,7 +208,7 @@ function insertionSort(arr: IObservedArray, indexMap: IndexMap, from: number, to
   }
 }
 
-function quickSort(arr: IObservedArray, indexMap: IndexMap, from: number, to: number, compareFn: (a: IIndexable | Primitive, b: IIndexable | Primitive) => number): void {
+function quickSort(arr: IObservedArray, indexMap: IndexMap, fromIndex: number, toIndex: number, compareFn: (a: IIndexable | Primitive, b: IIndexable | Primitive) => number): void {
   let thirdIndex = 0, i = 0;
   let v0, v1, v2;
   let i0, i1, i2;
@@ -220,14 +219,14 @@ function quickSort(arr: IObservedArray, indexMap: IndexMap, from: number, to: nu
 
   // tslint:disable-next-line:no-constant-condition
   while (true) {
-    if (to - from <= 10) {
-      insertionSort(arr, indexMap, from, to, compareFn);
+    if (toIndex - fromIndex <= 10) {
+      insertionSort(arr, indexMap, fromIndex, toIndex, compareFn);
       return;
     }
 
-    thirdIndex = from + ((to - from) >> 1);
-    v0 = arr[from];       i0 = indexMap[from];
-    v1 = arr[to - 1];     i1 = indexMap[to - 1];
+    thirdIndex = fromIndex + ((toIndex - fromIndex) >> 1);
+    v0 = arr[fromIndex];       i0 = indexMap[fromIndex];
+    v1 = arr[toIndex - 1];     i1 = indexMap[toIndex - 1];
     v2 = arr[thirdIndex]; i2 = indexMap[thirdIndex];
     c01 = compareFn(v0, v1);
     if (c01 > 0) {
@@ -249,11 +248,11 @@ function quickSort(arr: IObservedArray, indexMap: IndexMap, from: number, to: nu
         v2 = vtmp; i2 = itmp;
       }
     }
-    arr[from] = v0;   indexMap[from] = i0;
-    arr[to - 1] = v2; indexMap[to - 1] = i2;
+    arr[fromIndex] = v0;   indexMap[fromIndex] = i0;
+    arr[toIndex - 1] = v2; indexMap[toIndex - 1] = i2;
     vpivot = v1;      ipivot = i1;
-    lowEnd = from + 1;
-    highStart = to - 1;
+    lowEnd = fromIndex + 1;
+    highStart = toIndex - 1;
     arr[thirdIndex] = arr[lowEnd]; indexMap[thirdIndex] = indexMap[lowEnd];
     arr[lowEnd] = vpivot;          indexMap[lowEnd] = ipivot;
 
@@ -283,12 +282,12 @@ function quickSort(arr: IObservedArray, indexMap: IndexMap, from: number, to: nu
         }
       }
     }
-    if (to - highStart < lowEnd - from) {
-      quickSort(arr, indexMap, highStart, to, compareFn);
-      to = lowEnd;
+    if (toIndex - highStart < lowEnd - fromIndex) {
+      quickSort(arr, indexMap, highStart, toIndex, compareFn);
+      toIndex = lowEnd;
     } else {
-      quickSort(arr, indexMap, from, lowEnd, compareFn);
-      from = highStart;
+      quickSort(arr, indexMap, fromIndex, lowEnd, compareFn);
+      fromIndex = highStart;
     }
   }
 }
--- a/packages/runtime/src/binding/ast.ts
+++ b/packages/runtime/src/binding/ast.ts
@@ -7,8 +7,6 @@ import { IConnectableBinding } from './connectable';
 import { ISignaler } from './signaler';
 import { ValueConverterResource } from './value-converter';
 
-// tslint:disable:no-empty
-
 /**
  * StrictAny is a somewhat strongly typed alternative to 'any', in an effort to try to get rid of all 'any''s
  * It's not even remotely foolproof however, and this can largely be attributed to the fact that TypeScript imposes
@@ -154,11 +152,10 @@ export function isLiteral(expr: IsExpressionOrStatement): expr is IsLiteral {
   return (expr.$kind & ExpressionKind.IsLiteral) === ExpressionKind.IsLiteral;
 }
 export function arePureLiterals(expressions: ReadonlyArray<IsExpressionOrStatement>): expressions is IsLiteral[] {
-  const len = expressions && expressions.length || 0;
-  if (len === 0) {
+  if (expressions.length === 0) {
     return true;
   }
-  for (let i = 0; i < len; ++i) {
+  for (let i = 0; i < expressions.length; ++i) {
     if (!isPureLiteral(expressions[i])) {
       return false;
     }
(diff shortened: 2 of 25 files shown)


packages/runtime/src/binding/array-observer.ts
```diff
@@ -93,7 +93,7 @@ function observeSplice(this: IObservedArray, start: number, deleteCount?: number
   }
   const indexMap = o.indexMap;
   if (deleteCount > 0) {
-    let i = start || 0;
+    let i = isNaN(start) ? 0 : start;
```
fkleuver: Nice one

packages/runtime/src/binding/ast.ts
```diff
@@ -347,7 +347,7 @@ export class ValueConverter implements IExpression {
     if (signals === undefined) {
       return;
     }
-    const signaler = locator.get(ISignaler) as ISignaler;
+    const signaler = locator.get(ISignaler);
```
fkleuver: Thanks for picking this up. The casting is still a remnant from before the typing fixes in DI so we should be able to get rid of all these casts now :)

packages/runtime/src/binding/binding-behavior.ts
```diff
@@ -20,6 +20,7 @@ export const BindingBehaviorResource: IResourceKind<IBindingBehaviorSource, IBin
     return `${this.name}:${name}`;
   },
 
+  // tslint:disable-next-line:no-reserved-keywords
```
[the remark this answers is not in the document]
BBosman: I'm not sure. `from` has the same issue and having mixed Pascal case and Camel case in interfaces and classes just looks "weird".

Switching `"no-reserved-keywords": true,` to `"no-reserved-keywords": [true, {"allow-quoted-properties": true}],` in `tslint.json` and quoting those reserved words in the interfaces would already reduce the number of `tslint:disable:no-reserved-keywords` by a lot.

Renaming `type` arguments in functions to something more specific, like `bindingBehaviorType` in this specific case would reduce it even more.

That would only leave actual class implementation needing a `tslint:disable`, which after a quick count would be about three.

Whichever route we take, I think it would be best to do that in a follow-up PR.

packages/runtime/src/binding/ast.ts
BBosman: These are the only ones the linter picked up for now. I'm sure we'll see more pop up once we start making all those `any` typings still left more specific. 😄

packages/runtime/src/binding/binding-behavior.ts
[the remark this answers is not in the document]
fkleuver: I don't think `type` and `from` are subject to the same need for consistency in that sense. `type` is always in fact a constructor function and we're already naming it `ctor`, `Ctor`, `Type` and `Class` in various places. Since calling it like `new type()` is valid (it is by all means and purposes a class), it's fair to say it's eligible for the same naming convention as classes thus CamelCase.

We could might as well name it `$type` or some such to ensure it's not a reserved keyword (not my pref though). There are several legitimate ways to prevent that, so I don't think disabling the warning is a necessity here. I like either `Ctor` or `Type` in any case. What would you suggest?

As for `from` that has a different meaning entirely. It used to be `src` (as in sourceExpression) but we renamed it to `from` and `to` to add a little semantic meaning to instructions. It doesn't necessarily have to stay that way.

Agree it's fine to do in a next PR, all I'm really saying is I don't think it should be suppressed because it can be fixed.

packages/runtime/src/binding/binding-behavior.ts
```diff
@@ -7,10 +7,10 @@ export interface IBindingBehaviorSource {
 
 export type IBindingBehaviorType = IResourceType<IBindingBehaviorSource>;
 
-export function bindingBehavior(nameOrSource: string | IBindingBehaviorSource) {
-  return function<T extends Constructable>(target: T) {
+export function bindingBehavior(nameOrSource: string | IBindingBehaviorSource): <T extends Constructable>(target: T) => T & IResourceType<IBindingBehaviorSource> {
```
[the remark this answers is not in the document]
BBosman: I've been experimenting with this just now, but I must admit I'm a bit stuck.

Looking at this code in `binding-command.ts`:
```ts
export interface IBindingCommand {
  compile($symbol: IAttributeSymbol): TargetedInstruction;
  handles($symbol: IAttributeSymbol): boolean;
}

export type IBindingCommandType = IResourceType<IBindingCommandSource, IBindingCommand>;

export function bindingCommand(nameOrSource: string | IBindingCommandSource): any {
  return function<T extends Constructable>(target: T): T & IResourceType<IBindingCommandSource, IBindingCommand> {
    return BindingCommandResource.define(nameOrSource, target);
  };
}
```
Because of the `any` I decided to try and mimic the setup from `custom-element.ts`. My first step was to simplify the `bindingCommand` function to this:
```ts
export function bindingCommand(nameOrSource: string | IBindingCommandSource): any {
  return function<T extends Constructable>(target: T): T & IBindingCommandType {
    return BindingCommandResource.define(nameOrSource, target);
  };
}
```
Then I created a `BindingCommand` type to encapsulate the `Decoratable ` as you described:
```ts
type BindingCommand = <T extends Constructable>(target: Decoratable<IBindingCommand, T>) => Decorated<IBindingCommand, T> & IBindingCommandType;
```
If I now switch the `any` to the newly created `BindingCommand` stuff starts breaking at all `@bindingCommand` use sites with this missing `inject` error:
```
Unable to resolve signature of class decorator when called as an expression.
  Type 'Function & { new (...args: unknown[]): any; readonly prototype: Required<IBindingCommand> & Required<Constructable<{}>>; } & IResourceType<IBindingCommandSource, IBindingCommand>' is not assignable to type 'typeof OneTimeBindingCommand'.
    Property 'inject' is missing in type 'Function & { new (...args: unknown[]): any; readonly prototype: Required<IBindingCommand> & Required<Constructable<{}>>; } & IResourceType<IBindingCommandSource, IBindingCommand>'.
```

I'm probably missing something obvious, but at the moment I just don't see it. 😿

BBosman:
Updated the PR.

Changes:
- Rebased
- Removed most references of `// tslint:disable:no-reserved-keywords` by renaming the *offenders* (as suggested by @fkleuver)
- Re-used `findOriginalEventTarget` from `self-binding-behavior` in `event-manager.ts` instead of redefining it inline. (less code duplication and fixes a couple of linting warnings)
- A few more `any` -> "a more specific type" changes.

I'd like to leave the suggested *"special typings in for decorators"* changes for a follow up PR, as I need to dig a bit deeper on that. (pointers appreciated, see comment above).

Happy to make additional changes on review feedback, but otherwise this is done for now. 😃

packages/runtime/src/binding/binding-behavior.ts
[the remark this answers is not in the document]
BBosman: Figured it out I think. Needed to switch  `IBindingCommandType ` from a `type` to an `interface` and give it an `inject: Function[];` property.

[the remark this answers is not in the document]
BBosman: One quick question though. Looking at `custom-element.ts`, where I copied this from, I see 
```ts
type CustomElementDecorator = <T extends Constructable>(target: Decoratable<ICustomElement, T>) => Decorated<ICustomElement, T> & ICustomElementType;
```
Shouldn't the arguments of `Decoratable` and `Decorated` be switched around as their signature is `<TOptional, TRequired>`? I would expect `T` to be optional and `ICustomElement` required.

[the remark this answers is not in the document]
fkleuver: Sorry @BBosman I didn't get a notification for the question you asked 3 days ago, so I had no idea you had an open question! Otherwise I would have replied on that much sooner. Feel free to just ping me in any comment in the future to make sure I don't miss it for whatever reason.

I should point out that, by looking at `binding-command` more closely, that was actually a bad example since it doesn't use the `Decoratable`  and `Decorated` interfaces. My bad!

I could explain the point of these decorators and how they work here but I think it's better if I just write it up properly in some comments so the rest of the world will understand too. Don't worry about it for now and I will ping you when the PR is out :)

packages/runtime/src/binding/ast.ts
```diff
@@ -154,8 +154,8 @@ export function isLiteral(expr: IsExpressionOrStatement): expr is IsLiteral {
   return (expr.$kind & ExpressionKind.IsLiteral) === ExpressionKind.IsLiteral;
 }
 export function arePureLiterals(expressions: ReadonlyArray<IsExpressionOrStatement>): expressions is IsLiteral[] {
-  const len = expressions && expressions.length || 0;
-  if (len === 0) {
+  const len = expressions && expressions.length;
+  if (len === undefined || len === 0) {
```
fkleuver: @BBosman Gave this a bit more thought and I think, for future reference, that I would actually prefer this:
```ts
if (expressions === undefined || expressions.length === 0) {
    return true;
}
```
Or in this particular instance, perhaps even this:
```ts
if (expressions.length === 0) {
    return true;
}
```
I'd like to move away from all coercions at some point and only use strict equality. If the number of necessary checks become too many, we could further constrain the preconditions (documented at the API boundary with JSDoc comments) and simply let it fail if these are not met.

Reason is to reduce the number of silent failures and enforce more strict pre/postconditions and invariants across the code base, and make sure nothing works or fails by accident and everything is intentional.

In any case, that doesn't change anything for this PR. This is fine for now. I just wanted to point it out

packages/runtime/src/binding/element-observation.ts
```diff
@@ -123,7 +123,7 @@ const defaultHandleBatchedChangeFlags = BindingFlags.fromFlushChanges | BindingF
 
 interface IInternalInputElement extends IInputElement {
   matcher?: typeof defaultMatcher;
-  model?: any;
+  model?: Primitive | IIndexable;
```
fkleuver: I should point out that I only learned about / fully understood `unknown` *after* I introduced this `Primitive | IIndexable` "pattern". I never liked it much, only thought of it as slightly better than `any` and very much a temporary solution

Again, it's ok for now (because as I said it's still better than `any`) but I think somehow introducing `unknown` in these cases would still be better going forward

packages/runtime/src/binding/event-manager.ts
```diff
@@ -1,24 +1,20 @@
 import { DI, IDisposable } from '@aurelia/kernel';
 import { DOM, INode } from '../dom';
+import { findOriginalEventTarget } from './resources/self-binding-behavior';
```
fkleuver: Good catch, but can you please switch them around so that `self-binding-behavior` uses the function defined here instead?
Reason is twofold:
1) From a coherent code organization perspective, `event-manager` you could argue is where the event management related logic is centralized.
2) I actually rewrote the one that is declared here recently because it had a few mistakes in it. I re-read the relevant parts of the HTML spec / browser specific documentation to make sure. I haven't done that for the one in `self-binding-behavior` so that one is probably broken.

packages/runtime/src/binding/ast.ts
BBosman: @fkleuver In that case can't we just drop the entire `if` here?

The `for` loop will do nothing as the array is empty and then still `return true;`.

packages/runtime/src/binding/element-observation.ts
BBosman: Good to know.

I'll do about experimenting in the near future with changing the `Primitive` type alias to point to `unknown` and see what pops up and if we can iteratively/easily "fix" it.

packages/runtime/src/binding/ast.ts
fkleuver: Good question. It's a consistency and a performance thing. In many situations the method may do additional processing and/or return something different based on whether there's stuff to process or not. Even when that stuff has no effect, preventing potential additional checks and such by returning early is a perf gain.

This is often how the specs operate as well "if nothing to process, return early", even when it could technically let the loop achieve the same behavior. It makes more explicit that there is absolutely nothing to do if there are no items.

It makes no difference here - there would be no impact on performance or behavior, so you're technically right. But I think letting the loop do the short circuit is not necessarily intuitive (especially when debugging) so I'm on the side of how the specs tend to do it.

I don't feel particularly strongly about it however, I just think doing it like this improves code clarity without being a trade-off.

BBosman:
Switched `findOriginalEventTarget`. Could use a final spot check on that.

packages/runtime/src/binding/event-manager.ts
```diff
@@ -6,8 +6,22 @@ export interface IEventWithStandardPropagation extends Event {
...
+  // old composedPath
+  deepPath?(): EventTarget[];
+
+  // current spec
+  composedPath?(): EventTarget[];
+};
+
 //Note: path and deepPath are designed to handle v0 and v1 shadow dom specs respectively
-function findOriginalEventTarget(event: any): IEventTargetWithLookups {
+export function findOriginalEventTarget(event: Event & CompatibleEvent): EventTarget {
```
fkleuver: Oh man. I just noticed it was the other bloody way around apparently. The one in `self-binding-behavior` is correct. The `CompatibleEvent` type made me realize that (doesn't this produce linting warnings itself by the way?).

And it should be internal, so:
```ts
/*@internal*/
export function findOriginalEventTarget(event: Event & CompatibleEvent): EventTarget {
  return (event.composedPath && event.composedPath()[0]) || (event.deepPath && event.deepPath()[0]) || (event.path && event.path[0]) || event.target;
}
```
Apologies for the trouble!

BBosman: No problem. Made the changes. 😄

packages/runtime/src/binding/ast.ts
BBosman: Makes sense. 👍

PR #265, opened 2018-11-08
feat(di): add transient and singleton decorators

fkleuver:
# Pull Request

## 📖 Description

This PR adds the `@transient` and `@singleton` decorators "back" (they already existed in vCurrent and were gone). The primary methods are exposed on the `DI` object for consistency with `inject` and `Resource.define`. They can be used as decorators and as expressions. Some examples:

```ts


// 1. declaration + decorator
@transient
class Foo { }

// 2. declaration + decorator expression
@transient()
class Foo { }

// 3. declaration + member expression
class Foo { }
DI.transient(Foo);

// 4. declaration + function expression (1. decorator)
class Foo { }
transient(Foo);

// 5. declaration + function expression (2. decorator expression)
class Foo { }
transient()(Foo);

// 6. class expression + member expression (returns class with strongly typed static register)
const Foo = DI.transient(class {});

// 7. class expression + function expression (returns class with strongly typed static register)
const Foo = transient(class {})();
```

We probably wouldn't want to encourage usage 4, 5 or 7. For inline declarations usage 3 and 6 would be the recommended way.

Other than that, usage 1. vs 2. would depend both on preference/style/standard as well as potentially needing any options we may add in the future (in which case usage 2 is needed)

### 🎫 Issues

No related issues

## 👩‍💻 Reviewer Notes

They simply put a static register method on the class that returns a resolver directly. It's how I've seen it done elsewhere with vNext DI by @EisenbergEffect 

Furthermore I've applied overload signatures to give alternative docs/examples based on the kind of invocation (something I've done with other decorators in templating as well)

## 📑 Test Plan

Two simple unit tests are added for each that verify both decorator and decorator expression. The actual registrations and such are already verified by other tests, so didn't bother spending too much time on combinations etc.

## ⏭ Next Steps

This PR is in the light of end user experience, so follow up steps would involve improving typings, method signatures and jsdoc comments. No concrete plans for those however, so anyone is free to pick that up as well.

### DIFF
--- a/packages/kernel/src/di.ts
+++ b/packages/kernel/src/di.ts
@@ -76,6 +76,10 @@ export interface IResolverBuilder<T> {
   aliasTo(destinationKey: Key<T>): IResolver;
 }
 
+export type RegisterSelf<T extends Constructable> = {
+  register(container: IContainer): IResolver<InstanceType<T>>;
+};
+
 // Shims to augment the Reflect object with methods used from the Reflect Metadata API proposal:
 // https://www.typescriptlang.org/docs/handbook/decorators.html#metadata
 // https://rbuckton.github.io/reflect-metadata/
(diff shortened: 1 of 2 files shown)


bigopon:
I'm not sure about 4, as it's either an error for class or it wont matter at all for normal function.

About 5, it's very nice. 👍

fkleuver:
Lol you're right, I switched it the wrong way around. Correct and completed now

PR #266, opened 2018-11-08
chore: update devDependencies and remove package level package-lock.json files

BBosman:
# Pull Request

## 📖 Description

I was working on linting fixes and noticed that a more recent version of `tslint-microsoft-contrib` and `typescript` were available and I wondered if they would fix/introduce new linting warnings, so I decided to update them. And while I was updating I decided to update them all while I was there.

Mostly minor and patch releases, but a few majors.

### 🎫 Issues

n/a

## 👩‍💻 Reviewer Notes

Notable changes:
- `@types/karma` from `^1.7.5` to `^3.0.0`, but we're also already running `karma@^3`.
- `commitizen` from `^2.10.1` to `3.0.4`.
- `husky` from `^0.14.3` to `1.1.3`. This also resulted in `commitmsg` being removed from `scripts` and a new `husky` top element was introduced as a replacement.

Another "big" change is that I removed the package level `package-lock.json` files. The default behaviour of `lerna bootstrap --ci --hoist` is to not create package level `package-lock.json` files, so my assumption is that somebody accidentally ran `lerna bootstrap` without those arguments at one time. The removel is a seperate commit, so I can easily revert it if there is a reason to have package level `package-lock.json` files.

## 📑 Test Plan

What I tested:
- `npm run commit` still works.
- `git commit` messages are still validated through `husky`.
- `npm run build` still runs without errors.
- `npm run test` still runs without failures.
- `npm run lint` still runs with the same warnings.
- `npm run refresh` still works.
- `npm run init` works on a fresh clone of this branch.

## ⏭ Next Steps

n/a


fkleuver:
> We'll have to try and see whether CircleCI and the nightly build agree with this, but I don't see why they wouldn't.

Big whoops. I just realized this will actually break the publishing scripts because they'll be looking for package-locks to update which don't exist anymore.

Can you change this PR from `master` to the `deps` branch I just created for you? Then I'll merge it in and I'll clean up the scripts real quick, and run a verification on CI as well.
You can try fixing them too if you're interested - it's completely safe to run all the bump/publish scripts locally, because they'll just fail with auth errors and such. Everything is reversible by `git reset HEAD~1` etc if it commits stuff.

In either case, let's move it to the `deps` branch in this repo for CI verification.

BBosman:
Switched to `deps` and I'll leave fixing the publishing scripts to you. I'll dig into those eventually, but one step at a time. 😉 

I don't blame you regarding `lerna` and `package-json.lock`, as `lerna` itself is inconsistent and still a bit rough around the edges.

For example: `lerna bootstrap --ci --hoist` won't generate a local `package-lock.json`, but a `lerna add [npmpackagename] --dev --scope=@aurelia/[localpackagename]` will. 
So after doing that I just kill the newly created package level `package-lock.json` and run a `lerna bootstrap --hoist` to update the root `package-lock.json` and all is peachy again. 😄

fkleuver:
> So after doing that I just kill the newly created package level package-lock.json

How about adding a `.npmrc` with `package-lock=false` then, just like we've got in the examples and e2e tests?

PR #267, opened 2018-11-08
chore: update devDependencies and remove package level package-lock.json files

fkleuver:
# Pull Request

## 📖 Description

This is a follow-up to @BBosman's pr https://github.com/aurelia/aurelia/pull/266 to make sure it all still works on CircleCI and to fix the publish script which would still try to modify the non-existing lockfiles.

### 🎫 Issues

Follow-up to https://github.com/aurelia/aurelia/pull/266

## 👩‍💻 Reviewer Notes

In this PR I added `.npmrc` files to the packages themselves so that they won't generate new lockfiles when adding new dependencies via lerna.

I also modified the publish script to no longer try to modify the lockfiles.

## 📑 Test Plan

CircleCI speaks for itself - that was the main thing I wanted to verify here :)

## ⏭ Next Steps

Executed the same steps as @BBosman  in his PR https://github.com/aurelia/aurelia/pull/266

Also ran `npm run publish:dev` locally and verified the version bump happens correctly.


fkleuver:
@BBosman are you behind the addition of `.npmrc` or do you see other issues with that?

BBosman:
@fkleuver I don't have any working knowledge of having having`package-lock=false` in an `.npmrc`. To be honest I didn't know it was an option. 😄 But the idea is sound and I don't for-see any issues with it. (And I made a mental note for myself to do this in our work mono-repo on Monday as well 😉)

And if I'm not mistaken, these changes also allow us to remove all packages from the top `package.json` that are only used in child `package.json` files. They only need to be in the top level `package-lock.json`. But this would need some testing, as I'm not entirely sure that doesn't have side effects. Definitely something for a follow up PR.

fkleuver:
Great, thanks for your thoughts :) I think this is a really nice improvement (a -100k LOC improvement) that makes the repo more lightweight and future packages easier to add/update.

> And if I'm not mistaken, these changes also allow us to remove all packages from the top package.json that are only used in child package.json files. They only need to be in the top level package-lock.json

I'm sure there is some redundancy here but I'm also a bit cautious with having deps exclusively in a package-lock.json. That doesn't sound very transparent and might confuse people in the future.

I don't think there are that many though. Most of the test related stuff for example can both be run on a package level and on a root level (not necessarily by lerna). Are there any tangible benefits from cleaning up the root package-lock further?

PR #268, opened 2018-11-10
fix(binding): add null check to check for existing behaviorKey

fkleuver:
# Pull Request

## 📖 Description

Fixes an issue with binding behaviors that had no test coverage yet. It surfaced when @Alexander-Taran tried to run the krausest benchmark. When re-binding a binding behavior it only checks if the behaviorKey is undefined, but it needs to check for null as well since that's what it's set to during unbind.

### 🎫 Issues

n/a

## 👩‍💻 Reviewer Notes

It's just one line of code :)

## 📑 Test Plan

Verified the fix by running the krausest benchmark. Breaks no existing tests. Need to add additional tests (was already on my todo) for repeater mutations + binding behaviors after the lifecycle PR is merged in, since the test would break right after merging otherwise.

## ⏭ Next Steps

Add the appropriate tests

### DIFF
--- a/packages/runtime/src/binding/ast.ts
+++ b/packages/runtime/src/binding/ast.ts
@@ -241,7 +241,7 @@ export class BindingBehavior implements IExpression {
     if (!behavior) {
       throw Reporter.error(RuntimeError.NoBehaviorFound, this);
     }
-    if (binding[behaviorKey] !== undefined) {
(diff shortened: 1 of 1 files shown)


fkleuver:
You mean strict mode?

bigopon:
Yes

PR #269, opened 2018-11-10
WIP docs(testing): explain the integration testing strategy for the jit

fkleuver:
# Pull Request

## 📖 Description

This readme is meant to provide some background and context to the combinations of (generated and non-generated) template compiler integration tests in the jit, so folks can understand what's going on without having to reverse engineer those huge files.

Will continue with this document in the next couple of days. Meantime feedback is welcome as to whether this is actually helpful or not.


EisenbergEffect:
Can we move this into the docs/engineering folder?

fkleuver:
We could, but my idea was to have links in the engineering docs pointing to those located elsewhere. That would keep things centralized but also having docs close to the pieces they relate to makes it easier to quickly find related information. Or so I'm thinking. What do you think about that?

EisenbergEffect:
I'd prefer to point it back the other way, by keeping the full doc with the rest of the documentation and dropping a readme near the tests that points to it. It's not a huge deal though. We can move it around later if we have some other need.

PR #277, opened 2018-11-12
chore: script linting fixes

BBosman:
# Pull Request

## 📖 Description

Linting fixes for `scripts`.

Mostly syntactic changes. Left all `any` references and things requiring larger rework for a later time.

### 🎫 Issues

- #249 

## 👩‍💻 Reviewer Notes

These changed need some close scrutiny, as I did some basic testing, but I've never used these scripts *in anger*, so I don't know all the edge cases.

## 📑 Test Plan

I must admit that the changes have had limited testing as there are no unit tests for them. Ran the scripts and as far as I can tell they still work.

## ⏭ Next Steps

Eventually fix the rest of the linting issues within the `scripts` folder, but as the remaining issues require rework and testing the code is a bit cumbersome/difficult it's not high on my priority list.


scripts/test-lib.ts
```diff
@@ -332,53 +332,52 @@ function $getAllPropertyDescriptors(proto: Object, filter: (pd: PropertyDescript
...
+  arrays: [(() => T1)[], ((arg1: T1) => T2)[], ((arg1: T1, arg2: T2) => T3)[], ((arg1: T1, arg2: T2, arg3: T3) => T4)[], ((arg1: T1, arg2: T2, arg3: T3, arg4: T4) => T5)[], ((arg1: T1, arg2: T2, arg3: T3, arg4: T4, arg5: T5) => T6)[], ((arg1: T1, arg2: T2, arg3: T3, arg4: T4, arg5: T5, arg6: T6) => T7)[], ((arg1: T1, arg2: T2, arg3: T3, arg4: T4, arg5: T5, arg6: T6, arg7: T7) => T8)[], ((arg1: T1, arg2: T2, arg3: T3, arg4: T4, arg5: T5, arg6: T6, arg7: T7, arg8: T8) => T9)[], ((arg1: T1, arg2: T2, arg3: T3, arg4: T4, arg5: T5, arg6: T6, arg7: T7, arg8: T8, arg9: T9) => T10)[], ((arg1: T1, arg2: T2, arg3: T3, arg4: T4, arg5: T5, arg6: T6, arg7: T7, arg8: T8, arg9: T9, arg10: T10) => T11)[], ((arg1: T1, arg2: T2, arg3: T3, arg4: T4, arg5: T5, arg6: T6, arg7: T7, arg8: T8, arg9: T9, arg10: T10, arg11: T11) => T12)[]],
```
fkleuver: oh man :laughing:

fkleuver:
Wow. Did tslint complain about this? I was sure these were all ignored..

BBosman:
They're ignored when running `npm run lint`, but for some reason they weren't in my IDE.

So when I went looking to the scripts to learn what they did I noticed some missing semicolons, so I fixed those. And some double instead of single quotes, so I fixed those. And then I decided to order the imports. And then... And then... And then I ended up with this PR. 😁

fkleuver:
> And then... And then... And then

Tell me about it :)

Well I wouldn't want to go through another several days trial-and-error marathon of trying to get the whole workflow to work, so as long as the only changes are in typings (no actual functional js changes) I don't see any issues.

Did the space injections in the test suite and cartesian join functions generic parameters make the code more readable to you? I deliberately formatted it like that for compactness to improve readability. But I'm one of those that likes thick, dense code so lots of it fits on a screen. I can deal with the spaces as long as it's more readable for others.

BBosman:
My assumption is that the current set of linting rules is there for a reason, so my preference is to fix instead of suppress.

But I do agree that in this case the extra whitespace might actually be doing the opposite for readability.

Want me to reverse and suppress those specific whitespace fixes?

fkleuver:
> My assumption is that the current set of linting rules is there for a reason, so my preference is to fix instead of suppress.

I fully agree with the notion of fix instead of suppress.

We haven't necessarily evaluated and hand picked every rule deliberately. It's based on the microsoft-contrib recommended rules, most of which is targeted at regular teams with their share of juniors, etc. We're all experienced here so some of the rules just don't make sense for this repo (`no-bitwise` is a good example there)

tslint is meant to make it easier for people to dive in and understand the code due to the familiarity etc etc. Some code just doesn't lend itself very well for that in the first place, and different measures are needed to try and improve the situation :)

> Want me to reverse and suppress those specific whitespace fixes?

If you agree about the readability then yes please.

BBosman:
@fkleuver Reverted and suppressed those whitespace changes/warnings.

fkleuver:
Thank you :) rest looks fine to me, appreciate you taking a look at it!

PR #280, opened 2018-11-13
chore: disable promise-function-async linting rule

### LINKED ISSUE #273: [RFC] async usage?
# 💬 RFC

I just noticed c262ab2295c75547a9ad14f2536cfcef52eb4aef by @fkleuver in #250 with a description of `revert aurelia-cli incompatible async change`.

Does that mean we can't use `async` in general, or just in this use case?

## 🔦 Context

The reason I'm asking this is because I added it specifically in #262 to fix a `promise-function-async` linting error. I don't have any issues with the revert, but I am curious about the rationale 😊.

If we can't use `async` in general, we should disable the `promise-function-async` rule in `tslint.json`, otherwise we should `tslint:disable-next-line` those specific changes.

## 💻 Examples

n/a.

BBosman:
# Pull Request

## 📖 Description

Disable the `promise-function-async` rule for `tslint` for now.

Paraphrasing @fkleuver from https://github.com/aurelia/aurelia/issues/273#issuecomment-437924643:

> The Esprima parser version that ships with aurelia-cli only understands ES2015.
> Later syntax causes aurelia-cli to fail in the transform phase. 
> Until aurelia-cli is updated (and other popular bundlers!), we need to avoid ES2015 incompatible syntax altogether.

As we're not allowed to use `async` for now, it's counterproductive to have the linter complain about it missing.

### 🎫 Issues

- Related to #249 
- Fixes #273 

## 👩‍💻 Reviewer Notes

Disabling the rule altogether (for now) is a better solution than sprinkling the code with `// tslint:disable` statements imho. 

## 📑 Test Plan

Ran:
- `npm run lint` and it didn't warn about `promise-function-async` issues anymore.

## ⏭ Next Steps

Hope that Esprima and other bundlers start supporting > ES2015 syntax soon. 😜

### DIFF
--- a/tslint.json
+++ b/tslint.json
@@ -91,7 +91,7 @@
     "no-unused-expression": true,
     "no-use-before-declare": false,
     "no-with-statement": true,
-    "promise-function-async": true,
+    "promise-function-async": false, // See https://github.com/aurelia/aurelia/issues/273
     "promise-must-complete": true,
     "radix": true,
     "react-this-binding-issue": true,


fkleuver:
TLDR: I agree, let's turn off this rule

Just a few small corrections/additions for future readers:
- Esprima (and aurelia-cli) does support >ES2015 syntax, it's just going wrong with TypeScript. So my statement there is a bit off
- @huochunpeng Pointed out to me yesterday that the problem likely lies elsewhere, still needs investigation

Regardless, I think it's good to turn off this rule. As @StrahilKazlachev already pointed out when async is transpiled to earlier ES versions it gets very bloated. I don't think of this as a huge concern but it still is a valid one.
With or without async, we generally want to avoid promise-returning code in the framework as much as possible in the first place as a means to keep things simple and easy to reason about.

Issue #182, opened 2018-09-17
Decouple renderer


bigopon:
At the moment, it seems to me `Renderer` has very weak connection with the rest of runtime, and only strongly coupled with instructions list. It seems to make sense to me that we make those instructions at externally configurable, together with a plug-able renderer for those instructions to depend on. This way, we can have default combo of compiler and renderer for web platform. To target different platform, we can provide different combo instead of our current structure. This requires some changes in the way how `runtime` understands the basic behavior, or we can even make all behavior externally configurable, like .. there is no such thing as custom element. It's a renderer thing.
@EisenbergEffect @fkleuver

fkleuver:
Is this different from the `@renderStrategy` decorator I added a while ago?

bigopon:
I think it could be more than that. By separating out the renderer and its infrastructure, we can achieve more than render strategy. Atm there is a fixed set of core instructions, but those instructions are relevant to only renderer and a few more. Maybe should be configurable in bindingcommand resource way instead of patching the prototype way

fkleuver:
The only difference between `@renderStrategy` and `@bindingCommand` is in the storage mechanism; the binding commands are registered and resolved exclusively via DI, whereas render strategy is registered with DI, resolved once, and then stored on the prototype. This is for performance because the renderer is such a hot path.
Are you looking for the flexibility of having different sets of render strategies for different component contexts? That's the only difference there is atm from a functional point of view.

bigopon:
The renderer contract with rendering engine at the moment is quite simple: only a render method that takes a renderable, targets, definition, host and replacement parts. From that joint, we can probably break the renderer out of runtime and let it sit in its own package, or merge with JIT. Doing so we have a plug-able renderer, and it could make more sense for renderer and compiler to sit together in same package. Then registering different targeted instruction type will be made configurable, unlike what we have at the moment, always assume there is a fixed set of core instruction type

EisenbergEffect:
I don't think it can move to the JIT because we need it at runtime for AOT.

bigopon:
When we move it out of runtime, there is nothing stopping us from having different renderers for different target. JIT, AOT, each of them can have their own renderer, or they can share the same renderer. say AOT targeting web will share the same renderer with JIT-web etc

EisenbergEffect:
[...]

Issue #165, opened 2018-09-13
View compiler extension point


bigopon:
At the moment, we have very nice view compilation model, thanks to binding command resource from @fkleuver . However, there is something still lacking: a way to handle an arbitrary unknown binding command - attribute combination.

```html
<div
  background-color.style='bgColor'
  style.color='color'
  selected.class='selected'
  class.selected='selected'
  ref='div'></div>
```

With the above usage, declaring new binding command does not always give the ability to handle the scenario. I think we may want to introduce `inspectAttribute`, that will be called before anything else to give a chance to have those highly dynamic usage a solution. Our default implementation will be a noop function, or a function that handle `ref` only

@fkleuver @EisenbergEffect

fkleuver:
Keep in mind we're completely parsing every attribute and attribute value. This has a huge impact on larger applications with a lot of css classes and whatnot, and is a non-negligible contributor to vCurrent's startup time.
And of course when people start intercepting `inspectAttribute` this can easily go wrong in all sorts of ways via little mistakes.

What I'd like to try instead is allow user to define custom patterns in a similar way to how they can define routes, using a NFA of sorts (similar to `route-recognizer`). Not entirely sure about the pattern syntax yet but I thought about [re2](https://github.com/google/re2/wiki/Syntax) (or a subset thereof) which is basically a subset of regex that is faster/easier to parse than normal regex.

I'd propose doing it via `@bindingSyntax` decorators, and it would look roughly like this:

```ts
@bindingSyntax('([^\.]+)\.([^\.]+)') // vCurrent dot binding syntax - generic

@bindingSyntax('([^\.]+).*\.([^\.]+)') // vNext binding syntax - generic
```

```ts
export interface IAttributeInspector {
    inspectAttribute(resources: IResourceDescriptions, ...parts: string[]); InspectedAttribute | null;
}

@attributeInspector('([a-zA-Z0-9\.\-_])\.([a-z\-]+))')
export class DefaultAttributeInspector implements IAttributeInspector {
    public inspectAttribute(resources: IResourceDescriptions, target: string, command: string); InspectedAttribute | null {
       // return null if the command is unknown, and let the next inspector try
    }
}
```

Registering more than one inspector would essentially merge the patterns. Priority would be determined based on most specific to least specific match, etc.

Such that:

```ts
@attributeInspector('style\.([a-z\-]+))')
```

Will take precedence over the default:

```ts
[...]

PR #281, opened 2018-11-13
renderer refactor

fkleuver:
# Pull Request

## 📖 Description

This PR extracts the instruction-specific processing of the renderer out into one class per instruction, called `InstructionRenderers` (name could be better).
What's left in the renderer itself is simple aggregation logic in the constructor that puts all registered instruction renderers into a `instructionRenderers` object map where the key is the instruction type (just like in the previous renderer), as well as the platform-agnostic loops that distribute the targets and instructions to the instruction renderers.

### html-renderer object
The instruction renderers reside in html-renderer.ts and are exposed via an `HtmlRenderer` object that can be passed into DI. It contains the registrations for the renderers. The `all` DI injection function is used to inject a list of all registered instruction renderers into the `Renderer` class.

In effect, there is no more browser-specific code in the renderer or rendering engine and implementing alternative platforms like pixi or NativeScript should be a matter of registering a different set of instruction renderers (which may need to construct alternative bindings that use different observers, in some cases).

### instructionRenderer decorator
There is a new decorator, `@instructionRenderer` which replaces the previous renderer extension point `@renderStrategy`.
The `renderStrategy` instruction type is also gone as it is no longer needed. `@instructionRenderer` can be used for both alternative environments as well as for simply extending the browser environment with custom rendering for custom `@bindingCommand`s.

The `@instructionRenderer` takes the instruction type (string) as an argument which becomes the instruction type string that this renderer will be able to handle.

### 🎫 Issues

Directly related to: https://github.com/aurelia/aurelia/issues/182
Loosely related to: https://github.com/aurelia/aurelia/issues/165

## 👩‍💻 Reviewer Notes

Example can be seen in `packages/jit/test/unit/binding-command.spec.ts` where there is a custom bindingCommand `"keyup"` with a corresponding same-name `instructionRenderer`, (was previously `renderStrategy`)

### Renderer is instantiated/resolved differently
`Renderer` is no longer constructed per template but instead registered via DI as a singleton. The underlying assumption is that you only need one set of renderers per Aurelia instance (this seems like a pretty safe assumption to make). To facilitate this the `RenderContext` is passed around via an additional method parameter instead of stored in the renderer, and the `RenderingEngine` uses DI to resolve the singleton `Renderer` instance.

Render instructions also use DI to get only those specific dependencies that they need, such as parser/observer locator, etc. The hydrate instructions that need the renderer itself, use the RenderContext to resolve it.

### InstructionType is set by the decorator
`@instructionRenderer` replaces the constructor with a wrapper that sets the instruction type on the instance. This is effectively the same thing as `this.instructionType = 'x'`, but saves users that trouble (setting it on the instance is better for perf than setting it on the prototype because it's read very often)

### An addition to BasicConfiguration
The `HtmlRenderer` object is added to the `BasicConfiguration` object in the JIT so for end users nothing additional needs to be done. Using a different renderer would mean using a different configuration that we would expose.

### Fix in DI
The `all` function wasn't working correctly (called the wrong method), fixed that as well.

## 📑 Test Plan

A few small changes were needed to make all existing tests pass again. No additional tests were added because no functionality was added (a small `@instructionRenderer` specific test was already present in the JIT in the form of `@renderStrategy`)

## ⏭ Next Steps

- Implement an additional renderer, such as pixi renderer or NativeScript (@bigopon could pick this up if he wants to, that would be a nice test to see if the extension point is good enough)
- Proceed with completing the template compiler (surrogates) and adding an extension point there as well
- Implement rendering of replace-parts?


bigopon:
> refactor(renderer): decouple instruction-specific processing from the renderer

that commit **ლ(´ڡ`ლ)**

fkleuver:
@bigopon @EisenbergEffect This is ready for review. All tests pass and everything seems to be working.

It kind of feels like "this was too easy" for the nature of the change / I'm missing something, but it works so I wouldn't know what. Related template compiler changes are for a follow-up PR in any case.

Let me know if anything is missing.

bigopon:
@fkleuver Lgtm. This is how i hoped it to be when mentioning decoupling the renderer. Not sure if this opened the opportunity for us to move the renderer to a level lower, aka completely separate all infra of instruction renderers. Atm, custom element and custom attribute are core assumptions, they could be hard to be separated. The main benefit of that is when doing integration with other platform, like pixi, phaser, native script or react natice, we may need very different custom element and custom attribute impl

fkleuver:
> Not sure if this opened the opportunity for us to move the renderer to a level lower, aka completely separate all infra of instruction renderers

Can you be more specific on what you mean by "infra"?

> Atm, custom element and custom attribute are core assumptions, they could be hard to be separated. The main benefit of that is when doing integration with other platform, like pixi, phaser, native script or react natice, we may need very different custom element and custom attribute impl

If you mean we may need different lifecycle methods such as `hydrate`, `bind` and `attach` with different behavior - those aren't tied to anything except from the decorators. Decorators could technically apply different lifecycles with other params, or we could add extra decorators for those situations.

The only true "assumption" made by the runtime is the existence of the `ICustomElement` and `ICustomElementType` (ditto for attribute) interfaces. They only describe very high-level method signatures and properties. If those would need to differ, we might as well have an alternative RenderingEngine implementation (that's the free benefit we have with everything that has an exported interface symbol).

And if you're talking about having alternative bindings, observers etc - that's precisely what this PR covers.

I'm all for decoupling things a bit more because there's still a high dependency density in areas that only need small parts of certain high-level interfaces, but I'm also wary of over-architecting the compiler+renderer section of the framework.

bigopon:
Maybe dependency could have been a better word.

`Renderer` <-- (1) Instruction renderer <-- (2) binding <-- (3) observer locator <-- (4) observer

Infra of instruction renderer covers (2) (3) & (4), and for that, it could be nice if (1) ... (4) could be separated from `Renderer` as Renderer now needs no knowledge about those. This could also affect compiler. For example:

```js
export class TextBindingRenderer {
  static createInstruction(prop, srcOrExpr) {
    return { };
  }
  
  ....rendering related code
}
```

Renderers will be responsible for creating their own instruction / render them.

fkleuver:
> as Renderer now needs no knowledge about those

I think I'm misunderstanding. The renderer currently has no knowledge of those. The only notable interfaces it knows about are `IRenderable` (which could be narrowed down to `IBindables & IAttachables`) and `TemplateDefinition` (which could be narrowed down to `{ instructions: ITargetedInstruction[] }`). It certainly doesn't know about concrete bindings, the observer locator, or observers.

> Renderers will be responsible for creating their own instruction / render them.

Creating their own instruction how? The compiler currently creates those based on the parsed html syntax (elements, attributes, text nodes) and the available resources. That is (and should probably stay) a JIT concern.
Decoupling the instruction-specific compilation logic from the template compiler so it can be told to create different instructions for different renderers is the next step (view compiler extension point).

I don't think we need much more beyond that to be able to support other runtimes?

fkleuver:
> let them have a static method to create the compatible instruction

What would be in there? The logic that's currently in the instruction classes? Can you give an example

>  so maybe they (renderer and instruction renderer) dont need to sit in the same package

Well it's already in its own `html-renderer.ts` file. I did that very deliberately to hint at it being ripe for going to its own package. But I think it would need to go together with HTML-specific pieces of DOM and observation into some `html` package. This is something I was working on previously and this (as well as the next PR) are pretty much the missing pieces that should allow us to extract html out into its own package.

bigopon:
> What would be in the

```js
// 1
export class TextBindingInstruction implements ITextBindingInstruction {
  public type: TargetedInstructionType.textBinding = TargetedInstructionType.textBinding;
  constructor(public from: string | Interpolation) {}
 }

// 2
@inject(IExpressionParser, IObserverLocator)
@instructionRenderer(TargetedInstructionType.textBinding)
/*@internal*/
export class TextBindingRenderer implements IInstructionRenderer {
  private parser: IExpressionParser;
  private observerLocator: IObserverLocator;
   constructor(parser: IExpressionParser, observerLocator: IObserverLocator) {
    this.parser = parser;
    this.observerLocator = observerLocator;
  }
   public render(context: IRenderContext, renderable: IRenderable, target: INode, instruction: ITextBindingInstruction): void {
    const next = target.nextSibling;
    DOM.treatAsNonWhitespace(next);
    DOM.remove(target);
    let bindable: MultiInterpolationBinding | InterpolationBinding;
    const expr = ensureExpression(this.parser, instruction.from, BindingType.Interpolation);
    if (expr.isMulti) {
      bindable = new MultiInterpolationBinding(this.observerLocator, expr, next, 'textContent', BindingMode.toView, context);
    } else {
      bindable = new InterpolationBinding(expr.firstExpression, expr, next, 'textContent', BindingMode.toView, this.observerLocator, context, true);
    }
    addBindable(renderable, bindable);
  }
}
```

Could be merged into 1 like this:
```ts
// 2
@inject(IExpressionParser, IObserverLocator)
@instructionRenderer(TargetedInstructionType.textBinding)
/*@internal*/
export class TextBindingRenderer implements IInstructionRenderer {

  public static instruction(from: string | Interpolation): ITextBindingInstruction {
    return { from, type: TargetedInstructionType.textBinding };
  }

  private parser: IExpressionParser;
  private observerLocator: IObserverLocator;
   constructor(parser: IExpressionParser, observerLocator: IObserverLocator) {
    this.parser = parser;
    this.observerLocator = observerLocator;
  }
   public render(context: IRenderContext, renderable: IRenderable, target: INode, instruction: ITextBindingInstruction): void {
    // ...
    // ...
  }
}
```

fkleuver:
Man. Why didn't you just say "move constructor logic from the instruction classes into the renderers, remove instruction classes" 🤣 
It kinda makes sense. I added those classes partly for performance reasons but primarily to simplify the construction from the compiler point of view. The renderers could be doing it. However, that would couple the compiler to the instruction renderers and I'm not sure I like that idea..

bigopon:
That process could also trigger restructuring related to compiler, renderer, JIT & runtime organization 😜

So better describe it a bit lengthier

fkleuver:
I see where you're coming from. I think this is not the right PR to be discussing the instruction creation and template compiler integration. I very deliberately stopped here to prevent another lifecycle-magnitude PR. We can always nitpick a bit more on the template compiler PR and make a few last minute tweaks to the renderers from there imo

fkleuver:
@EisenbergEffect fixed :)

Issue #254, opened 2018-11-02
[RFC] Use TypeScript strict compiler option?


timfish:
# 💬 RFC

Firstly I'd like to say 👍 for choosing TypeScript. It makes it much easier to contribute without requiring intimate knowledge of the entire codebase. 

Admittedly, we haven't gone back over large legacy TypeScript projects and enabled strict mode, but when we start a new project, or refactor into modules, we now always default to enabling the TypeScript strict compile option.

This is a shortcut for enabling: 
```
noImplicitAny
noImplicitThis 
alwaysStrict 
strictNullChecks
strictFunctionTypes 
strictPropertyInitialization
```

We've found it a great way to improve code quality as `strictNullChecks` specifically forces developers to consider all the places where something might be undefined. 

Maybe you've already considered this and decided that the performance overhead of extra null checks is not worth it? You almost certainly wouldn't want it enabled for tests.

If strict mode is too cumbersome for Aurelia core, its still worth using 'nullable' type annotations for public API's (ie. `ISomeElement | undefined`) so that consumers of Aurelia using strict mode know exactly what return values they have to cater for.

Once the vNext changes have settled down, I can devote some time to work on this, I just didn't want to do that if you've already discounted it usage!

fkleuver:
Thank you for opening this issue, I think it's good to have a dedicated discussion about the typings.

I think I speak for most of the core team (myself included) when saying: the stricter, the better.

Not only does it help find and prevent bugs, it's also a form of documentation. I for one am a big advocate of self-documenting code: if you can explain logic by picking the right variable names and organizing the code in a certain manner, then that's better than by doing it via comments.

The drawback of TS/JS is that, unlike with a compiled language such as C#, more verbose ways of writing something often leads to worse performance. The most performant code in JavaScript also tends to be the least readable code. So wherever we can improve that situation via stricter typings (and JS doc comments for intellisense), we definitely should.

> Maybe you've already considered this and decided that the performance overhead of extra null checks is not worth it?

We actually left it out because the code was changing so frequently and in such large volumes that strict mode would create too much overhead with refactoring. We needed to iterate fast without having to change along a ton of typings all the time.

In the mean time I've been relying on unit and integration tests to pound the code with as many combinations of nonsense as possible, in order to discover potential edge cases and patch them up where needed.

All that said, I think now is actually an okay time to start looking at strictness of typings in the less volatile areas of the code.

[...]

PR #282, opened 2018-11-14
Strict properties

fkleuver:
# Pull Request

## 📖 Description

This PR is an initial step in preparation for turning on the tsconfig.json options `strictPropertyInitialization` and `strictNullChecks` (the latter must be turned on in conjunction with the former).

Principles I applied:
1. No constructor short-hands; all properties are declared on class level (not required by strict property rule)
2. No declaration short-hands; properties are exclusively assigned inside the constructor (not required by strict property rule)
3. All properties are assigned a value during construction (required by strict property rule)
4. Similar properties on similar classes are declared and assigned in the same order as much as possible (not required by strict property rule)
5. Properties that are assigned conditionally and are difficult to keep in the same order per instance of the same class, are assigned a default value eagerly

I added 1. and 2. for consistency and improved debugging experience. It also makes properties easier to document and gives better (sense of) control over the order of assignment. It makes the code look more verbose, but the transpiled JS should be unaffected by this duplication because property declarations are removed anyway.

Point 4. reduces the learning curve when trying to understand multiple similar classes.
Point 5. is in many cases a minor perf improvement for the browser JIT because fewer variations of hidden classes need to be generated

Apart from that, this PR is also just meant as a motivating example for how to approach the rather large effort of enabling strict property initialization, to get the ball rolling, and perhaps to get the discussion going as well (I'm curious what @timfish thinks about this approach)

### 🎫 Issues

Related: https://github.com/aurelia/aurelia/issues/254

## 👩‍💻 Reviewer Notes

I went over all classes in the kernel and in the jit. No functional changes apart from property assignment order.
Pinging @timfish and @BBosman to glance over as well, since they brought this topic to the foreground from the tsconfig / tslint perspectives, respectively.

## 📑 Test Plan

If all CI checks pass, it should be good

## ⏭ Next Steps

The AST, all `IBinding` implementations and binding-resources (most of binding, really) seem like good next candidates since they're quite stable. If @timfish or @BBosman is interested in helping out that would be great. 

The observers have somewhat non-obvious interplay with typings and properties (my bad), and I think it's better if I take a look at that myself, but any ideas are of course welcome.

### DIFF
--- a/packages/jit/src/attribute-parser.ts
+++ b/packages/jit/src/attribute-parser.ts
@@ -2,11 +2,17 @@ import { DI } from '@aurelia/kernel';
 import { Char } from './common';
 
 export class AttrSyntax {
-  constructor(
-    public readonly rawName: string,
-    public readonly rawValue: string,
-    public readonly target: string,
-    public readonly command: string | null) { }
+  public readonly rawName: string;
+  public readonly rawValue: string;
+  public readonly target: string;
+  public readonly command: string | null;
+
+  constructor(rawName: string, rawValue: string, target: string, command: string | null) {
+    this.rawName = rawName;
+    this.rawValue = rawValue;
+    this.target = target;
+    this.command = command;
+  }
 }
 
 export interface IAttributeParser {
--- a/packages/jit/src/binding-command.ts
+++ b/packages/jit/src/binding-command.ts
@@ -71,7 +71,12 @@ export interface OneTimeBindingCommand extends IBindingCommand {}
 @bindingCommand('one-time')
 export class OneTimeBindingCommand implements IBindingCommand {
   public static inject: Function[] = [IExpressionParser];
-  constructor(private parser: IExpressionParser) {}
+
+  private parser: IExpressionParser;
+  constructor(parser: IExpressionParser) {
+    this.parser = parser;
+  }
+
   public compile($symbol: IAttributeSymbol): TargetedInstruction {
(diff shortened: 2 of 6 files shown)


BBosman:
I like what I'm seeing and I'm willing to give it a try on the AST.

There are a few instances where it isn't obvious what the default value should be, like:
```ts
public connect: IExpression['connect'];
```

What would you suggest as default value in this case?

fkleuver:
Whenever there is a method signature declared on a property and it is not assigned in the constructor, 99% chance I've assigned `PLATFORM.noop` on the prototype somewhere down below in the same file. I'm doing this in various observers as well.

The motivation there is performance and library size. It's less code to assign the same function to the prototype, and it's more performance to reuse the same function multiple times. However, the latter is in theory and I'm not sure if there's actually much tangible gain to it in practice.

The problem with directly assigning `PLATFORM.noop` in the instance is that you then get a function per instance instead of one shared across all instances of a prototype, which leads to more properties defined and thus a higher memory usage. Again, it's incredibly small, so perhaps not worth it.

Perhaps a regular empty function with only a `return;` statement is the way to go.

EisenbergEffect:
@fkleuver Recent merged require conflict resolution on this PR.

fkleuver:
@EisenbergEffect fixed

EisenbergEffect:
I'm down with this approach. While certain things are slightly less convenient, it provides greater clarity and consistency which I think are crucial, esp. when some of this can affect perf in hot code paths (something we always have to be sensitive of in a framework). @fkleuver Ready for me to merge this?

fkleuver:
Yep ready

PR #285, opened 2018-11-15
test(kernel): enable nodejs tests and comment out non-compilable ts

fkleuver:
# Pull Request

## 📖 Description

Initial step towards adding nodejs to the tested runtimes. Took the kernel as an easy package to start with.

Enabling this in the runtime and jit, even with shimming variables and conditionally disabling certain tests, will require a bit more work. So this is not yet enabled on CircleCI or added to the root-level lerna scripts.

## 👩‍💻 Reviewer Notes

I disabled a few DI tests that apparently just can't be compiled by typescript. `ts-loader` in webpack uses some interesting tricks to force TS to transpile it anyway. I'm not entirely sure if there was any value in those tests in the first place but I'm not removing them just in case.

Any feedback/ideas are welcome


fkleuver:
Hmm getting a few weird errors on CI. Back to the drawing board with this one :)

PR #272, opened 2018-11-12
chore(runtime): decorate bindingBehavior

BBosman:
# Pull Request

## 📖 Description

Add `Decorate` and `Decorated` typeguards to `@bindingBehavior`.

### 🎫 Issues

Related to https://github.com/aurelia/aurelia/pull/262#discussion_r230879707 and #264.

## 👩‍💻 Reviewer Notes

Just `@bandingBehaviour` for now, to validate my interpretation of #264. 

## 📑 Test Plan

Ran:
- `npm run lint` without introducing new warnings
- `npm run build` without errors
- `npm run test` without test failures

## ⏭ Next Steps

Once this is done, I'll start rolling this out to other decorators.

Issue #210, opened 2018-10-07
Create test placeholders / instructions for testing


bigopon:
So I thought about our tests, and decided to create this. Things are working and looking good, but we need way more tests. One way to solve it is to ask @fkleuver to do everything, but it will probably be too much and take long time for him to finish everything alone, so I think I'll walk through every missing tests and create placeholder for them, together with some specs, such as what should be tested, what scenarios should be validated. Then we can easily split them up into individual issues and label `test + help-wanted`. That way, we can invite community members to share fun building Aurelia. Thoughts ? @EisenbergEffect @fkleuver

fkleuver:
Love the idea! I'm very happy that you're proposing this :)

EisenbergEffect:
Sounds like a great idea to me too!

fkleuver:
Just so I don't forget (and others know what's still to-do), some notes on which things I believe need significantly more test coverage:
### Kernel
- DI jitRegister
- DI interpret
- DI childContainers

### Runtime - Binding
- ComputedObserver (this also still needs to be fixed)
- LengthObserver
- DirtyChecker
- **all** binding resources (complex integration test combinations with other binding and templating resources)
- AST (specifically BindingBehavior and ValueConverter)
- Call
- EventManager
- ExpressionParser (the runtime one)
- LetBinding
- Listener
- Ref

### Runtime - Templating
- Compose (complex integration test combinations with other resources and specifically async)
- Replaceable
- With (complex scope situations such as replace-parts)
- createElement (complex combinations of inputs)
- Lifecycle (exotic usage of hooks with/without async task registrations, flush/bind/others getting queued from within hooks, verifying the only-detach-roots behavior with mocks)
- Aurelia (unit tests that verify the various non-obvious task/start/stop combinations)

PR #286, opened 2018-11-15
refactor: add typeguards to decorators and rename Source to Definition

BBosman:
# Pull Request

## 📖 Description

Decorated `@bindingCommand`, `@valueConverter`, `@customAttribute` and `@customElement` with `Decorate` and `Decorated` typeguards.

Also added a base `IResourceDefinition` interface per the discussion at https://github.com/aurelia/aurelia/pull/272#issuecomment-438000257.

### 🎫 Issues

Follow up to #272.

## 👩‍💻 Reviewer Notes

- This includes the next step in the `Source` to `Definition` rename. I think I got all occurrences but a check couldn't hurt.
- All decorators define their `ISomething` and `ISomethingType` interfaces in the same file, with the exception of `customElement` and `customAttribute`. I decided to move those types to their corresponding files as well for consistency. Is this the right approach?
- Added and reverse engineered `IValueConverter` from current usage in the codebase, so a check if it's valid would be appreciated.
- Did an `any` -> `unknown` conversion on all `@valueConverters` as a result of the previous item.

## 📑 Test Plan

*Works on my machine*™️ Lets hope CircleCI agrees 😜.

## ⏭ Next Steps

n/a.

### DIFF
--- a/packages/jit/src/binding-command.ts
+++ b/packages/jit/src/binding-command.ts
@@ -1,4 +1,4 @@
-import { Constructable, IContainer, IRegistry, Registration, Writable } from '@aurelia/kernel';
+import { Constructable, Decoratable, Decorated, IContainer, IRegistry, Registration, Writable } from '@aurelia/kernel';
 import {
   BindingType,
   CallBindingInstruction,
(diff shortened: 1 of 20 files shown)


packages/jit/src/binding-command.ts
```diff
@@ -48,8 +51,8 @@ export const BindingCommandResource: IResourceKind<IBindingCommandSource, IBindi
     return Type.kind === this;
   },
 
-  define<T extends Constructable>(nameOrSource: string | IBindingCommandSource, ctor: T): T & IBindingCommandType {
-    const description = typeof nameOrSource === 'string' ? { name: nameOrSource, target: null } : nameOrSource;
+  define<T extends Constructable>(nameOrDefinition: string | IBindingCommandDefinition, ctor: T): T & IBindingCommandType {
```
fkleuver: Since this is also public API (users can call this directly on a declared or inline class), I think for consistency here we'll want to have the same overload signatures as the decorator that calls this.

packages/runtime/src/templating/custom-attribute.ts
```diff
@@ -1,30 +1,51 @@
...
+import { $hydrateAttribute, IRenderingEngine } from './lifecycle-render';
+
+type OptionalHooks = ILifecycleHooks & Omit<IRenderable, Exclude<keyof IRenderable, '$mount' | '$unmount'>>;
+type RequiredLifecycleProperties = Readonly<Pick<IRenderable, '$scope'>> & IState;
+
+type CustomAttributeStaticProperties = Pick<AttributeDefinition, 'bindables'>;
+
+export type CustomAttributeConstructor = Constructable & CustomAttributeStaticProperties;
+
+export interface ICustomAttribute extends Partial<IChangeTracker>, IBindScope, ILifecycleUnbindAfterDetach, IAttach, OptionalHooks, RequiredLifecycleProperties {
```
fkleuver: I agree this is a better place for these types :)

packages/jit/src/binding-command.ts
BBosman: As far as i know you can't put overloads in an object literal. Or did you mean we should put the overloads on the interface?

BBosman:
Made the requested tweaks.

packages/runtime/src/binding/value-converter.ts
```diff
@@ -1,19 +1,24 @@
-import { Constructable, IContainer, Registration, Writable } from '@aurelia/kernel';
-import { IResourceKind, IResourceType } from '../resource';
+import { Constructable, Decoratable, Decorated, IContainer, Registration, Writable } from '@aurelia/kernel';
+import { IResourceDefinition, IResourceKind, IResourceType } from '../resource';
 
-export interface IValueConverterSource {
-  name: string;
+export interface IValueConverter {
+  toView(input: unknown, ...args: unknown[]): unknown;
+  fromView?(unknown: string): unknown;
```
fkleuver: Wait what? 😆

BBosman: Doh! 😬 

Funny that it passed all automated tests though. 😊

packages/jit/src/binding-command.ts
BBosman: Might as well include it in this PR, as I'm touching all those files anyway.

BBosman: @fkleuver Done.

packages/runtime/src/binding/value-converter.ts
fkleuver: Test coverage on binding resources is crap :) I still have [20-30k tests to write](https://github.com/aurelia/aurelia/issues/210#issuecomment-438812664) for runtime/jit

EisenbergEffect:
Thanks for the detailed work @BBosman !

PR #287, opened 2018-11-15
refactor: fix errors and warnings reported by "lgtm"

BBosman:
# Pull Request

## 📖 Description

Fixes almost all errors and warnings as reported by [lgtm](https://lgtm.com/projects/g/aurelia/aurelia/alerts/?mode=list)

### 🎫 Issues

Related to #249 

## 👩‍💻 Reviewer Notes

These could use some scrutiny:
- `examples/jit-browserify-ts/gulpfile.js`
- `packages/runtime/src/binding/target-accessors.ts`

## 📑 Test Plan

n/a

## ⏭ Next Steps

n/a


packages/runtime/src/binding/target-accessors.ts
```diff
@@ -124,7 +124,7 @@ export class StyleAttributeAccessor implements StyleAttributeAccessor {
           }
         }
       } else if ((<string>newValue).length) {
-        const rx = /\s*([\w\-]+)\s*:\s*((?:(?:[\w\-]+\(\s*(?:"(?:\\"|[^"])*"|'(?:\\'|[^'])*'|[\w\-]+\(\s*(?:^"(?:\\"|[^"])*"|'(?:\\'|[^'])*'|[^\)]*)\),?|[^\)]*)\),?|"(?:\\"|[^"])*"|'(?:\\'|[^'])*'|[^;]*),?\s*)+);?/g;
+        const rx = /\s*([\w\-]+)\s*:\s*((?:(?:[\w\-]+\(\s*(?:"(?:\\"|[^"])*"|'(?:\\'|[^'])*'|[\w\-]+\(\s*(?:[^"](?:\\"|[^"])*"|'(?:\\'|[^'])*'|[^\)]*)\),?|[^\)]*)\),?|"(?:\\"|[^"])*"|'(?:\\'|[^'])*'|[^;]*),?\s*)+);?/g;
```
fkleuver: Nice!!

packages/runtime/src/binding/computed-observer.ts
```diff
@@ -22,7 +22,7 @@ export function computed(config: ComputedOverrides): PropertyDecorator {
   };
 }
 
-const noProxy = !(typeof Proxy !== undefined);
+const noProxy = !(Proxy !== undefined);
```
fkleuver:
```suggestion
const noProxy = !(typeof Proxy !== 'undefined');
```
This will throw a reference error if `Proxy` really isn't defined otherwise

BBosman: Using `typeof` triggers another linting error so thats why I solved it this way, but this is a valid reason for suppressing.

Just pushed the change.

BBosman:
@fkleuver Made the changes.

`lgtm` still reports one (new) issue, but I think it's a false positive?

fkleuver:
> lgtm still reports one (new) issue, but I think it's a false positive?

I'm baffled by it..
image

It looks fine to me. Double checked in the browser console..
Maybe worth reporting to LGTM (though I didn't see a straightforward channel for that)

Let's pretend it's not there for now :P

EisenbergEffect:
Wow. Some really good catches here. Thanks @BBosman !

PR #289, opened 2018-11-15
refactor(runtime): changes to Binding for strictPropertyInitialization

BBosman:
# Pull Request

## 📖 Description

Follow up to #282, with a focus on Binding.

### 🎫 Issues

Extends #282.

## 👩‍💻 Reviewer Notes

Tried to adhere to the principles of #282. Also snuck in a few typing/linting fixes.

## 📑 Test Plan

n/a

## ⏭ Next Steps

See #282.

### DIFF
--- a/packages/runtime/src/binding/binding-context.ts
+++ b/packages/runtime/src/binding/binding-context.ts
@@ -24,11 +24,13 @@ export class InternalObserversLookup {
 export class BindingContext implements IBindingContext {
   [key: string]: ObservedCollection | StrictPrimitive | IIndexable;
 
-  public readonly $synthetic: true = true;
+  public readonly $synthetic: true;
 
   public $observers: ObserversLookup<IOverrideContext>;
 
   private constructor(keyOrObj?: string | IIndexable, value?: ObservedCollection | StrictPrimitive | IIndexable) {
+    this.$synthetic = true;
+
     if (keyOrObj !== undefined) {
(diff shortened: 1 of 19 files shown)


fkleuver:
Overall this is good. Just the usual nitpick..
So I noticed you appear to be aligning the property declaration order with the constructor argument order (at least in the bindings), and those that are not constructor arguments come either below and/or above or something.

This is really a big nitpick, but for consistency (and to make clearer to readers which things are common and which things are variable), I'd prefer a property ordering that doesn't necessarily align with the constructor parameter order:
`$nextBind` and `$prevBind`  always on top. Then `$state` and `$lifecycle` right below. Those 4 properties should exist on all `IBinding` implementations.
The rest of the properties aren't as straight forward so I guess those don't really matter for now. I might have to take a look at that myself at a later point in time, potentially rename some (for example the interpolation property in `Interpolation` could also just be `sourceExpression`).

I also had plans to move the bindings together either in the same file or in the same folder, reorganize the observers, etc, but that's not really important now. Just to give you a bit of context on what I had in mind.

BBosman:
Did some re-ordering as requested. 

Went for those `$vars` first and as the order of the rest was up for grabs I *decided* on alphabetical for now 😊.

fkleuver:
Thanks! I'm down with alphabetical for now. Other than that no remarks, looks good :)

PR #290, opened 2018-11-15
refactor: remove no-reserved-keywords suppressions and fix most of them

BBosman:
# Pull Request

## 📖 Description

Removed and fixed almost all `no-reserved-keywords` suppressions.

There are 8 left, of which 7 are valid suppressions imho. Number 8 is in `ITargetedInstruction` and its kin, but when I started fixing those it was turtles all the way down, so I'd like to leave that one as a separate issue for now.

### 🎫 Issues

One more for #249 😄 

## 👩‍💻 Reviewer Notes

Just the usual checks and maybe validate that the remaining 7 are indeed valid suppressions.

## 📑 Test Plan

n/a

## ⏭ Next Steps

Fix `ITargetedInstruction` and its kin.

### DIFF
--- a/packages/debug/src/reporter.ts
+++ b/packages/debug/src/reporter.ts
@@ -9,8 +9,7 @@ const enum MessageType {
 
 interface IMessageInfo {
   message: string;
-  // tslint:disable-next-line:no-reserved-keywords
-  type: MessageType;
+  messageType: MessageType;
 }
 
 export const Reporter: typeof RuntimeReporter = {...RuntimeReporter,
@@ -18,7 +17,7 @@ export const Reporter: typeof RuntimeReporter = {...RuntimeReporter,
     const info = getMessageInfoForCode(code);
 
     // tslint:disable:no-console
-    switch (info.type) {
+    switch (info.messageType) {
       case MessageType.debug:
         console.debug(info.message, ...params);
         break;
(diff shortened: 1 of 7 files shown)


packages/debug/src/reporter.ts
```diff
@@ -9,16 +9,15 @@ const enum MessageType {
 
 interface IMessageInfo {
   message: string;
-  // tslint:disable-next-line:no-reserved-keywords
-  type: MessageType;
+  Type: MessageType;
```
fkleuver: I think I prefer `messageType` here, as to not confuse it with `Type` properties that are constructor types

fkleuver:
The suppression for the keyword `get` is okay for now. We have been discussion the option of changing this to something like `resolve` (which is what other DI frameworks often do as well), but that's also a backwards compat matter (for which we may introduce a backward compat package actually). It's probably best left out of scope for this PR though

BBosman:
Renamed `Type` to `messageType ` as requested. I agree it removes ambiguity.

BBosman:
On a side-note. Am I the only on getting `Sorry, description could not be provided due to the internal error: Error: rule named no-var-self is not found in the metadata` errors when trying to view the `codeclimate` issues?

EisenbergEffect:
@BBosman There are conflicts. If you can resolve those, we'll get this in. Great work!

BBosman:
@EisenbergEffect Rebased, but it's now dependent on #294 to fix the build.

PR #291, opened 2018-11-15
feat(aurelia): add startup api shorthand

fkleuver:
# Pull Request

## 📖 Description

Adds a simple shorthand for starting an aurelia app without having to manually instantiate the app root. Especially useful when dependencies need to be injected into the app root.

## 👩‍💻 Reviewer Notes

It's just a little branch-off: https://github.com/aurelia/aurelia/compare/startup-api?expand=1#diff-0bcdaa7de1d39a267573c522d86a0850R30

## 📑 Test Plan

Simple integration test that demonstrates the idea: https://github.com/aurelia/aurelia/compare/startup-api?expand=1#diff-dbeceac1edfda3ca8d0d46ee8d0e6be2R14


EisenbergEffect:
What's weird about this case....is that the normal element registration registers things as transient in the container, which doesn't seem quite right for the app root. Thoughts?

fkleuver:
Well that's an interesting one. It's only resolved during `.app` which you generally only call once for a given Aurelia instance. Should you for whatever reason want to instantiate multiple Aurelia's with the same app (which should work since you can pass a different host each time) - and really, I doubt anyone would want this, but it's possible - then you would probably want a different app instance for each Aurelia instance as well.
In any other case it has no effect since it's only ever resolved once anyway. So I don't think the app needs special treatment apart from the special treatment it already gets from the Aurelia instance, honestly.

EisenbergEffect:
Merging from now but I'm thinking we'll still want to revisit the whole "startup" story again at a later point. Not critical at this junction though.

PR #296, opened 2018-11-17
view compiler extension point (part 1/2)

fkleuver:
# Pull Request

## 📖 Description

Introduces a new decorator `@attributePattern` that allows you to register a custom pattern for parsing attribute names.

The only special keyword is `PART` which means "match any number of characters that does not occur in the provided symbol list" ~"match any number of characters out of `-_$+abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789`". Any characters in the pattern that are not included in this list will automatically be interpreted as symbols (or part separators).~

You place this decorator on a class, with one or more patterns per decorator. Then you add methods to the class corresponding to the patterns in the decorator - those methods will be called with the interpretation result if a pattern for that class is matched.

The interpretation result is the raw name+value and an array of strings (parts). You need to return an `AttrSyntax`.

It's quite simple:

### Default aurelia syntax
```ts
@attributePattern({ pattern: 'PART.PART', symbols: '.' })
export class DotSeparatedAttributePattern {
  public ['PART.PART'](rawName: string, rawValue: string, parts: string[]): AttrSyntax {
    return new AttrSyntax(rawName, rawValue, parts[0], parts[1]);
  }
}
```

### 🎫 Issues

Directly related to: https://github.com/aurelia/aurelia/issues/165

## 👩‍💻 Reviewer Notes

See [`@attributePattern({ pattern: 'PART.PART', symbols: '.' }, { pattern: 'PART.PART.PART', symbols: '.' })`](https://github.com/aurelia/aurelia/pull/296/files#diff-bda6980752cccfd5f1db2d37f30249fcR398) for what represents the current templating syntax

See [`@attributePattern({ pattern: ':PART', symbols: ':' })`](https://github.com/aurelia/aurelia/pull/296/files#diff-bda6980752cccfd5f1db2d37f30249fcR413) and [`@attributePattern({ pattern: '@PART', symbols: '@' })`](https://github.com/aurelia/aurelia/pull/296/files#diff-bda6980752cccfd5f1db2d37f30249fcR424) for the newly introduced shorthands

## 📑 Test Plan

Testing for the alternative syntax is not very extensive. A few basic unit tests exist to verify that the discussed use cases are supported.

Since this also replaces the previous attribute parsing logic, the test coverage for the normal dot syntax is in fact quite extensive since all the template compiler integration tests hit it.

## ⏭ Next Steps

- Performance improvements might be needed (marked by TODO's in comments)
- Some cleanup and/or rework of the template compiler and/or binding commands might be appropriate to better utilize this new feature
- Perhaps `@bindingCommand` may be deprecated and the attribute patterns should implement a compile method (one per command, etc)?

A follow-up PR should include some finalizing touches directly pertaining to the template compiler.

### DIFF
--- a/packages/jit/src/attribute-parser.ts
+++ b/packages/jit/src/attribute-parser.ts
@@ -1,7 +1,6 @@
 import { all, DI, inject } from '@aurelia/kernel';
 import { AttrSyntax } from './ast';
 import { IAttributePattern, IAttributePatternHandler, Interpretation, ISyntaxInterpreter } from './attribute-pattern';
-import { Char } from './common';
 
 export interface IAttributeParser {
   parse(name: string, value: string): AttrSyntax;
@@ -22,15 +21,11 @@ export class AttributeParser implements IAttributeParser {
     this.cache = {};
     const patterns: AttributeParser['patterns'] = this.patterns = {};
     attrPatterns.forEach(attrPattern => {
-      const patternOrPatterns = attrPattern.$patterns;
-      interpreter.add(patternOrPatterns);
-      if (Array.isArray(patternOrPatterns)) {
-        patternOrPatterns.forEach(pattern => {
-          patterns[pattern] = attrPattern as unknown as IAttributePatternHandler;
-        });
-      } else {
-        patterns[patternOrPatterns] = attrPattern as unknown as IAttributePatternHandler;
-      }
+      const defs = attrPattern.$patternDefs;
+      interpreter.add(defs);
+      defs.forEach(def => {
+        patterns[def.pattern] = attrPattern as unknown as IAttributePatternHandler;
+      });
     });
   }
 
--- a/packages/jit/src/attribute-parser.ts
+++ b/packages/jit/src/attribute-parser.ts
@@ -1,19 +1,7 @@
-import { DI } from '@aurelia/kernel';
+import { all, DI, inject } from '@aurelia/kernel';
+import { AttrSyntax } from './ast';
+import { IAttributePattern, ISyntaxInterpreter } from './attribute-pattern';
 import { Char } from './common';
-
-export class AttrSyntax {
-  public readonly rawName: string;
-  public readonly rawValue: string;
-  public readonly target: string;
-  public readonly command: string | null;
-
-  constructor(rawName: string, rawValue: string, target: string, command: string | null) {
-    this.rawName = rawName;
-    this.rawValue = rawValue;
-    this.target = target;
-    this.command = command;
-  }
-}
 
 export interface IAttributeParser {
   parse(name: string, value: string): AttrSyntax;
@@ -23,11 +11,26 @@ export const IAttributeParser = DI.createInterface<IAttributeParser>()
   .withDefault(x => x.singleton(AttributeParser));
 
 /*@internal*/
+@inject(ISyntaxInterpreter, all(IAttributePattern))
 export class AttributeParser implements IAttributeParser {
+  private interpreter: ISyntaxInterpreter;
   private cache: Record<string, [string, string]>;
+  private patterns: Record<string, IAttributePattern>;
 
-  constructor() {
+  constructor(interpreter: ISyntaxInterpreter, attrPatterns: IAttributePattern[]) {
+    this.interpreter = interpreter;
     this.cache = {};
+    const patterns = this.patterns = {};
+    attrPatterns.forEach(attrPattern => {
+      const patternOrPatterns = attrPattern.$patternOrPatterns;
+      if (Array.isArray(patternOrPatterns)) {
+        patternOrPatterns.forEach(pattern => {
+          patterns[pattern] = attrPattern;
+        });
+      } else {
+        patterns[patternOrPatterns] = attrPattern;
+      }
+    });
   }
 
   public parse(name: string, value: string): AttrSyntax {
@@ -35,6 +38,7 @@ export class AttributeParser implements IAttributeParser {
     let command: string;
     const existing = this.cache[name];
     if (existing === undefined) {
+      const interpretation = this.interpreter.interpret(name);
       let lastIndex = 0;
       target = name;
       for (let i = 0, ii = name.length; i < ii; ++i) {
--- a/packages/jit/src/attribute-parser.ts
+++ b/packages/jit/src/attribute-parser.ts
@@ -1,6 +1,6 @@
 import { all, DI, inject } from '@aurelia/kernel';
 import { AttrSyntax } from './ast';
-import { IAttributePattern, ISyntaxInterpreter } from './attribute-pattern';
+import { IAttributePattern, IAttributePatternHandler, Interpretation, ISyntaxInterpreter } from './attribute-pattern';
 import { Char } from './common';
 
 export interface IAttributeParser {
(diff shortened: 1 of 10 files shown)


fkleuver:
@EisenbergEffect @bigopon It's working now. Please have a look and let me know what you think

packages/jit/src/attribute-pattern.ts
```diff
@@ -0,0 +1,431 @@
...
+
+@attributePattern('PART.PART', 'PART.PART.PART')
+export class DotSeparatedAttributePattern implements DotSeparatedAttributePattern {
+  public static register: IRegistry['register'];
+
+  public ['PART.PART'](rawName: string, rawValue: string, parts: string[]): AttrSyntax {
+    return new AttrSyntax(rawName, rawValue, parts[0], parts[1]);
+  }
+
+  public ['PART.PART.PART'](rawName: string, rawValue: string, parts: string[]): AttrSyntax {
```
bigopon: Why is this prebuilt?

packages/jit/src/attribute-pattern.ts
```diff
@@ -0,0 +1,431 @@
...
+      }
+      return;
+    }
+    let currentState = this.rootState;
+    const types = new SegmentTypes();
+    const segments = this.parse(patternOrPatterns, types);
+    const len = segments.length;
+    i = 0;
+    while (i < len) {
+      segments[i].eachChar(ch => {
```
bigopon: Maybe move this arrow fn outside of this loop.

packages/jit/src/attribute-pattern.ts
```diff
@@ -0,0 +1,431 @@
...
+    }
+
+    return result;
+  }
+}
+
+function validatePrototype(handler: IAttributePatternHandler, patternOrPatterns: string | string[]): void {
+  const patterns = Array.isArray(patternOrPatterns) ? patternOrPatterns : [patternOrPatterns];
+  for (const pattern of patterns) {
+    // note: we're intentionally not throwing here
```
bigopon: maybe 1 more line explaining what would happen when nothing matched? later it can be turned into doc easier than this state

packages/jit/src/attribute-parser.ts
```diff
@@ -23,36 +11,39 @@ export const IAttributeParser = DI.createInterface<IAttributeParser>()
...
+  constructor(interpreter: ISyntaxInterpreter, attrPatterns: IAttributePattern[]) {
+    this.interpreter = interpreter;
     this.cache = {};
+    const patterns: AttributeParser['patterns'] = this.patterns = {};
+    attrPatterns.forEach(attrPattern => {
+      const patternOrPatterns = attrPattern.$patterns;
+      interpreter.add(patternOrPatterns);
+      if (Array.isArray(patternOrPatterns)) {
+        patternOrPatterns.forEach(pattern => {
+          patterns[pattern] = attrPattern as unknown as IAttributePatternHandler;
```
bigopon: I'm not sure how to feel about this cast. It seems random and it could obscure the code analysis.

packages/kernel/src/di.ts
```diff
@@ -254,7 +260,8 @@ Foo.register(container);
   // tslint:enable:jsdoc-format
   singleton<T extends Constructable>(target: T & Partial<RegisterSelf<T>>): T & RegisterSelf<T> {
     target.register = function register(container: IContainer): IResolver<InstanceType<T>> {
-      return Registration.singleton(target, target).register(container, target);
+      const registration = Registration.singleton(target, target);
+      return registration.register(container, target);
```
bigopon:
```suggestion
      return container.register(registration);
```

bigopon: I remember @EisenbergEffect suggested this once. When I was doing the `root` feature for Aurelia class

packages/jit/src/attribute-pattern.ts
```diff
@@ -0,0 +1,431 @@
...
+  public rootState: State;
+  private initialStates: State[];
+
+  constructor() {
+    this.rootState = new State(null);
+    this.initialStates = [this.rootState];
+  }
+
+  public add(pattern: string): void;
+  public add(patterns: string[]): void;
```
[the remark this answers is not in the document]
fkleuver: Nope but it's nicer api imo

packages/jit/src/attribute-pattern.ts
fkleuver: To make some unit tests pass that depended on this behavior from the previous attribute parser =)

packages/jit/src/attribute-parser.ts
fkleuver: I know how I feel about it. It sucks. This is a TypeScript problem. You can't combine an interface with an index signature with an interface with any other kind of signature. I'm not sure how else to solve it.. :(

packages/jit/src/attribute-pattern.ts
fkleuver: why?

packages/jit/src/attribute-pattern.ts
fkleuver: Yeah but why here? The idea is to properly document the API endpoints, so in this case the decorators once it's done and stuff.
As for what happens when nothing is matched, that's simple (and same as it's always been): the attribute will be treated as a normal attribute that has no command.

packages/jit/src/attribute-pattern.ts
bigopon: fn creation cost evasion. I can see it's just few states, maybe ignore.

fkleuver: oh you're right, i didn't realize this was a possibility given the scoping..

packages/jit/src/attribute-parser.ts
bigopon: If the type is `A | B` and it's casted to `B`, then it can be understood that from that point on, `B` is the type. Casting to unknown is really a way to tell `well, yeah it's B`

BBosman: Using `attrPattern as IAttributePattern|IAttributePatternHandler as IAttributePatternHandler` works and is a bit more type-safe.

packages/kernel/src/di.ts
```diff
@@ -621,11 +634,12 @@ export class Container implements IContainer {
     let current: Container = this;
 
     while (current !== null) {
-      const resolver = current.resolvers.get(key);
+      let resolver = current.resolvers.get(key);
 
       if (resolver === undefined) {
         if (current.parent === null) {
-          return this.jitRegister(key, current).resolve(current, this);
+          resolver = this.jitRegister(key, current)
```
BBosman: Linting police :p
Missing `;`.

packages/jit/src/index.ts
```diff
@@ -1,4 +1,5 @@
 export * from './attribute-parser';
```
BBosman: Shouldn't `ast` be in here as well, as the types it contains aren't marked as `internal` and were exported before.

fkleuver: Yep, good catch :)

packages/kernel/src/di.ts
fkleuver: Boils down to the same thing. `container.register` will end up calling `register` on the registration, which then calls `container.registerResolver`.
`registration.register` directly calls `container.registerResolver`.
However calling `container.registerResolver` directly won't work because it's of the wrong type. So this is the shortest route supported by the API afaik.

fkleuver:
Let's get it in then @EisenbergEffect

PR #297, opened 2018-11-18
refactor: more any to strict typing conversions

BBosman:
# Pull Request

## 📖 Description

More linting fixes. This time its mostly `any` to more strict type conversions.

### 🎫 Issues

Related to #249.

## 👩‍💻 Reviewer Notes

Notable changes:
- Mostly `any` conversions where I could upstream already existing types.
- ~~In `ITargetedInstruction` the `type` property will be renamed to `Type` in a future PR, but as far as I know there are no such plans for `from`, so I suppressed those warnings.~~
- Disabled the `no-reserved-keywords` linting rule, see discussion below.

## 📑 Test Plan

Depend on CircleCI.

## ⏭ Next Steps

- Check types used by Scope (https://github.com/aurelia/aurelia/pull/297#discussion_r234462754)
- Maybe undo some renames that were done just to satisfy `no-reserved-keywords`.
- And more, see #249.

### DIFF
--- a/packages/runtime/src/templating/resources/compose.ts
+++ b/packages/runtime/src/templating/resources/compose.ts
@@ -45,9 +45,9 @@ export class Compose {
     };
 
     this.properties = instruction.instructions
-      .filter((x: any) => !composeProps.includes(x.to))
+      .filter((x: ITargetedInstruction & {to?: string}) => !composeProps.includes(x.to))
       .reduce(
-        (acc, item: any) => {
+        (acc, item: ITargetedInstruction & {to?: string}) => {
           if (item.to) {
             acc[item.to] = item;
           }
--- a/packages/runtime/src/templating/resources/with.ts
+++ b/packages/runtime/src/templating/resources/with.ts
@@ -1,8 +1,8 @@
 import { inject, IRegistry } from '@aurelia/kernel';
 import { Scope } from '../../binding/binding-context';
 import { IRenderLocation } from '../../dom';
-import { IView, IViewFactory, State } from '../../lifecycle';
-import { LifecycleFlags } from '../../observation';
+import { IBindScope, IView, IViewFactory, State } from '../../lifecycle';
+import { IBindingContext, LifecycleFlags } from '../../observation';
 import { bindable } from '../bindable';
 import { ICustomAttribute, templateController } from '../custom-attribute';
 
@@ -12,7 +12,8 @@ export interface With extends ICustomAttribute {}
 export class With {
   public static register: IRegistry['register'];
 
-  @bindable public value: any = null;
+  // TODO: this type is incorrect (it can be any user-provided object), need to fix and double check Scope.
+  @bindable public value: IBindScope | IBindingContext = null;
 
   private currentView: IView = null;
 
--- a/packages/runtime/src/templating/view.ts
+++ b/packages/runtime/src/templating/view.ts
@@ -59,13 +59,13 @@ export class View implements IView {
     this.$bind = lockedBind;
   }
 
-  public release(flags: LifecycleFlags): any {
+  public release(flags: LifecycleFlags): boolean {
     this.isFree = true;
     if (this.$state & State.isAttached) {
       return this.cache.canReturnToCache(this);
     }
 
-    return this.$unmount(flags);
+    return !!this.$unmount(flags);
   }
 }
 
--- a/packages/jit/src/binding-command.ts
+++ b/packages/jit/src/binding-command.ts
@@ -28,7 +28,7 @@ export interface IBindingCommand {
 
 export interface IBindingCommandDefinition extends IResourceDefinition { }
 
-export interface IBindingCommandType extends IResourceType<IBindingCommandDefinition, IBindingCommand, Class<IBindingCommand, IIndexable>> { }
+export interface IBindingCommandType extends IResourceType<IBindingCommandDefinition, IBindingCommand, Class<IBindingCommand>> { }
 
 export interface IBindingCommandResource extends
   IResourceKind<IBindingCommandDefinition, IBindingCommand, Class<IBindingCommand>> { }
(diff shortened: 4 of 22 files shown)


packages/runtime/src/templating/resources/with.ts
```diff
@@ -12,7 +12,7 @@ export interface With extends ICustomAttribute {}
 export class With {
   public static register: IRegistry['register'];
 
-  @bindable public value: any = null;
+  @bindable public value: IBindScope | IBindingContext = null;
```
fkleuver: The only assumption we can make about this type is that it will be an object. For example:
```ts
export class MyViewModel {
  public person = { name: 'Bob' }
}
```
```html
<div with.bind="person">
  ${name}
</div>
```

So this might be one of the very rare situations where `IIndexable` is actually the most appropriate type.
That said, `IIndexable` introduces `any`. And changing that to `unknown` is something which I believe we can't do quite yet. Perhaps introduce an `IUnknownIndexable` as a temporary intermediate? Thoughts?

packages/runtime/src/templating/resources/compose.ts
```diff
@@ -45,9 +45,9 @@ export class Compose {
     };
 
     this.properties = instruction.instructions
-      .filter((x: any) => !composeProps.includes(x.to))
+      .filter((x: TargetedInstruction & {to?: string}) => !composeProps.includes(x.to))
```
fkleuver: If we're duck-typing this I'd rather use `ITargetedInstruction` because `TargetedInstruction` is an intersection type of all `ITargetedInstruction` derivations (making it a "bigger" dependency).

packages/runtime/src/templating/view.ts
```diff
@@ -59,13 +59,13 @@ export class View implements IView {
...
 
-  public release(flags: LifecycleFlags): any {
+  public release(flags: LifecycleFlags): boolean {
     this.isFree = true;
     if (this.$state & State.isAttached) {
       return this.cache.canReturnToCache(this);
     }
 
-    return this.$unmount(flags);
+    return this.$unmount(flags) as boolean;
```
fkleuver:
```suggestion
    return !!this.$unmount(flags);
```
Would that work? Then we're not lying about the returned type as much :)

packages/runtime/src/templating/resources/with.ts
BBosman: I chose those types, because that's what `Scope.fromParent` on line 47 expects it to be. And as that's the only place the class that `value` is used... Does that mean we need to change `fromParent` as well?

BBosman:
@fkleuver Made a couple of the requested changes, added an inline reply to one of them and reverted the `from` suppressions. It felt a bit *dirty* suppressing them, but as the talk so far had only been around renaming `type`, I'd made the assumption that `from` was final. Good to know it's up for discussion.

For background, [this is the list](https://github.com/Microsoft/TypeScript/issues/2536) that the tslint rule was based on, which does mark `from` as a reserved word in strict mode.

packages/runtime/src/templating/resources/with.ts
fkleuver: Oh. You're actually raising a really important point now. I never realized this particular scenario when I rewrote `Scope`, `OverrideContext` and `BindingContext`. I need to look into that.
Goes to show how important strict typing is as this might expose potential issues that otherwise might not have surfaced until it bit an end user. Good stuff :)

fkleuver:
> For background, this is the list that the tslint rule was based on, which does mark from as a reserved word in strict mode.

No kidding.. Oo thanks for that link. Reminder to self: check the ES spec instead of MDN next time.

Then we definitely should change it. We want every class/object to be safe to destructure without things breaking in strict mode.

packages/runtime/src/templating/resources/with.ts
BBosman: So leave it as `IBindScope | IBindingContext` for now? Or go the `IUnknownIndexable` way? My preference would be option 1, because that is a valid representation of how it's currently implemented and then "fix" it when you look into it a bit more.

BBosman:
You're looking at ES. Apparently TS has some additional things marked as reserved. See the Typescript sourcecode: https://github.com/Microsoft/TypeScript/blob/master/src/compiler/types.ts#L274

But looking at https://github.com/Microsoft/tslint-microsoft-contrib/issues/652 they're thinking about deprecating that entire rule in favour of TSLint core's [variable-name](https://palantir.github.io/tslint/rules/variable-name/) with `ban-keywords` anyway, so that might not be a bad way to go here as well.

packages/runtime/src/templating/resources/with.ts
fkleuver:
```suggestion
  // TODO: this type is incorrect (it can be any user-provided object), need to fix and double check Scope
  @bindable public value: IBindScope | IBindingContext = null;
```
Yeah that's okay. I suggested a TODO comment so i don't forget

fkleuver:
> so it would mean only switching off the no-reserved-keywords rule in our case.

Would that still ban strict mode reserved keywords? In that case I'm for it.

BBosman:
> Would that still ban strict mode reserved keywords? In that case I'm for it.

Yes, imho it would. But not by the linter, but by the Typescript compiler.

For example:
```ts
function test(interface: string) {
    
}

export class TestClass {
  public interface: string = '';

  public test(interface: string) {

  }
}
```
This would not compile, as it errors out with
```
Line 1: [ts] Identifier expected. 'interface' is a reserved word in strict mode. Modules are automatically in strict mode. [1214]
Line 8: [ts] Identifier expected. 'interface' is a reserved word in strict mode. Class definitions are automatically in strict mode. [1213]
```

BBosman:
@fkleuver I added the TODO comment and disabled the `no-reserved-keywords` linting rule (and removed its suppressions), so I think this is ready for a final review.

Also, with the above in mind, do we want to *undo* some of the recent renames (i.e. `type` to `Type`, `type` to `messageType`, `from` -> `fromIndex`, and others) we've done as a result of that rule in a future PR?

fkleuver:
So you're saying `from` and `type` are no longer forbidden with the new setup?

BBosman:
> So you're saying `from` and `type` are no longer forbidden with the new setup?

Yes. They're valid ES/TS, but the `no-reserved-keywords` rule warns, because it *could* indicate an error and therefore recommends not using them to avoid the possibility of it being an error.

PR #299, opened 2018-11-19
fix(lifecycle): temporary solution for infinite flush loop on $detach

fkleuver:
# Pull Request

## 📖 Description

This is a fix for an infinite loop reported by @jwx while working on the router. This is not the "final" fix (see comments), but it is sufficient for the time being. Improvement in this area will come anyway when the necessary more complex integration tests are in place.

## 👩‍💻 Reviewer Notes

N/A

## 📑 Test Plan

Break no existing tests. Follow up asap with proper integration tests (it would be somewhat duplicate effort to add a quick-n-dirty test right now)

## ⏭ Next Steps

Beef up the integration tests as was already in the planning

### DIFF
--- a/packages/runtime/src/lifecycle.ts
+++ b/packages/runtime/src/lifecycle.ts
@@ -754,6 +754,14 @@ export class Lifecycle implements ILifecycle {
         current.flush(flags);
         current = next;
       } while (current !== marker);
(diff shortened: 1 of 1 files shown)


fkleuver:
Closing this and commiting to `jwx-router` instead

PR #301, opened 2018-11-19
refactor: partially revert renames due to no-reserved-keywords

BBosman:
# Pull Request

## 📖 Description

Now that the decision is made to disable the `no-reserved-keywords` linting rule (See #297) it's better to revert some of the changes made just to satisfy that linting rule, when their original names better suited their purpose.

### 🎫 Issues

Part of #249, follow up to #297 and partially reverts #262 and #290.

## 👩‍💻 Reviewer Notes

Straight up renames.

Will increase the number of linting errors until #297 is merged and `no-reserved-keywords` is actually disabled.

## 📑 Test Plan

Trust in CircleCI

## ⏭ Next Steps

See #249.

### DIFF
--- a/packages/debug/src/reporter.ts
+++ b/packages/debug/src/reporter.ts
@@ -9,7 +9,7 @@ const enum MessageType {
 
 interface IMessageInfo {
   message: string;
-  messageType: MessageType;
+  type: MessageType;
 }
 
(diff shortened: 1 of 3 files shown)


fkleuver:
The order doesn't matter from what I understand, just that the other one disables a rule that would trigger extra linting errors to show up here.

PR #305, opened 2018-11-19
perf(all): shorten au-marker to au-m

fkleuver:
# Pull Request

## 📖 Description

Shortens the `au-marker` element name to `au-`.

This makes templates quite a bit shorter, thereby easier to read and smaller in size (for pre-compiled templates).
Originally the idea was to rename it to `au`, but that wouldn't make it compliant with WebComponents, hence the dash. There's no rule that a custom element name cannot end with a dash, and something like `a-u` just seems very awkward.
It's a bit weird ending with a dash, but I actually quite like it. It stands out.

Wonder what others think @EisenbergEffect @bigopon ?

### 🎫 Issues

N/A

## 👩‍💻 Reviewer Notes

N/A

## 📑 Test Plan

Verified that a custom element with this name can be created via `customElements.define`

## ⏭ Next Steps

N/A

### DIFF
--- a/packages/examples/dbmonster-a/app-config.ts
+++ b/packages/examples/dbmonster-a/app-config.ts
@@ -22,8 +22,8 @@ export const appConfig: ITemplateDefinition = {
           cache: "*",
           template: `
             <tr>
-              <td class="dbname"><au-marker class="au"></au-marker> </td>
-              <td class="query-count"><au-marker class="au"></au-marker> </td>
+              <td class="dbname"><au- class="au"></au-> </td>
+              <td class="query-count"><au- class="au"></au-> </td>
               <td class="au"></td>
             </tr>
           `,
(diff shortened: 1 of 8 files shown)


EisenbergEffect:
Hmm. While it may be technically ok, I feel funny about it. What if we just add one more letter: `au-m`?

fkleuver:
Thought about that too, was curious what you'd think about this first :)
Done. It doesn't really stop being a little odd tbh. I wish it could just be `au`. But also don't want to take even the miniscule risk that that might bite us in some distant future either.
You think `au-m` is ok?

bigopon:
Only Fred could suggest it 😩

EisenbergEffect:
I think au-m is ok. Shall I merge?

fkleuver:
> Only Fred could suggest it

Someone has to be the weirdo right ;)

fkleuver:
> I think au-m is ok. Shall I merge?

I'm cool with it. Still prefer it because it's half the amount of text.

Issue #197, opened 2018-09-27
Improve templating test coverage


fkleuver:

fkleuver:
Closing this because there is no clear scope. Improving test coverage is an overall ongoing thing anyway.

PR #306, opened 2018-11-20
replaceable template controller + jit refactor + remove lib.dom.d.ts dep

fkleuver:
# Pull Request

## 📖 Description

- Includes (preliminary) replacement typings for `lib.dom.d.ts` (that dependency is removed) to fully rid the jit and runtime of DOM specific references. This was more of a discovery / exposure exercise to get a better picture of HTML specific code. Once the html specific logic sits in its own package, `lib.dom.d.ts` can come back.
- Implements `replaceable` template controller. This showed a few limitations in the existing template compiler design which motivated the full rewrite. It's not in the final state that I have in mind yet, alternative runtimes are next - they will drive the rest of the work on this area.
- For `replaceable`, I added a "backdoor" to scope traversal, allowing you to reach outside of an element's scope via `parentScope`. This is internal API however and even for `replaceable` it's guarded from going further than 1 level up, with help of `LifecycleFlags`
- The rewrite of the template compiler boils down to cleaning up the symbol classes and moving most of their behavior to `template-binder` (still due for a better name..). This makes the `template-compiler` fully runtime agnostic as it only deals with symbols and instructions.

### 🎫 Issues

Directly related to https://github.com/aurelia/aurelia/issues/197
Loosely related to https://github.com/aurelia/aurelia/issues/89

### Misc
- Enables the `LengthObserver`: https://github.com/aurelia/aurelia/pull/306/files#diff-442b9e1f4993be33ec2440e362e7e364R10
- Adds the attach/detach flags to start/stop task

### Test related
- Test scenarios are accumulated in [packages/runtime/test/integration/runtime.spec.ts](https://github.com/aurelia/aurelia/compare/test-coverage?expand=1#diff-c3914f9c46f091970560b58bd3bd3107)
- Test scenarios are accumulated in [packages/runtime/test/integration/runtime.spec.ts](https://github.com/aurelia/aurelia/compare/test-coverage?expand=1#diff-c3914f9c46f091970560b58bd3bd3107)
- The test builder is in [packages/runtime/test/integration/test-builder.ts](https://github.com/aurelia/aurelia/compare/test-coverage?expand=1#diff-b2e32df525baf76eefc9cfc57f6ce308)

### Replaceable
- Compilation: https://github.com/aurelia/aurelia/pull/306/files#diff-6311bbb5fb758683f5e5c11c3a0f20a5R499
- Scope traversal: https://github.com/aurelia/aurelia/pull/306/files#diff-e97e1a37b14825b9c9e2b4a186809c2aR87 (parentScope is passed in via parent bindables, `Aurelia` passes in null)
- 3 new flags: https://github.com/aurelia/aurelia/pull/306/files#diff-f4f7a420bc2504ec2ddb90e3e28afd70R34
- flag is stored on binding so it can pass it through to AST: https://github.com/aurelia/aurelia/pull/306/files#diff-462d7124d7584037488f5af73fa5ab7aR44

## 📑 Test Plan

- For replaceable about 1k additional generated tests are added. Apart from replaceable this is largely a refactor, so I'm relying on existing tests.
- Some tests utilities (and utilization thereof) have been added but this needs to be built upon further.

## ⏭ Next Steps

- Separate html-specific functionality into separate packages and implement alternative runtimes


fkleuver:
Yeah I know, that applies to a much larger scope though and that's what #89 is for. It's easier to build up a proper suite starting with a single file, I'll split it up eventually.

Addressing your suggestion specifically, however, let's thing about this a bit more deeply because I don't think it's that simple. There are multiple ways in which tests can be grouped and surely we don't want to end up with thousands of files.

packages/jit/src/semantic-model.ts
```diff
@@ -473,17 +483,42 @@ export class ElementSymbol {
...
     this._isSlot = false;
     this._isLet = false;
-    this._node = syntax.node;
+    const node = this._node = syntax.node;
     this._syntax = syntax;
     this._name = this.node.nodeName;
     this._isCustomElement = false;
     this._isLifted = false;
+    // TODO: improve DOM typings and clean up this mess etc
+    if (node.nodeType === NodeType.Element && (<NodeWithAttributes>node).hasAttribute('replace-part')) {
```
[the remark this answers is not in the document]
fkleuver: The code itself is a bit ugly/hacky, but I'm planning on doing another big refactor pass when truly all features are implemented and AOT efforts have started. Then the semantic model will also need more work.
But yeah I generally don't see any reason to not support this (or anything else) on non-template elements since it's very easy to just wrap them.

packages/runtime/src/templating/lifecycle-bind.ts
```diff
@@ -40,10 +40,13 @@ export function $bindAttribute(this: Writable<ICustomAttribute>, flags: Lifecycl
 }
 
 /*@internal*/
-export function $bindElement(this: Writable<ICustomElement>, flags: LifecycleFlags): void {
+export function $bindElement(this: Writable<ICustomElement>, flags: LifecycleFlags, parentScope: IScope | null): void {
```
bigopon: where will this `parentScope` come from? I'm unable to follow this

fkleuver: From any parent bindable that has this element as its child. Everything already passes the `scope` in, but this `$bind` was previously just ignoring that scope (not even declaring the argument). So it was already available.

packages/jit/test/integration/template-compiler.kitchen-sink.spec.ts
```diff
@@ -999,4 +999,173 @@ describe(spec, () => {
...
+
+    au.app({ host, component });
+
+    au.start();
+
+    expect(host.textContent).to.equal('def');
+
+  });
+
+  it(`replaceable/template - same part multiple times`, () => {
```
bigopon: I think this, albeit unrealistic, is a nice capability.

packages/jit/test/integration/template-compiler.kitchen-sink.spec.ts
```diff
@@ -999,4 +999,173 @@ describe(spec, () => {
...
+    au.app({ host, component });
+
+    au.start();
+
+    expect(host.textContent).to.equal('abcabc');
+
+  });
+
+  // TODO: fix this scenario
+  xit(`replaceable/template - parent template controller`, () => {
```
bigopon: This seems like another case of unavailable knowledge of template controller template. Maybe we can have a realization phase to help with this? Also that will help bring vnext compliation on par with vCurrent, where we have `beforeCompile` and `afterCompile` hook 
@fkleuver  @EisenbergEffect

fkleuver: Actually working on this feature exposed a few flaws in the way I structured semantic model / template compiler. I already knew it was rough around the edges, but it's getting really messy now so I need to rework a couple things potentially from scratch.
However, I would really like to get surrogate compilation working first because I suspect that will become yet another hackfest to integrate properly with everything else. I prefer cleaning up a mess once over doing it twice.. :-)

packages/runtime/src/binding/target-accessors.ts
```diff
@@ -1,5 +1,6 @@
 import { IIndexable } from '@aurelia/kernel';
-import { DOM, IHTMLElement, INode } from '../dom';
+import { DOM } from '../dom';
+import { IHTMLElement, INode, IElement } from '../dom.interfaces';
```
BBosman: Named imports must be alphabetized.

packages/runtime/src/binding/resources/self-binding-behavior.ts
```diff
@@ -1,12 +1,12 @@
 import { IRegistry, Reporter } from '@aurelia/kernel';
-import { INode } from '../../dom';
+import { INode, IEvent } from '../../dom.interfaces';
```
BBosman: Named imports must be alphabetized.

packages/runtime/src/binding/observer-locator.ts
```diff
@@ -1,5 +1,6 @@
 import { DI, inject, Reporter } from '@aurelia/kernel';
-import { DOM, IHTMLElement, IInputElement } from '../dom';
+import { DOM } from '../dom';
+import { IHTMLElement, IElement } from '../dom.interfaces';
```
BBosman: Named imports must be alphabetized.

packages/runtime/src/binding/observer-locator.ts
```diff
@@ -9,7 +10,7 @@ import {
 import { getArrayObserver } from './array-observer';
 import { createComputedObserver } from './computed-observer';
 import { IDirtyChecker } from './dirty-checker';
-import { CheckedObserver, ISelectElement, SelectValueObserver, ValueAttributeObserver } from './element-observation';
+import { CheckedObserver, ISelectElement, IInputElement, SelectValueObserver, ValueAttributeObserver } from './element-observation';
```
BBosman: Named imports must be alphabetized.

packages/runtime/src/binding/event-manager.ts
```diff
@@ -1,39 +1,21 @@
...
-  path?: EventTarget[];
-
-  // old composedPath
-  deepPath?(): EventTarget[];
-
-  // current spec
-  composedPath?(): EventTarget[];
-};
+import { DOM } from '../dom';
+import { INode, IManagedEvent, IEventTarget, IEventListenerOrEventListenerObject } from '../dom.interfaces';
```
BBosman: Named imports must be alphabetized.

packages/runtime/src/binding/element-observation.ts
```diff
@@ -8,6 +8,7 @@ import { IEventSubscriber } from './event-manager';
 import { IObserverLocator } from './observer-locator';
 import { SetterObserver } from './property-observation';
 import { targetObserver } from './target-observer';
+import { INode, IHTMLInputElement, IMutationObserver, IHTMLSelectElement, IHTMLOptionElement } from '../dom.interfaces';
```
BBosman: Named imports must be alphabetized.

packages/kernel/src/platform.ts
```diff
@@ -1,6 +1,12 @@
+import { IWindowOrWorkerGlobalScope, IWindow } from './interfaces';
```
BBosman: Named imports must be alphabetized.

packages/jit/src/template-compiler-2.ts
```diff
@@ -0,0 +1,59 @@
...
+      new SymbolPreprocessor(model),
+      new NodePreprocessor(model),
+      new InstructionBuilder(model)
+    ];
+
+    visitors.forEach(visitor => {
+      target.accept(visitor);
+    });
+
+    return <TemplateDefinition>target.definition;;
```
BBosman: A `;` too many.

packages/jit/src/template-compiler.ts
```diff
@@ -24,10 +24,13 @@ import {
   TextBindingInstruction,
   ToViewBindingInstruction,
   TwoWayBindingInstruction,
-  ViewCompileFlags
+  ViewCompileFlags,
+  IText,
```
BBosman: Named imports must be alphabetized.

packages/jit/src/semantic-model.ts
```diff
@@ -1,10 +1,10 @@
 import { Immutable, IServiceLocator, PLATFORM } from '@aurelia/kernel';
-import { BindingMode, CustomAttributeResource, CustomElementResource, DOM, HydrateTemplateController, IAttributeDefinition, IBindableDescription, IExpressionParser, IResourceDescriptions, ITemplateDefinition, TargetedInstruction } from '@aurelia/runtime';
+import { BindingMode,  buildTemplateDefinition, CustomAttributeResource, CustomElementResource, DOM, HydrateTemplateController, IAttributeDefinition, IBindableDescription, IElement, IExpressionParser, IHydrateElementInstruction, IResourceDescriptions, ITemplateDefinition, TargetedInstruction, NodeType, INode, IHTMLTemplateElement } from '@aurelia/runtime';
```
BBosman: Named imports must be alphabetized.

BBosman:
It would be nice if the *auto-import* feature of Visual Studio Code would automagically alphabetise them. 😉

packages/jit/src/semantic-model-2.ts
```diff
@@ -0,0 +1,1435 @@
+import { all, DI, inject, PLATFORM, Reporter, Immutable } from '@aurelia/kernel';
+import { AttributeDefinition, BindingType, CustomAttributeResource, CustomElementResource, DOM, IAttr, IChildNode, IExpressionParser, IHTMLElement, IHTMLSlotElement, IHTMLTemplateElement, Interpolation, IResourceDescriptions, IsExpressionOrStatement, ITemplateDefinition, IText, NodeType, TemplateDefinition, IDocumentFragment, IElement, BindingMode, IBindableDescription, IBind, IHydrateElementInstruction, IHydrateTemplateController, IHydrateAttributeInstruction, TextBindingInstruction, HydrateTemplateController, IPropertyBindingInstruction, OneTimeBindingInstruction, FromViewBindingInstruction, TwoWayBindingInstruction, ToViewBindingInstruction, IsBindingBehavior, HydrateAttributeInstruction, TargetedInstruction, HydrateElementInstruction } from '@aurelia/runtime';
+import { AttrSyntax } from './ast';
+import { IAttributePattern, IAttributePatternHandler, Interpretation, ISyntaxInterpreter } from './attribute-pattern';
+import { BindingCommandResource, IBindingCommand } from './binding-command';
+
+// tslint:disable-next-line:no-any
+const emptyArray = <any[]>PLATFORM.emptyArray;
```
BBosman: Can't this be `<unknown[]>PLATFORM.emptyArray;` and then do some explicit casting where it's used?

bigopon:
For alphabetizing imports, guess we can use vscode auto format for that. The conbi Alt Shift O does it

packages/jit/test/integration/template-compiler.kitchen-sink.spec.ts
fkleuver: @bigopon What do you think about the new visitors approach as a means to achieve this sort of phased separation?

packages/jit/test/integration/template-compiler.kitchen-sink.spec.ts
fkleuver: Well I don't know about unrealistic, but perhaps nonsensical is a better word. In the end though I'm just shooting at the code with as weird tests as possible to make it as robust as I can, so you get these sorts of tests inevitably

packages/jit/test/integration/template-compiler.kitchen-sink.spec.ts
bigopon: My main concern with this is perf of compilation, beside that, this is a very nice approach

fkleuver: You might be surprised to hear that this approach actually creates significantly fewer objects and performs fewer operations on the DOM than the previous approach:
- The symbols are held together via linked lists (previous I used arrays)
- Symbols have much fewer properties
-,The `ElementSyntax` layer is gone
- New `HTMLTemplateElement`s are only created when there are multiple template controllers on the same element, or when the element is not a template. Otherwise, it reuses the one you declared in the DOM
- Attributes are not migrated across template elements during lifting because the symbol layer holds this information already

There are more (specialized) classes and I'll likely be removing a few. I was a bit overzealous with that in the initial design.

What will negatively impact perf:
- Call stack is twice as deep due to the visitors.
- There are three passes, and thus three loops through everything, instead of two

For the call stack depth:
From my earlier experiment with the AST it appeared that even in `connect` and `evaluate` that only gave a 5-10% perf hit (in extremely efficient code that performs no DOM or any other expensive operations).
In the template compiler where much more expensive work happens, this will likely be in the orders of 1-2%. I think this is an acceptable trade-off for architectural soundness and improved extensibility and such. It's laying the groundwork for more advanced capabilities.

For the multiple passes:
I don't know how this impacts perf. We may need to somehow bench it. I could probably consolidate 3 passes into 2 passes however. I just think it's easier to start with proper design and architecture and optimize perf later down the road, as opposed to starting with incredibly optimized code and then restructuring it to be able to add new features (problem I ran into which motivated this rework)

bigopon: Yes, worst case scenario: merge everything into 2, or 1 if possible 🤷‍♂️

EisenbergEffect:
This might take a little while...I'll start in on it tomorrow. Expect a few days probably.

fkleuver:
@bigopon @EisenbergEffect Apart from the review/PR notes being outdated and a handful of tests failing, this PR is close to completion.

The semantic model is now in a state where it provides true interoperability. If you look at the template compiler, it has no HTML specific processing or traversal in it. It's a plain dumb symbol-to-instruction translation.

The template binder has a bit more responsibility than what I tried to give it, but it just didn't turn out to be practical to split up parsing from lifting.

In other words, I've rewritten the template compiler from scratch and after a few huge detours and experiments I ended up with.. not exactly what I hoped to end up with (it's just too much rearranging and re-linking for what I was trying to do) but still quite a bit better than what it was before. I feel it's much easier to extend and tinker with in its current state.

A couple of things are still in a "need to further improve" state (as indicated with comments) but it works, and it works quite well I believe.

Performance wise there is less processing going on during compilation than before and fewer objects are created, but there are still a few things that need tweaks. Another nice little gain is all template elements in the original markup are reused instead of blindly lifted. The metadata model adds a fair amount of overhead during the start of compilation and this needs some caching etc.

Please take a look at:
- template-binder.ts
- template-compiler.ts
- template-factory.ts
- metadata-model.ts (very hacky, but concept stands)

packages/jit/src/debugging.ts
```diff
@@ -0,0 +1,117 @@
+import { IHTMLElement, IHTMLTemplateElement, INode, ITemplateDefinition, NodeType, TargetedInstruction, TargetedInstructionType } from '@aurelia/runtime';
```
EisenbergEffect: This seems out of place. Is there any way to move this to the debug package and handle this in a similar way to the Reporter and Tracer?

fkleuver: It is kind of out of place. Problem is that debug package doesn't depend on the jit, and making it depend on the jit would force the jit on people who only want runtime debugging.
Since nothing in the jit depends on this, it's just a loose export that can easily be tree-shaken afaik unless you explicitly use it.
The "clean" way would be to have `@aurelia/debug-runtime`, `@aurelia/debug-jit`, etc. While clean, that also feels a bit bloated though. So I don't know. This seemed like a decent middleground for the time being. What do you think?

packages/jit/src/template-binder.ts
```diff
@@ -0,0 +1,917 @@
...
+  return marker as IHTMLElement;
+}
+
+const invalidSurrogateAttribute = {
+  'id': true,
+  'part': true,
+  'replace-part': true
+};
+
+export class TemplateBinder {
```
EisenbergEffect: I understand the derivation of "binder" in this context but I wonder that it might be confusing since it's not actually related to data-binding. Is there another term that we could use instead?

packages/jit/src/debugging.ts
EisenbergEffect: Let's leave it here for now and rely on tree shaking. If we find that we add more and more jit-specific debug code, then we can evaluate whether we need more packages.

packages/jit/src/metadata-model.ts
```diff
@@ -0,0 +1,218 @@
...
+    const commands: MetadataModel['commands'] = this.commands = {};
+    const element = CustomElementResource.name;
+    const attribute = CustomAttributeResource.name;
+    const command = BindingCommandResource.name;
+    // TODO: make a less hacky way to do this
+    const container = (resources as unknown as { context: IContainer & { resolvers: Map<unknown, IResolver> } }).context;
+    let type: string;
+    let name: string;
+    let key: unknown;
+    const keys = container.resolvers.keys();
```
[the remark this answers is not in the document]
fkleuver: Yeah, I actually thought about taking that a step further and having a specialized `ResourceRegistry` of sorts, where only these string based resources are registered. Would give the metadata far fewer keys to loop through which can make a big difference in larger apps, I imagine. I couldn't see a really clean good-feeling solution just yet, hence I left it like this for the time being. What do you think about a specialized container / registry for resources?

packages/jit/src/template-binder.ts
fkleuver: I agree. It is hard to avoid putting template/attribute parsing together with dom preprocessing and symbol binding, in summary this thing just has a few too many responsibilities.

I tried to separate those concerns out into visitors earlier. Doable, sure, but I question the benefit considering how much code and overhead that adds. That's something we can get away with in AOT, but JIT maybe not so much.

So what are all the steps involved:
- Gather all registered resources and provide an easy lookup (`metadata-model`)
- Parse the html or wrap an existing node in a template (`template-factory`)
- Find and map runtime resource associated with element (`template-binder`)
- Interpret the binding syntax (`attribute-pattern`) and use it to parse attributes into attribute syntax nodes (`attribute-parser`)
- Find and map runtime resource associated with attribute target and binding command (`template-binder`)
- Determine the correct expression parsing mode (`template-binder`) and parse the attribute values (`expression-parser`)
- Glue everything together in a `ElementSymbol` with `AttributeSymbol`s, etc (`template-binder`)
- Lift, replace, rearrange DOM nodes in case of replace-parts and template controllers (`template-binder`)

So in the strict sense of the word it certainly "databinds" the DOM to the symbol tree, but you could also say it "builds", "parses" or "hydrates" the symbol tree (seems there is no single word to accurately describe what it does). Perhaps it "synthesizes" a new template from the old one, "enriched" with symbols.

I can come up with plenty of alternative but all of them seem like a lie. "precompiler" might come closer but then what are we going to call the methods. Do you have any ideas? Maybe @bigopon ?

packages/jit/src/metadata-model.ts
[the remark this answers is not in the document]
EisenbergEffect: I'd rather not have a specialized container if we can help it. We had something like that in vCurrent and it always bothered me. If this is the only real place it's needed, I'd prefer to leave this as is for now.

packages/jit/src/template-binder.ts
EisenbergEffect: TemplateLexer ?

packages/jit/src/metadata-model.ts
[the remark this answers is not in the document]
fkleuver: It's not even really "needed" here, it's just a perf optimization and for clean encapsulation. I could turn it into a lazy loaded model that just talks to the container when it needs something. That might be more efficient anyway.

packages/jit/src/template-binder.ts
bigopon: @EisenbergEffect Nice idea, also maybe `TemplateSymbolizer`?

packages/jit/src/template-factory.ts
```diff
@@ -0,0 +1,74 @@
...
+  private template: IHTMLTemplateElement;
+
+  constructor() {
+    this.template = DOM.createTemplate();
+  }
+
+  public createTemplate(markup: string): IHTMLTemplateElement;
+  public createTemplate(node: INode): IHTMLTemplateElement;
+  public createTemplate(input: unknown): IHTMLTemplateElement;
+  public createTemplate(input: string | INode): IHTMLTemplateElement {
```
bigopon: The swaps are nice. Thought it would be a nice candidate for little comment

packages/jit/src/template-factory.ts
```diff
@@ -0,0 +1,74 @@
...
+    if (typeof input === 'string') {
+      const template = this.template;
+      template.innerHTML = input;
+      const node = template.content.firstElementChild;
+      if (node.nodeName !== 'TEMPLATE') {
+        this.template = DOM.createTemplate();
+        return template;
+      }
+      template.content.removeChild(node);
```
bigopon: This is insufficient, as template string may look like this
```html
<template>
  ...
</template>
<this-stays></this-stays>
```

fkleuver: I'm not entirely sure what you mean. You're saying `<this-stays></this-stays>` should come along?

packages/jit/src/template-factory.ts
fkleuver: Look about 20 lines up :)

packages/jit/src/template-factory.ts
bigopon: In case the first parse is a template string that looks like this:
```html
<template>
  ...
</template>
<this-stays></this-stays>
```
The template of `TemplateFactory` will keep `<this-stays/>` on the second parse. and if the second parse input is an element that is not a `<template/>`, `<this-stays/>` will be returned instead. The logic here is nice but feels very puzzling.

fkleuver: Nvm the above, I misread the line you commented on.

Yeah good catch, but what should we do in such a case anyway? Only the first child is returned as well (cannot return more than 1 node after all..)

bigopon: Correction, it will look like this

fail

>  but what should we do in such a case anyway

I'm not sure but maybe it's a good candidate for a comment, maybe?

Test:

```
class TemplateFactory {
  constructor() {
    this.template = document.createElement('template');
  }
  createTemplate(input) {
    if (typeof input === 'string') {
      var template = this.template;
      template.innerHTML = input;
      var node = template.content.firstElementChild;
      if (node.nodeName !== 'TEMPLATE') {
        this.template = document.createElement('template');
        return template;
      }
      template.content.removeChild(node);
      return node;
    }
    if (input.nodeName !== 'TEMPLATE') {
      var template = this.template;
      this.template = document.createElement('template');
      template.content.appendChild(input);
      return template;
    }
    if (input.parentNode !== null) {
      input.parentNode.removeChild(input);
    }
    return input;
  }
}
var a = new TemplateFactory();
a.createTemplate('<template></template><this-stays></this-stays>');
a.createTemplate(document.createElement('not-returned'))
```

packages/jit/src/template-factory.ts
fkleuver: Oh.. you're commenting on the wrong line bruh

packages/jit/src/template-factory.ts
fkleuver: What about now?

bigopon: Nice comments. The middle I think it can be simplified to 
```js
  if (input.nodeName !== 'TEMPLATE') {
      template.content.appendChild(input);
      return template;
  }
```

fkleuver: yep you're right, changed it

packages/debug/src/reporter.ts
```diff
@@ -12,7 +12,61 @@ interface IMessageInfo {
...
+  name: 'marker',
+  params: PLATFORM.emptyArray,
+  depth: -1,
+  prev: null,
+  next: null
+};
+class TraceInfo implements ITraceInfo {
+  public static head: ITraceInfo = marker;
+  public static tail: ITraceInfo = marker;
+  public static stack: ITraceInfo[] = [];
```
BBosman: Should be initialized in the `constructor`? See #282.

packages/jit/src/attribute-pattern.ts
```diff
@@ -494,6 +494,24 @@ export class DotSeparatedAttributePattern implements DotSeparatedAttributePatter
   }
 }
 
+export interface RefAttributePattern extends IAttributePattern {}
```
BBosman: General question, shouldn't interfaces generally start with an `I`?

packages/jit/src/index.ts
```diff
@@ -4,7 +4,5 @@ export * from './attribute-pattern';
 export * from './binding-command';
 export * from './common';
 export * from './configuration';
-export * from './element-parser';
 export * from './expression-parser';
-export * from './semantic-model';
 export * from './template-compiler';
```
BBosman: Don't `./template-binder`, `./debugging`, `metadata-model` and `template-factory` need to be exported here?

packages/runtime/src/binding/binding-context.ts
```diff
@@ -82,47 +89,80 @@ export class BindingContext implements IBindingContext {
...
     return observers;
   }
 }
 
 export class Scope implements IScope {
   public readonly bindingContext: IBindingContext | IBindScope;
   public readonly overrideContext: IOverrideContext;
+  // parentScope is strictly internal API and mainly for replaceable template controller.
+  // NOT intended for regular scope traversal!
+  /*@internal*/public readonly parentScope: IScope | null;
```
BBosman: Maybe do a find-replace-all `/*@internal*/` with `/** @internal */` across the whole PR, as a follow up to #337?

packages/runtime/src/dom.interfaces.ts
```diff
@@ -0,0 +1,340 @@
...
+  readonly parentStyleSheet: IStyleSheet | null;
+}
+
+export interface IStyleSheetList {
+  readonly length: number;
+  item(index: number): IStyleSheet | null;
+  [index: number]: IStyleSheet;
+}
+
+export type IShadowRootMode = 'open' | 'closed';
```
BBosman: I wouldn't have expected a `type` to start witn an `I`?

packages/runtime/src/dom.interfaces.ts
```diff
@@ -0,0 +1,340 @@
...
+  removeProperty(propertyName: string): string;
+  setProperty(propertyName: string, value: string | null, priority?: string | null): void;
+  [index: number]: string;
+}
+
+export interface IElementCSSInlineStyle {
+  readonly style: ICSSStyleDeclaration;
+}
+
+export type IInsertPosition = 'beforebegin' | 'afterbegin' | 'beforeend' | 'afterend';
```
BBosman: I wouldn't have expected a `type` to start witn an `I`?

packages/runtime/src/dom.interfaces.ts
```diff
@@ -0,0 +1,340 @@
...
+  attributeFilter?: string[];
+  attributeOldValue?: boolean;
+  attributes?: boolean;
+  characterData?: boolean;
+  characterDataOldValue?: boolean;
+  childList?: boolean;
+  subtree?: boolean;
+}
+
+export type IMutationRecordType = 'attributes' | 'characterData' | 'childList';
```
BBosman: I wouldn't have expected a `type` to start with an `I`?

packages/debug/src/reporter.ts
fkleuver: These are static properties, and unlike C# there is no `static constructor` in TypeScript unfortunately

packages/runtime/src/dom.ts
```diff
@@ -1,160 +1,62 @@
...
+declare var document: IDocument;
+declare var MutationObserver: Constructable & IMutationObserver;
+declare var Element: IElement;
+declare var HTMLElement: IHTMLElement;
+declare var SVGElement: ISVGElement;
 
 // tslint:disable:no-any
 export const DOM = {
-  createDocumentFragment(markupOrNode?: string | IElement): IDocumentFragment {
+  createDocumentFragment(markupOrNode?: unknown): IDocumentFragment {
```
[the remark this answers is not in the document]
fkleuver: See https://github.com/aurelia/aurelia/pull/306#discussion_r240387403

packages/runtime/src/dom.interfaces.ts
fkleuver: See https://github.com/aurelia/aurelia/pull/306#discussion_r240385900

packages/runtime/src/dom.interfaces.ts
fkleuver: See https://github.com/aurelia/aurelia/pull/306#discussion_r240385900

packages/runtime/src/dom.ts
```diff
@@ -1,160 +1,62 @@
...
-export interface INodeObserver {
-  disconnect(): void;
-}
+declare var document: IDocument;
+declare var MutationObserver: Constructable & IMutationObserver;
+declare var Element: IElement;
+declare var HTMLElement: IHTMLElement;
+declare var SVGElement: ISVGElement;
 
 // tslint:disable:no-any
```
BBosman: Does this need to be a global `disable`? There aren't many `any` uses in here, so I'd rather make them explicit with a `disable-next-line`.

fkleuver: Huh, didn't realize that one was still there. You're right. I fixed the 3 occurrences of it as well.

packages/runtime/src/dom.ts
```diff
@@ -1,160 +1,62 @@
...
+      fragment.appendChild(markupOrNode);
+      return fragment;
     }
-    return DOM.createTemplate(markupOrNode as string).content as IDocumentFragment;
+    return DOM.createTemplate(markupOrNode).content;
   },
-  createTemplate(markup?: string): IElement {
-    if (markup === undefined) {
-      return document.createElement('template') as IElement;
+  createTemplate(markup?: unknown): IHTMLTemplateElement {
```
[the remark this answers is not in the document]
BBosman: Valid point. Once we have alternative runtime implementations we can get a better idea of the typing constraints we'll need to support other implementations in a typesafe manner.

packages/debug/src/reporter.ts
BBosman: Doh... Missed the `static` part. 😊

packages/runtime/src/dom.interfaces.ts
BBosman: Maybe wrap them in a `namespace`?

fkleuver: That could be a pretty neat solution actually. The only "problem" that would then introduce is (I realized this just now), is all the `lib.dom.d.ts` interfaces are also not prefixed with `I`. It would then feel weird to have a double deviation from the normal dom types (namespace *and* prefix).

packages/jit/src/attribute-pattern.ts
```diff
@@ -494,6 +494,24 @@ export class DotSeparatedAttributePattern implements DotSeparatedAttributePatter
...
+
+@attributePattern(
+  { pattern: 'ref', symbols: '' },
+  { pattern: 'ref.PART', symbols: '.' }
```
bigopon: Probably we will need `{ pattern: 'PART.ref', symbols: '.' }` as well.

fkleuver: How does that work in vCurrent?
Btw, this is just pro forma. The template compiler doesn't actually do anything with ref binding commands yet. I'm not sure what it's supposed to do with them. Could you clarify?

bigopon: behavior (read element /atribute) reference will be stored on element, via namespace `au`. Then ref binding (NameBinding) will look up from there based on the attribute before the command `ref`. So we need that, at least for compat reason:
```html
<template>
  <div custom-attr="aaa" custom-attr.ref="customAttrViewModel"></div>
  <custom-el view-model.ref="customElViewModel"></custom-el>
</template>
```

fkleuver: Would you mind creating an issue briefly describing this behavior like you did now? Then I'll pick that up in a follow up pr

PR #309, opened 2018-11-21
refactor: linting fixes for IIndexable

BBosman:
# Pull Request

## 📖 Description

This is a WIP.

This changes `IIndexable` from `any` to `unknown` and fixes call sites, propagating out to the API boundaries. It also switches from `IIndexable` to base `object` in a few places where we're not actually interested type-wise that it is indexable, just that it's an object. 

### 🎫 Issues

Part of #249.

## 👩‍💻 Reviewer Notes

Is this the right way to go? Especially interested in feedback from @fkleuver as he had some ideas on where he wanted to go with this, but open to feedback from all. 👍

(And all these changes are just for one `any` -> `unknown` linting warning fix 😁)

## 📑 Test Plan

CircleCI

## ⏭ Next Steps

See #249.

### DIFF
--- a/packages/runtime/src/binding/array-observer.ts
+++ b/packages/runtime/src/binding/array-observer.ts
@@ -1,4 +1,3 @@
-import { IIndexable, Primitive } from '@aurelia/kernel';
 import { ILifecycle } from '../lifecycle';
 import { CollectionKind, ICollectionObserver, IndexMap, IObservedArray, LifecycleFlags } from '../observation';
 import { collectionObserver } from './collection-observer';
@@ -142,7 +141,7 @@ function observeReverse(this: IObservedArray): ReturnType<typeof nativeReverse> 
 
 // https://tc39.github.io/ecma262/#sec-array.prototype.sort
 // https://github.com/v8/v8/blob/master/src/js/array.js
-function observeSort(this: IObservedArray, compareFn?: (a: IIndexable | Primitive, b: IIndexable | Primitive) => number): IObservedArray {
+function observeSort(this: IObservedArray, compareFn?: (a: unknown, b: unknown) => number): IObservedArray {
   const o = this.$observer;
   if (o === undefined) {
     return nativeSort.call(this, compareFn);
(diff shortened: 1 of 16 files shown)


packages/runtime/src/binding/property-observation.ts
```diff
@@ -49,18 +49,18 @@ export interface SetterObserver extends IPropertyObserver<IIndexable, string> {}
 export class SetterObserver implements SetterObserver {
   public subscribe: (subscriber: IPropertySubscriber) => void;
   public unsubscribe: (subscriber: IPropertySubscriber) => void;
-  public obj: IIndexable;
+  public obj: object;
```
fkleuver: I think this is one out of less than a handful of places in the framework where `IIndexable` is actually the most appropriate type :)
Also look at `observer-locator` - some observers are never created without particular type assertions. In those cases the `obj` property of the observer can be typed accordingly. Off the top of my head this only applies to `Observer` and `SetterObserver` though

packages/runtime/src/binding/array-observer.ts
```diff
@@ -168,7 +168,7 @@ function observeSort(this: IObservedArray, compareFn?: (a: IIndexable | Primitiv
 }
 
 // https://tc39.github.io/ecma262/#sec-sortcompare
-function sortCompare(x: IIndexable | Primitive, y: IIndexable | Primitive): number {
+function sortCompare(x: object | Primitive, y: object | Primitive): number {
```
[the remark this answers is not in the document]
BBosman: I think I can get away with `unknown` and duck-typing it to something with a `toString` method below.

packages/runtime/src/binding/target-accessors.ts
```diff
@@ -76,17 +75,17 @@ export class DataAttributeAccessor implements DataAttributeAccessor {
...
 }
 
-export interface StyleAttributeAccessor extends IBindingTargetAccessor<IHTMLElement, 'style', string | IIndexable> {}
+export interface StyleAttributeAccessor extends IBindingTargetAccessor<IHTMLElement, 'style', string | object> {}
 
 @targetObserver()
 export class StyleAttributeAccessor implements StyleAttributeAccessor {
-  public currentValue: string | IIndexable;
-  public defaultValue: string | IIndexable;
+  public currentValue: string | object;
```
fkleuver:
```suggestion
  public currentValue: string | Record<string, string>;
```
In this unique scenario, we do actually know precisely what type this can (or is allowed to) be.

packages/runtime/src/binding/target-accessors.ts
```diff
@@ -230,45 +229,45 @@ ClassAttributeAccessor.prototype.doNotCache = true;
...
   }
 
-  public setValueCore(value: Primitive | IIndexable): void {
+  public setValueCore(value: unknown): void {
     this.obj[this.propertyKey] = value;
   }
 }
 
-export interface PropertyAccessor extends IBindingTargetAccessor<IIndexable, string, Primitive | IIndexable> {}
+export interface PropertyAccessor extends IBindingTargetAccessor<object, string> {}
```
fkleuver: Can this be `IIndexable` instead of `object`?

packages/runtime/src/templating/create-element.ts
```diff
@@ -49,7 +49,7 @@ export class RenderPlan {
   }
 }
 
-function createElementForTag(tagName: string, props?: IIndexable, children?: ArrayLike<ChildType>): RenderPlan {
+function createElementForTag(tagName: string, props?: object, children?: ArrayLike<ChildType>): RenderPlan {
```
fkleuver: Can we try something like `Record<string, string | ITargetedInstruction>` here?

fkleuver:
> (And all these changes are just for one any -> unknown linting warning fix 😁)

Well.. that's what it seemed like on the surface, but the rest might as well have been `any`. It wasn't that much better. Now it's getting there, though :)

BBosman:
Made the requested changes.

fkleuver:
@EisenbergEffect let's get this merged

PR #310, opened 2018-11-22
refactor: add Sonarts linting rules

BBosman:
# Pull Request

## 📖 Description

This is just an experiment based on a discussion with @zewa666 and @fkleuver. 

It appears that Code Climate runs the Sonarts linting rules. The problem is that we only see those issues once we've created a PR. It would be beneficial to see those issues earlier in the process, so this PR brings those rules into our own ruleset.

One of my goals was to make this PR *violation neutral*, so ~~it does not introduce new linting errors.~~ I did cheat a bit ~~though, as i suppressed a few~~ and added `// TODO: fix` style comments.

### 🎫 Issues

Related to #249.

## 👩‍💻 Reviewer Notes

Notable things:
- The rule with the most violations is `cognitive-complexity`, which has a default value of 15. After a discussion with @fkleuver I set it to 30 for now and ~~suppressed~~ *marked* the higher ones with `// TODO: fix` style comments.
- I suppressed some `no-ignored-initial-value` as it doesn't like arguments named `this`.
- I disables the `no-identical-functions` and `no-duplicate-string` rules on `tslint.json` level, as fixing those would lead to less readable/maintainable code imho.

## 📑 Test Plan

Trust CircleCI and I'm curious what Code Climate says.

## ⏭ Next Steps

Decide if we want to use this, and if so, with what rule configuration.

### DIFF
--- a/packages/jit/src/expression-parser.ts
+++ b/packages/jit/src/expression-parser.ts
@@ -91,6 +91,14 @@ export function parseCore(input: string, bindingType?: BindingType): IExpression
 }
 
 /*@internal*/
+// JUSTIFICATION: This is performance-critical code which follows a subset of the well-known ES spec.
+// Knowing the spec, or parsers in general, will help with understanding this code and it is therefore not the
+// single source of information for being able to figure it out.
+// It generally does not need to change unless the spec changes or spec violations are found, or optimization
+// opportunities are found (which would likely not fix these warnings in any case).
+// It's therefore not considered to have any tangible impact on the maintainability of the code base.
+// For reference, most of the parsing logic is based on: https://tc39.github.io/ecma262/#sec-ecmascript-language-expressions
+// tslint:disable-next-line:no-big-function cognitive-complexity
 export function parse<TPrec extends Precedence, TType extends BindingType>(state: ParserState, access: Access, minPrecedence: TPrec, bindingType: TType):
   TPrec extends Precedence.Unary ? IsUnary :
   TPrec extends Precedence.Binary ? IsBinary :
@@ -183,7 +191,6 @@ export function parse<TPrec extends Precedence, TType extends BindingType>(state
           } else if (state!.currentToken === Token.EOF) {
             throw Reporter.error(SyntaxError.ExpectedIdentifier, { state });
           }
-          continue;
         } else if (state.currentToken & Token.AccessScopeTerminal) {
           const ancestor = access & Access.Ancestor;
           result = ancestor === 0 ? $this : ancestor === 1 ? $parent : new AccessThis(ancestor);
@@ -673,6 +680,7 @@ function parseInterpolation(state: ParserState): Interpolation {
  */
 function parseTemplate(state: ParserState, access: Access, bindingType: BindingType, result: IsLeftHandSide, tagged: boolean): TaggedTemplate | Template {
   const cooked = [state.tokenValue as string];
+  // TODO: properly implement raw parts / decide whether we want this
   //const raw = [state.tokenRaw];
   consume(state, Token.TemplateContinuation);
   const expressions = [parse(state, access, Precedence.Assign, bindingType)];
--- a/packages/jit/src/semantic-model.ts
+++ b/packages/jit/src/semantic-model.ts
@@ -276,6 +276,7 @@ export class AttributeSymbol implements IAttributeSymbol {
     return this._isProcessed;
   }
 
+  // TODO: Reduce complexity (currently at 60)
   constructor(
     semanticModel: SemanticModel,
     $element: ElementSymbol,
--- a/packages/jit/src/template-compiler.ts
+++ b/packages/jit/src/template-compiler.ts
@@ -116,11 +116,10 @@ export class TemplateCompiler implements ITemplateCompiler {
         // Doesn't make sense for these properties as they need to be unique
         const name = $attr.target;
         if (name !== 'id' && name !== 'part' && name !== 'replace-part') {
+          // tslint:disable-next-line:no-small-switch
           switch (name) {
             // TODO: handle simple surrogate style attribute
             case 'style':
-              attrInst = new SetAttributeInstruction($attr.rawValue, name);
-              break;
             default:
               attrInst = new SetAttributeInstruction($attr.rawValue, name);
           }
@@ -310,18 +309,15 @@ export class TemplateCompiler implements ITemplateCompiler {
           }
         }
       }
-      // plain attribute on a custom element
-      if ($attr.onCustomElement) {
-        // bindable attribute
-        if ($attr.isElementBindable) {
-          const expression = parser.parse($attr.rawValue, BindingType.Interpolation);
-          if (expression === null) {
-            // no interpolation -> make it a setProperty on the component
-            return new SetPropertyInstruction($attr.rawValue, $attr.to);
-          }
-          // interpolation -> behave like toView (e.g. foo="${someProp}")
-          return new InterpolationInstruction(expression, $attr.to);
+      // plain bindable attribute on a custom element
+      if ($attr.onCustomElement && $attr.isElementBindable) {
+        const expression = parser.parse($attr.rawValue, BindingType.Interpolation);
+        if (expression === null) {
+          // no interpolation -> make it a setProperty on the component
+          return new SetPropertyInstruction($attr.rawValue, $attr.to);
         }
+        // interpolation -> behave like toView (e.g. foo="${someProp}")
+        return new InterpolationInstruction(expression, $attr.to);
       }
       {
         // plain attribute on a normal element
--- a/packages/runtime/package.json
+++ b/packages/runtime/package.json
@@ -76,6 +76,7 @@
     "tslib": "^1.9.3",
     "tslint": "^5.11.0",
     "tslint-microsoft-contrib": "^5.2.1",
+    "tslint-sonarts": "^1.8.0",
     "typescript": "^3.1.6",
     "webpack": "^4.25.1"
   }
--- a/packages/runtime/src/binding/array-observer.ts
+++ b/packages/runtime/src/binding/array-observer.ts
@@ -24,7 +24,8 @@ function observePush(this: IObservedArray): ReturnType<typeof nativePush> {
   this.length = o.indexMap.length = len + argCount;
   let i = len;
   while (i < this.length) {
-    this[i] = arguments[i - len]; o.indexMap[i] = - 2;
+    this[i] = arguments[i - len];
+    o.indexMap[i] = - 2;
     i++;
   }
   o.callSubscribers('push', arguments, LifecycleFlags.isCollectionMutation);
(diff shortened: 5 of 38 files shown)


BBosman:
Funny.... This actually kills Code Climate... ☹️

packages/runtime/src/binding/binding-context.ts
```diff
@@ -21,14 +21,16 @@ export class InternalObserversLookup {
   }
 }
 
+type BindingContextValue = ObservedCollection | StrictPrimitive | IIndexable;
```
fkleuver: I like this idea

packages/runtime/src/binding/array-observer.ts
```diff
@@ -226,43 +236,64 @@ function quickSort(arr: IObservedArray, indexMap: IndexMap, from: number, to: nu
...
     thirdIndex = from + ((to - from) >> 1);
-    v0 = arr[from];       i0 = indexMap[from];
-    v1 = arr[to - 1];     i1 = indexMap[to - 1];
-    v2 = arr[thirdIndex]; i2 = indexMap[thirdIndex];
+    v0 = arr[from];
+    i0 = indexMap[from];
+    v1 = arr[to - 1];
+    i1 = indexMap[to - 1];
+    v2 = arr[thirdIndex];
+    i2 = indexMap[thirdIndex];
```
fkleuver: Hmm. Not sure about this. The point here (and with other array observer methods) of having the index map on the same line as the array operation, is to shift focus away from the index map / emphasize that all array operations are exactly duplicated on the index map. Since all operations on the index map are identical, keeping the two on the same line effectively keeps "one operation per line" and I believe that to be more readable. But maybe that's just me.
@bigopon  @EisenbergEffect thoughts?

packages/runtime/package.json
```diff
@@ -43,7 +43,8 @@
     "access": "public"
   },
   "dependencies": {
-    "@aurelia/kernel": "0.3.0"
+    "@aurelia/kernel": "0.3.0",
+    "tslint-sonarts": "^1.8.0"
```
fkleuver: This should be dev dependency.

packages/jit/src/template-compiler.ts
```diff
@@ -264,74 +265,71 @@ export class TemplateCompiler implements ITemplateCompiler {
...
+      // plain attribute on a normal element
+      const expression = parser.parse($attr.rawValue, BindingType.Interpolation);
+      if (expression === null) {
+        // no interpolation -> do not return an instruction
+        return null;
+      }
+      // interpolation -> behave like toView (e.g. id="${someId}")
+      return new InterpolationInstruction(expression, $attr.to);
+    }
+}
```
fkleuver: What happened here? :P looks like lots of changes but they aren't changes. I think this brace is misplaced?

packages/jit/src/template-compiler.ts
```diff
@@ -84,6 +85,7 @@ export class TemplateCompiler implements ITemplateCompiler {
       case NodeType.Text:
         const expression = this.exprParser.parse((<Text>$el.node).wholeText, BindingType.Interpolation);
         if (expression === null) {
+          // tslint:disable-next-line:no-extra-semicolon
```
fkleuver: I think we should disable this rule in tslint.json because I absolutely disagree with it in every case. So that would just be a suppression on every occurrence. Explicit semicolon insertion > automatic semicolon insertion, especially when the `while` loop has no code block in it.

packages/jit/src/semantic-model.ts
```diff
@@ -276,6 +276,7 @@ export class AttributeSymbol implements IAttributeSymbol {
     return this._isProcessed;
   }
 
+  // tslint:disable-next-line:cognitive-complexity // TODO: Reduce complexity
```
fkleuver: I think we shouldn't suppress something when there is an intent to fix it. A TODO with the current complexity is good enough I think. We don't want to get or create the illusion that the code base is in a better state than it really is :)

packages/jit/src/expression-parser.ts
```diff
@@ -673,21 +673,16 @@ function parseInterpolation(state: ParserState): Interpolation {
  */
 function parseTemplate(state: ParserState, access: Access, bindingType: BindingType, result: IsLeftHandSide, tagged: boolean): TaggedTemplate | Template {
   const cooked = [state.tokenValue as string];
-  //const raw = [state.tokenRaw];
```
fkleuver: I'd rather leave these in, perhaps with a TODO comment along the lines of "properly implement raw parts / decide whether we want this". Because this is a slight deviation from the spec and it's important that we either fix it or make an explicit, documented decision not to.

packages/jit/src/expression-parser.ts
```diff
@@ -91,6 +91,7 @@ export function parseCore(input: string, bindingType?: BindingType): IExpression
 }
 
 /*@internal*/
+// tslint:disable-next-line:no-big-function cognitive-complexity
```
fkleuver:
```suggestion
// JUSTIFICATION: This is performance-critical code which follows a subset of the well-known ES spec.
// Knowing the spec, or parsers in general, will help with understanding this code and it is therefore not the
// single source of information for being able to figure it out.
// It generally does not need to change unless the spec changes or spec violations are found, or optimization
// opportunities are found (which would likely not fix these warnings in any case).
// It's therefore not considered to have any tangible impact on the maintainability of the code base.
// For reference, most of the parsing logic is based on: https://tc39.github.io/ecma262/#sec-ecmascript-language-expressions
// tslint:disable-next-line:no-big-function cognitive-complexity
```

packages/runtime/src/binding/computed-observer.ts
```diff
@@ -22,7 +22,7 @@ export function computed(config: ComputedOverrides): PropertyDecorator {
 }
 
 // tslint:disable-next-line:no-typeof-undefined
-const noProxy = !(typeof Proxy !== 'undefined');
+const noProxy = typeof Proxy === 'undefined';
```
fkleuver: 😆

packages/runtime/src/binding/element-observation.ts
```diff
@@ -111,11 +111,9 @@ export class ValueAttributeObserver implements ValueAttributeObserver {
 
   private flushFileChanges(): void {
     const currentValue = this.currentValue;
-    if (this.oldValue !== currentValue) {
-      if (currentValue === '') {
-        this.setValueCore(currentValue, this.currentFlags);
-        this.oldValue = this.currentValue;
-      }
+    if (this.oldValue !== currentValue && currentValue === '') {
```
fkleuver: Yep nice one. This is another one of those examples where there used to be an `else`, and it was removed without restructuring accordingly.

packages/runtime/src/binding/element-observation.ts
```diff
@@ -305,16 +303,18 @@ export interface IOptionElement extends IElement {
   value: string;
 }
 
+type SelectValueObserverValue = Primitive | IIndexable | UntypedArray;
```
fkleuver: Good idea :)

packages/runtime/src/binding/expression-parser.ts
```diff
@@ -1,12 +1,14 @@
 import { DI, PLATFORM, Reporter } from '@aurelia/kernel';
 import { AccessMember, AccessScope, CallMember, CallScope, ExpressionKind, ForOfStatement, Interpolation, IsBindingBehavior, PrimitiveLiteral } from './ast';
 
+type Expression = Interpolation | ForOfStatement | IsBindingBehavior;
```
fkleuver: Being very nit picky, but this isn't really an expression. `for .. of` is actually a [statement](https://tc39.github.io/ecma262/#sec-for-in-and-for-of-statements). I think `ExpressionOrStatement`, albeit more verbose, is the more accurate type here. Wouldn't be an issue if it's the only type we used, but we do differentiate between expressions and expressionOrStatements in various places.

On second thought, that's still wrong, because an interpolation (by our implementation) is neither an expression nor a statement, but an aurelia-specific construct.

Perhaps `BindingExpression` then?

packages/runtime/src/binding/target-observer.ts
```diff
@@ -31,12 +31,10 @@ function setValue(this: BindingTargetAccessor, newValue: Primitive | IIndexable,
 }
 
 function flush(this: BindingTargetAccessor, flags: LifecycleFlags): void {
-  if (flags & LifecycleFlags.doNotUpdateDOM) {
-    if (DOM.isNodeInstance(this.obj)) {
-      // re-queue the change so it will still propagate on flush when it's attached again
-      this.lifecycle.enqueueFlush(this).catch(error => { throw error; });
-      return;
-    }
+  if (flags & LifecycleFlags.doNotUpdateDOM && DOM.isNodeInstance(this.obj)) {
```
fkleuver:
```suggestion
  if ((flags & LifecycleFlags.doNotUpdateDOM) && DOM.isNodeInstance(this.obj)) {
```
I know this doesn't matter functionally, but bitwise operators aren't particularly well known by most developers so I'd like to keep the extra clarify.

packages/runtime/src/definitions.ts
```diff
@@ -292,6 +293,7 @@ export function buildTemplateDefinition(
   containerless?: boolean | null,
   shadowOptions?: { mode: 'open' | 'closed' } | null,
   hasSlots?: boolean | null): TemplateDefinition;
+  // tslint:disable-next-line:parameters-max-number cognitive-complexity // TODO: Reduce complexity
```
fkleuver: Same as with the semantic model, I'd prefer not to suppress stuff we intend to fix.
The parameter-max-number can stay however

packages/runtime/src/observation.ts
```diff
@@ -366,11 +366,6 @@ export interface IScope {
   readonly overrideContext: IOverrideContext;
 }
 
-export interface IObserversLookup<TObj extends IIndexable = IIndexable, TKey extends keyof TObj =
```
fkleuver: Hmrpf. I did leave a thing or two laying around didn't I..
I'd prefer to keep this however with a //TODO comment "currently unused, still need to fix the observersLookup type". The observers lookup is a messy aspect of the runtime I really still need to fix.

packages/runtime/src/templating/lifecycle-attach.ts
```diff
@@ -6,6 +6,7 @@ import { ICustomAttribute } from './custom-attribute';
 import { ICustomElement } from './custom-element';
 
 /*@internal*/
+// tslint:disable-next-line:no-ignored-initial-value
```
fkleuver: What's the ignored value here?

packages/runtime/src/binding/array-observer.ts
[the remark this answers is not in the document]
fkleuver: You mean just a bit more spacing? Not a bad idea actually..

fkleuver:
> Funny.... This actually kills Code Climate...

It doesn't understand the sonar-ts rules (same with the tslint-microsoft-contrib rules it seems). I wonder if there's a way we can tell code climate to ignore them.

@EisenbergEffect you happen to know more about this?

packages/runtime/src/templating/lifecycle-attach.ts
BBosman: That's `this`. I think the rule name is misleading in this case, as it's more a case of shadowing `this` than ignoring the initial value.

BBosman:
Made some changes @fkleuver:
- Removed `cognitive-complexity` suppressions and added their current complexity to the `TODO` comments.
- Globally disabled `no-extra-semicolon` and removed the individual suppressions.
- Made `sonarts-tslint` a `devDependency` instead of a regular `dependency`. Rookie mistake 😊 
- Reverted the *multiline* changes in `array-observer.ts` and added suppressions. Did increase the spacing for better alignment as suggested by @bigopon. 👍 
- Other requested changes.

BBosman:
@EisenbergEffect Rebased.

EisenbergEffect:
Do I need to disable some code climate settings after this? What was the conclusion on that?

fkleuver:
@EisenbergEffect See https://github.com/aurelia/aurelia/pull/314

PR #314, opened 2018-11-24
chore(codeclimate): disable tslint on codeclimate

fkleuver:
# Pull Request

## 📖 Description

The `sonarts` and `tslint-microsoft-contrib` custom tslint rules cause CodeClimate to crash. This PR disables tslint on CodeClimate following discussion around https://github.com/aurelia/aurelia/pull/310

### 🎫 Issues

Follow-up to: https://github.com/aurelia/aurelia/pull/310

## ⏭ Next Steps

We could create an secondary tslint.json config with just the rules that CodeClimate supports, and point codeclimate.yml to that one. This isn't particularly important since we already have a lint job on CI, but it is a potential point of improvement for the future

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -58,8 +58,7 @@ checks:
     enabled: false # default true
 plugins:
   tslint:
-    enabled: true
-    config: tslint.json
+    enabled: false
 exclude_patterns:
 # Exclude code that is not actually built and published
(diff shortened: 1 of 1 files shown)


BBosman:
I would have expected Code Climate to stop choking with this PR?

EisenbergEffect:
Maybe it needs to be merged? We'll see...

Issue #307, opened 2018-11-21
[RFC] [WIP] Dynamic routing


jwx:
# 💬 RFC

A proposal for a feature set where the router doesn't use traditional url/path routes.

## 🔦 Context

A lot of the trickier part of routing -- route recognizing, child routing, defining navigation statically -- is related to routes. By removing the routes, we can get rid of some of the issues as well.

## 💻 Examples

A fully functional, routeless router would have the following url
```
/viewport:component/=parameters (not sure about the parameters part yet)
```
Typical examples would be:
```
/default:start-page  -- just a simple start page in the default viewport
/master:users/detail:user/=id=123 -- a master-detail in two viewports
```
Now this can be simplified: 
* default viewport can be left out
* a single viewport application can also leave out the viewport
* components that are decorated `@viewport('master')` or has static variable `viewport = 'detail'` can also leave out the viewport
* components that are decorated `@parameters(['id'])` or has static variable `parameter = ['id']` can leave out parameter names
* and more to come further down

Same examples would then be:
```
/start-page  -- just a simple start page in the default viewport
/users/user/=123 -- a master-detail in two viewports
```
The key would be to get to decisions made by convention and as little configuration as possible. Whenever there's only one option left, the router will be able to make the decision. So, more ways to eliminate viewport options are
* the viewports can specify which components can be placed in 
```html
<viewport name="top" used-by="login,menu"></viewport>
<viewport name="properties" used-by="property*"></viewport>
<viewport name="main"></viewport>
```
After the above, you could just go
```html
<a href="#/login/file/property-file/=123">Show menu, a file and its properties with one click</a>
```
The router can also chime in
```typescript
router.setViewports({ top: [ 'login', 'menu' ], main: '*' });
```
But specification from the parent can always overrule conventions, router and viewport configuration and child decorators/static variables:
```html
<a href="#/main:property-major/properties:about-editing">Properties in main viewport and editing help in properties</a>
```
To be continued...

davismj:
Seems like this would shut the door on using the same component in two separate viewports.

fkleuver:
Could the `viewport` property be of type `string | string[]`?

EisenbergEffect:
I confess to be having a hard time understanding why we would invite a new string-based syntax for routes. For example, if we want a developer to be able to tell the router how to adhoc navigate to a specific component and viewport config, why not create a router API directly for that? If they need to generate links for that, why not just have use a custom attribute to configure the params for the same underlying router API?

[...]

PR #315, opened 2018-11-25
[WIP] Router

jwx:
# Pull Request

[WIP] Router

## 📖 Description

WIP of a Router. The ambition is to make it routeless. Test application can be found in `packages/router/test/e2e/modules`.


packages/router/src/router.ts
```diff
@@ -0,0 +1,335 @@
...
+
+  public back(): void {
+    this.historyBrowser.back();
+  }
+
+  public forward(): void {
+    this.historyBrowser.forward();
+  }
+
+  private closestScope(element: Element): Scope {
```
fkleuver: DOM traversal is something I'd *really* rather avoid. There are all kinds of scenarios where sections of the DOM tree are not connected and this would fall apart.

packages/router/src/history-browser.ts
```diff
@@ -0,0 +1,264 @@
...
+
+export interface INavigationInstruction extends IHistoryEntry, INavigationFlags { }
+
+export class HistoryBrowser {
+  public currentEntry: IHistoryEntry;
+  public historyEntries: IHistoryEntry[];
+  public historyOffset: number;
+  public replacedEntry: IHistoryEntry;
+
+  public history: History;
```
fkleuver: As we discussed I think these hard dependencies on DOM are acceptable for a first version, but we do want to start thinking about how we'll abstract these as we do eventually want to stop depending on the DOM directly

packages/router/src/history-browser.ts
```diff
@@ -0,0 +1,264 @@
...
+    const state = { ...this.history.state };
+    const { pathname, search } = this.location;
+    this.history.replaceState(state, null, `${pathname}${search}#/${path}`);
+  }
+
+  get titles(): string[] {
+    return (this.historyEntries ? this.historyEntries.slice(0, this.currentEntry.index + 1).map((value) => value.title) : []);
+  }
+
+  private pathChanged = (): void => {
```
fkleuver: With code such as this (which is rather critical to the router) imo it's worth the trouble to spend some extra time cleaning things up and commenting them. It might even make sense to have this in its own specialized class - it makes it easier to unit test as well. I'd love to see a couple of unit tests touching this code and verifying some of the use cases

packages/router/src/router.ts
```diff
@@ -0,0 +1,335 @@
...
+    this.historyBrowser = new HistoryBrowser();
+  }
+
+  public activate(options?: IRouterOptions): void {
+    if (this.isActive) {
+      throw new Error('Router has already been activated.');
+    }
+
+    this.isActive = true;
+    this.options = {
```
fkleuver: This looks funny.. :) what's going on here?

packages/router/src/router.ts
```diff
@@ -0,0 +1,335 @@
...
+    }
+
+    // We've gone via a redirected route back to same viewport status so
+    // we need to remove the added history entry for the redirect
+    if (!viewports.length && this.isRedirecting) {
+      this.historyBrowser.cancel();
+      this.isRedirecting = false;
+    }
+
+    let cancel: boolean = false;
```
fkleuver: Like with the `pathChanged` method, this is the kind of code that will cause the biggest issues on the long term and it needs extra care. Also, is there any particular reason why you're not using `async`? I think it would make the code more readable and easier to debug

packages/router/src/router.ts
```diff
@@ -0,0 +1,335 @@
...
+    return Promise.all(viewports.map((value) => value.canLeave()))
+      .then((promises: boolean[]) => {
+        if (cancel || promises.findIndex((value) => value === false) >= 0) {
+          cancel = true;
+          return Promise.resolve([]);
+        } else {
+          return Promise.all(viewports.map((value) => value.canEnter()));
```
fkleuver: I suppose you don't have to address this during this PR, but this is definitely something we'll want a proper lifecycle class for to manage all the transitions and edge cases

packages/router/src/router.ts
```diff
@@ -0,0 +1,335 @@
...
+      }).then(() => {
+        if (cancel) {
+          this.historyBrowser.cancel();
+        }
+        // tslint:disable-next-line:no-console
+        console.log('=========== ROUTER', this);
+      }).then(() => {
+        // TODO: Eliminate duplications
+        const viewportStates = this.rootScope.viewportStates();
+        this.historyBrowser.replacePath(viewportStates.join('/'));
```
fkleuver: This code confuses me a little. How does this work?

packages/router/src/scope.ts
```diff
@@ -0,0 +1,131 @@
...
+    // Need more ways to resolve viewport based on component name!
+    const comp = this.resolveComponent(component);
+    if (comp.viewport) {
+      name = comp.viewport;
+      return this.viewports[name];
+    }
+    return null;
+  }
+
+  public addViewport(name: string, element: Element, newScope: boolean): Viewport {
```
fkleuver: What is `newScope` for?

packages/router/src/scope.ts
```diff
@@ -0,0 +1,131 @@
...
+      }
+    }
+    if (element) {
+      if (viewport.scope && !viewport.scope.element) {
+        viewport.scope.element = element;
+      }
+      if (!viewport.element) {
+        viewport.element = element;
+        if (!viewport.element.children) {
+          this.renderViewport(viewport);
```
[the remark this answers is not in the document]
fkleuver: I would even argue to go for `async` / `await` with the router, since it's not performance critical code and it would benefit the readability. Would you agree @BBosman ?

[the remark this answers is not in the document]
BBosman: I agree that `async` / `await` improves readability significantly, but how does it relate to the recent discussion in #273 and #280, especially https://github.com/aurelia/aurelia/pull/280#issuecomment-438282961?

fkleuver:
All the nit picking in my comments aside, overall I think this is a very impressive piece of work for the first PR to the router. You've clearly thought things through.

The API for the most part makes sense at first glance. Stuff is missing obviously, but that's no surprise given the scope of this endeavor.

As we already discussed in chat at various points in time, my concerns are not so much with the general approach - you're staying close to the native browser API's which in itself is a big win. And I like your unique approach with the convention based viewport routing as an extra option.

My concern is the lack of tests, lack of encapsulation and lack of proper lifecycle management. The limitations will not surface until we're relatively far down the road and start getting into advanced scenarios with recursion, inter-dependent lifecycle states, complex conditional transitions, pipeline hooks (e.g. auth), async tasks/animations and such.

So while I think this PR is pretty close to good enough for an early first version, I would like to stress that the follow up steps (and the one after and the one after) should place significant emphasis on improving API clarity, unit tests, and overall code quality.

packages/router/src/scope.ts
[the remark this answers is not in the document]
fkleuver: The bug in aurelia-cli has been solved. That turned out to be caused by a strange accident with a release.
I still think we should keep the tslint warning off. *We* want to choose when to use async/await. 
I'm not very concerned about the perf hit of transpiling down to ES2015. That's only necessarily for IE which is.. typical.

packages/router/src/router.ts
jwx: This code is called from the `attached` method in the `viewport` custom element, so as long as `attached` traverses downwards (and it does, right?) there shouldn't be any problems.

packages/router/src/history-browser.ts
jwx: Yeah, definitely. But I'm thinking it can wait a bit.

packages/router/src/history-browser.ts
jwx: Comments and unit tests will definitely come. Will think about splitting into separate class.

packages/router/src/router.ts
jwx: Not sure what you mean...?

packages/router/src/router.ts
jwx: Let's discuss and focus on this a bit further down the line?

packages/router/src/router.ts
jwx: Short version: Navigation now (possibly) adds partial state (one or more viewports) to existing state (already populated viewports). This code fetches full state (all existing viewports) and rewrites the url so that refresh and browser button navigation will go to full state and not just last (possibly) partial state. This is also where "url simplification" will go, for example removing unnecessary `viewport-name:` from the url.

packages/router/src/router.ts
```diff
@@ -0,0 +1,335 @@
...
+  //         this.historyBrowser.cancel();
+  //       }
+  //     }).then(() => {
+  //       const viewports = Object.values(this.viewports).map((value) => value.description()).filter((value) => value && value.length);
+  //       this.historyBrowser.history.replaceState({}, null, '#/' + viewports.join('/'));
+  //     });
+  // }
+
+  public findRoute(entry: IHistoryEntry): IRoute {
+    return this.routes.find((value) => value.path === entry.path);
```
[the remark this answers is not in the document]
jwx: Yeah, just ignore this. It's part of the route setup that I'm not really developing in this PR. Sorry.

packages/router/src/scope.ts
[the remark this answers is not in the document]
jwx: Yeah, I should handle any returned promises. On it.

jwx:
The lack of tests, encapsulation and proper lifecycle management is something that'll come in future PRs. (I was advised to wrap it up. ;) )

Regarding your other concerns, at least recursion and pipeline hooks shouldn't be a problem at all. As for the others, we'll have to address it when we get there.

To feed my brain, could you give examples of "inter-dependent lifecycle states" and "complex conditional transitions"?

packages/router/src/router.ts
```diff
@@ -0,0 +1,335 @@
...
+    this.isActive = true;
+    this.options = {
+      ...{
+        callback: (navigationInstruction) => {
+          this.historyCallback(navigationInstruction);
+        }
+      }, ...options
+    };
+
+    this.historyBrowser.activate(this.options);
```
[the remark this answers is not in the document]
jwx: Yup.

packages/router/src/router.ts
fkleuver: It's not relevant how `attached` traverses, you simply cannot ever rely on DOM traversal to be a reliable way of finding components. You don't get any guarantees. Besides that, you also won't be able to traverse out of a shadow root so this will break when a custom element has `@shadowDOM`

packages/router/src/router.ts
fkleuver: Yep, just making note

packages/router/src/router.ts
```diff
@@ -0,0 +1,335 @@
...
+  public activate(options?: IRouterOptions): void {
+    if (this.isActive) {
+      throw new Error('Router has already been activated.');
+    }
+
+    this.isActive = true;
+    this.options = {
+      ...{
+        callback: (navigationInstruction) => {
+          this.historyCallback(navigationInstruction);
```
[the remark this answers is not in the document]
jwx: Yeah.

packages/router/src/router.ts
jwx: So it'll be like we discussed on Discord: someone, preferably with intimate knowledge of the vNext lifecycles, should eventually rewrite this specific part... :P

packages/router/src/router.ts
jwx: I wrote a reply to this, but it seems to be gone. But yes, this needs to be documented and perhaps also clarified a bit. There's no real reason except that I'm more used to promises and that I've seen someone posting issues about `async/await`. Could easily be changed, though.

packages/router/src/router.ts
jwx: (I did however think that shadow DOM supported this type of traversal through `element.getRootNode().host`...?)

jwx:
I've pushed a new version where the test app has, among other things, recursive viewports and scope. @EisenbergEffect @davismj @fkleuver @bigopon

davismj:
I don't think we want a LinkHandler. This is just off the top of my head but I remember that being a bad design choice. The LinkHandler intercepts links before they go to the browser URL bar, which triggers a completely separate logical flow than if the user typed a URL in the URL bar. I'm not sure why we needed the LinkHandler in the first place but I would investigate building without that.

fkleuver:
> The LinkHandler intercepts links before they go to the browser URL bar, which triggers a completely separate logical flow than if the user typed a URL in the URL bar.

As far as I understood it, it basically does "by hand" what the browser would do otherwise. I agree and I think we should try to have everything go exclusively via history.

> A lot of the logic seems adapted from vCurrent. 

This is the case for 90% of the vNext code base (most of which has been fully refactored by now).

> If we're not going to learn the lessons of vCurrent, we should just submit PRs to vCurrent. 

Even if we took the vCurrent router and ported it to vNext as-is, clean up a few things and add some tests, we would already be in a better situation.

Regardless - vCurrent lacks the foundation in terms of lifecycle management and synchronous binding (among other things) that allow certain design approaches to be feasible in the first place. 

> Most of the code, even the methods, aren't commented.

I'd shift the focus more towards tests. Commenting code at this early stage would just lead to tremendous overhead in keeping it up-to-date without providing any benefit whatsoever apart from momentarily making the code easier to review.
Tests provide real tangible benefits right away and accomplish similar goals.

I think in the abstract I share @davismj 's concerns, but for different reasons. We already discussed this in chat. There need to be at least a couple high level integration tests. They play the role of specs as well, they prove that the code works, and makes for more meaningful review discussion where we don't have to wonder if something actually *works*, but rather if it *does the right thing*.

davismj:
> I'd shift the focus more towards tests. Commenting code at this early stage would just lead to tremendous overhead in keeping it up-to-date without providing any benefit whatsoever apart from momentarily making the code easier to review.

Three reasons why this isn't the full story:

1. If you don't comment the code before you write it, you're probably not going to comment it after you're done. I've got 100k lines of vCurrent to show you if you don't believe me.
2. Comments force you to think through your code. It's one thing to write ```activate() { // activate all the things }```. It's another thing to say `/** activate() responsible for turning on all event handlers and other functionality that might require disposal before another router is activate`, at which point you realize (a) your code doesn't turn anything on that requires disposal and (b) there's only one router.
3. Commenting code, at least methods, in a library or framework in which you expect others to be involved respects maintainers' time. Uncommented code says you couldn't be bothered to document your thoughts and you expect others to figure it out on their own time.

fkleuver:
> If you don't comment the code before you write it, you're probably not going to comment it after you're done.

This is a nasty assumption you're making based on the behavior of the *vast* majority of developers, and hard to argue, as I can only say "wait and see".

> Comments force you to think through your code. 

I do wholeheartedly agree with this point. I have experienced this myself too. However, so does writing tests. I think testing and commenting are on par when it comes to this point.

> Commenting code, at least methods, in a library or framework in which you expect others to be involved respects maintainers' time. Uncommented code says you couldn't be bothered to document your thoughts and you expect others to figure it out on their own time.

I also fully agree. It needs to be commented at some point.

On the same token, however, comments very often become an escape hatch for covering poorly written code. Code should first and foremost attempt to document itself. Clearly named variables (that might be lengthy), well-organized branching, recognizeable patterns all contribute to this.


On the first point, I would *really* like to stress the following:

I am making a very deliberate choice in my work on vCurrent to delay the commenting as much as possible *so that I only have to do it once*. We have linting rules and automated jobs and checks and everything in place to throw errors as soon as we flip those switched.

And we *will* come at a point where we're all going into documentation mode, stop writing code, and comment the whole thing front-to-back until there are no more black boxes.

> I've got 100k lines of vCurrent to show you if you don't believe me.

I've got 50k tests (95% which would fail in vCurrent) to show you that tests accomplish more important things early on.

davismj:
> On the same token, however, comments very often become an escape hatch
for covering poorly written code.

@jods4 made this comment once, but its a [slippery slope logical fallacy](
https://en.wikipedia.org/wiki/Slippery_slope). Commented code doesn't lead
to bad code. In fact, it often leads to better code. The converse is a more
realistic scenario. Uncommented code doesn't force you to write good code.
In fact, it often leaves you with uncommented and bad code. I'd rather have
commented bad code.

> This is a ~~nasty~~ reasonable assumption you're making based on the
behavior of the *vast* majority of developers

> I've got 50k tests (95% which would fail in vCurrent) to show you that
tests accomplish more important things early on.

No, you have 50k tests that can show me how important tests are. They don't
prove, somehow, that comments are less important. Your tests cannot
demonstrate how easy or difficult the code is to maintain.

fkleuver:
> Commented code doesn't lead to bad code. 

>  I'd rather have commented bad code.

Fair. I'll take that point.

> They don't prove, somehow, that comments are less important.

I should note that I intentionally used the same kind of flawed logic here. I'm happy to take this point back if you take back that vCurrent is somehow proof that we wouldn't go back and comment things at a later point in time.

jwx:
No, we don't want a link handler and hopefully that can be removed by refactoring. This link handler does however not cause any problems with the navigation flow. https://github.com/aurelia/aurelia/issues/307#issuecomment-442262667

All the code is designed and written from scratch with the exception of the link handler which is based on the vCurrent one when it comes to capturing the events. I've used `activate`, `deactivate` and `isActive` long before I saw the vCurrent code. If they should prove to be unnecessary, they will be removed in clean up.

Regardless of past vCurrent experiences, code vs comments vs tests vs thinking and other discussions, before _my code_ goes into production it is cleaned up and at least reasonably clear, reasonably commented, has a reasonable amount of tests and is very thought through. While that still might not mean that it's good enough or that you like what it's doing, it will at least tick those boxes of professional standards.

davismj:
My main issue with comments is that it makes it difficult for me to jump in and review.

fkleuver:
Folks, I think we should get this merged in. While it does not follow @davismj 's specs, it does provide a working router with decent test coverage (decent, as in good numbers but low density).

Follow-up work could be divided in roughly 3 areas:
1. PR that further completes the vision @jwx had in mind
2. PR that implements @davismj 's specs
3. PR that improves testing and/or overall code quality

I think everyone's efforts would be spent most efficiently and effectively if I focused on point 2, and @jwx on point 1. I would also put effort in point 3 (as would @jwx )

As a third person looking at this from a distance I see two compatible visions and I would personally like to help out in making that work. For that, this PR needs to be merged. I think I can offer @davismj (and others) some piece of mind by working in parallel on implementing the specs while @jwx can peacefully continue working on his idea, and then eventually we'll meet up somewhere in the middle. That's when conflicts and discussions might heat up, but they will likely also be more fruitful at that point.

I certainly wouldn't care for this useful work to forever sit in limbo in the midst of endless theorizing over what might and might not work in the long run.

Let's work iteratively on this.

@jwx @EisenbergEffect @davismj can you guys get behind that?

EisenbergEffect:
I'm on board with that. I'd like to see this merged in so we can have a definitive marker to iterate forward from. If it's anything like binding/templating, we'll take this through quite a few iterations before we land it where we want it.

I can review and merge if we're ready for that.

PR #320, opened 2018-11-27
[Do not merge 🚫]: basic ssr example

bigopon:
Runnable app inside `packages/examples/test-ssr`. Only basic, just return what ever the request url is. server runs at 8081


bigopon:

PR #321, opened 2018-11-27
update package.json and package-lock.json

fkleuver:
# Pull Request

## 📖 Description

To address https://github.com/dominictarr/event-stream/issues/116 (and because it's just that time again), hereby updating all deps and re-generated a fresh lockfile so that it's all clean and up-to-date again :)

## 👩‍💻 Reviewer Notes

Plain and simply updated all devDependencies in the root-level package.json as well as the individual package jsons. Deleted the lockfile and did a fresh install.


package.json
```diff
@@ -69,30 +69,30 @@
     "karma-source-map-support": "^1.3.0",
     "karma-sourcemap-loader": "^0.3.7",
     "karma-webpack": "^3.0.5",
-    "lerna": "^3.4.3",
+    "lerna": "^3.5.0",
```
BBosman: Shouldn't `lerna.json` be updated to match this version? Although it appears it hasn't been updated for some time.

fkleuver: Great catch! I updated it.

PR #324, opened 2018-11-29
refactor: standardise on "as" type casts

BBosman:
# Pull Request

## 📖 Description

Typescript has two ways of type assertion:
1. `variable as type`
2. `<type>variable`

In the codebase both ways are used interchangeably. For consistency it would be better to pick one and stick to it. After discussing it with @fkleuver we decided to go for the `as` style type assertion, because:
- When using TypeScript with JSX, only as-style assertions are allowed. ([Reference](https://www.typescriptlang.org/docs/handbook/basic-types.html))
- It reduces confusion when used in conjunction with Generics.

So this PR turns on the related linting rule and changes all type assertions to use the `as` style.

### 🎫 Issues

Related to #249.

## 👩‍💻 Reviewer Notes

Migration was done using the `--fix` option of `tslint`.

## 📑 Test Plan

CircleCI

## ⏭ Next Steps

See #249.

### DIFF
--- a/packages/debug/src/reporter.ts
+++ b/packages/debug/src/reporter.ts
@@ -35,7 +35,7 @@ export const Reporter: typeof RuntimeReporter = {...RuntimeReporter,
   error(code: number, ...params: unknown[]): Error {
     const info = getMessageInfoForCode(code);
     const error = new Error(info.message);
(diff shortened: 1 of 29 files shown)


EisenbergEffect:
Excellent!

PR #327, opened 2018-11-30
[Do not merge 🚫]: basic libui integration

bigopon:
Basic integration with [libui](https://github.com/andlabs/libui), via [libui-node](https://github.com/parro-it/libui-node)


bigopon:
It's still difficult to implement a full integration (template controller : if, repeat etc) as there is no easy way to traverse the DOM in `libui-node` yet. Issue here https://github.com/parro-it/libui-napi/issues/52

bigopon:

PR #330, opened 2018-12-01
[Do not merge 🚫]: rendering integration, test blessed integration

bigopon:
In this integration I tried doing more during mapping of template nodes to native nodes. It sounds good and easy for plain attributes. There is no different between this and other integrations beside few main steps.

Runnable app at https://github.com/bigopon/aurelia-vnext-blessed


bigopon:

PR #336, opened 2018-12-08
refactor(runtime): strictNullChecks fixes

BBosman:
# Pull Request

## 📖 Description

Refactoring `runtime` to adhere more to `strictNullChecks`.

### 🎫 Issues

Part of #254.

## 👩‍💻 Reviewer Notes

This is just a partial fix, but I decided to get some of the low hanging fruit out of the way first, before going for the bigger changes.

## 📑 Test Plan

Trust in CircleCI.

## ⏭ Next Steps

Do the rest and turn the options on by default in the `tsconfig` files.

### DIFF
--- a/packages/runtime/src/binding/ast.ts
+++ b/packages/runtime/src/binding/ast.ts
@@ -1127,7 +1127,7 @@ export class BindingIdentifier implements IExpression {
     this.name = name;
   }
 
-  public evaluate(flags: LifecycleFlags, scope: IScope, locator: IServiceLocator): string {
+  public evaluate(flags: LifecycleFlags, scope: IScope, locator: IServiceLocator | null): string {
     return this.name;
   }
   public connect(flags: LifecycleFlags, scope: IScope, binding: IConnectableBinding): void {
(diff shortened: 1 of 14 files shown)


packages/runtime/src/binding/ast.ts
```diff
@@ -1197,8 +1197,8 @@ export class Interpolation implements IExpression {
     this.assign = PLATFORM.noop as () => unknown;
     this.parts = parts;
     this.expressions = expressions === undefined ? PLATFORM.emptyArray : expressions;
-    this.isMulti = expressions.length > 1;
-    this.firstExpression = expressions[0];
+    this.isMulti = this.expressions.length > 1;
+    this.firstExpression = this.expressions[0];
```
bigopon: Imo, it should be
```js
expressions = expressions || PLATFORM.emptyArray;
```

fkleuver: The API contract specifies "array or undefined", why use (slower) coercion instead of strict equality?

bigopon: We still need some Minimizer friendliness. So it can be still strict, just with expression

fkleuver: I think we can tackle minimizer friendliness with AOT. There are a lot of parts in the framework that still have a long way to go in this area.

PR #338, opened 2018-12-10
refactor(kernel): DI typing

BBosman:
# Pull Request

## 📖 Description

Changes all `any` types in `di.ts` to a more concrete type.

### 🎫 Issues

Part of #249.

## 👩‍💻 Reviewer Notes

Part reverse engineering, part guessing 😊, so I assume it will need some tweaking.

## 📑 Test Plan

CircleCI.

## ⏭ Next Steps

See #249.

### DIFF
--- a/packages/kernel/src/di.ts
+++ b/packages/kernel/src/di.ts
@@ -6,11 +6,11 @@ import { IResourceType } from './resource';
 
 const slice = Array.prototype.slice;
 
-export type ResolveCallback<T = any> = (handler?: IContainer, requestor?: IContainer, resolver?: IResolver) => T;
+export type ResolveCallback<T = unknown> = (handler?: IContainer, requestor?: IContainer, resolver?: IResolver) => T;
 
 export type Key<T> = InterfaceSymbol<T> | Primitive | IIndexable | Function;
 
-export type InterfaceSymbol<T> = (target: Injectable<T>, property: string, index: number) => any;
+export type InterfaceSymbol<T> = (target: Injectable<T>, property: string, index: number) => unknown;
 
 export interface IDefaultableInterfaceSymbol<T> extends InterfaceSymbol<T> {
   withDefault(configure: (builder: IResolverBuilder<T>) => IResolver): InterfaceSymbol<T>;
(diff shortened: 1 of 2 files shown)


packages/kernel/src/di.ts
```diff
@@ -21,20 +21,20 @@ interface IResolverLike<TValue, TContainer> {
   getFactory?(container: TContainer): IFactory<TValue> | null;
 }
 
-export interface IResolver<T = any> extends IResolverLike<T, IContainer> { }
+export interface IResolver<T = InstanceType<Constructable>> extends IResolverLike<T, IContainer> { }
```
fkleuver: `InstanceType<Constructable>` is essentially a huge detour to `{}`. We might as well make it `{}` then :)

BBosman: If I do that I get a lot of `Type 'IResolver<{}>' is not assignable to type 'IResolver<T>'.
  Type '{}' is not assignable to type 'T'.` errors. So it seems to like the detour. 😊

fkleuver: Right, it resolves to `any` because that's the return type of the prototype of `Constructable`. 

image

I thought it was `{}` but I must have changed that myself a couple weeks ago then (likely for this very reason).

The main problem with the `Constructable` interface (or really the whole class-to-instance resolution in TypeScript) is that it relies on the return type of the prototype. Constraining that return type further than `any` almost always manages to break something.

What I'm thinking as a possible solution here is to give every DI-related interface 2 generic type arguments instead of 1. `TClass` and `TProto`, allowing you to specify (and infer) them directly instead of having to hack around the prototype property. Then we can leave the prototype to `any` and still have some strong typing mechanism that doesn't rely on `InstanceType`.

I've done this around the resources and although things got a bit too complex there to my liking, the concept does work. Just food for thought

BBosman: I reverted those `any` to `InstanceType<Constructable>` changes back to `any` as they don't add value.

I tried different solutions, but I have to admit that fixing those `any`'s is turning out to be above my proficiency level.

BBosman:
Superseded by #366.

PR #339, opened 2018-12-11
chore(examples): update gulp semver range to reflect 4.0.0 release

fkleuver:
# Pull Request

## 📖 Description

Gulp finally published `4.0.0` to their `latest` channel, so the `next` semver tag stopped working and broke the `aurelia-cli-ts` example. In addition, the `latest` now points to `4.0.0` which breaks the browserify example. This PR fixes both issues.

### 🎫 Issues

Build broke

## 👩‍💻 Reviewer Notes

N/A

## 📑 Test Plan

CircleCI tests should pass

## ⏭ Next Steps

N/A

### DIFF
--- a/examples/jit-aurelia-cli-ts/package.json
+++ b/examples/jit-aurelia-cli-ts/package.json
@@ -26,7 +26,7 @@
     "gulp-sourcemaps": "latest",
     "gulp-typescript": "latest",
     "gulp-watch": "latest",
-    "gulp": "next",
+    "gulp": "latest",
     "http-server": "latest",
     "minimatch": "latest",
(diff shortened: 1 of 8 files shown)


huochunpeng:
Should just use `^4.0.0` and `^3.9.1` in all next/latest cases.

fkleuver:
That could be said for all dependencies. I'm intentionally picking `latest` wherever possible to make for more realistic tests (many end users will do the same after all..). So we'll have a build breaking because of this now and then, and that's fine. It's just the CI build. Better than someone else's prod build right :)

huochunpeng:
I see.

PR #343, opened 2018-12-21
fix(template-binder): correctly traverse not-compilable node types

fkleuver:
# Pull Request

## 📖 Description

@jwx discovered this the hard way in his router test app, but apparently I broke the compiler when there are comments. Big oops. I added traversal for all node types back in and all should work again as it did before.

### 🎫 Issues

N/A

## 👩‍💻 Reviewer Notes

Added traversal for other node types and reused `bindChildren` in `template-binder`

## 📑 Test Plan

Added a few xml-based tests to verify all node type traversals don't cause the compiler to hang or crash. This is just a bit quick-n-dirty but it does the job for now

## ⏭ Next Steps

Add more complete tests for the different document types (html, xml, svg) that can be parsed by the browser as well as rendering for these.

### DIFF
--- a/packages/jit/src/template-binder.ts
+++ b/packages/jit/src/template-binder.ts
@@ -416,19 +416,7 @@ export class TemplateBinder {
       ++i;
     }
 
-    let childNode: INode = node.content.firstChild;
-    let nextChild: INode;
-    while (childNode !== null) {
-      switch (childNode.nodeType) {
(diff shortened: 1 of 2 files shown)


fkleuver:
Interesting. It reports the total coverage going down because the amount of code is reduced, so the relative coverage is less :p

PR #344, opened 2018-12-23
Reorganization of types

fkleuver:
# Pull Request

## 📖 Description

This PR contains some of the changes I made while working on the integrations. I've split it up into a separate PR to reduce the scope of the integrations PR a bit (otherwise just about every single file would be touched - again).

There are no functional changes, it's purely moving files around and rearranging the imports a bit. Things are a bit more logically grouped together imo. I hope it makes sense to others as well.

### 🎫 Issues

N/A

## 👩‍💻 Reviewer Notes

Summary:
- resource.ts moved to kernel because it had no dependencies on the runtime and doesn't really need to be coupled to it. `RuntimeCompilationResources` is more of a DI concern as well.
- All runtime resources moved under `resources` folder and organized similarly to how they would be organized in a regular Aurelia app.
- All observers (or rather, the stuff that directly manage data flow between viewmodel and view) are moved under `observers` folder
- Shortened binding behavior file names for consistence

## 📑 Test Plan

N/A - no functional changes

## ⏭ Next Steps

There are a few more things in this category that I'll be looking at next, such as some renames and cleaning up prototype assignments, but to limit the scope of a single PR I kept this one strictly to moving files around.


EisenbergEffect:
The only thing I'm not 100% sure about is the move of the resources to Kernel. However, it's mostly just interfaces and doesn't have any DOM or even UI implications, so it's probably ok. There are some people who will want to use the Kernel but not the rest of Aurelia, so I just want to make sure that we don't push a lot of UI framework type things in there.

Ready to merge?

fkleuver:
Yeah I agree it's borderline. I think however that the conceptual aspect of it, which is really just getting the type associated with a (string) key, doesn't have to be aurelia-specific. Which is why it could potentially have a different name as to make that clearer.

It's ready to merge.

Issue #87, opened 2018-08-20
Integration


bigopon:
The rendering targets that should be easily integrated with vNext, which I hope it will help shape the 
abstraction layer APIs:

* libui for native UI on Windows, Linux & Mac: https://github.com/parro-it/libui-node
* nativescript for native Androi, iOS: https://github.com/NativeScript/NativeScript
* pixi, three, phaser for graphical app

... to be updated

@EisenbergEffect @fkleuver

fkleuver:
You think AOT could play a role here? Replace HTML-specific api calls with the appropriate compatible other UI api calls? Or maybe have separate plugin packages instead and try to keep templating free of HTML as much as possible?

EisenbergEffect:
I think AOT can play a role, but it might be easier if it can be based on a known abstraction pattern. For example `DOM.appendChild(parent, child)` can be implemented to actually work at runtime without any AOT, but it's also possible to have an AOT process convert this into `parent.appendChild(child)` with 100% accuracy. It could also be converted to any other API such as `parent.add(child)` or `child.appendTo(parent)` or whatever the target API supports. But having the `DOM` APIs as known, distinguishable operations, enables us to get it working today without any AOT and later, reliably build an AOT mechanism to produce slightly more optimized code.

EisenbergEffect:
Things have changed a bit, but do you all think we can close this now that we've factored things out to separate rendering?

bigopon:
Yes. Its good for now

PR #348, opened 2018-12-25
runtime-html / jit-html refactor

fkleuver:
# Pull Request

## 📖 Description

- Moves all HTML-specific processing and references out of the runtime and jit and into runtime-html / jit-html.
- These html packages talk directly to `lib.dom.d.ts` types. The previous stub dom types are now gone from the runtime. The `INode` interface exposes no DOM properties or methods whatsoever, it's just an empty interface inheriting only from `Object`. The DOM abstraction takes care of operations.
- ObserverLocator takes an internal runtime-specific locator that it delegates target observer resolution to. Theoretically those inner locators could be wrapped which supports mixing, but perhaps this needs a cleaner way. Not sure yet.
- Various moving parts have an extra locator/factory layer in between to facilitate the separation of runtimes:
   - The projectors are resolved by `ProjectorLocator`
   - `CompiledTemplate` sits behind a `TemplateFactory` abstraction with runtime-specific implementations that give it the correct `NodeSequenceFactory`
- `Listener`, `SetAttribute` and `TextBinding` are moved to html runtime. The base runtime now essentially only deals with objects and things that look/behave like nothing more than objects.
- `encapsulationSource` is removed from the `$attach` lifecycle and replaced with `Node.getRootNode()` to get the shadow root
- `DOM.createNodeObserver` is removed and instead the projectors and `SelectValueObserver` (which are now in the html package) instantiate MutationObserver directly
- The `registerElementConfiguration` api as well as the configurations that it contained are removed from `EventManager` in favor of slightly more efficient logic in the html-specific observer-locator (which is the new extension point essentially). Now it's not as easy to add new events as it used to be, and if there is a real need we could add this facility back in, but it does not make a whole lot of sense to me with the current direction.
- Moved global `setTimeout`, `clearTimeout` etc to `PLATFORM` so they don't break in non-browser environments
- Changed a lot of types in bindings and observers from `INode` / `IIndexable` to `Object`
- Renamed `Observer` to `SelfObserver` and moved every observer to its own file

I probably forgot to mention several things. There has been a lot of shuffling things around :)
Will add more info later.

The generated JIT tests pass, so the concept works. As before, there are now a whole bunch of other tests to fix.

I'm quite content with how the runtime turned out but I don't like the jit yet. This is very difficult to split up without resorting to an observerlocator-like extension point, which can get ugly for the template compiler. I'll probably leave it like this for now, though, and as a next step first implement an additional runtime to see what kind of extension point (if any) it really needs.
I suppose a bit of duplicate JIT code for alternative runtimes isn't the end of the world. Don't want them to be overly dependent on one complex generic abstraction either..

### Important note
The `runtime-html` package is intended to be a dependency for most alternative runtimes that run in regular browser DOM (WebGL renderers like pixi, etc). So it's not a completely standalone package. This is why some stuff was moved there that you might not necessarily consider very html-specific, such as `compose`, `set-attribute`, etc. It should work even in JSDom.
I would guess that things like NS and UWP will not depend on `runtime-html` since they aren't even parsed the same way, and need some things done differently.

### 🎫 Issues

Related to https://github.com/aurelia/aurelia/issues/87

## 👩‍💻 Reviewer Notes

Just about everything changed or moved, so it's probably easier to get a picture of the new structure by pulling the branch locally and checking it out.

- Each of the 4 packages related to this PR (runtime, runtime-html, jit, jit-html) have a `configuration.ts` exposing the DI configuration. Those configurations are probably a good place to start looking, and just follow the references from there.

- Also made kernel compatible with nodejs runtime and added pure nodejs tests (mocha). A few tweaks to platform.ts were needed due to browser-specific assumptions

## ⏭ Next Steps

Implement an alternative runtime. I'm considering ThreeJS and NativeScript.


fkleuver:
@EisenbergEffect @bigopon I mentioned earlier that the `@aurelia/runtime` being runtime-agnostic would open the door to some interesting possibilities, such as using an arbitrary custom DOM implementation.

To show what I meant by that @bigopon, I implemented a very minimalistic js dom. This is primarily for testing purposes and to showcase that the runtime can be used in any js runtime (even server side) without doing anything special: https://github.com/aurelia/aurelia/pull/348/commits/5f45e30c10819e0723f421eb6e75ec12b83d4bd0#diff-f509d7fe6ad6559a0ad2277da72ca5efR55

All `@aurelia/runtime` tests except for the resource tests actually successfully run in plain NodeJS with mocha now. The only reason the resource tests don't work is because they still use the HTML DOM. I'll probably move those tests to `runtime-html` and add a few in the regular runtime with the `AuDOM` implementation.

It's relatively straightforward. There is a ~200 LOC (not counting some helper methods for syntactic sugar) `INode` implementation which is all that's needed to support most of the non-html runtime features (text bindings, if/else/repeat/with/replaceable, etc).

The rest of that file is mostly infrastructure boilerplate, implementations for:
- `IDOM` (smaller than the HTML impl)
- `IProjectorLocator` (just returns the one projector)
- `IElementProjector` (basically a copy of `HostProjector`)
- `INodeSequence` (very similar to the HTML `NodeSequence`)
- `INodeSequenceFactory` (smaller than the HTML impl)
- `IDOMInitializer` (just returns a new `AuDOM` instance as there is no need for a global `document`)
- `ITemplateFactory` (just a pass-through like the projector locator)
- `ITargetAccessorLocator` (noop since no special observers are needed)
- `ITargetObserverLocator` (noop since no special observers are needed)
- one `ITargetedInstruction` with an accompanying `IInstructionRenderer` for text bindings

Some of these implementations are practically empty. Not a lot is needed to make a custom runtime work, and there is lots of room for building on top of / customizing things.

I made `AuNode` similar to `Node` with previous/next sibling, childNodes etc, but that's just out of familiarity. In fact nothing in the runtime depends on any particular tree structure or properties.

So from a technical point of view I think it's pretty close to where it needs to be. But perhaps in terms of API contracts and coherence some more polishing work is needed. @EisenbergEffect generally has a better feel for that than I do, so any suggestions are welcome.

EisenbergEffect:
This is great to hear! Thanks for going the extra mile and proving that out @fkleuver !

fkleuver:
@EisenbergEffect @bigopon it's ready for final review. The only tests that are not fixed yet (they're commented out) are the `repeat` and `if` runtime integration tests. They are however backed up by jit integration tests (which cover the static scenarios and basic mutations) and a significant number of additional view integration tests.
I'd rather take my time to properly fix them and finish the test cleanup in a next PR, rather than rush it now. I've got a couple busy days upcoming. This PR also contains that shadow root fix as well as updated integration tests to verify them properly, but does not contain the local dependencies fix.
If you want to get it merged in I think that should be fine. If you prefer to let it wait until all tests are fully fixed, that's fine too. Let me know.

bigopon:
I really like the new look of whole thing, it came a lot closer to what was discussed about a separation of (a) renderer (done), (b) binding & observer infrastructure, with a bonus of similar structure in JIT. I can't say for sure but gut feeling is this is gonna be great for integrations. I'll re-implement integrations and validate this against them.

Overall: very nice work ❤️

PR #354, opened 2018-12-31
[WIP] Router parameters

jwx:
# Pull Request

## 📖 Description

WIP of parameters for the convention-based router. Test application can be found in packages/router/test/e2e/layouts. (Select Lobby followed by Contacts.)


packages/router/src/router.ts
```diff
@@ -49,10 +51,16 @@ export class Router {
   public historyBrowser: HistoryBrowser;
   public linkHandler: LinkHandler;
 
+  public navs: Object = {};
```
fkleuver: I know it's WIP but I'm seeing lots of `Object` all over the place. Let's refrain from using that. There's a dedicated type precisely for this scenario: `Record`. So this would be `Record<string, Nav>`. I think the earlier to get the typings right, the better. It will also make things easier for you in the end..

packages/router/src/resources/nav.ts
```diff
@@ -0,0 +1,36 @@
...
+  </nav>
+  <ul if.bind="routes" class="nav-level-\${level}">
+    <li repeat.for="route of routes" class="\${route.active} \${route.hasChildren}">
+      <a if.bind="route.link && route.link.length" href="\${route.link}">\${route.title}</a>
+      <a if.bind="!route.link || !route.link.length" click.delegate="route.toggleActive()" href="">\${route.title}</a>
+      <au-nav if.bind="route.children" routes.bind="route.children" level.bind="level + 1" containerless></au-nav>
+    </li>
+  </ul>
+</template>` })
+export class NavCustomElement {
```
fkleuver: I think I actually like the idea of having some built-in navigation custom element. There is sufficient low-level API available to make this much more flexible and configurable if need be. In any case having a working default (even if it doesn't suit everyone's needs) would probably be appreciated by community. 👍

packages/router/src/history-browser.ts
```diff
@@ -144,7 +147,14 @@ export class HistoryBrowser {
     });
   }
 
-  public replacePath(path: string, fullStatePath: string): void {
+  public replacePath(path: string, fullStatePath: string, entry: INavigationInstruction): void {
+    if (entry.index !== this.currentEntry.index) {
+      // TODO: Store unresolved in localStorage to set if we should ever navigate back to it
```
fkleuver: I see the benefit of using localStorage but if you do work on that, please make sure to allow users to disable it as well. LocalStorage lives forever, and there are various situations where a user really wouldn't want that.

packages/router/src/nav-route.ts
```diff
@@ -0,0 +1,99 @@
...
+    } else {
+      return this.linkName(this.components);
+    }
+  }
+
+  private handleChange(): void {
+    if (this.link && this.link.length) {
+      this.active = this._active();
+    } else {
+      this.active = (this.active === 'nav-active' ? 'nav-active' : (this.activeChild() ? 'nav-active-child' : ''));
```
fkleuver: Is this a magic string for state or is this a css class name?

packages/router/src/nav.ts
```diff
@@ -0,0 +1,40 @@
+import { CustomElementResource, ICustomElementType } from '@aurelia/runtime';
+import { NavRoute } from './nav-route';
+import { Router } from './router';
+
+export interface INavRoute {
+  components: string | ICustomElementType | Object;
```
fkleuver: Again, I know it's WIP so maybe you've already got plans, but I just want to stress I *really dislike* `Object` for this kind of thing. That's so very vCurrent, we got to stop doing that :p

packages/router/src/parser.ts
```diff
@@ -0,0 +1,61 @@
...
+  merged: string[] | Object;
+}
+
+export function parseQuery(query: string): IParsedQuery {
+  const parameters = {};
+  const list = [];
+  if (!query || !query.length) {
+    return { parameters: parameters, list: list };
+  }
+  const params = query.replace('+', ' ').split('&');
```
fkleuver: A tiny 5-word or so comment on what the various delimiters represent would be really great here. I know (or at least I think I do) that `+` separates view port targets but others and future readers won't. Perhaps put them in a `enum` with descriptive names since you're using this in various places?

packages/router/src/nav-route.ts
jwx: Just class(es) added for the developer to style.

packages/router/src/history-browser.ts
jwx: Yeah, but certain scenarios (for example where the browser is stressed with multiple navigations through browser buttons) might suffer without it. It's rare, so it should definitely be possible to turn off/clear out.

packages/router/src/parser.ts
jwx: I've got the delimiters as variables so that they can changed by developers who want a different format.

packages/router/src/router.ts
jwx: I tend to throw `Object` in when I'm not sure of whether the code stays (that way) or not. But you're right, I need to do better (at least before PRing).

packages/router/src/resources/nav.ts
jwx: Yeah, I think so to. And I think it'd be possible to find the right balance between flexibility, useability and cost.

packages/router/src/parser.ts
fkleuver: But right now they are hard coded magic strings..

jwx: Oh, sorry, I just read your comment and didn't actually look at your code (in a hurry). This isn't part of my own syntax, but standard parsing of the querystring parameters. Come to think of it, the replace might be unnecessary. Just threw it in out of habit, I think. Will review tomorrow.

PR #312, opened 2018-11-23
[Do not merge] Tracking pixi integration

bigopon:
PR for visualizing what changes made to runtime, for PIXI renderer.

PR #361, opened 2019-01-03
Pixi

fkleuver:
# Pull Request

## 📖 Description

Goal of this PR is to have something visual/moving (a little more interesting than a +/- button) with as little as possible infrastructure boilerplate. And to get the first non-html runtime package in the running.

It's is a very rough initial pixi implementation that only supports single level nested sprites, but it does so by using only custom elements. I'd like to try and incrementally involve the renderer or even compiler extension points in working towards full support for the various APIs, nesting, and more efficient ways to watch/observe/synchronize, but only when necessary. I prefer to save that kind of infrastructure work for integrations where it's really needed.

I was initially trying to use slots but that didn't work just yet, that's something I need to fix in the compiler. It uses `replaceable` at the moment.

### 🎫 Issues

Follow-up to https://github.com/aurelia/aurelia/pull/312

## 👩‍💻 Reviewer Notes

Fixed a couple issues I spotted along the way:
- Small fix in `Binding` to force `fromView` into the right direction when the change doesn't come from the DOM.
- It seems I overlooked `ref` binding on custom elements, fixed it, need more tests for that as well.

Can run the test app by cloning the repo, followed by `npm ci && npm run build && cd examples/jit-pixi-webpack-ts && npm install && npm run watch`

## 📑 Test Plan

N/A, still early WIP. I'm thinking about adding mostly JIT integration tests at a later point and do assertions on the pixi objects to see that everything is mapped correctly. The visuals are tricky to test.

## ⏭ Next Steps

- Support full nesting
- Reuse `DisplayObject` properties/bindings
- Find cleaner way to map bindings
- Write tests


fkleuver:
@bigopon I put it in the review notes just here:

> Can run the test app by cloning the repo, followed by npm ci && npm run build && cd examples/jit-pixi-webpack-ts && npm install && npm run watch

bigopon:
@fkleuver I'd prefer to have a link to online playable 😛

fkleuver:
@bigopon  https://dist-kwlismwcbb.now.sh/

PR #363, opened 2019-01-04
tslint warnings as errors + strict tsconfig option

fkleuver:
# Pull Request

## 📖 Description

And so it begins.. :-)

This might take a while.

@BBosman feel free to make commits to this branch as well if you want to. I'll make sure to push whenever I'm done and pull before I start.

PR #366, opened 2019-01-08
refactor: enable ban-types linting rule and fix violations

BBosman:
# Pull Request
## 📖 Description

Brings in the `ban-types` changes from #363 and fixes all related violations.  I also included the (remaining) changes from #338, as they were partially overlapping and would just introduce merge conflicts.

~~As requested on Discord by @fkleuver I set the `defaultSeverity` to `error`~~. I explicitly marked `completed-docs`, `cognitive-complexity` and `no-any` as `warning`. That last one shouldn't be in there, but the remaining `any`'s are above my proficiency level, so I'll leave those for somebody else to fix 😊,  after which the exception can be undone.  

### 🎫 Issues

Related to #249.
Follow up to the aborted #363 and #338.

## 👩‍💻 Reviewer Notes

A lot of typing changes all over the map as a result. Mostly reverse engineering A few could maybe be tightened up a bit more based on implementation knowledge that I don't (yet) have, so feel free to give suggestions. 😄 

## 📑 Test Plan

Trust CI.

## ⏭ Next Steps

See #249. And moving more parts from #363 over.

Also we need to figure out why CircleCI has "invalid" linting issues so we can switch `defaultSeverity` *back* to `error`.


fkleuver:
Excellent! Also a merge conflict hot-spot upcoming.. :)

fkleuver:
@EisenbergEffect You're the mastermind behind DI, could you double check if @BBosman got the typings correct for `dependencies` / `staticDependencies` / `dynamicDependencies` in the `Factory` and `Invoker`? I would say they need to be of type "key", right? (which isn't really a type.. since that's essentially `string | Constructable | InterfaceSymbol | Resolver` etc etc.. but I'm not sure

EisenbergEffect:
@BBosman Same as the other PR. Router PR made things go 💥 Do you mind fixing this up?

EisenbergEffect:
Ok, other PR is now merged.

fkleuver:
I think type of `key` and type of `deps` should just be `unknown`. It can be `Constructable`, `InterfaceSymbol` or `Resolver` which are useful types because they are generic, but it can also be `string` (in fact anything that's not falsey-ish or `Object` is a valid key). So a type can never reliably be inferred from it. 

I "do it anyway" in the `IServiceLocator`  interface's `get` method because it's just too useful for end users to not do it. But they still need to explicitly cast string-key resources. Inside DI there's therefore no real value to it imo.

The only reason I explicitly specified `inject` as an array of `InterfaceSymbol` at various components in the other PR is because they are specified explicitly by us, so we know that's what they are. We can't know this inside DI.

Same goes for the `register` contract. As @EisenbergEffect mentioned the other day, the existing constraints are just a pain as they require you to always export resources with `static register` or a merged declaration of `IResourceType`. I think good API documentation + good errors is the best thing we can do there.

BBosman:
I rebased this on master.

- Rebasing triggered some new "generic" linting violations, which I fixed along the way.
- I included the (remaining) changes from #338, they were partially overlapping and would just introduce more merge conflicts.
- As requested on Discord by @fkleuver I set the `defaultSeverity` to `error`. I explicitly marked `completed-docs`, `cognitive-complexity` and `no-any` as `warning`. That last one shouldn't be in there, but the remaining `any`'s are above my proficiency level, so I'll leave those for somebody else to fix 😊,  after which the exception can be undone.  
- Not all feedback on this PR and #338 has been incorporated, as there is still some discussion on what types should be used and where. Getting this PR in a reviewable state (which it now is) should make it possible for @EisenbergEffect to give some insight based on his DI knowledge.

No clue why linting fails on CircleCI. Those errors don't show up locally and making the suggested changes actually breaks it locally.

EisenbergEffect:
@BBosman Did I mess this up again by merging the other lint PR?

BBosman:
@EisenbergEffect No you didn't, I did. By avoiding `ts` files touched by this PR in the other I thought we wouldn't have conflicts, but I forgot about touching the `package.json` files in both. 😊 

I just rebased and it's ready to go.

packages/router/src/history-browser.ts
```diff
@@ -47,7 +47,7 @@ export class HistoryBrowser {
   private isReplacing: boolean;
   private isRefreshing: boolean;
 
-  private ignorePathChange: Function;
+  private ignorePathChange: ((value?: {} | PromiseLike<{}>) => void) | null;
```
jwx: This should've been `ignorePathChange?`.

packages/router/src/viewport.ts
```diff
@@ -46,7 +46,7 @@ export class Viewport {
   private readonly router: Router;
 
   private clear: boolean;
-  private elementResolve: Function;
+  private elementResolve: ((value?: void | PromiseLike<void>) => void) | null;
```
jwx: This should've been `elementResolve?`.

jwx:
This is really helpful, @BBosman. Thanks for doing this.

BBosman:
Now that the tests are also linted, `ban-types` also triggers there. Took the liberty to add a last minute commit to fix those.

@EisenbergEffect Ready to go imho.

fkleuver:
@EisenbergEffect Let's get this in

fkleuver:
@BBosman Did a quick last pass but looks good. A nice follow up might be to give `InterfaceSymbol` the `unknown` type argument as a default, so the whole thing doesn't need to be written out in full everywhere.

PR #373, opened 2019-01-12
chore(benchmarks): add some initial microbenchmark scripts

fkleuver:
# Pull Request

## 📖 Description

While looking at binding / lifecycles, I was experimenting with using [function scope for the lifecycle](https://github.com/aurelia/aurelia/commit/a81144ae296b9fd890b74e9f737406338f2d0741#diff-5bffb21f0a04d038c5027956f70bab9bR671) instead of a class, thinking the elimination of `this.` accessor might improve performance (in addition to making it smaller when minified).
Then I added some microbenchmarks to verify that, and it turns out it actually made everything twice as slow (*after* removing the tracer calls). So I reverted that change, but I'm keeping the benchmark :)

The idea is that it imports both the local build and the one currently on npm `@dev` and runs the benchmark in both, to compare the local changes to the previous nightly.

Of course there are the tracer calls to account for (they're removed from the bundle) so they need to be removed locally with some `if \(Tracer\.enabled\).*\n` replace-all regex before running a benchmark.

The results of the ones currently in place:

`container.register(BasicConfiguration)` (that's a couple dozen registrations fyi)
image

`container.get(SomeDep)` x 10 (with 10 pre-registered deps), so multiply ops/sec by 10 to get the "real" ops/sec
image

`lifecycle.beginBind()` + `lifecycle.enqueueBound(component)` (where `component.bound()` is a noop + `lifecycle.endBind()` x 10
image
(tracer calls are still present locally here, so that's the 20% hit - with function scope, this went down to 2mln ops/sec)

### 🎫 Issues

N/A

## 👩‍💻 Reviewer Notes

See scripts/benchmarks

## 📑 Test Plan

N/A

## ⏭ Next Steps

Gradually add some microbenchmarks as I come across components where I feel it makes sense.


EisenbergEffect:
Love, love, love this. So happy to see this. @fkleuver is rocking vNext :)

PR #376, opened 2019-01-12
feat(runtime): fix+test the computed observer

fkleuver:
# Pull Request

## 📖 Description

This PR fixes (and verifies) computed observation

### 🎫 Issues

N/A (was an outstanding TODO)

## 👩‍💻 Reviewer Notes

- Removed `GetterController` and merged it into `GetterObserver`, reducing the overall call stack depth and complexity of the implementation.
- Removed `DirtyChecker` fallback when `Proxy` is not available, since this falls under IE11 compatibility which I think shouldn't be clogging the code too much right now. Better tackle that framework-wide as a whole later on, taking a good look at the best polyfill options etc.
- Separated batched and synchronous dependencies, making computed observation in principle synchronous like the rest (DOM and array changes are batched anyway) and removing the need for queueing the computed observer itself in the lifecycle.
- Various small tweaks and fixes regarding problematic prototype calls on natives. See [this huge getter](https://github.com/aurelia/aurelia/compare/binding-2#diff-d864b1975b5c08f204f8abd7be245b2cR289) for an idea of the sort of stuff that works.

## 📑 Test Plan

- About 265 generated tests verify all the various combinations of inputs and assert them *reasonably* well (this can and should improve before considering computed observer production ready since there are tons of additional potential edge cases)
- One monstrous test verifies that a big nested tree with various collection, merge, spread, stringify operations with conditional dependencies also works

## ⏭ Next Steps

- Test and improve the dirty checker (this would make observation more or less complete from runtime perspective)
- Implement real proxy observation (as an opt-in replacement for descriptor observation)
- Improve/test/integrate lifecycle better with async tasks
- Improve HTML aware observation

### DIFF
--- a/packages/runtime/src/observation/computed-observer.ts
+++ b/packages/runtime/src/observation/computed-observer.ts
@@ -1,7 +1,7 @@
-import { PLATFORM, Reporter } from '@aurelia/kernel';
+import { PLATFORM, Reporter, Tracer } from '@aurelia/kernel';
 import { ILifecycle } from '../lifecycle';
 import {
-  IBindingTargetAccessor,
+  IBatchedSubscribable,
   IBindingTargetObserver,
   IObservable,
   IPropertySubscriber,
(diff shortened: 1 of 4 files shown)


EisenbergEffect:
That 'getter' is awesome 🤣

fkleuver:
Yup

PR #377, opened 2019-01-13
Dirty checker / ticker

fkleuver:
# Pull Request

## 📖 Description

- Added a ticker (based on pixi ticker but some irrelevant things removed and with slight perf tweaks) to `PLATFORM` that uses a global RAF loop, to serve as a generic scheduling utility.
- Modified the dirty checker to use this ticker instead of inline `setTimeout` calls
- Added a configuration point that allows you to adjust the dirty checker frequency, as well as disable it entirely (or throw/warn when a property is dirty checked)

### 🎫 Issues

N/A (was an open TODO)

## 👩‍💻 Reviewer Notes

- See dirty-checker.ts and platform.ts
- Also added acknowledgements for pixi (as well as inferno/mobx that were still due).

## 📑 Test Plan

- Added tests to verify the various options as well as that the dirty checker updates neither too late nor too early
- No specific tests for the ticker (it's being tested implicitly, will add more tests at a later point in time when it's used by more components if it starts deviating more from pixi's ticker)

## ⏭ Next Steps

- Potentially add some binding behaviors to control dirty checking behavior on a more granular level
- Utilize the ticker for other things where scheduling makes sense, such as mounting (have to experiment with this)


packages/runtime/src/observation/dirty-checker.ts
```diff
@@ -10,54 +10,101 @@ export interface IDirtyChecker {
...
+   * - A value of 6 will result in 10 dirty checks per second
+   */
+  framesPerCheck: 6,
+  /**
+   * Default: `false`
+   *
+   * Disable dirty-checking entirely. Properties that cannot be observed without dirty checking
+   * or an adapter, will simply not be observed.
+   */
+  disable: false,
```
bigopon: `disabled`?

fkleuver: yep agree

EisenbergEffect:
@fkleuver There's one more thing I'd like to do here before merging. Because the ticker is more than just inspired by Pixi, at least at this point, I'd like to include their full license into the ack file. Can you append it to the bottom of the file below what you've got now? If we change our stuff significantly later, we can remove that and keep the general ack we have above. I just want to be above the board and make sure we do a good job here.

Issue #76, opened 2018-08-19
Proxies for array.length and array[index] observation


fkleuver:
I have been experimenting with this but it will be more involved than simply adding it on top of the existing observation infrastructure (in which case it breaks). Creating an issue for this because I've been procrastinating on this and I would welcome any suggestions.

bigopon:
In order to have full observation, normal code needs to work on proxied array, instead of the raw array. So I think it's the matter of a documentation: if `array.length / array[index]` usage needs to be observable, then wrap the array with our provided `ProxyObserver`, some pseudo code:

```js
export class App {
  // items = []; // this can't work
  items = ProxyObserver.for([]);
  
  get stuff() {
    return this.items.filter(item => item.isStuff)
  }

  removeAllStuff() {
    this.items.length = 0; // this works instead of this.items.splice(0);
  }
}
```

fkleuver:
This actually already works under the hood without explicitly needing to do that, I believe I just need to hook up the length observer still.

bigopon:
Implicit observation is supported on read access. It's the write that isn't easily supported. I think we either have to force wrapping or use splice. Using splice is a much easier/safer way to teach/understand than forced wrapping.

PR #381, opened 2019-01-14
initial work for proxy observation

fkleuver:
# Pull Request

## 📖 Description

This is an initial stab at implementing proxy-based observation as a flexible opt-in (that is, it can be toggled on multiple levels via flags, but the easiest way is via the `@customElement` decorator). This is a preliminary API of course just to try things out.

~~The integration with `observer-locator` is pretty quick-n-dirty and it doesn't catch on every level just yet. Need to find the right places to replace/trap objects.~~

At least observing array length and index setters seems to work when the observation activates at the right level, so seems to be on the right track.

### 🎫 Issues

Related to (but broader than) https://github.com/aurelia/aurelia/issues/76

## 👩‍💻 Reviewer Notes

- Most file changes have to do with a single parameter change that adds lifecycle to hydration so the scope creation can be intercepted. Majority of changes are actually test files.
- Interesting bits to look at: [`binding-context.ts`](https://github.com/aurelia/aurelia/pull/381/files#diff-fb20f0198f8c451e96b4f6c194154bd2R81), [`observer-locator.ts`](https://github.com/aurelia/aurelia/pull/381/files#diff-41934ef11565bd232010c1e170bb633bR44) and of course [`proxy-observer.ts`](https://github.com/aurelia/aurelia/pull/381/files#diff-629432392ecd37dc6a52c15e75ae57beR1)

Would strongly recommend using VS code PR review tool to weed out the tests and stuff

## 📑 Test Plan

- Mostly planning to reuse existing tests and see if they all pass with proxy based observation as well
- Also need to add some extra tests for observation of things that weren't possible before,such as array length and index setter
- A little benchmark would be in place as well

## ⏭ Next Steps

Continue improving this and proceed with lifecycle improvements.

### DIFF
--- a/packages/runtime-html/test/behavior-assistance.ts
+++ b/packages/runtime-html/test/behavior-assistance.ts
@@ -9,6 +9,7 @@ import {
   IRenderable,
   IRenderingEngine,
   ITargetedInstruction,
+  LifecycleFlags as LF,
   TargetedInstructionType
 } from '@aurelia/runtime';
 import { InstanceProvider } from '../../runtime/src/rendering-engine';
@@ -48,7 +49,7 @@ export function hydrateCustomElement<T>(Type: Constructable<T>, ctx: HTMLTestCon
 
   const renderingEngine = container.get(IRenderingEngine);
   const projectorLocator = container.get(IProjectorLocator);
-  element.$hydrate(0, dom, projectorLocator, renderingEngine, host);
+  element.$hydrate(LF.none, dom, projectorLocator, renderingEngine, host);
 
   return { element, parent };
 }
--- a/packages/runtime-html/test/behavior-assistance.ts
+++ b/packages/runtime-html/test/behavior-assistance.ts
@@ -48,7 +48,7 @@ export function hydrateCustomElement<T>(Type: Constructable<T>, ctx: HTMLTestCon
 
   const renderingEngine = container.get(IRenderingEngine);
   const projectorLocator = container.get(IProjectorLocator);
-  element.$hydrate(dom, projectorLocator, renderingEngine, host);
+  element.$hydrate(0, dom, projectorLocator, renderingEngine, host);
 
   return { element, parent };
 }
--- a/packages/runtime/test/templating/view.spec.ts
+++ b/packages/runtime/test/templating/view.spec.ts
@@ -423,7 +423,7 @@ describe('View', () => {
       const rootTemplate: ITemplate<AuNode> = {
         renderContext: null as any,
         dom: null as any,
-        render(renderable) {
+        render( renderable) {
           const text = AuNode.createText();
           const wrapper = AuNode.createTemplate().appendChild(text).appendChild(childWrapper);
           const nodes = new AuNodeSequence(dom, wrapper);
@@ -447,7 +447,7 @@ describe('View', () => {
 
       // - Round 1 - bind
 
-      const scope1 = Scope.create(BindingContext.create(propName, propValue1));
+      const scope1 = Scope.create(0, BindingContext.create(0, propName, propValue1));
       if (lockScope1) {
         sut.lockScope(scope1);
       }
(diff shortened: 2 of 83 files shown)


fkleuver:
@EisenbergEffect @bigopon I would like to get some early feedback on this since this is going to have potentially huge impact on the course of the framework, and we should think carefully about how to control this via the API.
It's not a pretty nor very flexible implementation and needs many more tests. It's most likely not particularly efficient, doesn't catch everywhere yet and probably breaks in some edge cases. 
Still, at the basic level it seems to work.

EisenbergEffect:
@fkleuver Is the idea that someone can choose which version they want on a system or per component basis depending on their startup vs. repaint speeds? Also, would be always use proxy for array length and index change, regardless of the setting? (Haven't read the code yet but wanted to just get this first to mind question out.)

packages/runtime-html/test/behavior-assistance.ts
```diff
@@ -48,7 +48,7 @@ export function hydrateCustomElement<T>(Type: Constructable<T>, ctx: HTMLTestCon
 
   const renderingEngine = container.get(IRenderingEngine);
   const projectorLocator = container.get(IProjectorLocator);
-  element.$hydrate(dom, projectorLocator, renderingEngine, host);
+  element.$hydrate(0, dom, projectorLocator, renderingEngine, host);
```
BBosman: `LifecycleFlags.none` instead of `0`?

BBosman: I see there's more of these `0` calls in the PR. I'll leave it to just this comment instead of tagging them all individually. ;)

fkleuver: Yeah, there were about 800 of them to add (not all of which could be easily done with find/replace due to small differences) and I first wanted to see if it would work in the first place :) will fix this up

fkleuver:
@EisenbergEffect The idea is that you can choose between proxy vs. descriptor observation even on a binding level. That would be a bit verbose to configure, but possible. Most folks would probably want to set it on their root-level `app` or pass it via `aurelia.app({ ..., useProxies: true })` and forget about it. But since it is controlled by lifecycle flags, it can be changed at any level.
For length and index observation the idea is that you can use either proxies, the regular length observation (so listening for mutation calls) or dirty checking. I'm still meaning to add a binding behavior to say "dirty check this". But nothing "automatically" uses proxies. It *could* be that way, but I figured it would be safer to not do that just yet.

packages/runtime/test/templating/view.spec.ts
```diff
@@ -423,7 +423,7 @@ describe('View', () => {
       const rootTemplate: ITemplate<AuNode> = {
         renderContext: null as any,
         dom: null as any,
-        render(renderable) {
+        render( renderable) {
```
BBosman: Bonus space?

fkleuver: There was about 3h of mind-numbing test fixing, had to change a few things back and forth, did not pay attention to linting just yet :P

BBosman: As this was the only one I spotted during my visual check I'd consider that a good score for a 3 hour coding marathon ;)

fkleuver: Phew, haha

fkleuver:
I think this needs to be done in multiple steps. I can't go much further without properly fixing up and testing the lifecycles. @bigopon can you do a sanity check on this? @BBosman do you have any findings regarding typings/linting?

BBosman:
@fkleuver No new findings here. 👍

fkleuver:
Thanks @BBosman 
@EisenbergEffect I'm merging this with the intent of cleaning up and improving a couple things revolving around this as soon as I'm done with some lifecycle improvements. For now it is to avoid blocking others and avoiding a merge conflict hot spot.
I asked @bigopon for a sanity check and he pointed out (which I agree with) that proxies shouldn't be flowing through the AST. There needs to be a smarter mechanism for this, potentially something with getter traps, but will have to experiment with that.

The PR in its current form does not in any way affect the existing runtime as long as `useProxies` is not specified, so even though it is in a rough-ish state, it can not really do any harm either.

EisenbergEffect:
Sounds good @fkleuver I'm in full agreement with this approach.

PR #382, opened 2019-01-14
build: make linting the tests possible

BBosman:
# Pull Request

## 📖 Description

Created new `script`'s that run linting separately for `src` and `tests` and have the old `lint` script run both of those. This way you can do a quick lint on source or tests during development, while CircleCI will still enforce linting for the entire thing.

### 🎫 Issues

See #249.

## 👩‍💻 Reviewer Notes

- This PR adds the scrips, excludes some linting rules in the tests and fixes a whole lot of violations. 
- I also tweaked the scripts that generate tests, to write out linting compliant tests by fixing or ignoring (as they're in code generated by the ts compiler) those violations.

## 📑 Test Plan

CircleCI

## ⏭ Next Steps

See #249 and fix more of the linting errors in the test.


fkleuver:
Nice one with the generated tests :)
Thanks for fixing this up!

EisenbergEffect:
Wow @BBosman ! You are turning into a true vNext hero. Thank you so much for the time and effort here. Merging now, and also getting you added to the vNext team so you can create branches for future work. Sorry I didn't get that earlier. It somehow slipped my mind. Should be sorted in a minute or two.

PR #383, opened 2019-01-15
refactor: use substring in favor of substr

BBosman:
# Pull Request

## 📖 Description

From [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr):
> Warning: Although `String.prototype.substr(…)` is not strictly deprecated (as in "removed from the Web standards"), it is considered a **legacy function** and should be avoided when possible. It is not part of the core JavaScript language and may be removed in the future. If at all possible, use the `substring()` method instead.

And that's just what this PR does. 😈 

### 🎫 Issues

n/a

## 👩‍💻 Reviewer Notes

Switched all usages of `substr` to `substring`. And added `substr` to the `ban` linting rule to avoid future usage.

Note that the major difference between those two function is that the second argument of `substring` is the index into the string instead of the length of the characters to grab.

## 📑 Test Plan

CircleCI.

## ⏭ Next Steps

n/a


BBosman:
@EisenbergEffect Lets merge #366 before this one, once this one is approved. 😉

fkleuver:
@BBosman Looks like a merge conflict found its way in. If you could just fix that up then this should be good to go :)

BBosman:
@fkleuver I was expecting the `tslint.json` merge conflict. Rebased and good to go.

PR #386, opened 2019-01-16
build: update (dev)Dependencies

BBosman:
# Pull Request

## 📖 Description

Update devDependencies. Mostly patch-level updates. Only noteworthy is the major bump in `rollup` and related packages.

### 🎫 Issues

n/a

## 👩‍💻 Reviewer Notes

The `rollup` update is a major version bump, so upgrading could introduce breakage. I tested that `npm run dist:all` still works, but the generated `min` files differ slightly. I'm not a `rollup` expert, so I don't know if those matter. Having somebody take this branch for a spin is advisable. 

## 📑 Test Plan

CircleCI

## ⏭ Next Steps

n/a


fkleuver:
Yep looks good, I was thinking about this but you beat me to it :) thanks!

fkleuver:
Looks like you missed updating the package-lock file in the browserstack's package.json which caused the e2e build job to fail (this is the only job we're not running on each PR because I still need to fix a concurrency issue)
```
cd ./test/browserstack
npm ci
npm run build
npm ERR! cipm can only install packages when your package.json and package-lock.json or npm-shrinkwrap.json are in sync. Please update your lock file with `npm install` before continuing.
npm ERR! 
npm ERR! 
npm ERR! Invalid: lock file's ts-loader@5.3.2 does not satisfy ts-loader@^5.3.3
npm ERR! Invalid: lock file's webpack-cli@3.1.2 does not satisfy webpack-cli@^3.2.1
npm ERR! 
```

Don't forget to actually `npm install` when there is a lockfile because circleci uses `npm ci` and will fail when theyre out of sync

PR #387, opened 2019-01-16
refactor(jit-html): more linting fixes through test generation scripts

BBosman:
# Pull Request

## 📖 Description

A couple of small linting fixes through the test generation scripts that slipped through the cracks in #382.

### 🎫 Issues

Follow up to #382.

## 👩‍💻 Reviewer Notes

n/a

## 📑 Test Plan

CircleCI

## ⏭ Next Steps

n/a


fkleuver:
Looks good. Just a note in case you happened to be planning to do more stuff with the test generation: I'm intentionally keeping it typing-free to give the TS compiler as little work to do as possible, because many more of these will be added over time and the build time is already quite slow. But of course linting fixes don't affect that, so this is all good.

PR #388, opened 2019-01-17
feat(kernel): port EventAggregator to vNext

BBosman:
# Pull Request

## 📖 Description

Ports EventAggregator over from vCurrent. Basic port of both `src` and `test`, with linting and typing changes applied.

### 🎫 Issues

n/a

## 👩‍💻 Reviewer Notes

Notable changes:
- ~~Added a `static register` method for registration with DI, with test.~~
- Made the callback signature explicit `(data?: unknown, event?: string) => unknown` (was `Function`) and generic and gave it its own type `EventAggregatorCallback<T>` for ease of use and readability.
- You can subscribe by channel (`string`) or type (`Constructable`, was `any`) and added function overloads (with generics) to make the API clearer.
- Dropped the `aurelia-logger` dependency and switched to `Reporter` for logging/throwing errors.
- Added some missing tests to increase coverage.
- Many cosmetic linting fixes (almost all spacing related).

## 📑 Test Plan

CircleCI.

## ⏭ Next Steps

Decide if we want to make the EventAggregator it's own `@aurelia/eventaggregator` package, and if so, extract it from `@aurelia/kernel`.

### DIFF
--- a/packages/kernel/src/eventaggregator.ts
+++ b/packages/kernel/src/eventaggregator.ts
@@ -66,10 +66,6 @@ export class EventAggregator {
     this.messageHandlers = [];
   }
 
-  public static register(container: IContainer): IResolver<EventAggregator> {
-    return Registration.singleton(EventAggregator, this).register(container);
-  }
-
   /**
    * Publishes a message.
    * @param channel The channel to publish to.
--- a/packages/kernel/src/eventaggregator.ts
+++ b/packages/kernel/src/eventaggregator.ts
@@ -49,6 +49,11 @@ export interface Subscription {
   dispose(): void;
 }
 
+/**
+ * Enables loosely coupled publish/subscribe messaging.
+ * @param data The optional data published on the channel.
+ * @param event The event that triggered the callback. Only available on channel based messaging.
+ */
 export type EventAggregatorCallback = (data?: unknown, event?: string) => unknown;
 
 /**
@@ -74,25 +79,27 @@ export class EventAggregator {
 
   /**
    * Publishes a message.
-   * @param event The event or channel to publish to.
+   * @param channelOrType The event or channel to publish to.
    * @param data The data to publish on the channel.
    */
-  public publish(event: string | InstanceType<Constructable>, data?: unknown): void {
+  public publish(channel: string, data?: unknown): void;
+  public publish(type: InstanceType<Constructable>, data?: unknown): void;
+  public publish(channelOrType: string | InstanceType<Constructable>, data?: unknown): void {
     let subscribers: (EventAggregatorCallback | Handler)[];
     let i: number;
 
-    if (!event) {
+    if (!channelOrType) {
       throw new Error('Event was invalid.');
     }
 
-    if (typeof event === 'string') {
-      subscribers = this.eventLookup[event];
+    if (typeof channelOrType === 'string') {
+      subscribers = this.eventLookup[channelOrType];
       if (subscribers) {
         subscribers = subscribers.slice();
         i = subscribers.length;
 
         while (i--) {
-          invokeCallback(subscribers[i] as EventAggregatorCallback, data, event);
+          invokeCallback(subscribers[i] as EventAggregatorCallback, data, channelOrType);
         }
       }
     } else {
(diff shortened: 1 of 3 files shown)


fkleuver:
Thanks for doing this. Would you mind briefly highlighting the changes you made compared to vCurrent and add those to review notes, not only for now but also for future readers/reference?

We can probably keep the class directly, but we do need to think of the public API. A couple of nitpick-ish things come to mind when I look at this:
- I don't think `eventLookup` and `messageHandlers` should necessarily be public, but at least they should be readonly
- We can use signature overloading to make the API clearer (this is a problem in vCurrent) and somewhat clarify the API doc in the process (as you can now specialize the API documentation specifically per overload)
- We should use the reporter to throw any errors (there are still two regular errors thrown). It's okay to just throw a code 0 with a // TODO comment that includes the original error message (im planning to reorganize error codes with a search-all for todo comments :))
- For completeness sake / to get full coverage, it would be nice to have 3 extra unit tests: one that verifies `register` returns a singleton resolver, and two (multiplied by invalid inputs) that validate errors are thrown for invalid input events. I don't mind doing this if you prefer to leave the loopy tests to me, that's up to you :)

BBosman:
The current (migrated) tests expect `eventLookup` and `messageHandlers` to be public, so I'll need to do a bit of refactoring for that. I'll be AFK for most of the weekend, but will take a stab at it (and the other suggestions) early next week.

fkleuver:
In that case you can just mark them as `/**@internal*/` so they're available for the tests but hidden from the public API.

EisenbergEffect:
I don't think those were meant to be public, so definitely agree with @fkleuver 's advice there.

packages/kernel/src/eventaggregator.ts
```diff
@@ -0,0 +1,165 @@
...
+        while (i--) {
+          invokeCallback(subscribers[i] as EventAggregatorCallback, data, channelOrType);
+        }
+      }
+    } else {
+      subscribers = this.messageHandlers.slice();
+      i = subscribers.length;
+
+      while (i--) {
+        invokeHandler(subscribers[i] as Handler, channelOrType);
```
fkleuver: Isn't this supposed to take data as well @EisenbergEffect ?

packages/kernel/src/eventaggregator.ts
```diff
@@ -0,0 +1,165 @@
...
+    return Registration.singleton(EventAggregator, this).register(container);
+  }
+
+  /**
+   * Publishes a message.
+   * @param channelOrType The event or channel to publish to.
+   * @param data The data to publish on the channel.
+   */
+  public publish(channel: string, data?: unknown): void;
+  public publish(type: InstanceType<Constructable>, data?: unknown): void;
```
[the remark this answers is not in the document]
fkleuver: Not at all in publish? (this override is not publicly visible)

packages/kernel/src/eventaggregator.ts
```diff
@@ -0,0 +1,165 @@
...
+
+  /**
+   * Creates an instance of the EventAggregator class.
+   */
+  constructor() {
+    this.eventLookup = {};
+    this.messageHandlers = [];
+  }
+
+  public static register(container: IContainer): IResolver<EventAggregator> {
```
EisenbergEffect: Should remove the explicit DI-related code from this? Since singleton is the default auto-registration behavior, this will happen anyways.

packages/kernel/src/eventaggregator.ts
```diff
@@ -0,0 +1,165 @@
...
+   */
+  dispose(): void;
+}
+
+/**
+ * Enables loosely coupled publish/subscribe messaging.
+ * @param data The optional data published on the channel.
+ * @param event The event that triggered the callback. Only available on channel based messaging.
+ */
+export type EventAggregatorCallback = (data?: unknown, event?: string) => unknown;
```
EisenbergEffect: Perhaps something like this:

```TypeScript
export type EventAggregatorCallback<T = unknown> = (data?: T, event?: string) => unknown;
```

See below for why.

packages/kernel/src/eventaggregator.ts
```diff
@@ -0,0 +1,165 @@
...
+    }
+  }
+
+  /**
+   * Subscribes to a message channel or message type.
+   * @param channelOrType The event channel or event data type.
+   * @param callback The callback to be invoked when the specified message is published.
+   */
+  public subscribe(channel: string, callback: EventAggregatorCallback): Subscription;
+  public subscribe(type: Constructable, callback: EventAggregatorCallback): Subscription;
```
EisenbergEffect: With the updating callback typing, we can provide strong typing to the subscribe callback:

```TypeScript
public subscribe<T extends Constructable>(type: T, callback: EventAggregatorCallback<InstanceType<T>>): IDisposable;
```

packages/kernel/src/eventaggregator.ts
BBosman: This is copy/pasted code, which is found on numerous places in the codebase as is, so this would have to be changed there as well? Pinging @fkleuver as it's his code I copied. ;)

fkleuver: It actually wasn't the default auto-registration behavior up until I fixed this in my previous PR, but now it is indeed, and it can be removed from the other locations as well

BBosman: Will you do those other locations? I can do a *"search/replace"*, but you probably know all locations and permutations by heart. 😊

fkleuver: Yeah I'll fix that in one of the coming cleanup passes I'm doing, no worries :)

BBosman:
I think I addressed all comments, so it's ready for another review imho.

fkleuver:
If anything, it's definitely good enough as a first version
I thought about the file name, `event-aggregator` might be better in terms of consistency. But that's an easy and small change, and let's just get this in before the nightly :)

PR #390, opened 2019-01-17
chore(jit-html): fix unit tests

fkleuver:
# Pull Request

## 📖 Description

Fixes a regression in the DI registration unit tests caused by the proxy observer PR.

It slipped through the crack because I thought the tests on master were failing solely because of the package-lock issue with the browserstack tests.


fkleuver:
@EisenbergEffect Merging this without additional review due to the minuscule scope (only fixes 3 tests), it not affecting anything else nor changing anything from a typing/linting perspective, and all checks going green.

PR #391, opened 2019-01-17
chore(all): pin event-stream to safe version 3.3.4 in all package.jsons

fkleuver:
# Pull Request

## 📖 Description

This PR fixes an open security alert on the vcurrent e2e benchmark project which had the latest event-stream installed, and also pins the event-stream version to 3.3.4 (the latest safe version) on other packages that had it set to `latest`.


fkleuver:
@EisenbergEffect Merging this without additional review because there are no code changes, all checks go green and the security alert is resolved.

PR #393, opened 2019-01-17
ci: simplify config.yml and cleanup unused scripts

fkleuver:
# Pull Request

## 📖 Description

CircleCI needs some additional stuff soon (router e2e tests, coverage for other envs, etc) and also the jobs chaining turned out to be very inefficient (`npm ci` takes 12 seconds, whilst the job that does it blocks all other jobs for 1 minute and 10 seconds).

This PR merges some jobs and removes all commands (most of which weren't even reused in the first place). I thought initially this would make the CI configuration more scalable on the long term, but it turns out it just makes it inflexible and hard to maintain. So I simplified things.


fkleuver:
@EisenbergEffect Reduced the circleci config.yml from 700 lines to 450 lines and reduced the workflow run time from 8 minutes to 4 minutes, so that's a good step in preparation for the additional complexity and run time that is to come when the e2e test suites are further expanded.

There are only CI-related changes, so merging this now to see if the master-specific build job still works.

PR #394, opened 2019-01-18
Lifecycles preparation work

fkleuver:
# Pull Request

## 📖 Description

I decided to PR this separately to reduce the fairly large scope of the lifecycle work that comes next.

This PR contains preparation work in the form of moving/merging/splitting several types related to lifecycles:
- enums (`LifecycleFlags`, `ExpressionKind`, `State`, `Hooks`, `BindingMode`) are moved to [flags.ts](https://github.com/aurelia/aurelia/compare/lifecycles?expand=1#diff-303b6dde30be4663ab4f78bc1ff0330dR9) to reduce the amount of circular referencing (enums don't have dependencies on anything, making them suitable for putting together in one file)
- Extracted interfaces from the whole AST and placed those in top-level [ast.ts](https://github.com/aurelia/aurelia/compare/lifecycles?expand=1#diff-ae6b1c91312f135f57cc18335da7673dR1)
- Merged all standalone lifecycle hooks interfaces into [LifecycleHooks](https://github.com/aurelia/aurelia/compare/lifecycles?expand=1#diff-5bffb21f0a04d038c5027956f70bab9bR123) (reasoning being that hooks are optional anyway, so even if a component cannot have them by definition, the typings are still accurate and say "you need to check this")
- Removed redundant `.$kind` properties from AST prototype while I was at it

The AST interface work is meant to make it "cheaper" for more things to reference the AST without actually importing the classes (which would kill tree shaking capabilities for optimized AOT builds).

The lifecycle interface work is meant as a starting point for `continueWith`-like behavior in lifecycle tasks and further integration within the lifecycles, allowing jumps, cancellations, reordering and the likes.

### 🎫 Issues

N/A

## 👩‍💻 Reviewer Notes

N/A

## 📑 Test Plan

No functional changes, just fixed a few AST tests due to removing the `$kind` properties

## ⏭ Next Steps

Start with the functional changes for the lifecycle / binding improvements


fkleuver:
@bigopon Would you mind doing a sanity check on this

fkleuver:
Thanks @bigopon 

@BBosman Does this look ok from a typing/linting perspective?

fkleuver:
@BBosman I double checked but I believe no additional linting errors were introduced, but let me know if I missed anything and I'll fix it then. You said you were going to be afk this weekend so I'll just go ahead and merge so I can continue with the next bit.

@EisenbergEffect FYI

PR #395, opened 2019-01-18
fix(expression-parser): handle trailing comma correctly

fkleuver:
# Pull Request

## 📖 Description

I missed something in the spec: https://tc39.github.io/ecma262/#sec-array-initializer

> If an element is elided at the end of an array, that element does not contribute to the length of the Array.

### 🎫 Issues

N/A

## 👩‍💻 Reviewer Notes

N/A

## 📑 Test Plan

`$undefined` elements removed at the end of `ArrayLiteral`s in the tests

## ⏭ Next Steps

N/A

Move along people, nothing to see here.. :)


fkleuver:
@EisenbergEffect Merging this as it is just a tiny detail in the parser, only some code removed (which makes the project coverage go down by 0.01% but the parser stays at 100%)

Issue #369, opened 2019-01-08
[RFC] Routing with Aurelia Documentation


jwx:
# 💬 RFC
In addition to the "traditional url/path configured routing" the new router will have a viewport centric routing based on conventions. This RFC contains an early version of the first part of the "first contact document" for this non-traditional routing.
___
# Routing

Navigation in Aurelia is handled by viewports and components. In short, you put one or more viewports in your application and then you specify their content using links or code. Whenever someone navigates to a new ~~state of viewport components~~ route an entry will be added to the browser history, making it possible to use the browser's back, forward and refresh and get the right set of components. Let's take a look at an example. 

## Viewports

For the start view of our app, we're going to use two viewports:
```html
<!-- app.html -->
<body>
  <nav>
    <!-- We'll get back to this -->
  </nav>

  <viewport name="lists" used-by="authors,books" default="authors"></viewport>
  <viewport name="content" default="about"></viewport>
</body>
```
The first viewport we'll give the name _lists_ and specify that it's only used by the components/custom elements _Authors_ and _Books_. We also set _Authors_ as the default component, meaning that if nothing is specifically set for this viewport, it will display the _Authors_ component. For the second viewport we won't specify anything except giving it the name _content_ and setting the default to the _About_ component. Convention will then allow all components not used by another viewport to use it. Note: While it's not necessary to explicitly name a viewport (it will then be named _default_ by convention), it is recommended. Especially if you have more than one viewport in your application.

That's it for setting up the viewports. Now let's look at how to navigate and change the components in them.

## Navigation

Navigation can be triggered in html through `a`nchor tags or in code with the router's `goto` method. (There are more navigation methods available in the router, but we'll stick to `goto` for now.) A navigation segment in it's basic form consists of three parts: which _component_ to show, in which _viewport_ to show the component and what _parameters_ to pass into the component, like this
```html
"component@viewport(parameters)"
```
or this
```typescript
router.goto(Component, 'viewport', parameters);
```
But sometimes you don't want to pass _parameters_ to the _component_, so that part is optional. And if the router can deduce where a component should go by itself, for example based on the viewport settings we did earlier, _viewport_ can also be left out. So if you're using only one viewport, or have done a configuration similar to the one we did, you don't have to think about the viewport when navigating. Our `nav` part from above could then be
```html
  <nav>
    <ul>
      <li><a href="about">About</a></li>
      <li><a href="authors+about">Authors</a></li>
      <li><a href="books+about">Books</a></li>
[...]

Issue #367, opened 2019-01-08
Router demos


jwx:
# 💬 Demo

An issue for various router demos. Question, thoughts and opinions are always welcome.

jwx:
Following up on the demo in the [Router parameters PR](https://github.com/aurelia/aurelia/pull/362) I've made a small addition to demonstrate one of the ways to alter the "url syntax".

![navigation example - customized syntax](https://user-images.githubusercontent.com/3307996/50804279-8b972b80-12ed-11e9-85c0-6af4d708d05b.gif)

The links in the files have now changed to
`app`
```html
<li><a href="authors/about">Authors</a></li>
<li><a href="books/about">Books</a></li>
<li><a href="about">About</a></li>
```
`authors` (and `books` is the same)
```html
<a href="author/${author.id}">${author.name}</a>
```
`author` (and `book` is the same)
```html
<li repeat.for="book of author.books"><a href="book/${book.id}">${book.title}</a></li>
```
There's still no route configuration, but the custom syntax has been hooked into the router like this
`app`
```typescript
    this.router.activate({
      transformFromUrl: (path, router) => {
        if (!path.length) {
          return path;
        }
        if (path.startsWith('/')) {
          path = path.slice(1);
        }
        // Fetch components for the "lists" viewport
        const listsComponents = router.rootScope.viewports.lists.options.usedBy.split(',');
        const states = [];
        const parts = path.split('/');
        while (parts.length) {
          const component = parts.shift();
          const state: IComponentViewportParameters = { component: component };
          // Components in "lists" viewport can't have parameters so continue
          if (listsComponents.indexOf(component) >= 0) {
            states.push(state);
            continue;
          }
          // It can have parameters, but does it?
          if (parts.length > 0) {
            state.parameters = { id: parts.shift() };
          }
          states.push(state);
        }
        return states;
      },
      transformToUrl: (states: IComponentViewportParameters[], router) => {
        const parts = [];
        for (const state of states) {
          parts.push(state.component);
          if (state.parameters) {
            parts.push(state.parameters.id);
          }
        }
        return parts.join('/');
      }
    });
```
[...]

PR #396, opened 2019-01-19
Router: local components for direct routing

jwx:
# Pull Request

## 📖 Description

Adding support for local components/dependencies for direct routing. Even though the components are local for other components and resolved according to the viewport's context, there's no need for child routers or router configuration.

- chore(router): update type in test app
- chore(router): update types
- chore(router): fix rebase issues
- test(router): add local components to test app
- chore(router): delete unused files and code & reorganize test components
- chore(router): delete old files
- refactor(router): change entering timing & add viewport render and state
- chore(viewport): fix a small slip-up [Fred Kleuver]
- chore(viewport): cleanup a leftover [Fred Kleuver]
- refactor(router): update closest & add container influence on load
- refactor(router): add container to viewport

Example usages (from the demo below) can be found [here](https://github.com/aurelia/aurelia/issues/367#issuecomment-455612499).

### 🎫 Issues

Related to https://github.com/aurelia/aurelia/issues/307, https://github.com/aurelia/aurelia/issues/369 and https://github.com/aurelia/aurelia/issues/367#issuecomment-455612499.

## 👩‍💻 Reviewer Notes

A demo can be found here: https://dist-9r70rzqlm.now.sh/
In the demo, the `book` and `author` both link to their own version of `information`. The `information` components are local to `book` and `author` components, exist in each of their folders and have the same file name, custom element name and class name. Despite these collisions, the components are resolved according to the context of the viewport.

## 📑 Test Plan

- Added tests from @fkleuver that deals with local dependencies

## ⏭ Next Steps

- Increase test density
- Clean up code
- Re-visit scoped viewports


packages/router/src/resources/viewport.ts
```diff
@@ -1,10 +1,10 @@
-import { Constructable, InterfaceSymbol } from '@aurelia/kernel';
-import { bindable, CustomElementResource, INode, LifecycleFlags } from '@aurelia/runtime';
+import { Constructable, IContainer, InterfaceSymbol, Writable } from '@aurelia/kernel';
+import { bindable, createRenderContext, CustomElementResource, IDOM, IElementTemplateProvider, INode, IRenderContext, IRenderingEngine, ITemplate, LifecycleFlags, TemplateDefinition } from '@aurelia/runtime';
 import { Router } from '../router';
 import { IViewportOptions, Viewport } from '../viewport';
 
 export class ViewportCustomElement {
-  public static readonly inject: ReadonlyArray<InterfaceSymbol|Constructable> = [Router, INode];
+  public static readonly inject: ReadonlyArray<InterfaceSymbol<unknown> | Constructable> = [Router, INode, IRenderingEngine];
```
fkleuver:
```suggestion
  public static readonly inject: ReadonlyArray<InterfaceSymbol|Constructable> = [Router, INode, IRenderingEngine];
```
Not sure why this was changed back, but this means the same and is terser

packages/router/src/resources/viewport.ts
```diff
@@ -45,10 +84,10 @@ export class ViewportCustomElement {
     if (this.element.hasAttribute('no-history')) {
       options.noHistory = true;
     }
-    this.viewport = this.router.addViewport(this.name, this.element, options);
+    this.viewport = this.router.addViewport(this.name, this.element, (this as any).$context.get(IContainer), options);
```
fkleuver: If you add this to the top of the file:

```ts
export interface ViewportCustomElement extends ICustomElement<Element> { }
```

Then you don't need to do these casts and you get strong typing for the properties you need.

packages/router/src/router.ts
```diff
@@ -46,7 +47,7 @@ export interface IRouteSeparators {
 }
 
 export class Router {
-  public static readonly inject: ReadonlyArray<InterfaceSymbol> = [IContainer];
+  public static readonly inject: ReadonlyArray<InterfaceSymbol<unknown>> = [IContainer];
```
fkleuver:
```suggestion
  public static readonly inject: ReadonlyArray<InterfaceSymbol> = [IContainer];
```

packages/router/src/router.ts
```diff
@@ -254,7 +255,18 @@ export class Router {
...
+          if (canEnter) {
+            return value.enter();
+          } else {
+            return false;
+          }
+        }
+        // TODO: Deal with redirects
+        return true;
+      }));
       if (results.findIndex((value) => value === false) >= 0) {
```
fkleuver:
```suggestion
      if (results.some(result => result === false)) {
```

packages/router/src/router.ts
```diff
@@ -287,9 +297,12 @@ export class Router {
 
     this.replacePaths(instruction);
 
-    if (!keepHistoryEntry) {
+    // Remove history entry if no history viewports updated
+    if (!instruction.isFirst && !updatedViewports.reduce((accumulated: boolean, current: Viewport) => !current.options.noHistory || accumulated, instruction.isFirst)) {
```
fkleuver:
```suggestion
    if (!instruction.isFirst && updatedViewports.every(viewport => viewport .options.noHistory)) {
```

packages/router/src/router.ts
```diff
@@ -287,9 +297,12 @@ export class Router {
...
       this.historyBrowser.pop().catch(error => { throw error; });
     }
+
+    updatedViewports.forEach((value) => value.finalizeContentChange());
```
fkleuver:
```suggestion
    updatedViewports.forEach(viewport => {
      viewport .finalizeContentChange()
    });
```

Makes it easier to debug.

packages/router/src/router.ts
```diff
@@ -519,30 +532,21 @@ export class Router {
   private ensureRootScope(): void {
     if (!this.rootScope) {
       const aureliaRootElement = this.container.get(Aurelia).root().$host;
-      this.rootScope = new Scope(this, aureliaRootElement as Element, null);
+      this.rootScope = new Scope(this, aureliaRootElement as Element, aureliaRootElement.$customElement.$context, null);
```
fkleuver:
```suggestion
      const root = this.container.get(Aurelia).root();
      this.rootScope = new Scope(this, root.$host as Element, root.$context, null);
```
(and of course remove the line above)

packages/router/src/router.ts
```diff
@@ -519,30 +532,21 @@ export class Router {
...
-          break;
-        }
-        i++;
-        el = el.parentElement;
-      }
-      if (i < closest) {
-        closest = i;
-        scope = sc;
+    const el = closestCustomElement(element);
+    let container = (el as any).$customElement.$context.get(IContainer);
```
fkleuver: If you return `CustomElementHost` from `closestCustomElement` then you don't need to do this casting.

packages/router/src/scope.ts
```diff
@@ -19,6 +20,7 @@ export interface IFindViewportsResult {
 
 export class Scope {
   public element: Element;
+  public container: IRenderContext;
```
fkleuver: Can we just call this `context` please, for consistency with the rest of the framework?

packages/router/src/scope.ts
```diff
@@ -186,60 +189,30 @@ export class Scope {
...
+        scope = new Scope(this.router, element, container, this);
         this.router.scopes.push(scope);
       }
 
-      viewport = this.viewports[name] = new Viewport(this.router, name, element, this, scope, options);
+      viewport = this.viewports[name] = new Viewport(this.router, name, element, container, this, scope, options);
     }
-    if (element) {
-      viewport.setElement(element, options);
+    if (element || container) {
```
fkleuver: This seems confusing. Do you mean `element && container`? If not, could you clarify why only one of them needs to be defined?

packages/router/src/scope.ts
```diff
@@ -186,60 +189,30 @@ export class Scope {
...
-  public removeViewport(viewport: Viewport): number {
-    if (viewport.scope) {
-      this.router.removeScope(viewport.scope);
+  public removeViewport(viewport: Viewport, element: Element, container: IRenderContext): number {
+    if ((!element && !container) || viewport.remove(element, container)) {
+      if (viewport.scope) {
+        this.router.removeScope(viewport.scope);
+      }
+      // tslint:disable-next-line:no-dynamic-delete
+      delete this.viewports[viewport.name];
```
fkleuver:
```suggestion
      Reflect.deleteProperty(this.viewports, viewport.name);
```
Then you can remove the suppression.

packages/router/src/scope.ts
```diff
@@ -299,46 +272,25 @@ export class Scope {
     if (this.viewport) {
       parents.unshift(this.viewport.description(full));
     }
-    let viewport: Viewport = this.parent.closestViewport(this.element.parentElement);
+    let viewport: Viewport = this.parent.closestViewport((this.container as any).parent);
```
fkleuver: Not a fan of parent container traversal but it's better than DOM traversal. Mind declaring a type for this to avoid `any` casting? e.g. `type ChildContainer = IContainer & { parent?: ChildContainer }`

packages/router/src/utils.ts
```diff
@@ -0,0 +1,9 @@
+export function closestCustomElement(element: Element): Element {
```
fkleuver:
```suggestion
export function closestCustomElement(element: CustomElementHost): CustomElementHost{
```

packages/router/src/viewport.ts
```diff
@@ -124,11 +141,19 @@ export class Viewport {
...
+    if (this.container !== container) {
+      this.container = container;
+    }
+
     if (!this.component && !this.nextComponent && this.options.default) {
       this.router.addProcessingViewport(this, this.options.default);
     }
   }
 
+  public remove(element: Element, container: IRenderContext): boolean {
```
fkleuver: This doesn't actually remove anything. Do you mean `canRemove`?

packages/router/src/viewport.ts
```diff
@@ -304,26 +354,81 @@ export class Viewport {
...
+    const container = this.container || this.router.container;
+    if (typeof component !== 'string') {
+      return container.get(IContainer).get<IRouteableCustomElement>(component);
+    } else {
+      return container.get(IContainer).get<IRouteableCustomElement>(CustomElementResource.keyFrom(component));
+    }
   }
 
-  private addComponent(component: ICustomElement): void {
+  private freshState(): void {
```
fkleuver: Maybe `clearState` is a better name for this?

packages/router/src/viewport.ts
```diff
@@ -148,7 +173,7 @@ export class Viewport {
     return result;
   }
 
-  public canEnter(): Promise<boolean> {
+  public async canEnter(): Promise<boolean> {
     if (this.clear) {
       return Promise.resolve(true);
```
fkleuver: The method is already async, so you can just `return true`

packages/router/src/viewport.ts
```diff
@@ -176,6 +201,29 @@ export class Viewport {
     return result;
   }
 
+  public async enter(): Promise<boolean> {
+    // tslint:disable-next-line:no-console
+    console.log('Viewport enter', this.name);
+
+    if (this.clear) {
+      return Promise.resolve(true);
```
fkleuver: The method is already async, so you can just `return true`

packages/router/src/viewport.ts
```diff
@@ -176,6 +201,29 @@ export class Viewport {
...
+    }
+
+    if (!this.nextComponent) {
+      return Promise.resolve(false);
+    }
+
+    if (this.nextComponent.enter) {
+      const merged = mergeParameters(this.nextParameters, this.nextInstruction.query, (this.nextContent as IRouteableCustomElementType).parameters);
+      this.nextInstruction.parameters = merged.parameters;
```
fkleuver: What is the difference between `.parameters` and `.list`? Could there be a better name for them?

packages/router/src/viewport.ts
```diff
@@ -304,26 +354,81 @@ export class Viewport {
...
+  private freshState(): void {
+    this.options = {};
+
+    this.content = null;
+    this.parameters = null;
+    this.instruction = null;
+    this.component = null;
+  }
+
+  private async loadComponent(component: ICustomElementType | string): Promise<void> {
```
fkleuver: It would be clearer to let the names match up with the lifecycles, e.g. `hydrateComponent`, `bindComponent`, etc

fkleuver:
I'm really happy you got this all to work properly. Looks like some serious effort went into this.

There needs to be a dedicated component that offers some cleaner API and encapsulation for dealing with all the various swapping and lookup scenarios (instead of traversing up the container tree), but your code has the benefit of being imperative so it would be easy enough to refactor. I have some ideas on how to make some of these bits terser and more reusable, so architecturally I think we're good for the goal it accomplishes.
From a functional point of view I'd also say LGTM.

I added a few comments/suggestions that I would appreciate if you can try to address. It doesn't have to be perfect (I can also help you fix a few things if you're too short on time) but some things you know best, so you'd need to do a first pass.

Also, I'm not sure which pieces of commented out code you're still going to use. Could you add some brief clarifying comment on why you're leaving a piece of commented-out code in place?

Lastly, the test coverage obviously. I don't care about the percentage, but the `scope.ts` in particular seems like it's quite poorly tested. If its API still going to change a lot then integration tests are probably the best investment of time, as they are more refactor proof. If the API is fit for its purpose however then some unit tests would probably be faster, easier, and serve the added purpose of making it easier for others to understand what it's meant to do.

jwx:
@fkleuver I'll address your comments (and possibly also do some clean up of old code) tomorrow. I'm going to take another look at scoped viewports next which will probably result in both api changes and added tests for scope.

fkleuver:
Alright, we can drop the coverage thing for this PR then to prevent over-scoping it. Most important thing is to eliminate some casts that are better solved in other ways as I pointed out in comments (this should help you as well down the road) and improve a couple method / variable names to reflect what they actually do/mean. I don't want to introduce additional typing/linting errors into the master branch, even if it's WIP. These things are generally easy to fix (and if not, you can ask for help).

packages/router/src/scope.ts
jwx: You want me to change that everywhere?

fkleuver: If it's typed `IRenderContext` then yeah. If it's `IContainer` then it's container. If semantically speaking `IContainer` is what you really need/want, just `context.get(IContainer)` and pass that to the next thing. That's the "safe cast" from context to container.

packages/router/src/resources/viewport.ts
jwx: Thanks.

packages/router/src/router.ts
jwx: Doh!

packages/router/src/router.ts
jwx: Doh!

packages/router/src/scope.ts
jwx: Not sure that's my preference, but on the other hand I don't feel that much about it. Fixed all the other deletes as well.

packages/router/src/utils.ts
jwx: `CustomElementHost` doesn't have `parentElement` so I kept `Element` for the parameter and cast to `CustomElementHost` where needed.

packages/router/src/viewport.ts
jwx: It will eventually. (And if it won't, then I'll change it.)

packages/router/src/viewport.ts
jwx: I had that first, but it didn't feel right since it's not clearing the entire state, just a small subset of it. No real objection to change it back, though, if you prefer it?

packages/router/src/viewport.ts
jwx: Initially they weren't matched up with the lifecycle actions, and they might go back to that. But if they should end up staying matched. I'd be happy to change it.

packages/router/src/scope.ts
jwx: It might be changed. I added a TODO.

packages/router/src/viewport.ts
jwx: Is `.namedParameters` and `.parameterList` better and clear enough?

fkleuver: is `namedParameters` a map of key-values and `parameterList` a raw list of the values?

jwx: Yes.

packages/router/src/scope.ts
fkleuver: It takes some getting used to, I didn't particularly like it either, but I'm starting to see its elegance. It is kind of the way forward with the latest ES which we want to try to embrace.

packages/router/src/utils.ts
fkleuver: It has a generic type argument that it merges itself with, so you can use `CustomElementHost<Element>` to achieve the same effect and remove the need for casting

packages/router/src/viewport.ts
fkleuver: Then I suppose it's clear enough (at least for me) :)

packages/router/src/viewport.ts
fkleuver: If something is inconsistent due to some temporary state or "might change" then please add a note or TODO comment clarifying that. I sometimes leave stuff like this as well for the same reason but you'll never see me do that without leaving some kind of indicator (if you do see me do that, you can also tell me I need to change it).
It is not just for you and me, also for other readers, and also for our own shitty memories.

packages/router/src/viewport.ts
fkleuver: I refer to the same comment as with the `delete()` method, please either change or add a TODO clarifying the circumstance under which they will change.

packages/router/src/viewport.ts
jwx: Renamed it to `clearState`. It'll do.

fkleuver:
Merging this now so you don't get semi-blocked on the next things you're working on. Handful of nitpicks can be addressed later.

PR #398, opened 2019-01-20
lifecycle refactor step 1/?

fkleuver:
# Pull Request

## 📖 Description

Main purpose of this PR is to allow the repeater (and any other resources) to bind its child views during the `binding()` hook instead of during `bound()` in order to keep the timings of the different kinds of binding-related operations closer to each other.

To accomplish this, I adjusted the rendering process to put bindings in a separate list from bindable components. Those bindings are then bound before the `binding()` hook is invoked, so that within the `binding()` hook you can safely assume the component itself to already be in a "bound" state. Child components are still going to be bound *after* the `binding()` hook, so that you can also safely assume all parent components to both be in a bound state and to have had their `binding()` hooks already invoked.

On the side, I also added a register recursion guard to the container's `register()` method and added an `isClass` check so that if you pass in a regular class without a static `register` method, it's automatically registered as a singleton (consistent with vCurrent). This is to fix a test that decided to fail suddenly.

## 👩‍💻 Reviewer Notes

- `$bindableHead`/`Tail`, ... and `$attachableHead`/`Tail` linked lists are merged into `$componentHead`/`Tail`. These are always used together anyway.
- Bindings are stored in `$bindingHead`/`Tail` so they can be invoked separately (earlier)
- Standalone lifecycle hook interfaces for binding and attaching merged into `IComponent`, removing about 10 exported interfaces and simplifying the API

A lot more little details to get this all to work but the changes all boil down to merging bindable/attachable into component, and adding bindings to the renderables separately.

This all appears to have resulted in a performance gain of around 3-4%

## 📑 Test Plan

- No functionality removed or added, so existing tests passing should be sufficient

## ⏭ Next Steps

Further optimize/unify lifecycle related types where it makes sense, before proceeding with the task-related work

### DIFF
--- a/packages/runtime-html/src/resources/custom-elements/compose.ts
+++ b/packages/runtime-html/src/resources/custom-elements/compose.ts
@@ -16,8 +16,7 @@ import {
   IViewFactory,
   LifecycleFlags,
   TargetedInstruction,
-  TemplateDefinition,
-  ICustomElementType
+  TemplateDefinition
 } from '@aurelia/runtime';
 import { createElement, RenderPlan } from '../../create-element';
 
@@ -25,21 +24,6 @@ const composeSource: ITemplateDefinition = {
   name: 'au-compose',
   containerless: true
 };
-
-
-export interface IViewportComponent<T extends INode = INode> {
-  viewport?: string;
-  component: string | Partial<ICustomElementType<T>>;
-}
-
-export interface Books extends ICustomElement<HTMLElement> {}
-export class Books implements Books {
-}
-
-const viewport: IViewportComponent = {
-  component: Books
-}
-
 
 const composeProps = ['subject', 'composing'];
 
--- a/packages/runtime-html/src/resources/custom-elements/compose.ts
+++ b/packages/runtime-html/src/resources/custom-elements/compose.ts
@@ -16,7 +16,8 @@ import {
   IViewFactory,
   LifecycleFlags,
   TargetedInstruction,
-  TemplateDefinition
+  TemplateDefinition,
+  ICustomElementType
 } from '@aurelia/runtime';
 import { createElement, RenderPlan } from '../../create-element';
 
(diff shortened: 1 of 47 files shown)


bigopon:
Nice explanation, I think
> Child components are still bound ...

Can be changed to
>Child components are still going to be bound ...

About the change in this PR, it's great.

fkleuver:
Yeah good point. It's difficult to speak in the appropriate past/future tense when that overlaps with both state and methods in the framework 🤣

fkleuver:
@bigopon I'm about done with this pass. Could you do another quick check and see if it all makes sense? In particular the merge of bindable/attachable into component

fkleuver:
Thanks @bigopon 

@EisenbergEffect I'm going to merge this in order to proceed but I would recommend you to take a look when you have time as there are some potential points for discussion about how to split things up and when the various hooks really need to be timed. `binding` now means "the bindings are bound but the children are not yet", instead of "nothing is bound yet".

This makes `binding` more like vCurrent's `bind` and less like `beforeBind` (one might ask what do you need `beforeBind` for when there is `created`, since those two come right after each other with no state changes in between on the component itself)

These changes are definitely not set in stone yet but in the overall context of vNext they seem to make things work better and as an added bonus there is one property less to render, improving proxy performance by about 10%

EisenbergEffect:
I'll review this in full tomorrow or Monday. Based on the explanation, it sounds like good changes though. It's funny that you mention `created` vs. `binding`. I was thinking of getting rid of `created` for the same reason but now that you've evolved `binding` to be a bit different, that actually gives `created` a reason to continue to exist.

packages/runtime-html/src/resources/custom-elements/compose.ts
```diff
@@ -25,6 +26,21 @@ const composeSource: ITemplateDefinition = {
   containerless: true
 };
 
+
+export interface IViewportComponent<T extends INode = INode> {
+  viewport?: string;
+  component: string | Partial<ICustomElementType<T>>;
+}
+
+export interface Books extends ICustomElement<HTMLElement> {}
```
EisenbergEffect: @fkleuver What's happening here?

fkleuver: @EisenbergEffect I was typing something out for @jwx on Discord and forgot to delete that snippet haha :) completely unrelated to this PR. But I think you're looking at a specific commit instead of the latest changes of the PR because I removed it later

EisenbergEffect: Oops! You're right! Thanks!

PR #401, opened 2019-01-21
Repeater keyed mode

fkleuver:
# Pull Request

## 📖 Description

This is an initial stab at keyed mode for the repeater using [longest increasing subsequence](https://en.wikipedia.org/wiki/Longest_increasing_subsequence) to find the optimal way to reorder nodes in the DOM.

I'm doing this before the rest of the lifecycle work because I'd like to take into account the need for lower-level `$nodes` operations when finalizing the lifecycles, as this is an important use case for many plugin authors need.

### 🎫 Issues

N/A

## 👩‍💻 Reviewer Notes

It would take some time to explain exactly what's happening. I will add more clarifying comments in the code at a later point in time when, paired with the lifecycle work, things are sure to actually work (and thus comments probably won't go outdated in 1-2 weeks).

Files of interest:
- repeat.ts (keyed logic)
- ast.ts (a few tweaks in binding behavior and ForOfStatement to make this work)
- collection observers (minor tweaks in tracking deleted items to make this work)
- "real" keyed mode not yet implemented (it tracks by index, which is guaranteed to be accurate with normal mutations but will cause everything to get removed and added again on assignment, so that's still a todo)

## 📑 Test Plan

- Duplicated most repeater unit and integration tests with keyed mode, to ensure all edge cases work as well with keyed mode as they do otherwise, adding up to a total of 15k additional tests.
- Deep lifecycle integration tests are still pending
- Ehh.. jsdom and firefox seem to give up at 76k tests. They just crash with memory exceptions. Ouch? I suppose I should go hunting for memory leaks in the tests

## ⏭ Next Steps

- Really the lifecycle this time (and lots of tests) :)
- A benchmark for this would also be interesting


packages/runtime/src/resources/custom-attributes/repeat.ts
```diff
@@ -184,6 +202,182 @@ export class Repeat<C extends ObservedCollection = IObservedArray, T extends INo
...
+              }
+            }
+          },
+          $nextMount: null
+        };
+
+        $lifecycle.enqueueMount(operation as IMountableComponent);
+
+        $lifecycle.endDetach(flags);
+        $lifecycle.endAttach(flags);
```
bigopon: These 2 lines are very nice part of Lifecycle 👍

packages/runtime/src/resources/custom-attributes/repeat.ts
```diff
@@ -184,6 +202,182 @@ export class Repeat<C extends ObservedCollection = IObservedArray, T extends INo
...
+        $lifecycle.endBind(flags);
+        if (views.length !== mapLen) {
+          // TODO: create error code and use reporter with more informative message
+          throw new Error(`viewsLen=${views.length}, mapLen=${mapLen}`);
+        }
+      }
+
+      if (this.$state & (State.isAttached | State.isAttaching)) {
+        const { location } = this;
+        const seq = longestIncreasingSubsequence(indexMap);
```
bigopon: Maybe have a comment here for why we need to perform this

packages/runtime/src/observation/map-observer.ts
```diff
@@ -64,7 +64,7 @@ const observe = {
       let i = 0;
       for (const entry of $this.keys()) {
         if (indexMap[i] > -1) {
-          indexMap.deletedItems.push(entry);
+          indexMap.deletedItems.push(indexMap[i]);
```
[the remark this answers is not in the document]
fkleuver: To get the index of the deleted item instead of the actual item. Easier to work with, performs better

PR #402, opened 2019-01-22
WIP: Fetch client migration

Vheissu:
# Pull Request

## 📖 Description

This is an initial migration of the existing Aurelia Fetch Client package. Minimal changes were required to port this. The migration work has exposed a need for an exportable DOM object which allows calling native API's such as CustomEvent which @fkleuver is aware of.

### 🎫 Issues

N/A

## 👩‍💻 Reviewer Notes

- The original Fetch client relied on an exported `DOM` class which allows methods only available in specific environments (browsers, non-browsers, native devices) to be safely called and used. Specifically, the creation and dispatching of events in a browser. Initially, an empty DOM object has been created which does nothing.
- I had to add `libs` to the `tsconfig.json` file for the runtime package because I needed interfaces only available by providing libs. If this is wrong, please change.
- The tests are currently written in the old Jasmine format and need to be changed.

## 📑 Test Plan

Get migrated tests to work.

## ⏭ Next Steps

Get tests working, create an example application in the `examples` folder showcasing the Fetch client in use.


Vheissu:
@fkleuver You mentioned on Discord you had a regex to convert the tests to work with our current test setup.

fkleuver:
@Vheissu Well, the tests are pretty much converted now but those retry tests fail miserably for some reason. I've commented them out for now, ran out of time (and patience) ;p

fkleuver:
Merging this with the TODO in mind that the last few tests need to be fixed, typings need to be cleaned up and probably refactored to async

PR #404, opened 2019-01-24
perf(all): pre-declare variables used in loops

fkleuver:
# Pull Request

## 📖 Description

Just a quick in-between PR to make a few small tweaks across the project that needed to happen for a while

## 👩‍💻 Reviewer Notes

- Replaced `delete obj.property` with `Reflect.deleteProperty(obj, property)` for consistency
- Pre-declare variables with `let` instead of declaring them with `const` on each iteration in loops
- Replace `Object.defineProperty` with `Reflect.defineProperty` for consistency
- Eliminate a few locally declared variables where they wouldn't have any perf benefit

### DIFF
--- a/packages/kernel/src/di.ts
+++ b/packages/kernel/src/di.ts
@@ -564,16 +564,23 @@ export class Container implements IContainer {
       // Most likely cause is trying to register a plain object that does not have a
       // register method and is not a class constructor
     }
+    let current: IRegistry | Record<string, IRegistry>;
+    let keys: string[];
+    let value: IRegistry;
+    let j: number;
+    let jj: number;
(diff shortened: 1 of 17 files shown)


fkleuver:
@BBosman Do these changes seem OK to you?

BBosman:
Looks OK. 👍

The *pre-declare variables* change will take some getting used to, as it makes the code look overly verbose, but I support the rationale behind the change.

fkleuver:
Thanks! What some others do is add those variables as optional parameters to the functions to save some lines, but I'm not sure if that necessarily makes things better. Food for thought.
Merging this now.

PR #407, opened 2019-01-26
ci: update package versions before merging

fkleuver:
# Pull Request

## 📖 Description

Update the versions before merging to the channel branch to keep the channel-specific branches in sync with what's published to npm, and also fix the nightly


fkleuver:
@EisenbergEffect Merging this right away since I just need to try and see if it works

PR #408, opened 2019-01-27
misc / patch mode

fkleuver:
# Pull Request

## 📖 Description

In this PR:
- A very misleading branch name (still no lifecycle/task rework)
- Typing fixes on the projectors
- `BindingStrategy` with the options `getterSetter`, `proxies` and `patch` (as well as `keyed` which can be combined with either of the 3). `patch` is the new one: instead of getters/setters or proxies, nothing is observed, and you manually call `$patch` which does a diff through the observer tree and updates everything whenever you tell it to. *it kinda works, but it needs more work* (just like proxies)
- To facilitate debugging sessions, I added a general `LiveTraceWriter` implementation to the `debug` package so you don't necessarily have to implement your own. Instead of nuking the console with json, it performs some basic stringification of arguments (lifecycle flags, gets resource/constructor names, etc). 

Tracer has some options that allow you to control which kinds of things are written to the console in a so-so level of granularity (I grouped things by what you might want to analyze as a whole, rather than putting 80 flags in there)
See a demonstration in `test/kitchen-sink`, but this is what it boils down to:
```ts
TraceConfiguration.register();
Tracer.enabled = true;
Tracer.enableLiveLogging({
  di: false,
  jit: false,
  rendering: false,
  lifecycle: true,
  binding: true,
  attaching: true,
  mounting: true,
  observation: true
});
```

This helped me a ton in figuring out where things went wrong / were missing in implementing the patch strategy and I hope it will help others as well.

- Two new pretty cool test apps:
  - `test/kitchen-sink`, is a very low level configured app with a little bit of this and that. It has no particular purpose other than being a playground for testing new and low level api's.
  - `test/1kcomponents`, is basically inferno's 1kcomponents ported to aurelia (see inferno's here: https://infernojs.github.io/inferno/1kcomponents/).

In trying to port from that React-style to Aurelia I unintentionally discovered a pretty interesting way to build an Aurelia app as a one-liner:

[test/1kcomponents/app.js](https://github.com/aurelia/aurelia/blob/419e2e3c9bec12ed3fdcc6864974ec5176394ad9/test/1kcomponents/app.js#L99)

Here is inferno's original:

https://github.com/infernojs/inferno/blob/master/docs/1kcomponents/app.js

- Several fixes in proxy observation
  - array indexers are now observed properly (when part of the binding expression) (observing them when they are NOT part of the binding expression is still a todo)
  - properties of objects in arrays are now also observed when changed from within a view model
 
- Fixes SVG analyzer integration again (for the demo)


fkleuver:
@BBosman could you take a quick look as well?

packages/aot/package.json
```diff
@@ -31,12 +31,13 @@
...
-    "test-firefox": "cross-env TS_NODE_PROJECT=\"../../scripts/tsconfig.json\" karma start ../../scripts/karma.conf.ts --package=aot --browsers=FirefoxHeadless --single-run --coverage",
-    "test:watch": "cross-env TS_NODE_PROJECT=\"../../scripts/tsconfig.json\" karma start ../../scripts/karma.conf.ts --package=aot --browsers=ChromeHeadlessOpt --coverage",
-    "test:watch:verbose": "cross-env TS_NODE_PROJECT=\"../../scripts/tsconfig.json\" karma start ../../scripts/karma.conf.ts --package=aot --browsers=ChromeHeadlessOpt --coverage --reporter=mocha",
-    "test:debugger": "cross-env TS_NODE_PROJECT=\"../../scripts/tsconfig.json\" karma start ../../scripts/karma.conf.ts --package=aot --browsers=ChromeDebugging",
-    "test:debugger:verbose": "cross-env TS_NODE_PROJECT=\"../../scripts/tsconfig.json\" karma start ../../scripts/karma.conf.ts --package=aot --browsers=ChromeDebugging --reporter=mocha",
+    "~test": "cross-env TS_NODE_PROJECT=\"../../scripts/tsconfig.json\" karma start ../../scripts/karma.conf.ts --package=aot",
```
BBosman: I'm missing similar changes in `fetch-client/package.json`?

packages/debug/src/configuration.ts
```diff
@@ -1,17 +1,18 @@
-import { IContainer, Reporter as RuntimeReporter, Tracer as RuntimeTracer } from '@aurelia/kernel';
+import { IContainer, Reporter as RuntimeReporter, Tracer as RuntimeTracer, Registration } from '@aurelia/kernel';
 import { enableImprovedExpressionDebugging } from './binding/unparser';
-import { Reporter, Tracer } from './reporter';
+import { Reporter } from './reporter';
+import { Tracer } from './tracer';
 
 export const DebugConfiguration = {
-  register(container: IContainer): void {
+  register(container?: IContainer): void {
```
BBosman: Why is `container` optional in both classes? That doesn't match the `IRegistry` interface.

packages/debug/src/configuration.ts
```diff
@@ -1,17 +1,18 @@
-import { IContainer, Reporter as RuntimeReporter, Tracer as RuntimeTracer } from '@aurelia/kernel';
+import { IContainer, Reporter as RuntimeReporter, Tracer as RuntimeTracer, Registration } from '@aurelia/kernel';
```
BBosman: `ordered-imports`

packages/debug/src/configuration.ts
fkleuver: The main reason for this is because this one doesn't *need* a container. Making it optional allows you to call the debug configuration before the container is created so you can trace the container creation itself

packages/runtime/src/observation/patch-properties.ts
```diff
@@ -0,0 +1,37 @@
+import { LifecycleFlags } from '../flags';
+import { AccessorOrObserver, CollectionKind, ICollectionObserver, IObserversLookup } from '../observation';
+
```
BBosman: As these functions are not marked `external` I would have expected them to be exported from `index.ts` as well.

fkleuver: good catch, they're supposed to be internal

packages/jit-html/test/integration/repeater-if-else.spec.ts
```diff
@@ -528,37 +594,52 @@ describe(spec, function() {
...
+                public display: boolean;
+              }
+            )
+          ]
+        },
+        class App {
+          public items: any[];
+          public display: boolean;
+          public count: number;
+          created() {
```
BBosman: `public created() {`

packages/jit-html/test/integration/kitchen-sink.spec.ts
```diff
@@ -0,0 +1,360 @@
...
+    expect(host.textContent).to.equal('Hello!');
+    au.stop();
+    expect(host.textContent).to.equal('');
+    au.start();
+    expect(host.textContent).to.equal('Hello!');
+    au.stop();
+    expect(host.textContent).to.equal('');
+  });
+
+
```
BBosman: `no-consecutive-blank-lines`

packages/jit-html/test/integration/kitchen-sink.spec.ts
```diff
@@ -0,0 +1,360 @@
...
+    au.app({ host, component });
+
+    au.start();
+
+    expect(host.textContent).to.equal('bar');
+  });
+});
+
+
```
BBosman: `no-consecutive-blank-lines`

packages/jit-html/test/integration/deep-bindables.spec.ts
```diff
@@ -23,189 +24,203 @@ describe(spec, function () {
...
+        disableTracing();
+      }
+
+      if ((strategy & BindingStrategy.proxies) > 0) {
+        expect(calls['SetterObserver'].length).to.equal(0, 'calls[\'SetterObserver\'].length');
+      } else {
+        expect(calls['ProxyObserver'].length).to.equal(0, 'calls[\'ProxyObserver\'].length');
+      }
+      const names = ['ProxyObserver', 'ProxySubscriberCollection', 'SelfObserver', 'SetterObserver'];
+      for (const {} of names) {
```
BBosman: This triggers `no-empty-destructuring`. Maybe comment it out for now as well?

packages/runtime/src/templating/lifecycle-render.ts
```diff
@@ -45,36 +45,49 @@ export interface ILifecycleRender {
 
 /** @internal */
 export function $hydrateAttribute(this: Writable<ICustomAttribute>, flags: LifecycleFlags, parentContext: IServiceLocator): void {
```
BBosman: `// tslint:disable-next-line:no-ignored-initial-value`

packages/runtime/src/templating/lifecycle-render.ts
```diff
@@ -45,36 +45,49 @@ export interface ILifecycleRender {
...
   if (this.$hooks & Hooks.hasCreated) {
-    this.created(flags);
+    bindingContext.created(flags);
   }
   if (Profiler.enabled) { leave(); }
   if (Tracer.enabled) { Tracer.leave(); }
 }
 
 /** @internal */
 export function $hydrateElement(this: Writable<ICustomElement>, flags: LifecycleFlags, parentContext: IServiceLocator, host: INode, options: IElementHydrationOptions = PLATFORM.emptyObject): void {
```
BBosman: `// tslint:disable-next-line:no-ignored-initial-value`

packages/runtime/src/resources/custom-attribute.ts
```diff
@@ -32,9 +32,11 @@ import {
 } from '../templating/lifecycle-attach';
 import {
   $bindAttribute,
+  $patch,
   $unbindAttribute
 } from '../templating/lifecycle-bind';
 import { $hydrateAttribute } from '../templating/lifecycle-render';
+import { Bindable } from '../templating/bindable';
```
BBosman: `ordered-imports`

packages/runtime/src/lifecycle.ts
```diff
@@ -553,9 +553,7 @@ export class Lifecycle implements ILifecycle {
   /** @internal */public $nextFlush: IChangeTracker;
   /** @internal */public flush: IChangeTracker['flush'];
   /** @internal */public $nextConnect: IConnectableBinding;
-  /** @internal */public connect: IConnectableBinding['connect'];
-  /** @internal */public $nextPatch: IConnectableBinding;
-  /** @internal */public patch: IConnectableBinding['patch'];
+  /** @internal */public connect: IConnectableBinding['connect'];;
```
BBosman: `;;`?

packages/runtime/src/templating/lifecycle-render.ts
fkleuver: huh what's that for?

packages/debug/src/tracer.ts
```diff
@@ -0,0 +1,554 @@
...
+};
+
+const toString = Object.prototype.toString;
+function flagsText(info: ITraceInfo, i: number = 0): string {
+  if (info.params.length > i) {
+    return stringifyLifecycleFlags(info.params[i] as LifecycleFlags);
+  }
+  return 'none';
+}
+function _ctorName(obj: Instance): string {
```
BBosman: Code Climate complains about this function with "Avoid too many return statements within this function." and I kinda agree.

Suggested alternative:
```ts
  let name: string;
  if (obj === undefined) {
    name = 'undefined';
  } else if (obj === null) {
    name = 'null';
  } else if (obj.constructor !== undefined) {
    if (obj.constructor.description) {
      name = `Resource{'${obj.constructor.description.name}'}`;
    } else {
      name = obj.constructor.name;
    }
  } else if (typeof obj === 'string') {
    name = `'${obj}'`;
  } else {
    name = toString.call(obj);
  }
  return name;
```

packages/debug/src/tracer.ts
```diff
@@ -0,0 +1,554 @@
+import { ILiveLoggingOptions, ITraceInfo, ITraceWriter, PLATFORM, Tracer as RuntimeTracer } from '@aurelia/kernel';
+import { IScope, LifecycleFlags } from '@aurelia/runtime';
+
+declare var console: {
```
BBosman: This is declared both here and in `reporter.ts`. Can it be consolidated somehow?

packages/runtime-html/src/observation/observer-locator.ts
```diff
@@ -118,7 +121,7 @@ export class TargetObserverLocator implements ITargetObserverLocator {
         const prefix = propertyName.slice(0, 5);
```
BBosman: This logic is repeated on #177. Abstract it to a private helper method?

BBosman:
Left a bunch of comments. Mostly nits.

packages/runtime/src/templating/lifecycle-render.ts
BBosman: It thinks you're re-assigning `this`.

packages/aot/package.json
BBosman: I noticed when reviewing the `fetch-client` PR, but made the assumption you'd fix it in here, so I'd made a mental node to check. 😄

packages/debug/src/configuration.ts
BBosman: Sounds like a useful code comment. 😉

packages/runtime/src/binding/binding.ts
```diff
@@ -193,19 +195,18 @@ export class Binding implements IPartialConnectableBinding {
...
     if (this.$state & State.isBound) {
       flags |= this.persistentFlags;
       this.sourceExpression.connect(flags | LifecycleFlags.mustEvaluate, this.$scope, this);
     }
     if (Tracer.enabled) { Tracer.leave(); }
   }
 
-  public patch(flags: LifecycleFlags): void {
-    if (Tracer.enabled) { Tracer.enter('Binding.patch', slice.call(arguments)); }
+  public $patch(flags: LifecycleFlags): void {
```
bigopon: Yeah this renaming is better I think

packages/runtime/src/flags.ts
```diff
@@ -14,6 +14,59 @@ export enum BindingMode {
   default  = 0b1000
 }
 
+export const enum BindingStrategy {
```
bigopon: What did you think of `ObservationStrategy`?

packages/runtime/src/observation/self-observer.ts
```diff
@@ -52,11 +56,14 @@ export class SelfObserver implements SelfObserver {
   public getValue(): unknown {
     return this.currentValue;
   }
+  public getValueDirect(): unknown {
```
bigopon: This may make changing strategy get things out of sync?

packages/runtime/src/observation/setter-observer.ts
```diff
@@ -44,4 +51,8 @@ export class SetterObserver implements SetterObserver {
       }
     }
   }
+  public $patch(flags: LifecycleFlags): void {
+    this.callSubscribers(this.obj[this.propertyKey], this.currentValue, this.persistentFlags | flags);
```
bigopon: In a future PR for a friendlier minification, can `this.obj`, `this.propertyKey` be stored locally?

packages/runtime/src/observation/target-observer.ts
```diff
@@ -53,6 +54,19 @@ function flush(this: BindingTargetAccessor, flags: LifecycleFlags): void {
   if (Tracer.enabled) { Tracer.leave(); }
 }
 
+function patch(this: BindingTargetAccessor, flags: LifecycleFlags): void {
+  if (Tracer.enabled) { Tracer.enter(this['constructor'].name, '$patch', slice.call(arguments)); }
+
+  const newValue = this.getValue();
+  if (this.currentValue !== newValue) {
```
bigopon: This makes me wonder if there could be a way to introduce update prevention callback in view model, instead of observer. It could be pretty powerful.

packages/runtime/src/observation/self-observer.ts
fkleuver: Maybe. I'm not even sure if we necessarily want to support changing strategy on a live component. Re-rendering is better.

packages/runtime/src/flags.ts
```diff
@@ -14,6 +14,59 @@ export enum BindingMode {
...
+   * Can be combined with either `getterSetter`, `proxies` or `patch`.
+   */
+  keyed        = 0b1000
+}
+
+const mandatoryStrategy = BindingStrategy.getterSetter | BindingStrategy.proxies | BindingStrategy.patch;
+
+export function ensureValidStrategy(strategy: BindingStrategy | null | undefined): BindingStrategy {
+  if ((strategy & mandatoryStrategy) === 0 || strategy === BindingStrategy.keyed) {
+    // TODO: probably want to validate that user isn't trying to mix proxy/patch, getterSetter/patch, getterSetter/proxy
```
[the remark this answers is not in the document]
fkleuver: Thing is we should throw if user is trying to mix strategy but we also need a flag guard to prevent one strategy from leaking into another component's lifecycle that has its own strategy defined. Currently there is no separation yet for this. You put a strategy on something, everything below it will be that. So that's a todo

packages/runtime/src/observation/target-observer.ts
fkleuver: Can you elaborate?

packages/runtime/src/templating/bindable.ts
```diff
@@ -71,6 +54,55 @@ export function bindable<T extends InstanceType<Constructable & Partial<WithBind
...
 }
 
-export type WithBindables = { bindables: Record<string, IBindableDescription> };
+interface IFluentBindableBuilder {
+  add(config: IBindableDescription): IFluentBindableBuilder;
+  add(propertyName: string): IFluentBindableBuilder;
+  get(): Record<string, IBindableDescription>;
+}
+
+export const Bindable = {
```
bigopon: This looks kinda cool ✨

packages/runtime/src/flags.ts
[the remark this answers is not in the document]
bigopon: I see, nice plan and cool explanation!

packages/runtime/src/observation/target-observer.ts
fkleuver: How does `shouldComponentUpdate` work then? User can return a boolean or? The observers generally already guard against unneeded updates

packages/runtime/src/observation/setter-observer.ts
fkleuver: Since it's internal API I'm actually thinking about using terser or AOT to simply shorten these property names on the build. Not sure yet

packages/runtime/src/flags.ts
fkleuver: I had that first but I changed my mind again. Honestly I can't decide.

packages/debug/src/tracer.ts
fkleuver: Actually yeah, thanks for pushing on that :) I'm calling the reporter from the tracer instead now

packages/runtime-html/src/observation/observer-locator.ts
fkleuver: Yeah I put it in a separate function now with a little cache while we're at it

PR #410, opened 2019-01-28
Prepare store

fkleuver:
# Pull Request

## 📖 Description

This PR prepares the `@aurelia/store` package for @Vheissu


fkleuver:
The stage is yours @Vheissu  :)

PR #412, opened 2019-01-30
Update (dev)Dependencies

BBosman:
# Pull Request

## 📖 Description

Another round of (dev)Dependency updates.

I noticed a lot of dependencies where out of date and a couple where even out of sync with the rest of the repository, so I went over all `package.json` files manually to bring them up to current.

### 🎫 Issues

n/a

## 👩‍💻 Reviewer Notes

Mostly patch and minor releases. Notable changes:
- Major bump for `event-stream`. They dropped `flatmap-stream` in favour of their own implementation.
- Major bump for `karma`. They dropped Support for Node 6.
- Major bump for `ts-node`. They removed JavaScript output cache.
- Major bump for `gulp`. We were already using `gulp@^4`, but missed a few.
- Fixed up a few `gulpfile.js` files due to upgrading to `gulp@^4`.
- Also updated all `package-lock.json` files.

## 📑 Test Plan

CircleCI and ran the "touched" `gulp` and `npm` tasks/scripts to verify they still worked

## ⏭ Next Steps

Try to not let them get out of sync again. 😉


EisenbergEffect:
@fkleuver If you want to install that bot for this repo that submits PRs for this stuff, please feel free to go ahead and do that.

fkleuver:
We kind of need the master branch tests to run on PRs as well because currently CircleCI will not catch problems with the browserstack e2e tests until they hit master. When everything can be verified on PR, we should be good to have a bot for this.

@BBosman Thanks for updating and bringing the deps in sync again. Just keep the above in mind please. It's really important to verify that the e2e still work as well. Last time we updated the deps we went 8 days without a nightly.

PR #413, opened 2019-02-01
chore(e2e): fix browserstack

fkleuver:
# Pull Request

## 📖 Description

This PR fixes the browserstack test job and runs it again on each PR, instead of only on merges to master and before releases, to prevent more regressions from slipping through.

Also fixes a regression in two-way binding I introduced in an earlier PR and only surfaced when running the e2e tests.

Lastly, pins ajv to 6.8.1 due to a regression in 6.9.0. This can probably be removed again soon, but I suppose it can't do any harm to have it around unnecessarily either, so I'll leave it for now.

## 👩‍💻 Reviewer Notes

- Removed the recursion guard from `subscriber-collection` as it turned out to be too naive (it would raise false positives on legitimate recursive scenarios)
- Added `allowPublishRoundtrip` flag to `value-attribute-observer`s `handleEvent` (which fixes the two-way binding regression)
- Disabled edge, FF and safari e2e tests because somehow webdriver managed to break for these 3

## 📑 Test Plan

- e2e tests are back in the game
- fixed some value attribute observer tests related to fixing the regression

## ⏭ Next Steps

- Look into SauceLabs or some other way to get the other browser tests working again
- Add more e2e tests (any volunteers?)


fkleuver:
@bigopon @EisenbergEffect there's not much to review here since the changes are only related to ci / browserstack, and for the rest it's just some removed code (no new code). Therefore I'm merging this right away so the nightly can stop failing.

PR #414, opened 2019-02-03
Route recognizer

fkleuver:
# Pull Request

## 📖 Description

This is an "as-is" port of `aurelia-path` and `route-recognizer`. Just to get the ball rolling. There is no integration yet.

## 👩‍💻 Reviewer Notes

- I added all `aurelia-path` functions even though only 2 of them are needed by `route-recognizer`. I added them as loose functions to the kernel so anything can use them, and they'll be tree-shaken if not used.
- Made an attempt to strongly type the `QueryParams` object from `aurelia-path` which proved to be a challenge.. @EisenbergEffect might want to chime in here as I believe you wrote them originally?
- `route-recognizer` is functionally still the same, which means it's not *entirely* ripe for use by vNext just yet (see next steps). The changes are:
  - Replaced interfaces with classes (and objects with constructor calls) to make it all a bit more explicit
  - Inlined all loose functions to reduce the total amount of code and make it just a little bit easier to follow what happens from start to end. But also to make it more refactor-friendly, which is the next step.
- @jwx perhaps some of these aurelia-path utilities with regards to parsing query params might come in handy already?

## 📑 Test Plan

Migrated the tests from vCurrent and just cleaned things up a bit in terms of typings and linting. No additional tests added (that's a todo, especially for route-recognizer)

## ⏭ Next Steps

- Refactor `route-recognizer` to allow "navigate to route" by path in addition to name (tiny change, but needs lots of tests)
- Make a start with configured routes and see how/where that could best be integrated with dynamic routes

### DIFF
--- a/packages/kernel/src/path.ts
+++ b/packages/kernel/src/path.ts
@@ -1,0 +1,269 @@
+function trimDots(ary: string[]): void {
+  const len = ary.length;
+  let i = 0;
+  let part: string;
+  for (; i < len; ++i) {
+    part = ary[i];
+    if (part === '.') {
+      ary.splice(i, 1);
+      i -= 1;
+    } else if (part === '..') {
+      // If at the start, or previous value is still ..,
+      // keep them so that when converted to a path it may
+      // still work when converted to a path, even though
+      // as an ID it is less than ideal. In larger point
+      // releases, may be better to just kick out an error.
+      if (i === 0 || (i === 1 && ary[2] === '..') || ary[i - 1] === '..') {
+        continue;
+      }
+      if (i > 0) {
+        ary.splice(i - 1, 2);
+        i -= 2;
+      }
+    }
+  }
+}
+
+/**
+ * Calculates a path relative to a file.
+ *
+ * @param name The relative path.
+ * @param file The file path.
+ * @return The calculated path.
+ */
+export function relativeToFile(name: string, file: string): string {
+  const fileParts = file && file.split('/');
+  const nameParts = name.trim().split('/');
+
+  if (nameParts[0].charAt(0) === '.' && fileParts) {
+    //Convert file to array, and lop off the last part,
+    //so that . matches that 'directory' and not name of the file's
+    //module. For instance, file of 'one/two/three', maps to
+    //'one/two/three.js', but we want the directory, 'one/two' for
+    //this normalization.
+    const normalizedBaseParts = fileParts.slice(0, fileParts.length - 1);
+    nameParts.unshift(...normalizedBaseParts);
+  }
+
+  trimDots(nameParts);
+
+  return nameParts.join('/');
+}
+
+/**
+ * Joins two paths.
+ *
+ * @param path1 The first path.
+ * @param path2 The second path.
+ * @return The joined path.
+ */
+export function join(path1: string, path2: string): string {
+  if (!path1) {
+    return path2;
+  }
+  if (!path2) {
+    return path1;
+  }
+  const schemeMatch = path1.match(/^([^/]*?:)\//);
+  const scheme = (schemeMatch && schemeMatch.length > 0) ? schemeMatch[1] : '';
+  path1 = path1.slice(scheme.length);
+
+  let urlPrefix: string;
+  if (path1.indexOf('///') === 0 && scheme === 'file:') {
+    urlPrefix = '///';
+  } else if (path1.indexOf('//') === 0) {
+    urlPrefix = '//';
+  } else if (path1.indexOf('/') === 0) {
+    urlPrefix = '/';
+  } else {
+    urlPrefix = '';
+  }
+
+  const trailingSlash = path2.slice(-1) === '/' ? '/' : '';
+
+  const url1 = path1.split('/');
+  const url2 = path2.split('/');
+  const url3 = [];
+
+  for (let i = 0, ii = url1.length; i < ii; ++i) {
+    if (url1[i] === '..') {
+      url3.pop();
+    } else if (url1[i] !== '.' && url1[i] !== '') {
+      url3.push(url1[i]);
+    }
+  }
+
+  for (let i = 0, ii = url2.length; i < ii; ++i) {
+    if (url2[i] === '..') {
+      url3.pop();
+    } else if (url2[i] !== '.' && url2[i] !== '') {
+      url3.push(url2[i]);
+    }
+  }
+
+  return scheme + urlPrefix + url3.join('/') + trailingSlash;
+}
+
+const encode = encodeURIComponent;
+const encodeKey = (k: string) => encode(k).replace('%24', '$');
+
+// these types are reverse-engineered from aurelia-path.. don't shoot the messenger..
+type Scalar = null | undefined | boolean | number | string;
+export interface IQueryParams {
+  [key: string]: Scalar | this | (string | string[] | this)[];
+}
+type ComplexQueryParams = Scalar | IQueryParams | (string | string[] | IQueryParams)[];
+type SimpleQueryParams = string | IQueryParams | (string | IQueryParams)[];
+
+/**
+ * Recursively builds part of query string for parameter.
+ *
+ * @param key Parameter name for query string.
+ * @param value Parameter value to deserialize.
+ * @param traditional Boolean Use the old URI template standard (RFC6570)
+ * @return Array with serialized parameter(s)
+ */
+function buildParam(key: string, value: ComplexQueryParams, traditional?: boolean): string[] {
+  let result = [];
+  if (value === null || value === undefined) {
+    return result;
+  }
+  if (Array.isArray(value)) {
+    for (let i = 0, l = value.length; i < l; i++) {
+      if (traditional) {
+        result.push(`${encodeKey(key)}=${encode(value[i] as string)}`);
+      } else {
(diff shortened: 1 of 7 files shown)


packages/kernel/src/path.ts
```diff
@@ -0,0 +1,269 @@
...
+  }
+
+  return scheme + urlPrefix + url3.join('/') + trailingSlash;
+}
+
+const encode = encodeURIComponent;
+const encodeKey = (k: string) => encode(k).replace('%24', '$');
+
+// these types are reverse-engineered from aurelia-path.. don't shoot the messenger..
+type Scalar = null | undefined | boolean | number | string;
```
BBosman: Isn't this the same as `Primitive`?

packages/kernel/src/path.ts
```diff
@@ -0,0 +1,269 @@
...
+/**
+ * Recursively builds part of query string for parameter.
+ *
+ * @param key Parameter name for query string.
+ * @param value Parameter value to deserialize.
+ * @param traditional Boolean Use the old URI template standard (RFC6570)
+ * @return Array with serialized parameter(s)
+ */
+function buildParam(key: string, value: ComplexQueryParams, traditional?: boolean): string[] {
+  let result = [];
```
BBosman: Implicit `any`, so: `let result: string[] = [];`

packages/kernel/src/path.ts
```diff
@@ -0,0 +1,269 @@
...
+
+/**
+ * Generate a query string from an object.
+ *
+ * @param params Object containing the keys and values to be used.
+ * @param traditional Boolean Use the old URI template standard (RFC6570)
+ * @returns The generated query string, excluding leading '?'.
+ */
+export function buildQueryString(params?: IQueryParams, traditional?: boolean): string {
+  let pairs = [];
```
BBosman: Implicit `any`, so: `let pairs: string[] = [];`

bigopon:
> Refactor route-recognizer to allow "navigate to route" by path in addition to name (tiny change, but needs lots of tests)

Did you mean merging the 2 APIs: `navigate` (for path) and `navigateToRoute` (for name) together? If so, I'm not sure if it's doable, as for simple cases, path and name are indistinguishable.

packages/kernel/test/path.spec.ts
```diff
@@ -0,0 +1,277 @@
...
+describe('relativeToFile', () => {
+  it('can make a dot path relative to a simple file', () => {
+    const file = 'some/file.html';
+    const path = './other/module';
+
+    expect(relativeToFile(path, file)).to.equal('some/other/module');
+  });
+
+  it('can make a dot path relative to an absolute file', () => {
+    const file = 'http://durandal.io/some/file.html';
```
BBosman: Maybe do a search and replace in this file for `durandal` -> `aurelia`?

packages/router/src/route-recognizer.ts
```diff
@@ -0,0 +1,593 @@
...
+    return '';
+  }
+
+  public generate(params: Record<string, string>, consumed: Record<string, boolean>): string {
+    return '';
+  }
+}
+export type Segment = StaticSegment | DynamicSegment | StarSegment | EpsilonSegment;
+
+
```
BBosman: `no-consecutive-blank-lines`

packages/kernel/src/path.ts
fkleuver: `Primitive` includes `symbol`, and that's that one type that can't join the party

packages/router/src/route-recognizer.ts
```diff
@@ -0,0 +1,593 @@
...
+      } else {
+        // Try to parse a star segment *whatever
+        match = part.match(/^\*(.+)$/);
+        if (match) {
+          segments.push(segment = new StarSegment(match[1]));
+          names.push(match[1]);
+          types.stars++;
+        } else if (part === '') {
+          segments.push(segment = new EpsilonSegment());
+          continue;
```
BBosman: Unnecassary `continue;` imho, It also triggers a `no-dead-store` on line 312.

packages/router/test/unit/route-recognizer.spec.ts
```diff
@@ -0,0 +1,275 @@
...
+      Object.keys(result[0].params).forEach((property) => {
+        if (params[property] === undefined) {
+          return;
+        }
+        expect(result[0].params[property].toUpperCase()).to.equal(params[property].toUpperCase());
+      });
+    });
+
+    it(`should generate - ${routeSpec.t}`, () => {
+      const { route, path, isDynamic, params } = routeSpec;
```
BBosman: `const { route, path, params } = routeSpec;`

packages/router/test/util.ts
```diff
@@ -0,0 +1,17 @@
+import {
```
BBosman: Why all the unused `import` statements?

fkleuver:
@bigopon I was thinking the path stuff could come in handy when we get around to implementing the various loader plugins that actually need to deal with file paths. Not sure where else to put these sorts of utilities then. Could take them out and postpone adding them back in to when we actually need it. Could also just leave them and decide before beta what we clean up etc. Either way is fine by me

packages/router/test/util.ts
fkleuver: I had some test utilities in there which I later removed, forgot to clean this up. Fixing.. :)

packages/kernel/test/path.spec.ts
fkleuver: yep

bigopon:
@fkleuver Yep, was just stating the obvious fact. For now, it's more important to progress and port over everything, so it's good where it is going to.

packages/router/src/route-recognizer.ts
```diff
@@ -0,0 +1,593 @@
...
+  public add(route: ConfigurableRoute|ConfigurableRoute[]): State | undefined {
+    if (Array.isArray(route)) {
+      route.forEach(r => {
+        this.add(r);
+      });
+      return;
+    }
+
+    let currentState = this.rootState;
+    const skippableStates = [];
```
BBosman: Implicit `any`, so: `const skippableStates: State[] = [];`

packages/router/src/route-recognizer.ts
```diff
@@ -0,0 +1,593 @@
...
+    let regex = '^';
+    const types = new TypesRecord();
+    const names = [];
```
BBosman: `const names: string[] = [];`

packages/router/src/route-recognizer.ts
```diff
@@ -0,0 +1,593 @@
...
+
+    let pathLen = normalizedPath.length;
+    if (pathLen > 1 && normalizedPath.charAt(pathLen - 1) === '/') {
+      normalizedPath = normalizedPath.slice(0, -1);
+      isSlashDropped = true;
+      --pathLen;
+    }
+
+    for (let i = 0; i < pathLen; ++i) {
+      const nextStates = [];
```
BBosman: `const nextStates: State[] = [];`

packages/router/src/route-recognizer.ts
fkleuver: Removing it doesn't break the tests for some reason, but route-recognizer wasn't particularly well tested in the first place. It is there to preserve the original flow. I'd want the tests to be a bit more thorough before changing the flow like that. I'll remove the assignment though, that's a good point.

BBosman:
Just a few minor nits.

Also, some `cognitive-complexity` violations, but as it's an "as-is" port I think those can be put on the back-burner for now.

fkleuver:
> > Refactor route-recognizer to allow "navigate to route" by path in addition to name (tiny change, but needs lots of tests)

> Did you mean merging the 2 APIs: navigate (for path) and navigateToRoute (for name) together? If so, I'm not sure if it's doable, as for simple cases, path and name are indistinguishable.

@bigopon No, I don't mean merging them. What I mean is letting the `name` of a route default to its `route` property, so you don't need to explicitly name your routes just to be able to navigate to them by "name". 
You get alot of this: `config: { route: 'foo', name: 'foo' }` just to be able to do `router.navigateToRoute('foo')` / route-href. We largely don't have this issue anymore with dynamic routing, but inherently it's still there with configured routes. I think it's overall a good default that would flow naturally with the way dynamic routes work.

packages/router/src/route-recognizer.ts
BBosman: Line 323 `expects` segment to have a value, so removing the assignment here can cause unintended side effects.

Just removing the `continue` doesn't change the flow does it?

fkleuver:
Thanks a ton @BBosman, lots of typings things I actually missed because there was just so much of it missing :)
I did actually increase the cognitive complexity by merging/inlining what was spread out across some separate functions before so that's a TODO to improve that again, but I intentionally wanted to have it on a big blank pile making it easier to see the whole picture, etc. I think (hope) you'll agree compared to the vCurrent route-recognizer this is a fair bit easier to follow :)

BBosman:
@fkleuver Agreed. Left a comment on your comment, but otherwise it's good to go from my perspective.

fkleuver:
As discussed @jwx merging this now, so the scope of follow up integration work can be limited to just that integration work

PR #416, opened 2019-02-04
Typing / Binding fixes

fkleuver:
# Pull Request

## 📖 Description

A small variety of fixes for things I ran into when trying to port `aurelia-validation` as well as based on feedback from @bigopon 

- Improved the typings for resource types: the returned value from `.define` on resources properly have both the static and instance properties of the resource type as well as the passed-in class merged (this eliminates the need for several casts, esp. in unit tests where these define calls are inlined a lot)
- Cleaned up a few unused flags and the connect queue lifecycle as it's currently not being used anywhere (queueing mechanism will come back in a different form)
- Fixes two-way binding between multiple custom elements
- Adds a change loopback guard that prevents `from-view` flowing back to the target and vice versa
- AST tries to get a reference to the observers directly and call `setValue` if possible, in order to preserve the flags
- Added an extra build script that builds the tests to catch test typing errors (not yet fully working due to type reference weirdness, still a semi WIP)

## 👩‍💻 Reviewer Notes

Files of interest:
- Resource type fix: https://github.com/aurelia/aurelia/pull/416/files#diff-86d414e957cf8bd4790349db418fd4b9R82
- AST changes: https://github.com/aurelia/aurelia/pull/416/files#diff-596e652716d84f1a9398b642928c418bR426
- Change loopback guard:
  - https://github.com/aurelia/aurelia/pull/416/files#diff-462d7124d7584037488f5af73fa5ab7aR74
  - https://github.com/aurelia/aurelia/pull/416/files#diff-f558c2b49aac6aa3bdfd2a7f317a1ed7R56
  - https://github.com/aurelia/aurelia/pull/416/files#diff-e9929181314489fffe6148fb5895c837R147

## 📑 Test Plan

- Fixed binding tests, noticed that indeed a lot of `callCount === 2` had to be changed to `callCount === 1` on `handleChanges`, so the guards are working
- Added only one simple `two-way` test as @bigopon said he had some specific ideas in mind for testing two-way

## ⏭ Next Steps

- Add more tests
- Look for more fixes/optimizations along the way while working on `validation`
- Add `primaryProperty` support for custom attributes


packages/runtime/src/observation/subscriber-collection.ts
```diff
@@ -116,26 +117,80 @@ function callPropertySubscribers(this: ISubscriberCollection<MutationKind.instan
...
+    subscriber.handleChange(newValue, previousValue, flags);
+  } else if ((ownFlags & LF.isPublishing) === 0) {
+    publisher[subscriber.id] = ownFlags | LF.isPublishing;
+
+    // Note: if the update flags for both directions are set, that means an observer's callSubscribers caused the update direction to switch
+    // back to the origin of the change.
+    // With this heuristic we stop this roundtrip a little earlier than vCurrent does (where the target or source is evaluated
+    // and compared again) and effectively make this a "purer" one-way update flow that prevents observable side-effects from
+    // flowing back the opposite direction.
+    if (((flags | ownFlags) & LF.update) === LF.update) {
```
bigopon: The comment is "both direction", but this check is about either direction, is this intentional?

fkleuver:
```ts
  update                    = 0b11,
  updateTargetInstance      = 0b01,
  updateSourceExpression    = 0b10,
```

If that expression evaluates to true, that means `ownFlags` and `flags` each have a different update direction, for example `callSubscribers` was called with `updateSourceExpression` but the stored flag is `updateTargetInstance`.

Crash course:
```ts
(((0b01 | 0b01) & 0b11) === 0b11) // false, it's === 0b01
(((0b10 | 0b10) & 0b11) === 0b11) // false, it's === 0b10
(((0b10 | 0b01) & 0b11) === 0b11) // true
(((0b01 | 0b10) & 0b11) === 0b11) // true
(((0b01 | 0b11) & 0b11) === 0b11) // true
```

packages/runtime/src/observation/subscriber-collection.ts
```diff
@@ -116,26 +117,80 @@ function callPropertySubscribers(this: ISubscriberCollection<MutationKind.instan
...
+    // We're not really expecting this to ever happen with the existing guards
+    // in place, but if for whatever reason a binding or observer manages to cause a
+    // (potential) infinite update loop, throwing is certainly preferable over a hanging app.
+
+    // Of course, false positives are possible and throwing this error also helps
+    // us weed those out and detect+handle them appropriately.
+    throw Reporter.error(650);
+    // TODO: create error code
+
+    // TODO: remove the isPublishing flag assigment and check from the bundled output, since it adds significant overhead
```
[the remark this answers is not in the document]
fkleuver: It means `index.es6.js`, `index.umd.js`, etc. The files that are published to npm and end up in end user app

packages/jit-html/test/integration/binding-commands.spec.ts
```diff
@@ -223,4 +223,60 @@ describe('binding-commands', () => {
       tearDown(au, lifecycle, host);
     }
   });
+
+  // two-way bindings between multiple custom elements
+  it('100', function () {
+    const App = CE.define(
+      {
+        name: 'app',
+        template: '${a}<foo b.two-way="a"></foo>',
```
bigopon: There should be a loop to test `bind` + default two way and from view here too. If you are busy with other stuff, maybe add a todo? I can enhance this area.

packages/runtime/src/binding/binding.ts
```diff
@@ -57,32 +58,27 @@ export class Binding implements IPartialConnectableBinding {
...
   public updateSource(value: unknown, flags: LifecycleFlags): void {
     flags |= this.persistentFlags;
-    this.sourceExpression.assign(flags | LifecycleFlags.updateSourceExpression, this.$scope, this.locator, value);
+    this.sourceExpression.assign(flags, this.$scope, this.locator, value);
   }
 
   public handleChange(newValue: unknown, _previousValue: unknown, flags: LifecycleFlags): void {
     if (Tracer.enabled) { Tracer.enter('Binding', 'handleChange', slice.call(arguments)); }
-    if (!(this.$state & State.isBound)) {
+    if ((this.$state & State.isBound) === 0) {
```
bigopon: 👍

packages/kernel/src/interfaces.ts
```diff
@@ -75,6 +75,15 @@ export type Class<T, C = IIndexable> = C & {
   new(...args: unknown[]): T;
 };
 
+// For resources, we want the 'constructor' property to remain on the instance type but we need to do that
+// with a separate type from Class, since that one is used for other things where this constructor property
+// would break the typings.
+// So, in lack of a better name.. we probably need to clean this up, but this is how it works for now.
+export type ConstructableClass<T, C = IIndexable> = C & {
```
bigopon: This is probably better way to handle this, it's been bugging me for sometimes

packages/runtime/src/binding/binding.ts
```diff
@@ -156,6 +153,7 @@ export class Binding implements IPartialConnectableBinding {
     }
     if (this.mode & fromView) {
       targetObserver.subscribe(this);
+      targetObserver[this.id] |= LifecycleFlags.updateSourceExpression;
```
bigopon: This feels better to me, no longer needs to "randomly" pass a flag in `setValue`, which is pretty off

PR #403, opened 2019-01-23
Dynamic routing: improved cancellation and queued browser history

jwx:
# Pull Request

## 📖 Description

Improving the cancellation routine and adding a queued browser history layer who's main job is to make sure the browser isn't throttling History events and actions.

- refactor(router): add container to viewport
- refactor(router): update closest & add container influence on load
- refactor(router): change entering timing & add viewport render and state
- chore(router): delete unused files and code & reorganize test components
- chore(router): remove old view code
- refactor(router): add cancel navigation & add waits to test app
- refactor(router): add navigation status to viewport
- feat(router): add api for get all viewports to router
- feat(router): add first entry flag to history browser
- chore(router): fix rebase issues
- feat(router): add queued browser history
- refactor(router): simplify tick in queued browser history
- chore(router): fix linting errors
- chore(router): housekeeping
- refactor(router): add queued browser history and appropriate awaits
- test(router): add some queued browser history tests

### 🎫 Issues

Related to https://github.com/aurelia/aurelia/issues/307, https://github.com/aurelia/aurelia/issues/369 and https://github.com/aurelia/aurelia/issues/367.

## 👩‍💻 Reviewer Notes

A demo with some explanations can be found here: https://dist-c255ljxb5.now.sh

## 📑 Test Plan

- Added some basic tests of queued browser history

## ⏭ Next Steps

- Increase test density
- Clean up code
- Re-visit scoped viewports
- Add a before navigation hook with accompanying redirect

Issue #192, opened 2018-09-25
Error Codes


EisenbergEffect:
For vNext, we'd like to have an *amazing* story around error messages. This involves:

* Verbose messages with guidance to help developers resolve the problem.
* Error codes that are easy to search and lookup in documentation.
* Error code ranges to classify and organize codes (inspired by HTTP codes).

EisenbergEffect:
From @fkleuver:

...we should probably discuss this in a separate issue, but overall I think we should at least have the equivalents of:

1xx/2xx for verbose logging (can be a single group)
4xx invalid user input like invalid DI key, nonexisting route, missing properties
5xx deep binding/templating errors such as undefined/null references that should be guarded

fkleuver:
Just some thoughts:

- Compile-time errors: AOT cannot compile/process, problem must be fixed by user
- Compile-time warnings: AOT can compile/process, but detected something that could cause runtime errors
- Compile-time tracing: detailed insights in the AOT resolve/transform/emit phases
- Runtime errors: violations of preconditions; invocations that are always invalid regardless of app state, such as passing `null` or `undefined` to `container.get()`, omitting the `name` property to various decorators, etc. These are detected and thrown by validation functions.
- Runtime errors: violations of invariants: same as preconditions, but these things are not checked/validated beforehand for performance reasons, usually things that "drop down" in an unhandled state such as invalid target/source context in binding, unresolvable observables, etc
- Runtime warnings: based on user configuration, certain things may warn (or even throw) where they would not by default do so, such as a warning when dirty checking is used somewhere.
- Runtime tracing: detailed logging of compiler / parser / DI / renderer flow by (probably AOT-injected) entry and exit logging methods on functions
- Unhandled exceptions: exceptions thrown by user code during lifecycles (either via rejected promises or synchronously) that bubble up to the framework code

timfish:
+1 for error codes. TypeScript does this which makes searching for help much easier. It also helped them do things like analyse which codes were searched for most and selectively improve the error messages to help guide users.

bigopon:
This is now resolved by #1226 The actual work of producing error code should be tracked else where.
Doc: https://docs.aurelia.io/reference/distributable
Error references https://docs.aurelia.io/reference/error-messages

PR #417, opened 2019-02-04
Dynamic routing: stateful viewports

jwx:
# Pull Request

## 📖 Description

This PR adds stateful component caching for viewports. (And also contains some refactoring.)

### 🎫 Issues

Related to https://github.com/aurelia/aurelia/issues/307, https://github.com/aurelia/aurelia/issues/369 and https://github.com/aurelia/aurelia/issues/367. Includes https://github.com/aurelia/aurelia/pull/403.

## 👩‍💻 Reviewer Notes

A demo with some explanations can be found here: https://dist-dgmnotq6y.now.sh

## 📑 Test Plan

- Added some basic tests of stateful components

## ⏭ Next Steps

- Increase test density
- Clean up code
- Re-visit scoped viewports


packages/router/src/history-browser.ts
```diff
@@ -1,7 +1,9 @@
+import { QueuedBrowserHistory } from './queued-browser-history';
 export interface IHistoryEntry {
   path: string;
   fullStatePath: string;
   index?: number;
+  firstEntry?: boolean;
```
davismj: is this not implicit in index?

packages/router/src/history-browser.ts
```diff
@@ -80,116 +83,125 @@ export class HistoryBrowser {
...
+    // let { pathname, search, hash } = this.location;
+    // state = this.history.state;
+    // await this.history.replaceState(state, null, `${pathname}${search}${newHash}`);
     // tslint:disable-next-line:promise-must-complete
     let wait = new Promise((resolve, reject): void => {
       this.ignorePathChange = resolve;
     });
-    this.history.go(-1);
+    await this.history.go(-1);
     await wait;
```
davismj: i'm never going to be able to figure out what this line is doing

packages/router/src/queued-browser-history.ts
```diff
@@ -0,0 +1,118 @@
...
+}
+
+export class QueuedBrowserHistory implements QueuedBrowserHistory {
+  public window: Window;
+  public history: History;
+
+  private readonly queue: QueueItem[];
+  private isActive: boolean;
+  private processingItem: QueueItem;
+  private callback: (ev?: PopStateEvent) => void;
```
davismj: I think we have the event aggregator available, lets use it instead of the callback pattern.

packages/router/src/queued-browser-history.ts
```diff
@@ -0,0 +1,118 @@
...
+    const promise: Promise<void> = new Promise((resolve) => {
+      _resolve = resolve;
+    });
+    this.queue.push({
+      object: object,
+      method: method,
+      parameters: parameters,
+      resolve: _resolve,
+    });
+    this.dequeue().catch(error => { throw error; });
```
davismj: Is there any function to a catch that throws?

packages/router/src/queued-browser-history.ts
```diff
@@ -0,0 +1,118 @@
...
+    return promise;
+  }
+
+  private async dequeue(delta?: number): Promise<void> {
+    if (!this.queue.length || this.processingItem !== null) {
+      return;
+    }
+    this.processingItem = this.queue.shift();
+    const method = this.processingItem.object[this.processingItem.method];
```
davismj: I have no idea what object or method are.

packages/router/src/viewport.ts
```diff
@@ -145,61 +123,67 @@ export class Viewport {
...
     }
 
-    const component: IRouteableCustomElement = this.component;
+    const component: IRouteableCustomElement = this.content.component;
     if (!component.canLeave) {
       return true;
     }
     // tslint:disable-next-line:no-console
-    console.log('viewport canLeave', component.canLeave(this.instruction, this.nextInstruction));
+    console.log('viewport canLeave', component.canLeave(this.content.instruction, this.nextContent.instruction));
```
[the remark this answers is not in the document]
BBosman: And to not spam the entire PR, this goes for the other `console.log` statements as well. 😉

packages/router/src/viewport.ts
```diff
@@ -210,80 +194,73 @@ export class Viewport {
...
   public finalizeContentChange(): void {
     this.previousViewportState = null;
   }
-  // TODO: Call this on cancel
-  public async restorePreviousContent(): Promise<void> {
-    if (this.entered) {
-      await this.nextComponent.leave();
-    }
+  public async abortContentChange(): Promise<void> {
+    this.nextContent.freeContent(this.element, this.options.stateful).catch(error => { throw error; });
```
BBosman: `await this.nextContent.freeContent(this.element, this.options.stateful);`?

packages/router/test/e2e/doc-example/src/app.ts
```diff
@@ -1,24 +1,40 @@
...
     <au-viewport name="lists" used-by="authors,books" default="authors"></au-viewport>
-    <au-viewport name="content" default="about"></au-viewport>
+    <au-viewport name="content" stateful default="about"></au-viewport>
     <au-viewport name="chat" used-by="chat" no-link no-history></au-viewport>
   </div>
 </template>
 ` })
 export class App {
-  constructor(private readonly router: Router) {
+  constructor(private readonly router: Router, authorsRepository: AuthorsRepository, private state: State) {
```
BBosman: `private readonly state: State`

packages/router/test/e2e/doc-example/src/components/about.ts
```diff
@@ -1,9 +1,45 @@
...
+In other words
+In other words, I scroll you
+</pre></div>
+<br>
+<input>
 </template>` })
-export class About {}
+@inject(State)
+export class About {
+  constructor(private state: State) { }
```
BBosman: `private readonly state: State`

packages/router/test/e2e/doc-example/src/components/authors/about-authors.ts
```diff
@@ -1,7 +1,20 @@
...
+<div>
 <h3>About authors</h3>
 <p>Authors write books. Books are good for you. End of story.</p>
+<div class="scrollbox">Space out!</div>
+</div>
 </template>` })
-export class AboutAuthors {}
+@inject(State)
+export class AboutAuthors {
+  constructor(private state: State) { }
```
BBosman: `private readonly state: State`

packages/router/test/e2e/doc-example/src/components/authors/author-details.ts
```diff
@@ -1,19 +1,31 @@
...
+<div class="scrollbox">All about the space, about the space, no truncate...</div>
+</div>
 </template>` })
+@inject(AuthorsRepository, State)
 export class AuthorDetails {
   public static parameters = ['id'];
 
   public author = {};
-  constructor(private readonly authorsRepository: AuthorsRepository) { }
+  constructor(private readonly authorsRepository: AuthorsRepository, private state: State) { }
```
BBosman: `private readonly state: State`

packages/router/test/e2e/doc-example/src/components/authors/author.ts
```diff
@@ -13,18 +15,25 @@ import { Information } from './information';
...
+@inject(Router, AuthorsRepository, State)
 export class Author {
   public static parameters = ['id'];
 
   public author: { id: number };
 
-  constructor(private readonly router: Router, private readonly authorsRepository: AuthorsRepository) { }
+  public hideTabs: boolean = false;
+
+  constructor(private readonly router: Router, private readonly authorsRepository: AuthorsRepository, private state: State) { }
```
BBosman: `private readonly state: State`

packages/router/test/e2e/doc-example/src/components/authors/authors.ts
```diff
@@ -12,10 +14,14 @@ import { AuthorsRepository } from '../../repositories/authors';
   </li>
 </ul>
 </template>` })
-@inject(AuthorsRepository)
+@inject(AuthorsRepository, State)
 export class Authors {
-  constructor(private readonly authorsRepository: AuthorsRepository) { }
+  constructor(private readonly authorsRepository: AuthorsRepository, private state: State) { }
```
BBosman: `private readonly state: State`

packages/router/src/queued-browser-history.ts
```diff
@@ -0,0 +1,118 @@
...
+    this.history = window.history;
+    this.queue = [];
+    this.isActive = false;
+    this.processingItem = null;
+    this.callback = null;
+  }
+
+  public activate(callback: (ev?: PopStateEvent) => void): void {
+    if (this.isActive) {
+      throw new Error('Queued browser history has already been activated.');
```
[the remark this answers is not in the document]
jwx: I should probably set up the error codes too then (if nothing else for testing purposes). What's the deal, @fkleuver? Can I just grab any codes I like?

packages/router/src/history-browser.ts
```diff
@@ -80,116 +83,125 @@ export class HistoryBrowser {
     this.isActive = true;
     this.options = { ...options };
 
-    window.addEventListener('popstate', this.pathChanged);
+    this.history.activate(this.pathChanged);
 
     return Promise.resolve().then(() => {
```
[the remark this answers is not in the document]
fkleuver: @davismj I'm on board with `await PLATFORM.nextTick()` or a similar construct as it allows us to redefine the temporal aspect of a "tick" as well as potentially do some deeper assertions in unit tests.
But that's not something for @jwx to solve in the router, that'd be a framework-wide find/replace I might do when I finally get all the lifecycle and ticking right

packages/router/src/queued-browser-history.ts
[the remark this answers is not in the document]
fkleuver: This is a todo on my part, I still need to reorganize the error code across the framework (see https://github.com/aurelia/aurelia/issues/192)
So @BBosman 's suggestion is my preference for now as it allows me to easily search all TODO's and tackle them all in one go when I get to it. I'd say just start with 2000 for the router and pick any number upwards of that. That gives you a safe range until I've organized it all.

packages/router/src/queued-browser-history.ts
```diff
@@ -0,0 +1,118 @@
...
+  }
+
+  private async dequeue(delta?: number): Promise<void> {
+    if (!this.queue.length || this.processingItem !== null) {
+      return;
+    }
+    this.processingItem = this.queue.shift();
+    const method = this.processingItem.object[this.processingItem.method];
+    // tslint:disable-next-line:no-console
+    console.log('DEQUEUE', this.processingItem.method, this.processingItem.parameters);
```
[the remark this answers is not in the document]
fkleuver: I think this is one where `Tracer` is better suited, since tracing is really what's happening here.

packages/router/src/router.ts
```diff
@@ -192,16 +195,16 @@ export class Router {
 
     if (!views && !Object.keys(views).length && !clearViewports) {
       this.processingNavigation = null;
+      this.processNavigations().catch(error => { throw error; });
```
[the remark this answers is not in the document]
jwx: I think there once was a reason, but can't remember. Changing it.

packages/router/src/history-browser.ts
fkleuver: Agree would be nice if you could add a comment here clarifying whether this is just a shorthand for `index === 0` or if there is actually some functional difference in any scenario

packages/router/src/scope.ts
```diff
@@ -28,7 +28,7 @@ export class Scope {
   public viewport: Viewport;
 
   public children: Scope[];
-  public viewports: Record<string, Viewport>;
+  public _viewports: Viewport[];
```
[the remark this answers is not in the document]
fkleuver: I would suggest changing the `viewports` record to `viewportLookup`, and `_viewports` to simply `viewports`

packages/router/src/scope.ts
```diff
@@ -210,17 +228,17 @@ export class Scope {
...
 
   public removeScope(): void {
     for (const child of this.children) {
       child.removeScope();
     }
-    for (const viewport in this.viewports) {
-      this.router.removeViewport(this.viewports[viewport], null, null);
+    for (const viewport in this.viewports()) {
+      this.router.removeViewport(this.viewports()[viewport], null, null);
     }
```
fkleuver: I would suggest:
```ts
const viewportLookup = this.viewports();
for (const name of viewportLookup) {
  const viewport = viewportLookup[name];
  this.router.removeViewport(viewport, null, null);
}
```

packages/router/src/viewport-content.ts
```diff
@@ -0,0 +1,185 @@
...
+}
+
+export interface IRouteableCustomElement extends ICustomElement {
+  canEnter?(nextInstruction?: INavigationInstruction, instruction?: INavigationInstruction): boolean | string | IComponentViewportParameters[] | Promise<boolean | string | IComponentViewportParameters[]>;
+  enter?(parameters?: string[] | Record<string, string>, nextInstruction?: INavigationInstruction, instruction?: INavigationInstruction): void | Promise<void>;
+  canLeave?(nextInstruction?: INavigationInstruction, instruction?: INavigationInstruction): boolean | Promise<boolean>;
+  leave?(nextInstruction?: INavigationInstruction, instruction?: INavigationInstruction): void | Promise<void>;
+}
+
+export const enum ContentStatuses {
```
fkleuver: I like where you're going with this. Just a nit on the naming, an enum should have a singular name rather than plural:
```ts
ContentStatus.loaded
```
Generally reads a bit smoother than
```ts
ContentStatuses.loaded
```

packages/router/src/history-browser.ts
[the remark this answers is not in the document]
jwx: I'll leave this as is for now.

packages/router/src/history-browser.ts
jwx: `index` might change to not reliably be `=== 0` for first entry. `firstEntry` should always be reliable. Added a comment.

packages/router/src/history-browser.ts
jwx: That method isn't finished yet, but for now I've renamed the `wait` to `goPathChangeTriggered`. Hopefully that'll be clear enough for now.

packages/router/src/queued-browser-history.ts
jwx: This way it matches `window.addEventListener('popstate', callback);` which I like. Also, this class only exists because the History API implementation in the browsers (or at least Chrome) isn't what it should be, so I'd like to keep it as "clean" as possible. So unless there are really strong feelings for changing this to use the event aggregator, I'd rather keep it as it is.

packages/router/src/queued-browser-history.ts
jwx: @davismj Because I forgot to remove it after adding it while testing some alternatives. Good catch. Thank you.

packages/router/src/queued-browser-history.ts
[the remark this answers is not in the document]
jwx: Oh, wait, I actually created the error codes as well. You want me to remove them?

packages/router/src/queued-browser-history.ts
fkleuver: @davismj The event aggregator is intended for cross-component communication. This queue is to deal with chrome's throttling of navigation actions and actually has nothing to do with any other component whatsoever. It's encapsulated and strictly internal. So, architecturally speaking the event aggregator wouldn't be the right tool for the job here as it would add complexity to no gain.

packages/router/src/queued-browser-history.ts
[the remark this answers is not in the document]
fkleuver: No it's fine you can leave error codes if you already created them. Only reason I haven't been creating them is because I want to do it efficiently all in one go at some point. So I just add the error message next to the temporary code in the comments

packages/router/src/queued-browser-history.ts
fkleuver: Yeah the naming could be a little better here. I think `processingItem` can be more direct as well, like `navigationAction` or some such

jwx:
@fkleuver I've resolved the conversations and the tests pass. Merge?

PR #425, opened 2019-02-07
Use Typescript

BBosman:
# Pull Request

## 📖 Description

Reap the benefits of the recent [Typescript 3.3](https://blogs.msdn.microsoft.com/typescript/2019/01/31/announcing-typescript-3-3/) update.

### 🎫 Issues

n/a

## 👩‍💻 Reviewer Notes

Notable changes:
- Removed the `T & any` hack from `Constructable`, as it's no longer needed.
- That triggered some typing errors that where previously masked due the use of `any`, so I fixed those.
- Removed some linting supressions that are no longer needed, as the false positives no longer trigger.
- Minor linting changes

## 📑 Test Plan

CircleCI

## ⏭ Next Steps

n/a

### DIFF
--- a/packages/kernel/src/di.ts
+++ b/packages/kernel/src/di.ts
@@ -102,6 +102,9 @@ if (!('getOwnMetadata' in Reflect)) {
     return function(target: Function): void {
       (target as IIndexable)[metadataKey] = metadataValue;
     };
+  } as (metadataKey: any, metadataValue: any) => {
+    (target: Function): void;
+    (target: Object, propertyKey: string | symbol): void;
   };
 }
 // tslint:enable:no-any ban-types
@@ -452,14 +455,14 @@ export class Resolver implements IResolver, IRegistration {
 }
 
 /** @internal */
-export interface IInvoker {
-  invoke(container: IContainer, fn: Constructable, dependencies: Key<unknown>[]): Constructable;
+export interface IInvoker<T = {}> {
+  invoke(container: IContainer, fn: Constructable<T>, dependencies: Key<unknown>[]): T;
   invokeWithDynamicDependencies(
     container: IContainer,
-    fn: Constructable,
+    fn: Constructable<T>,
     staticDependencies: Key<unknown>[],
     dynamicDependencies: Key<unknown>[]
-  ): Constructable;
+  ): T;
 }
 
 /** @internal */
@@ -482,7 +485,7 @@ export class Factory implements IFactory {
     return new Factory(Type, invoker, dependencies);
   }
 
-  public construct(container: IContainer, dynamicDependencies?: Key<unknown>[]): Constructable {
+  public construct(container: IContainer, dynamicDependencies?: Key<unknown>[]): {} {
     if (Tracer.enabled) { Tracer.enter('Factory', 'construct', [this.Type, ...slice.call(arguments)]); }
     const transformers = this.transformers;
     let instance = dynamicDependencies !== undefined
@@ -575,7 +578,7 @@ export class Container implements IContainer {
       if (isRegistry(current)) {
         current.register(this);
       } else if (isClass(current)) {
-        Registration.singleton(current, current).register(this);
+        Registration.singleton(current, current as Constructable).register(this);
       } else {
         keys = Object.keys(current);
         j = 0;
@@ -841,31 +844,31 @@ function buildAllResponse(resolver: IResolver, handler: IContainer, requestor: I
 /** @internal */
 export const classInvokers: IInvoker[] = [
   {
-    invoke<T extends Constructable, K>(container: IContainer, Type: T): K {
+    invoke<T>(container: IContainer, Type: Constructable<T>): T {
       return new Type();
     },
     invokeWithDynamicDependencies
   },
   {
-    invoke<T extends Constructable, K>(container: IContainer, Type: T, deps: Key<unknown>[]): K {
+    invoke<T>(container: IContainer, Type: Constructable<T>, deps: Key<unknown>[]): T {
       return new Type(container.get(deps[0]));
     },
     invokeWithDynamicDependencies
   },
   {
-    invoke<T extends Constructable, K>(container: IContainer, Type: T, deps: Key<unknown>[]): K {
+    invoke<T>(container: IContainer, Type: Constructable<T>, deps: Key<unknown>[]): T {
       return new Type(container.get(deps[0]), container.get(deps[1]));
     },
     invokeWithDynamicDependencies
   },
   {
-    invoke<T extends Constructable, K>(container: IContainer, Type: T, deps: Key<unknown>[]): K {
+    invoke<T>(container: IContainer, Type: Constructable<T>, deps: Key<unknown>[]): T {
       return new Type(container.get(deps[0]), container.get(deps[1]), container.get(deps[2]));
     },
     invokeWithDynamicDependencies
   },
   {
-    invoke<T extends Constructable, K>(container: IContainer, Type: T, deps: Key<unknown>[]): K {
+    invoke<T>(container: IContainer, Type: Constructable<T>, deps: Key<unknown>[]): T {
       return new Type(
         container.get(deps[0]),
         container.get(deps[1]),
@@ -876,7 +879,7 @@ export const classInvokers: IInvoker[] = [
     invokeWithDynamicDependencies
   },
   {
-    invoke<T extends Constructable, K>(container: IContainer, Type: T, deps: Key<unknown>[]): K {
+    invoke<T>(container: IContainer, Type: Constructable<T>, deps: Key<unknown>[]): T {
       return new Type(
         container.get(deps[0]),
         container.get(deps[1]),
@@ -896,12 +899,12 @@ export const fallbackInvoker: IInvoker = {
 };
 
 /** @internal */
-export function invokeWithDynamicDependencies<T extends Constructable, K>(
+export function invokeWithDynamicDependencies<T>(
   container: IContainer,
-  Type: T,
+  Type: Constructable<T>,
   staticDependencies: Key<unknown>[],
   dynamicDependencies: Key<unknown>[]
-): K {
+): T {
   let i = staticDependencies.length;
   let args: Key<unknown>[] = new Array(i);
   let lookup: Key<unknown>;
--- a/packages/runtime-html/src/dom.ts
+++ b/packages/runtime-html/src/dom.ts
@@ -405,7 +405,7 @@ export class NodeSequenceFactory implements NodeSequenceFactory {
   private readonly dom: IDOM;
   private readonly deepClone: boolean;
   private readonly node: Node;
-  private readonly Type: Constructable;
+  private readonly Type: Constructable<INodeSequence>;
 
   constructor(dom: IDOM, markupOrNode: string | Node) {
     this.dom = dom;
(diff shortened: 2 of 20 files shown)


BBosman:
No clue why `unit_test_node` on `CircleCI` is failing though.

fkleuver:
@BBosman There's probably a whole chunk of tests failing actually, I'm seeing this TS compiler error in the node tests:
SyntaxError: The requested module 'file:///home/circleci/repo/packages/runtime/dist/build/templating/lifecycle-render.js' does not provide an export named '$hydrateElement'

The browser tests will simply not be emitted if they fail to compile which results in those tests not running (that's where the -1.5% coverage comes from). It's probably a good idea to turn on the type checker in webpack. Anyway, the node tests do go through normal typescript compilation so they will actually tell you what's wrong when nothing else does.

BBosman:
@fkleuver I did see that error, but I don't understand why it's giving it.

The file `src/templating/lifecycle-render.ts` has this:
```ts
export function $hydrateElement(this: Writable<ICustomElement>, flags: LifecycleFlags, parentContext: IServiceLocator, host: INode, options: IElementHydrationOptions = PLATFORM.emptyObject): void {
  // ...
}
```

And `dist/build/templating/lifecycle-render.js` has this:
```ts
export function $hydrateElement(flags, parentContext, host, options = PLATFORM.emptyObject) {
  // ...
}
```

So it is there and being exported imho.

It is missing from `dist/templating/lifecycle-render.d.ts` though.

fkleuver:
@BBosman It's marked internal so it doesn't end up in the type defs, that's why it's imported via a relative import

fkleuver:
@BBosman there is a compiler error in `packages/runtime/test/resources/custom-element.resource.spec.ts` but that's not the thing causing this though. Just wanted to mention it. I'm trying to narrow down why it's complaining about `$hydrateElement`

fkleuver:
@BBosman Looks like this was a regression in `esm@3.2.1` which was fixed again in `esm@3.2.4`: somewhat related to [this](https://github.com/standard-things/esm/issues/722) (you currently have it on `esm@3.2.3`. That's the loader mocha uses.
Upgrading to `esm@3.2.4` should get rid of that strange `$hydrateElement` error. It does however expose the other compiler error I pointed out in my previous comment.

Now, suppressing that with `as any` like so: `expect(CustomElementResource.isType(type as any)).to.equal(false);` gets the node tests to fully run again. The problem with that, is that `CustomElementResource.isType` is very public API and shouldn't be causing type errors when passing in a constructor with the wrong prototype.

We might actually want to keep `Constructable` around for this purpose. The semantic difference between `Constructable` and `Class` is that `Constructable` does not have a strongly typed prototype, which is necessary in certain cases.

packages/kernel/src/di.ts
```diff
@@ -575,7 +578,7 @@ export class Container implements IContainer {
       if (isRegistry(current)) {
         current.register(this);
       } else if (isClass(current)) {
-        Registration.singleton(current, current).register(this);
+        Registration.singleton(current, current as Class).register(this);
```
fkleuver: Why is this cast needed?

packages/router/src/resources/viewport.ts
```diff
@@ -1,11 +1,11 @@
-import { Constructable, IContainer, InterfaceSymbol, Writable } from '@aurelia/kernel';
+import { Class, InterfaceSymbol, Writable } from '@aurelia/kernel';
 import { bindable, createRenderContext, CustomElementResource, ICustomElement, IDOM, IElementTemplateProvider, INode, IRenderContext, IRenderingEngine, ITemplate, LifecycleFlags, TemplateDefinition } from '@aurelia/runtime';
 import { Router } from '../router';
 import { IViewportOptions, Viewport } from '../viewport';
 
 export interface ViewportCustomElement extends ICustomElement<Element> { }
 export class ViewportCustomElement {
-  public static readonly inject: ReadonlyArray<InterfaceSymbol|Constructable> = [Router, INode, IRenderingEngine];
+  public static readonly inject: ReadonlyArray<InterfaceSymbol|Class> = [Router, INode, IRenderingEngine];
```
fkleuver: I introduced a new type `InjectArray` as a nice shorthand for this. Doesn't have to be this PR, just wanted to point out that it's there

packages/runtime-html/src/create-element.ts
```diff
@@ -21,7 +21,7 @@ const slice = Array.prototype.slice;
 
 export function createElement<T extends INode = Node>(
   dom: IDOM<T>,
-  tagOrType: string | Constructable,
+  tagOrType: string | Class,
```
fkleuver: Will this still work if you pass a regular function to it (I'm not even sure if that worked before, but I just got to think of it)

packages/runtime-html/src/dom.ts
```diff
@@ -434,7 +434,7 @@ export class NodeSequenceFactory implements NodeSequenceFactory {
   }
 
   public createNodeSequence(): INodeSequence {
-    return new this.Type(this.dom, this.node.cloneNode(this.deepClone));
+    return new this.Type(this.dom, this.node.cloneNode(this.deepClone)) as INodeSequence;
```
fkleuver: Can we use the generic type arg for this? e.g. `private readonly Type: Class<INodeSequence>`? That's kind of the point of the `Class` type :)

packages/runtime/src/resources/binding-behavior.ts
```diff
@@ -58,7 +57,7 @@ function define<T extends Constructable = Constructable>(this: IBindingBehaviorR
   Type.description = description;
   Type.register = register;
 
-  return Type;
+  return Type as T & IBindingBehaviorType<T>;
```
fkleuver: I might have overlooked this in my type refactor, but I think this cast can be moved up: `const Type = ctor as T & Writable<IBindingBehaviorType>;` so it's not needed here as well

packages/runtime/src/resources/custom-attribute.ts
```diff
@@ -181,7 +180,7 @@ function define<N extends INode = INode, T extends Constructable = Constructable
     proto.$nextDetached = null;
   }
 
-  return Type as ICustomAttributeType & T;
+  return Type as T & ICustomAttributeType<N, T>;
```
fkleuver: Same thing here, perhaps this can be moved up to the `const Type` declaration

packages/kernel/src/di.ts
BBosman: Also does not have a strongly typed prototype so it gave a typing error, so same as the rest.

I'm going to revert the `Constructable` -> `Class` change. It has more side-effects than I'd expected.

packages/runtime-html/src/create-element.ts
BBosman: No, it will give a type error. Changing it would have to be downstream as well, cause `createElementForType` has it as well.

packages/runtime/src/resources/binding-behavior.ts
BBosman: Doing that breaks the `Type.kind = BindingBehaviorResource;` on line 57.

```
Type '{ name: string; keyFrom: (this: IBindingBehaviorResource, name: string) => string; isType: <T>(this: IBindingBehaviorResource, Type: T & Partial<IBindingBehaviorType<Constructable<{}>>>) => Type is T & IBindingBehaviorType<...>; define: { ...; }; }' is not assignable to type 'IResourceKind<IBindingBehaviorDefinition, InstanceType<T> & IBindingBehavior, ConstructableClass<InstanceType<T> & IBindingBehavior, IIndexable<object>>>'.
  Types of property 'isType' are incompatible.
    Type '<T>(this: IBindingBehaviorResource, Type: T & Partial<IBindingBehaviorType<Constructable<{}>>>) => Type is T & IBindingBehaviorType<Constructable<{}>>' is not assignable to type '<T>(Type: T & Partial<IResourceType<IBindingBehaviorDefinition, InstanceType<T> & IBindingBehavior, ConstructableClass<InstanceType<T> & IBindingBehavior, IIndexable<object>>>>) => Type is T & ... 3 more ... & IResourceType<...>'.
      Type predicate 'Type is T & Partial<IResourceType<IBindingBehaviorDefinition, InstanceType<T> & IBindingBehavior, ConstructableClass<InstanceType<T> & IBindingBehavior, IIndexable<object>>>> & IBindingBehaviorType<...>' is not assignable to 'Type is T & object & { [key: string]: unknown; } & { new (...args: unknown[]): InstanceType<T> & IBindingBehavior & { constructor: IIndexable<object>; }; readonly prototype: InstanceType<T> & IBindingBehavior & { constructor: IIndexable<...>; }; } & IResourceType<...>'.
        Type 'T & Partial<IResourceType<IBindingBehaviorDefinition, InstanceType<T> & IBindingBehavior, ConstructableClass<InstanceType<T> & IBindingBehavior, IIndexable<object>>>> & IBindingBehaviorType<...>' is not assignable to type 'T & object & { [key: string]: unknown; } & { new (...args: unknown[]): InstanceType<T> & IBindingBehavior & { constructor: IIndexable<object>; }; readonly prototype: InstanceType<T> & IBindingBehavior & { constructor: IIndexable<...>; }; } & IResourceType<...>'.
          Type 'T & Partial<IResourceType<IBindingBehaviorDefinition, InstanceType<T> & IBindingBehavior, ConstructableClass<InstanceType<T> & IBindingBehavior, IIndexable<object>>>> & IBindingBehaviorType<...>' is not assignable to type '{ new (...args: unknown[]): InstanceType<T> & IBindingBehavior & { constructor: IIndexable<object>; }; readonly prototype: InstanceType<T> & IBindingBehavior & { constructor: IIndexable<object>; }; }'.
            Type 'IBindingBehavior & { constructor: unknown; }' is not assignable to type 'InstanceType<T> & IBindingBehavior & { constructor: IIndexable<object>; }'.
              Type 'IBindingBehavior & { constructor: unknown; }' is not assignable to type 'InstanceType<T>'. [2322]
```

BBosman:
@fkleuver The `esm` update indeed fixed the `export` issue.

I also reverted the `Constructable` -> `Class` change, as it had too many unintended side-effects.

> The problem with that, is that `CustomElementResource.isType` is very public API and shouldn't be causing type errors when passing in a constructor with the wrong prototype.
Isn't the purpose of an `isType` type guard to cause type errors when passing in something invalid?

fkleuver:
@BBosman Well, technically you can't pass in anything that's "invalid". There are no type constraints on what can be a custom element or custom attribute apart from it being constructable. What that method gives you is the ability to tell whether something is a certain resource type, and if it is, inside the `if` that resource will be strongly typed (as per the interface of that resource, merged with the class type you passed in)

fkleuver:
@BBosman sorry for not following up for a bit, is there something you still needed help with on this PR?

BBosman:
@fkleuver No problem. Life got in the way here, so haven't had time to look at it from my side either.

I just rebased on master and pushed the most recent version (and updated the PR description).

The only open issue is the failing `unit_test_node` due to the `isType` issue mentioned above. It's also the root cause of not being able to make the other casting changes requested above.

The problem is that I don't know how to fix it, even with your (extended) explanation, so some hands on help would be appreciated.

BBosman:
Minor update: Had to pin `chromedriver` to `~2.45.0`, as `2.46.0` requires a newer version of `Chrome` than the one that's running on CircleCI.

fkleuver:
For starters let's separate dependency updates from refactors, because dependency updates generally don't need much review and could have been merged a long time ago. If the refactor depends on it and you don't want to have to wait for review/merge, it's fine if you create a new branch off the previous one and have a second PR with everything included. Then, once the first is merged, you can merge again and the diff will be normalized to only the latest changes automatically

Do you think you can split it up so we can eliminate the deps from being a factor?

BBosman:
@fkleuver Split of the dependency updates and the `InjectArray` changes, so it's now down to the removal of the `Constructable`/`any` hack (and removing some no longer needed `tslint:disable` comments).

Unfortunately the problem still remains. I can fix it by using `unknown`, but I don't think that's the right solution:
```ts
function isType<T>(this: ICustomElementResource, Type: T & { kind: unknown }): Type is T & ICustomElementType {
```

BBosman:
@fkleuver I've found a solution that I'm happy with and that works.

If you could do another review when you've got a bit of time?

fkleuver:
@BBosman Overall it looks good but I'm not too fond of `InstanceType<Constructable>` since that's just a detour for its generic type argument (which is `{}` by default and we're trying to avoid this)

BBosman:
@fkleuver Valid comment. Did a bit of type refactoring and got rid of the `InstanceType<Constructable>`'s.

fkleuver:
Nice, thanks for fixing this up and apologies for the delay!

PR #426, opened 2019-02-08
DOM binding fix and misc

### LINKED ISSUE #421: [Bug] Remove Interpolation Expressions from Class Attributes During Compilation
# 🐛 Bug Report

After a template is compiled, any interpolation expression should be removed from the `class` attribute.

## 🤔 Expected Behavior

Given the following HTML:

```XML
<li repeat.for="a of activators" class="icon ${a.focused ? 'focused' : ''}">
...
</li>
```

When an `li` is associated with a data item with `a.focused === true`, we should see the following HTML in the DOM:

```XML
<li class="icon au focused">
...
</li>
```

## 😯 Current Behavior

At present, with the above template, we get the following in the DOM:

```XML
<li class="icon ${a.focused ? 'focused' : ''} au focused">
...
</li>
```

## 💁 Possible Solution

We need to ensure that our compiler's interpolation code for `class` removes the expression from the attribute's value during compilation. I'm not sure if this is happening anywhere else. The only place I've seen it is in the `class` attribute so far.

## 🔦 Context

At present, this doesn't appear to be causing any runtime issues. However, we want a cleaner DOM with final class values that are of the expected form. We don't want to expose the developer's expressions in the HTML.

### LINKED ISSUE #423: [Bug] Class Surrogates Not Updating
# 🐛 Bug Report

When placing a surrogate class binding on a template element, the value of the class does not consistently update when the bound view model property changes.

## 🤔 Expected Behavior

All bindings on the template element should be applied to the host element and should update when any bound properties change. Let's say we have a `my-element` component with the following template:

```xml
<template class="${controlPanel.isOpen ? 'control-panel-open' : ''}">
...
</template>
```

When `my-element` is used in HTML and `isOpen` is `true`, we should see the following in the DOM:

```xml
<my-element class="au control-panel-open"></my-element>
```

When the `isOpen` property changes to `false`, we should see the following in the DOM:

```xml
<my-element class="au"></my-element>
```

## 😯 Current Behavior

Currently, given the above setup, we always see the following in the DOM, even after `isOpen` changes to false:

```xml
<my-element class="au control-panel-open"></my-element>
```

## 💁 Possible Solution

Something's probably not fully worked out in the surrogate code.

fkleuver:
# Pull Request

## 📖 Description

- Fixes #423 (oldValue was not stored in the target observer so it never recognized the difference between an initial value and an empty string set as the next value)
- Adds `InjectArray` as a shorthand type for `ReadonlyArray<InterfaceSymbol | Constructable | string>`
- Register the app host as `INode` so it can be injected into the root component
- Removes coercion from change handler in SelfObserver and moves the callback to after `callSubscribers`, so updates are actually already propagated and/or queued when the callback is invoked and you can work with the results
- Fixes #421 by removing interpolations from attribute values during compilation

### 🎫 Issues

- Fixes #423 

## 👩‍💻 Reviewer Notes

Changes probably speak for themselves :)

## 📑 Test Plan

- Added ~2k tests to verify the various combinations of class binding initialization and updates

## ⏭ Next Steps

- Add more of these kinds of tests
- Look at the other bug reports :)

### DIFF
--- a/packages/jit-html/test/integration/html-attributes.spec.ts
+++ b/packages/jit-html/test/integration/html-attributes.spec.ts
@@ -1,0 +1,180 @@
+import { InjectArray } from '@aurelia/kernel';
+import { Aurelia, CustomElementResource as CE, ICustomElement, INode, LifecycleFlags as LF } from '@aurelia/runtime';
+import { expect } from 'chai';
+import { TestContext } from '../util';
+import { eachCartesianJoin } from './util';
+
+interface IThing {
+  enabled?: boolean;
+}
+type VM<T> = T & ICustomElement<Node> & {
+  el: Element;
+  parent?: VM<T>;
+};
+interface Spec {
+  t: string;
+}
+
+describe('html attribute', function() {
+  function setup() {
+    const ctx = TestContext.createHTMLTestContext();
+
+    const host = ctx.createElement('div');
+    const au = new Aurelia(ctx.container);
+    return { ctx, host, au };
+  }
+
+  describe('class', function () {
+    describe('on a surrogate', function () {
+      interface ThingSpec extends Spec {
+        createThing(): IThing;
+      }
+      interface ExpressionSpec extends Spec {
+        expression: string;
+        getExpectedClassNames(thing: IThing): string[];
+      }
+
+      const thingSpecs: ThingSpec[] = [
+        { t: '1', createThing() { return { enabled: false }; } },
+        { t: '2', createThing() { return { enabled: true  }; } },
+        { t: '3', createThing() { return {                }; } }
+      ];
+      const expressionSpecs: ExpressionSpec[] = [
+        {
+          t: '1',
+          expression: '${thing.enabled?\'class-1\':\'\'}',
+          getExpectedClassNames(thing: IThing) {
+            return thing && thing.enabled ? ['au', 'class-1'] : ['au'];
+          }
+        },
+        {
+          t: '2',
+          expression: '${thing.enabled?\'class-1\':\'class-2\'}',
+          getExpectedClassNames(thing: IThing) {
+            return thing && thing.enabled ? ['au', 'class-1'] : ['au', 'class-2'];
+          }
+        },
+        {
+          t: '3',
+          expression: '${thing.enabled?\'class-1\':\'\'} class-3',
+          getExpectedClassNames(thing: IThing) {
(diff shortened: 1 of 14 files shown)


packages/jit-html/test/integration/html-attributes.spec.ts
```diff
@@ -0,0 +1,209 @@
...
+            return thing && thing.enabled ? ['au', 'class-0', 'class-3', 'class-1'] : ['au', 'class-0', 'class-3'];
+          }
+        },
+        {
+          t: '8',
+          expression: 'class-0 ${thing.enabled?\'class-1\':\'class-2\'} class-3',
+          getExpectedClassNames(thing: IThing) {
+            return thing && thing.enabled ? ['au', 'class-0', 'class-3', 'class-1'] : ['au', 'class-0', 'class-2', 'class-3'];
+          }
+        },
```
bigopon: Let's have a todo to add test cases for following scenarios:

```
class-0 ${thing.enabled && 'enabled'}
```
mapped to `class-0 enabled` and `class-0 false` `class-0 0` `class-0 undefined` `class-0 null` `class-0 `

Similarly we can add test todo with other attributes as well, such as data attributes / svg element attributes

packages/runtime/src/observation/self-observer.ts
```diff
@@ -61,24 +52,28 @@ export class SelfObserver implements SelfObserver {
...
-
-        if (coercedValue !== undefined) {
-          this.currentValue = newValue = coercedValue;
+    if (newValue !== this.currentValue || (flags & LifecycleFlags.patchStrategy) > 0) {
+      if ((flags & LifecycleFlags.fromBind) === 0) {
+        const oldValue = this.currentValue;
+        flags |= this.persistentFlags;
+        this.currentValue = newValue;
+        this.callSubscribers(newValue, oldValue, flags | LifecycleFlags.allowPublishRoundtrip);
+        if (this.callback !== null) {
```
bigopon: Yeah this is a real dilemma, but it matches vCurrent timing, where subscribers are called before self callback.

packages/jit-html/test/integration/html-attributes.spec.ts
fkleuver: yeah agree, feel free to help me beef up tests here if you want, i think it should be pretty straightforward

fkleuver:
Cool, thanks for the review, let's go then

PR #429, opened 2019-02-10
switch from arrow to function expr in tests

fkleuver:
# Pull Request

## 📖 Description

- Replaces all `() =>` with `function ()` in tests as per mocha's recommendation (provides access to test context and better context isolation, improving garbage collection and lowering memory usage esp. in heavily looped tests)
- Adds a space between `function` and the opening paren as per recommendation of airbnb and many other style guides

### 🎫 Issues

N/A

## 👩‍💻 Reviewer Notes

Lots of files affected, but the only changes executed are:
- replace `(it|describe|before|after|beforeEach|afterEach)\((.*)\(\) => ` with `$1($2function() ` in `packages/*/test/**/*.ts` (with regex)
- replace `function()` with `function ()` in `**/*.ts` (without regex)

## 📑 Test Plan

N/A

## ⏭ Next Steps

N/A

### DIFF
--- a/packages/aot/test/unit/index.spec.ts
+++ b/packages/aot/test/unit/index.spec.ts
@@ -1,8 +1,8 @@
 import { expect } from 'chai';
 import { placeholder } from '../../src/index';
 
-describe('index', function() {
-  it('placeholder should be null', function() {
+describe('index', function () {
+  it('placeholder should be null', function () {
     expect(placeholder).to.equal(null);
   });
 });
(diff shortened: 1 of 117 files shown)


fkleuver:
@EisenbergEffect @bigopon @BBosman Wanted to do this separately from writing additional tests, so I'm merging this now so I can continue. It's easy enough to revert, however, so if there are objections to either of the 2 commits please do let me know and I'll revert it. But I think it's a good idea to go this way.

PR #430, opened 2019-02-10
WIP: lifecycles

fkleuver:
# Pull Request


EisenbergEffect:
With all the task usage, do we need to pool/recycle those? Any notion of how this will affect perf?

fkleuver:
This is just an initial version to get things working. Further down the line I'm thinking about having a task per hook to reduce the use of spread operators and such.

Effect on perf is negligible as long as things stay synchronous. It just passes the same singleton `LifecycleTask.done` around, checking the same boolean property. The aggregates (and arrays) are lazy, so there are no unnecessary objects being created.

I'm not sure about perf on async, there's not really a baseline we can compare that against. But when things are async, perf gets hit regardless because there are also the promise objects. But there is plenty of room for perf optimizations in this scenario for sure.

fkleuver:
Superseded by #464

PR #440, opened 2019-02-15
Update Typescript and other dev(Dependencies)

BBosman:
# Pull Request

## 📖 Description

Upgraded to [Typescript 3.3](https://blogs.msdn.microsoft.com/typescript/2019/01/31/announcing-typescript-3-3/).

### 🎫 Issues

This is part of #425, which had too many moving parts, so I'm splitting it up.

## 👩‍💻 Reviewer Notes

Notable changes:
- Updated the other (dev)Dependencies as well, while I was at it. (including updating the correspondig `package-lock.json` files)
- Pinned `chromedriver` to `~2.45.0`, as `2.46.0` requires a newer version of `Chrome` than the one that's running on CircleCI.
- Unpinned `ajv`, as epoberezkin/ajv#941 is fixed.

## 📑 Test Plan

CircleCI

## ⏭ Next Steps

Upgrade `Chrome` on CircleCI, so we can unpin `chromedriver`.


fkleuver:
@BBosman I was wondering where the decrease in coverage came from, so I ran the tests locally - for some reason there are 30 failing `au-compose` tests:
image

I am baffled by this.. why would these dependency upgrades have such side effect? :/ any idea what might be causing it?

(btw, we can completely remove ajv dependency.. it was only added to pin it in the firt place)

BBosman:
@fkleuver Weird. I tested locally with `npm run test-node` and that runs without failure, so I made the (false) assumption that `npm run test` would as well.

But it appears we must find the reason in whatever is different between those setups and got an upgrade within this PR.

(and I will remove `ajv` as dependency)

fkleuver:
@BBosman Well, running locally can be a bit tedious since it takes a few minutes. You can also just rely on CI. The coverage is supposed to stay identical. Fact that it went down is an immediate red flag that something is failing.

BBosman:
@fkleuver It keeps getting stranger. A fresh clone of the repo exhibits the same issue.

```shell
git clone https://github.com/aurelia/aurelia.git aurelia2
cd aurelia2
npm run bootstrap
npm run build
npm run test
```
Results in 30 failing tests.

So my PR is surfacing the issue, but not the root cause?

BBosman:
Both `test-firefox` and `test-node` work, so it has to be something `Chrome` specific.

fkleuver:
Yeah and the build on master still succeeds. I don't know what's going on here, I'll assume it's a weird fluke and if it breaks on master then we'll take it from there. Hunting for ghosts like this is not the most productive way to work on vNext I think you would agree :)
I'll just merge it

PR #444, opened 2019-02-17
Dynamic routing: RouteTable skeleton and extension points

jwx:
# Pull Request

## 📖 Description

This PR illustrates how the router's extension points are used to integrate traditional route table routes with the router. It contains a `RouteTable` skeleton that uses the extension points with a simple, hard coded example.

In summary, any route component, be it a route table or a dynamic routing syntax, can do whatever it wants and offer whatever API it wants towards developers as long as it translates to and from sets of viewport instructions when dealing with the router.

### 🎫 Issues

## 👩‍💻 Reviewer Notes

The `RouteTable` contains a simple, hard coded example that works (with limitations) for the application in `packages/router/test/e2e/doc-example` to illustrate the principles. Needless to say, it should eventually be removed.

## 📑 Test Plan

- Make sure tests still pass.

## ⏭ Next Steps

- Re-visit scoped viewports
- Finish refactoring to make router work solely with `ViewportInstruction` instead of strings

### DIFF
--- a/packages/router/src/route-table.ts
+++ b/packages/router/src/route-table.ts
@@ -2,7 +2,7 @@ import { Router } from './router';
 import { ViewportInstruction } from './viewport-instruction';
 
 /**
- * Class that handles routes configured into a route table
+ * Class that handles routes configured in a route table
  */
 export class RouteTable {
 
@@ -32,11 +32,11 @@ export class RouteTable {
     if (this.matchViewportInstructions(instructions, components)) {
       return 'welcome';
     }
-
     return instructions;
   }
 
-  // Check if instruction sets matches
+  // Check if instruction sets match. NOT COMPLETE CODE: it should also check viewport
+  // and, possibly, parameters (or their configuration)
   private matchViewportInstructions(instructionsA: ViewportInstruction[], instructionsB: ViewportInstruction[]): boolean {
     for (const instruction of instructionsA) {
       if (!instructionsB.find((i) => i.sameComponent(instruction))) {
--- a/packages/router/src/route-table.ts
+++ b/packages/router/src/route-table.ts
@@ -14,9 +14,7 @@ export class RouteTable {
    * @returns The viewport instructions for a found route or the route if not found.
    */
   public transformFromUrl = (route: string, router: Router): string | ViewportInstruction[] => {
-    if (route === 'welcome' || route === '/welcome') {
-      return [new ViewportInstruction('books'), new ViewportInstruction('about')];
-    }
+    // TODO: Implement route recognizing to transform a configured route to a set of viewport instructions
     return route;
   }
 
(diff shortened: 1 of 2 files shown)


packages/router/src/router.ts
```diff
@@ -44,10 +45,14 @@ export class Router {
   private processingNavigation: INavigationInstruction = null;
   private lastNavigation: INavigationInstruction = null;
 
-  constructor(public container: IContainer) {
+  private routeTable: RouteTable;
```
BBosman:
```ts
private readonly routeTable: RouteTable;
```

packages/router/src/route-table.ts
```diff
@@ -0,0 +1,32 @@
...
+
+  /**
+   * Check a route against the route table and return the appropriate viewport instructions.
+   *
+   * @param route The route to match.
+   * @param router The application router.
+   * @returns The viewport instructions for a found route or the route if not found.
+   */
+  public transformFromUrl = (route: string, router: Router): string | ViewportInstruction[] => {
+    // TODO: Implement route recognizing to transform a configured route to a set of viewport instructions
```
fkleuver: I still don't understand what makes this an extension point, when the code for it lives here. Perhaps what you're really meaning to do is define an interface and inject that interface into the router?

jwx: My terminology/point of view might be off. Regardless, when I talk about the extension point(s), I'm not talking about the `route-table.ts`. I'm talking about the _router_'s extension points [here](https://github.com/aurelia/aurelia/pull/444/files#diff-91a41028732ff4a368a53bc46649eb79R136) and [here](https://github.com/aurelia/aurelia/pull/444/files#diff-91a41028732ff4a368a53bc46649eb79R417) for implementing a custom syntax, a route table, a developer API or whatever addition is desired to the router's viewport instructions and own syntax. The `route-table.ts` is just a skeleton that's in this PR to illustrate how to use the extension points and to give you a start for the router integration.

fkleuver: An extension point, I would define that loosely as "a place in the code/api where you can extend/inject custom behavior". I can see the reasoning behind why you might call those two links extension points, but in my mind what you linked is *integration* points (where the router integrates with the route table).
As far as the `Router` is concerned, the *extension* point where it takes the (customizeable) `RouteTable` is [the Router's constructor](https://github.com/aurelia/aurelia/pull/444/files#diff-91a41028732ff4a368a53bc46649eb79R50).

Since that accepts a concrete class which is defined in the router package itself, it cannot be extended/changed from the outside. An end user would not be able to provide their own implementation because the router only accepts the concrete class you put there as an example.

I understand now that you're primarily offering this to me as a place where I can integrate the route-recognizer, and that's why I'm saying just make it an interface. Otherwise it's just confusing (it did confuse me, so yeah..).
An example belongs in an example folder or a unit test

davismj:
I realize I'm about a week late to the party.

I would say the `RouteTable` isn't responsible for transforming routes. It's responsible for registration of routes in a single place. It is an app wide singleton that a consumer can inject anywhere to begin adding, disabling, or enabling new or existing routes. It would be used by the Aurelia tools, e.g. the `@route` decorator, to register new route configurations.

PR #455, opened 2019-02-25
Dynamic routing: refactor viewport instruction, part 3

jwx:
# Pull Request

## 📖 Description

This PR contains the third part of the refactor towards viewport instruction and instruction resolver. It primarily removes the string parsing and handling from scope and replaces it with instruction resolver calls. It's one step closer having the router only deal with viewport instructions internally.

### 🎫 Issues

## 👩‍💻 Reviewer Notes

Ignore everything in packages/router/test/e2e/** since they are just my own test applications while working.

## 📑 Test Plan

- Make sure tests still pass.

## ⏭ Next Steps

- Review the extension points
- Re-visit scoped viewports

### DIFF
--- a/packages/router/src/instruction-resolver.ts
+++ b/packages/router/src/instruction-resolver.ts
@@ -98,43 +98,6 @@ export class InstructionResolver {
     return { clearViewports, newPath };
   }
 
-  public findViews(path: string): Record<string, string> {
-    const views: Record<string, string> = {};
-    // TODO: Let this govern start of scope
-    if (path.startsWith('/')) {
-      path = path.substring(1);
-    }
-    const sections: string[] = path.split(this.separators.sibling);
-
-    // TODO: Remove this once multi level recursiveness is fixed
-    // Expand with instances for all containing views
-    // const expandedSections: string[] = [];
-    // while (sections.length) {
-    //   const part = sections.shift();
-    //   const parts = part.split(this.separators.scope);
-    //   for (let i = 1; i <= parts.length; i++) {
-    //     expandedSections.push(parts.slice(0, i).join(this.separators.scope));
-    //   }
-    // }
-    // sections = expandedSections;
-
-    let index = 0;
-    while (sections.length) {
-      const view = sections.shift();
-      const scopes = view.split(this.separators.scope);
-      const leaf = scopes.pop();
-      const parts = leaf.split(this.separators.viewport);
-      // Noooooo?
-      const component = parts[0];
-      scopes.push(parts.length ? parts.join(this.separators.viewport) : `?${index++}`);
-      const name = scopes.join(this.separators.scope);
-      if (component) {
-        views[name] = component;
-      }
-    }
-    return views;
-  }
-
   public removeStateDuplicates(states: string[]): string[] {
     let sorted: string[] = states.slice().sort((a, b) => b.split(this.separators.scope).length - a.split(this.separators.scope).length);
     sorted = sorted.map((value) => `${this.separators.scope}${value}${this.separators.scope}`);
--- a/packages/router/src/instruction-resolver.ts
+++ b/packages/router/src/instruction-resolver.ts
@@ -39,39 +39,22 @@ export class InstructionResolver {
   }
 
   public parseViewportInstruction(instruction: string): ViewportInstruction {
-    let component, viewport, parameters, scope;
-    const [componentPart, rest] = instruction.split(this.separators.viewport);
-    if (rest === undefined) {
-      [component, ...parameters] = componentPart.split(this.separators.parameters);
-      if (component.endsWith(this.separators.ownsScope)) {
-        scope = true;
-        component = component.slice(0, component.length - 1);
-      }
-    } else {
-      component = componentPart;
-      [viewport, ...parameters] = rest.split(this.separators.parameters);
-      if (viewport.endsWith(this.separators.ownsScope)) {
-        scope = true;
-        viewport = viewport.slice(0, viewport.length - 1);
-      }
+    const instructions = instruction.split(this.separators.scope).map((scopeInstruction) => this._parseViewportInstruction(scopeInstruction));
+    for (let i = 0; i < instructions.length - 1; i++) {
+      instructions[i].nextScopeInstruction = instructions[i + 1];
     }
-    parameters = parameters.length ? parameters.join(this.separators.parameters) : undefined;
-    return new ViewportInstruction(component, viewport, parameters, scope);
+    return instructions[0];
   }
 
   public stringifyViewportInstruction(instruction: ViewportInstruction | string, excludeViewport: boolean = false): string {
     if (typeof instruction === 'string') {
-      return this.stringifyViewportInstruction(this.parseViewportInstruction(instruction), excludeViewport);
+      return this._stringifyViewportInstruction(instruction, excludeViewport);
     } else {
-      let instructionString = instruction.componentName;
-      if (instruction.viewportName && !excludeViewport) {
-        instructionString += this.separators.viewport + instruction.viewportName;
+      const instructions = [instruction];
+      while (instruction = instruction.nextScopeInstruction) {
+        instructions.push(instruction);
       }
-      if (instruction.parametersString) {
-        // TODO: Review parameters in ViewportInstruction
-        instructionString += this.separators.parameters + instruction.parametersString;
-      }
-      return instructionString;
+      return instructions.map((scopeInstruction) => this._stringifyViewportInstruction(scopeInstruction, excludeViewport)).join(this.separators.scope);
     }
   }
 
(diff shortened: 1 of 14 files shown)


jwx:
@fkleuver Everything but the same 30 unrelated tests pass so it'd be great if you could take a look.

jwx:
> Only thing I noticed worth mentioning is that you call `Reporter.write` on every `canEnter` and `canLeave` and I'm wondering if that isn't a bit _noisy_. Maybe only report when the result is `false`?

My intention is to remove it completely in a later PR. It's there purely for debugging purposes.

fkleuver:
It would be good to use tracer calls instead, like we do in the rest of the framework (kernel, runtime, jit, etc) because the `if (Tracer.enabled)` check makes it easy for rollup to remove the calls entirely from the production build.

packages/router/src/instruction-resolver.ts
```diff
@@ -38,40 +40,37 @@ export class InstructionResolver {
     return this.separators.clear;
   }
 
+  public parseViewportInstructions(instructions: string): ViewportInstruction[] {
```
fkleuver: Is this meant to be mutable? Can this be `ReadonlyArray<ViewportInstruction>`?

packages/router/src/instruction-resolver.ts
```diff
@@ -38,40 +40,37 @@ export class InstructionResolver {
...
+    if (instructions === null || instructions === '') {
+      return [];
```
fkleuver: We could return `PLATFORM.emptyArray` here if the function return value isn't meant to be mutable

packages/router/src/instruction-resolver.ts
```diff
@@ -182,4 +127,41 @@ export class InstructionResolver {
     }
     return strings.join(this.separators.sibling);
   }
+
+  private _parseViewportInstruction(instruction: string): ViewportInstruction {
```
fkleuver: Let's not have underscore convention for privates, the `private` keyword is already sufficient indicator for that

packages/router/src/instruction-resolver.ts
```diff
@@ -182,4 +127,41 @@ export class InstructionResolver {
...
+    let component, viewport, parameters, scope;
```
fkleuver: I'd prefer for these to be explicitly typed, e.g.
```ts
let component: SomeType;
let viewport: AnotherType;
// etc
```

packages/router/src/instruction-resolver.ts
```diff
@@ -182,4 +127,41 @@ export class InstructionResolver {
...
+    const [componentPart, rest] = instruction.split(this.separators.viewport);
+    if (rest === undefined) {
+      [component, ...parameters] = componentPart.split(this.separators.parameters);
+      if (component.endsWith(this.separators.ownsScope)) {
+        scope = true;
+        component = component.slice(0, -1);
```
fkleuver:
```suggestion
        component = component.slice(0, -this.separators.ownsScope.length);
```
This makes it a bit more obvious what's going on here and also makes it resilient to potentially having more than one character as a separator (which doesn't appear to be forbidden anywhere)

packages/router/src/instruction-resolver.ts
```diff
@@ -182,4 +127,41 @@ export class InstructionResolver {
...
+      }
+    } else {
+      component = componentPart;
+      [viewport, ...parameters] = rest.split(this.separators.parameters);
+      if (viewport.endsWith(this.separators.ownsScope)) {
+        scope = true;
+        viewport = viewport.slice(0, -1);
+      }
+    }
+    parameters = parameters.length ? parameters.join(this.separators.parameters) : undefined;
```
fkleuver: I think this is confusing. This looks like parameters is first an array and then changed into a string. I'd rather see two different variables in that case

packages/router/src/instruction-resolver.ts
```diff
@@ -182,4 +127,41 @@ export class InstructionResolver {
...
+    return new ViewportInstruction(component, viewport, parameters, scope);
+  }
+
+  private _stringifyViewportInstruction(instruction: ViewportInstruction | string, excludeViewport: boolean = false): string {
+    if (typeof instruction === 'string') {
+      return this.stringifyViewportInstruction(this.parseViewportInstruction(instruction), excludeViewport);
+    } else {
+      let instructionString = instruction.componentName;
+      if (instruction.viewportName && !excludeViewport) {
```
fkleuver: Can we use strict checking here? I'm not sure if `viewportName` can actually be null OR undefined OR empty string (in which case that's not good) and if it can't be, then this check should be more narrow imo

packages/router/src/router.ts
```diff
@@ -280,24 +269,28 @@ export class Router {
...
+      if (componentOrInstruction instanceof ViewportInstruction) {
+        if (!componentOrInstruction.viewport) {
+          // TODO: Deal with not yet existing viewports
+          componentOrInstruction.viewport = this.allViewports().find((vp) => vp.name === componentOrInstruction.viewportName);
+        }
+        this.addedViewports.push(componentOrInstruction);
+      } else {
+        if (typeof viewport === 'string') {
+          // TODO: Deal with not yet existing viewports
+          viewport = this.allViewports().find((vp) => vp.name === viewport);
```
fkleuver: Maybe cleaner to allow `allViewports()` to accept a predicate function? So you can do

```ts
viewport = this.allViewports(vp => vp.name === viewport);
```

Also, as I wrote that I noticed that `viewport` is another megamorphic type (can be ViewPort or string). Can we split that up please?

packages/router/src/router.ts
```diff
@@ -372,6 +365,8 @@ export class Router {
       nav = this.navs[name] = new Nav(this, name);
     }
     nav.addRoutes(routes);
+    // tslint:disable-next-line:prefer-object-spread
+    this.navs[name] = Object.assign({}, nav);
```
fkleuver:
```suggestion
    this.navs[name] = { ...nav };
```

That lint rule is there for a good reason :)

packages/router/src/viewport-content.ts
```diff
@@ -58,10 +62,89 @@ export class ViewportContent {
...
+    Reporter.write(10000, 'viewport canEnter', result);
+    if (typeof result === 'boolean') {
+      return Promise.resolve(result);
+    }
+    if (typeof result === 'string') {
+      return Promise.resolve([new ViewportInstruction(result, viewport)]);
+    }
+    return result as Promise<ViewportInstruction[]>;
+  }
+  public canLeave(nextInstruction: INavigationInstruction): boolean | Promise<boolean> {
```
fkleuver: Seems like this should have return type `Promise<boolean>` instead of `boolean | Promise<boolean>`?

fkleuver:
Since we're now in the refactoring / cleanup phases I'm going to be a bit more nit picky on the typings as well. I'm primarily doing that where the typings confuse me or don't convey the information that I'm hoping to get, which makes it more difficult for me to actually review the PR from a functional/architectural point of view. I hope you can make some improvements there.
Otherwise looks promising.

jwx:
> It would be good to use tracer calls instead, like we do in the rest of the framework (kernel, runtime, jit, etc) because the `if (Tracer.enabled)` check makes it easy for rollup to remove the calls entirely from the production build.

Okay, I'll look into doing that instead when I get there.

packages/router/src/instruction-resolver.ts
jwx: It's meant to be mutable.

packages/router/src/instruction-resolver.ts
jwx: It's meant to be mutable.

packages/router/src/router.ts
jwx: This gives me typing errors that I haven't figured out how to resolve. @fkleuver @BBosman

packages/router/src/router.ts
jwx: At least for now, I'd prefer not to add any parameters to `allViewports`. (And, if it had had a filter parameter, I would kind of half-expect to get the comment "Why is it called `allViewports` if it's not returning all viewports?")

I don't think `viewport` is megamorphic in the same sense as `parameters` in your earlier comment. In this case it's always the viewport that's passed into the method it's just that for one branch it needs to be a `Viewport` so if the value is a `string`, it's resolved and the type is changed.

Also, there's two TODO:s in this method so all in all, I'd prefer not to change anything here now if that's okay. @fkleuver

packages/router/src/router.ts
BBosman: Actually I'm amazed this works (and I think it doesn't). An object spread will only copy an object’s own enumerable properties, which explains the typing errors you got, as the missing functions are on the prototype and are thus not copied.

The same applies to `Object.assign` though. And some quick tests I did confirm that in both cases the functions are not copied.

So I think the basic logic has a flaw and Typescript is able to determine that in the case of the object spread, but not in the case of `Object.assign`. Another reason why object spread is preferable. 😉

For reference this is the error you're getting?
`Type '{ name: string; routes: NavRoute[]; router: Router; }' is missing the following properties from type 'Nav': addRoutes, addRoute`

Imho the way to fix this is to do something like this:
```ts
this.navs[name] = new Nav(nav.router, nav.name, nav.routes);
```
And add an (internal?) constructor to also accepts that `router` argument.

jwx: Thanks for the help! I went with your suggestion and changed the constructor to
```typescript
constructor(router: Router, name: string, routes: NavRoute[] = [])
```

packages/router/src/router.ts
fkleuver: I'm cool with not having the method as a parameter, that was just an idea.

For the language runtime a variable is megamorphic for as little as being able to take the shape of either a primitive or an object. And that in itself is something I'd really prefer to avoid. Few things are as confusing as a variable that can be either a string or an object.

For public API it's fine as long as their are clearly documented method overloads. It's kind of unavoidable then. But then internally you could do something like:

```ts
public publicMethod(thingName: string);
public publicMethod(obj: SomeClass);
public publicMethod(thingNameOrObj: string | SomeClass) {
  let obj: SomeClass;
  if (thingNameOrObj instanceof SomeClass) {
    obj = thingNameOrObj;
  } else {
    obj = resolveObj(thingNameOrObj);
  }

  // continue with strongly typed obj
}
```

PR #456, opened 2019-02-25
Always hoist

BBosman:
# Pull Request

## 📖 Description

Always `hoist` all packages through `lerna`.

### 🎫 Issues

n/a

## 👩‍💻 Reviewer Notes

I can't think of a situation where we don't want to have `lerna` host packages.

The `package.json` scripts have the `--hoist` argument, but that doesn't ensure hoisting when somebody manually does a `lerna bootstrap`. By adding it to `lerna.json` we can guarantee hosting.

## 📑 Test Plan

CircleCI.

## ⏭ Next Steps

n/a

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,4 +1,9 @@
 {
+  "command": {
+    "bootstrap": {
+      "hoist": "**"
+    }
+  },
   "lerna": "3.13.0",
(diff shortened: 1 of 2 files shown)


fkleuver:
Good call :)

PR #364, opened 2019-01-05
WIP feat(resources-html): add blur, focus custom attributes

bigopon:
# Pull Request

## 📖 Description

I was about to start a new app with vNext but realized not all necessary resources have been ported over. Doing it, also will be merging custom attributes: `blur` and `portal`

PR #460, opened 2019-03-03
feat(attr-binding): add class/style binding

bigopon:
# Pull Request

## 📖 Description

This PR is extracted from #364 to make the process easier. It adds support for class/style binding with the following syntax:
```html
<div class.my-class="shouldHaveMyClass">
<div my-class.class="shouldHaveMyClass">
<div style.background="bgValue">
<div background.style="bgValue">
<!-- vendor specific rules -->
<div -webkit-user-select.style="bgValue">
<div style.-webkit-user-select="bgValue">
<!-- custom properties -->
<div --my-custom.style="value">
<div style.--my-custom="value">
```
Class will be toggle based on truthy/falsy evaluation, while style will be set as is.

### 🎫 Issues

N/A

## 👩‍💻 Reviewer Notes

Ignore the `AttrAttributePattern` and `AttrBindingCommand`. They are there for future work related to following template:
```html
<div my-attr.attr="myAttr">
<div attr.my-attr="myAttr">

<!-- interoperability with web components that operate on attribute -->
<a-web-component disabled.attr="disabled">
<a-web-component attr.disabled="disabled">
```

## 📑 Test Plan

All tests are in two new files, with tests described before the loop.

## ⏭ Next Steps

Add `AttrAttributePattern` and `AttrBindingCommand`

### DIFF
--- a/packages/jit-html/src/attribute-pattern.ts
+++ b/packages/jit-html/src/attribute-pattern.ts
@@ -1,0 +1,51 @@
+import { attributePattern, AttrSyntax, IAttributePattern } from '@aurelia/jit';
+import { IRegistry } from '@aurelia/kernel';
+
+export interface AttrPattern extends IAttributePattern {}
+export class AttrPattern implements AttrPattern {
+  public static register: IRegistry['register'];
+
+  public ['attr.PART'](rawName: string, rawValue: string, parts: string[]): AttrSyntax {
+    return new AttrSyntax(rawName, rawValue, parts[1], 'attr');
+  }
+
+  public ['PART.attr'](rawName: string, rawValue: string, parts: string[]): AttrSyntax {
+    return new AttrSyntax(rawName, rawValue, parts[0], 'attr');
+  }
+}
+
+export interface StyleAttributePattern extends IAttributePattern {}
+export class StyleAttributePattern implements StyleAttributePattern {
+  public static register: IRegistry['register'];
+
+  public ['style.PART'](rawName: string, rawValue: string, parts: string[]): AttrSyntax {
+    return new AttrSyntax(rawName, rawValue, parts[1], 'style');
+  }
+
+  public ['PART.style'](rawName: string, rawValue: string, parts: string[]): AttrSyntax {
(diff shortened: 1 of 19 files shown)


EisenbergEffect:
I'm thinking that we should only support style.prop or prop.style, but not both. If we support both, we have to document both and then answer questions about which one is the right way, we get inconsistent code samples, etc. I'd vote for style.prop and class.prop.

bigopon:
I thought of that and I was not sure if not supporting both is a good way either. One syntax is natural reading (style.rule="expression") and one is consistency (rule.style="expression")

EisenbergEffect:
Why do you feel that rule.style is more consistent than style.rule?

bigopon:
My main reason is it's our common usage of binding syntax: `attribute.command="expression"`

I wouldn't defend it strongly either, was just bringing everything up on the table for discussion.

ghiscoding:
hello, I just wanted to give 2 cents on this as well, I also prefer the same as @EisenbergEffect which is `style.rule` because that is how we read it in plain HTML too (left to right). So if we take `<div style="height: 300px">`, we can read it as... the Style Height is 300px, that is `style.height='300px'`. 

Also, not that it's really important, but Angular uses that same syntax.

fkleuver:
I agree `style.rule` as well. Apart from that - this looks really promising. I especially like the fact that you've been able to utilize the existing infrastructure to add this extra stuff, it's a bit of proof that the extensibility points are practical :)

I see all your tests pass. Is there any chance you could get the linting job to pass as well?

bigopon:
@fkleuver I'll address that. About attr syntax, there isn't away to capture syntax with multiple parts easily and pass the information to binding command. for example:
```html
<div style.--custom-prop.two-way="customCss">
```
will requires `TwoWayStyleBindingCommand` beside `StyleBindingCommand` so that attr pattern can return `AttrSyntax` with `style-two-way` binding command to work properly. Should there be an easier way for that?

bigopon:
@fkleuver Done. Ready to go.

PR #463, opened 2019-03-10
Use verdaccio

fkleuver:
# Pull Request

## 📖 Description

This PR lets CircleCI use verdaccio to publish the packages to a local proxy, so that the e2e tests can run against the regular example apps which are installed from that proxy. This removes the need for the triple duplicated wdio test folders we had under `tests/wdio`, and thus 15 of the 20 example apps could be removed.

In addition, the publish-dev and publish-release scripts in the CircleCI config are trimmed down to only doing the publish (instead of also testing after), since things are already verified against a "properly" published package. This simplifies the CircleCI config as a whole.

### 🎫 Issues

N/A

## 👩‍💻 Reviewer Notes

- Reuses the verdaccio orb I created for vCurrent at https://github.com/aurelia/circleci/blob/master/src/common.yml
- Adds a few minor config flags to the bump/publish scripts to allow adding an arbitrary suffix to the version, as well as specifying a registry to publish to. These changes are under files in the `scripts` folder.

## 📑 Test Plan

From a testing perspective for the regular flow, no tests were added/removed. The e2e tests now require no special effort to run on CI

## ⏭ Next Steps

- There is a good chance the real release script is broken at this point after all the changes (and no release publish) in the last few months. This still needs to be verified. Also, it could still be useful to add tests back in specifically for targeting the `latest` channel
- Consider moving (or duplicating) the other apps under /test to /examples and include them in e2e tests. The downside of moving them, means we lose the capability to quickly and easily test local changes against those apps as the additional infra will be gone. The downside of duplicating them is, well, duplication :) need to think about this


fkleuver:
Since these are purely CI/workflow related changes and not related to framework code, I'll go ahead and merge without peer review so I can continue with other stuff. The failing tests have been failing for a while now, I'll be looking to fix those next.
Please do let me know if there are any questions/concerns about this direction.
CC @EisenbergEffect @bigopon @BBosman @jwx @Vheissu @rowellx68

PR #472, opened 2019-04-01
chore: pin typescript and harmonize renovate deps

BBosman:
# Pull Request

## 📖 Description

Pin typescript and fix some `renovate` warnings.

### 🎫 Issues

* https://github.com/Microsoft/TypeScript/issues/30662
* #471 

## 👩‍💻 Reviewer Notes

Due to Typescript introducing breaking changes in version `3.4.0`, for which there is no clean workaround, we need to stick to `~3.3.4000` for the foreseeable future.

Another (minor) change is that I moved some `devDependencies` to `dependencies` in the root `package.json`, as they're specified as `dependencies` in the individual packages, which was causing those versions to go out of sync as we have different upgrade policies for `devDependencies` and `dependencies`.

Once this is merged #471 should auto-update after the next run of the `renovate` job.

## 📑 Test Plan

CircleCI

## ⏭ Next Steps

Hope the `typescript` folks fix this one way or another.

### DIFF
--- a/examples/jit-pixi-webpack-ts/package.json
+++ b/examples/jit-pixi-webpack-ts/package.json
@@ -14,7 +14,7 @@
     "@aurelia/runtime-html": "file:../../packages/runtime-html",
     "@aurelia/runtime-pixi": "file:../../packages/runtime-pixi",
     "@aurelia/runtime": "file:../../packages/runtime",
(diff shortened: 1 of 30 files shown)


fkleuver:
Excellent, thanks for quickly fixing this up :)

PR #483, opened 2019-06-01
Dynamic routing: refactor navigator

jwx:
# Pull Request

## 📖 Description

This PR contains the first part of the refactor towards navigator and browser navigation. It primarily puts the new navigator in charge of queueing as well as informing the browser. It's one big step closer to not having the router relying on the browser.

### 🎫 Issues

## 👩‍💻 Reviewer Notes

Ignore everything in `packages/router/test/e2e/**` since they are just my own test applications while working. The 30 failing tests are not related to this PR or the router.

A working demo app can be found in `packages/router/test/e2e/doc-example/app` and [on now](https://dist-8flswg4j4.now.sh).

## 📑 Test Plan

- Make sure tests still pass.

## ⏭ Next Steps

- Review the extension points
- Re-visit scoped viewports


packages/router/.vscode/settings.json
```diff
@@ -1,6 +1,63 @@
+// Place your settings in this file to overwrite default and user settings.
```
fkleuver: Why did you add this file? 

<a id="ef7ca840-88f5-11e9-a7e3-ab5e107a4652" href="none" codereview-uuid-reserved-tag />

packages/router/src/browser-navigation.ts
```diff
@@ -13,37 +38,47 @@ interface QueueItem {
...
 }
 
-export class QueuedBrowserHistory implements QueuedBrowserHistory {
+export class BrowserNavigation implements INavigationStore, INavigationViewer {
   public window: Window;
   public history: History;
+  public location: Location;
+
+  public useHash: boolean;
+  public allowedNoOfExecsWithinTick: number;
```
fkleuver: What is an "Exec" and what is a "Tick"? Please add a comment

packages/router/src/browser-navigation.ts
```diff
@@ -13,37 +38,47 @@ interface QueueItem {
...
     this.isActive = false;
     this.currentHistoryActivity = null;
+    this.unticked = 0;
     this.callback = null;
     this.goResolve = null;
     this.suppressPopstateResolve = null;
   }
 
-  public activate(callback: (ev?: PopStateEvent) => void): void {
+  public activate(callback: (ev?: INavigationViewerEvent) => void): Promise<void> {
```
fkleuver: I think a method should either return a promise or accept a callback, but not both

packages/router/src/browser-navigation.ts
```diff
@@ -55,40 +90,44 @@ export class QueuedBrowserHistory implements QueuedBrowserHistory {
...
     if (!suppressPopstate) {
-      return this.enqueue(this, '_go', [delta], true);
+      const promise2 = this.enqueue(this, '_go', [delta], true);
+      this.dequeue().catch(error => { throw error; });
+      return promise2;
     }
     const promise = this.enqueue(this, 'suppressPopstate', [], true);
     this.enqueue(this.history, 'go', [delta]).catch(error => { throw error; });
+    this.dequeue().catch(error => { throw error; });
+    this.dequeue().catch(error => { throw error; });
```
fkleuver: Why 2x dequeue in a row?

packages/router/src/browser-navigation.ts
```diff
@@ -55,40 +90,44 @@ export class QueuedBrowserHistory implements QueuedBrowserHistory {
...
+    const { title, path } = state.NavigationEntry;
+    const promise = this.enqueue(this.history, 'pushState', [state, title, `#${path}`]);
+    this.dequeue().catch(error => { throw error; });
+    return promise;
   }
 
-  // tslint:disable-next-line:no-any - typed according to DOM
-  public async pushState(data: any, title: string, url?: string | null): Promise<void> {
-    return this.enqueue(this.history, 'pushState', [data, title, url]);
+  public async replace(state: INavigationState): Promise<void> {
```
fkleuver: Why not keep this named `replaceState`?

packages/router/src/browser-navigation.ts
```diff
@@ -113,6 +160,16 @@ export class QueuedBrowserHistory implements QueuedBrowserHistory {
     this.history.go(delta);
   }
 
+  private async _popState(resolve: ((value?: void | PromiseLike<void>) => void)): Promise<void> {
```
fkleuver: We shouldn't have underscore prefixed method names imo, better to have a more descriptive distinct name

packages/router/src/browser-navigation.ts
```diff
@@ -13,37 +38,47 @@ interface QueueItem {
...
+  public location: Location;
+
+  public useHash: boolean;
+  public allowedNoOfExecsWithinTick: number;
 
   private readonly queue: QueueItem[];
   private isActive: boolean;
   private currentHistoryActivity: QueueItem;
-  private callback: (ev?: PopStateEvent) => void;
+  private unticked: number;
```
fkleuver: unticked what?

packages/router/src/browser-navigation.ts
```diff
@@ -55,40 +90,44 @@ export class QueuedBrowserHistory implements QueuedBrowserHistory {
...
+    this.dequeue().catch(error => { throw error; });
+    return promise;
   }
 
-  // tslint:disable-next-line:no-any - typed according to DOM
-  public async replaceState(data: any, title: string, url?: string | null): Promise<void> {
-    return this.enqueue(this.history, 'replaceState', [data, title, url]);
+  public async pop(): Promise<void> {
+    const promise = this.enqueue(this, '_popState', []);
+    this.dequeue().catch(error => { throw error; });
```
fkleuver: I don't really like this. You're first storing the promise returned from `enqueue` (not awaiting it), then calling `dequeue` which returns a promise, but you're not awaiting that either. And then you're only returning the `enqueue` promise.

This is a huge red flag to me and is a similar vague kind of parallel asynchrony that's going on in vCurrent's router sometimes, and one of the main reasons for it being to incredibly prone to regressions and hard to test/debug.

Can/should these promises not be chained? And why make the method `async` when nothing is awaited?

packages/router/src/instruction-resolver.ts
```diff
@@ -94,10 +94,10 @@ export class InstructionResolver {
     return href;
   }
 
-  public shouldClearViewports(path: string): { clearViewports: boolean; newPath: string } {
+  public shouldClearViewports(path: string): { clear: boolean; newPath: string } {
```
fkleuver: Why did you shorten this name?

packages/router/.vscode/settings.json
jwx: I was opening the router folder directly when doing this refactor and wanted the settings to tag along. Should of course have ignored and then removed it. Gone now.

packages/router/src/navigator.ts
```diff
@@ -0,0 +1,240 @@
...
+    this.entries = null;
+    this.pendingNavigations = new Queue<INavigationInstruction>(this.processNavigations);
+
+    this.options = null;
+    this.isActive = false;
+  }
+
+  public activate(options?: INavigatorOptions): void {
+    if (this.isActive) {
+      throw Reporter.error(0); // TODO: create error code for 'Navigator has already been activated.'
```
fkleuver: Just throw an error with that message tbh. It was never a good idea of me to use these temporary `0` codes. We have too many of those now and they're actually becoming annoying lol

packages/router/src/navigator.ts
```diff
@@ -0,0 +1,240 @@
...
+    this.currentEntry.title = title;
+    return this.saveState();
+  }
+
+  get titles(): string[] {
+    return (this.entries ? this.entries.slice(0, this.currentEntry.index + 1).map((value) => value.title) : []);
+  }
+
+  public getState(): INavigatorState {
+    const state = { ...this.options.store.state };
```
fkleuver: Why a shallow clone instead of just returning the `state` itself?

packages/router/src/navigator.ts
```diff
@@ -0,0 +1,240 @@
...
+    const entries = (state.NavigationEntries || []) as IStoredNavigationEntry[];
+    const currentEntry = state.NavigationEntry as IStoredNavigationEntry;
+    return { state, entries, currentEntry };
+  }
+
+  public loadState(): void {
+    const state = this.getState();
+    this.entries = (state.entries || []);
```
fkleuver: Redundant parentheses

packages/router/src/navigator.ts
```diff
@@ -0,0 +1,240 @@
...
+    this.currentEntry = state.currentEntry;
+  }
+
+  public async saveState(push: boolean = false): Promise<void> {
+    const storedEntry = this.storableEntry(this.currentEntry);
+    this.entries[storedEntry.index] = storedEntry;
+    const state: INavigationState = {
+      'NavigationEntries': this.entries,
+      'NavigationEntry': storedEntry,
```
fkleuver: No need for string keys here since they're valid identifiers

packages/router/src/navigator.ts
```diff
@@ -0,0 +1,240 @@
...
+    };
+    if (push) {
+      return this.options.store.push(state);
+    } else {
+      return this.options.store.replace(state);
+    }
+  }
+
+  public storableEntry(entry: INavigationInstruction): IStoredNavigationEntry {
```
fkleuver: Shouldn't this be named `getStoredNavigationEntry`?

packages/router/src/navigator.ts
```diff
@@ -0,0 +1,240 @@
...
+
+      ...storableEntry } = entry;
+    return storableEntry;
+  }
+
+  public finalize(instruction: INavigationInstruction): void {
+    this.currentEntry = instruction;
+    if (this.currentEntry.untracked) {
+      if (instruction.fromBrowser) {
+        this.options.store.pop().catch(error => { throw error; });
```
fkleuver: Why not awaiting/chaining this?

packages/router/src/navigator.ts
```diff
@@ -0,0 +1,240 @@
...
+      }
+      this.currentEntry.index--;
+      this.entries[this.currentEntry.index] = this.storableEntry(this.currentEntry);
+      this.saveState().catch(error => { throw error; });
```
fkleuver: Why not awaiting/chaining this?

packages/router/src/navigator.ts
```diff
@@ -0,0 +1,240 @@
...
+      if (instruction.navigation.new) {
+        this.options.store.pop().catch(error => { throw error; });
+      } else {
+        this.options.store.go(-instruction.historyMovement, true).catch(error => { throw error; });
+      }
+    }
+    this.currentEntry.resolve();
+  }
+
+  private callback(entry: INavigationEntry, navigationFlags: INavigationFlags, previousEntry: INavigationEntry): void {
```
fkleuver: I don't think this is the right name for this method. Shouldn't this be something like `invokeCallback`?

packages/router/src/queue.ts
```diff
@@ -0,0 +1,85 @@
+import { PLATFORM, Reporter } from '@aurelia/kernel';
+
+export interface QueueItem<T> {
+  resolve?: ((value?: void | PromiseLike<void>) => void);
+  reject?: ((value?: void | PromiseLike<void>) => void);
+}
+
+export class Queue<T> {
```
fkleuver: What kind of queue is this? If it's a general-purpose queue it could live in the runtime or kernel, but it isn't.

Maybe give this a more precise name?

packages/router/src/queue.ts
```diff
@@ -0,0 +1,85 @@
...
+  public isActive: boolean;
+  public readonly pending: QueueItem<T>[];
+  public processing: QueueItem<T>;
+  public tickLimit: number;
```
fkleuver: What is `tickLimit`?

packages/router/src/router.ts
```diff
@@ -136,66 +143,64 @@ export class Router implements IRouter {
...
     }
-    this.historyBrowser.setHash(href);
+    this.goto(href).catch(error => { throw error; });
   }
 
-  public historyCallback(instruction: INavigationInstruction): void {
-    this.pendingNavigations.push(instruction);
-    this.processNavigations().catch(error => { throw error; });
+  public navigatorCallback = (instruction: INavigationInstruction): void => {
+    this.processNavigations(instruction).catch(error => { throw error; });
```
fkleuver: Yeah, I'm seeing a lot of these `.catch` to get rid of the "floating promise" thing but I really don't see that as a proper solution, because not awaiting or chaining them means they all run in parallel. That can't be fully intentional on so many places. Can you add justifying comments where the parallelism is indeed intentional and why? This will be very hard to understand for others int he future otherwise

packages/router/src/router.ts
```diff
@@ -136,66 +143,64 @@ export class Router implements IRouter {
...
       // }
     }
 
-    let path = instruction.path;
-    if (this.options.transformFromUrl && !fullStateInstruction) {
-      const routeOrInstructions = this.options.transformFromUrl(path, this);
-      // TODO: Don't go via string here, use instructions as they are
-      path = Array.isArray(routeOrInstructions) ? this.instructionResolver.stringifyViewportInstructions(routeOrInstructions) : routeOrInstructions;
-    }
+    let views, clearViewports;
```
fkleuver: One variable declaration per line, and explicit types please. This is not vCurrent :)

packages/router/src/browser-navigation.ts
jwx: Would 
// Limit no of executed actions within the same PLATFORM.Tick (due to browser limitation)
be enough?

fkleuver: What is an "executed action" in this context? A navigation?

jwx: A browser action through the History and Location api. Chrome (at least) ignores anything over a certain amount within a certain time.

packages/router/src/browser-navigation.ts
jwx: They are two different things. The `callback` is an option to the activation (the callback function) and the promise is to signal that the `BrowserNavigation` has finished activation (which includes triggering an initial (fake) popstate event).

packages/router/src/browser-navigation.ts
jwx: Didn't want to confuse them with the native `replace/pushState` methods. Want me to change it?

packages/router/src/browser-navigation.ts
jwx: This is refactored, but your original concerns may still apply. Take a look and let me know.

packages/router/src/instruction-resolver.ts
jwx: To make it more clear? ;) Not sure. I guess it just felt the `Viewport` part was redundant. Want me to change it back?

packages/router/src/navigator.ts
jwx: To avoid common mistakes with a memory store (but still retain the possibility to have actual objects stored). Might not add that much. Remove it?

packages/router/src/navigator.ts
jwx: It's pretty much a conversion and I tend to not `get` those. Do you want me to?

packages/router/src/navigator.ts
jwx: It's all thrown into a queue. But I've added awaits.

packages/router/src/router.ts
jwx: Most of these are queued so they won't be running in parallel anyway. I've awaited in some places, but here there's really no point since the router will only deal with one instruction at a time.

packages/router/src/router.ts
jwx: Ooops.

packages/router/src/queue.ts
jwx: I'd appreciate help with giving this a more precise name.

packages/__tests__/router/instruction-resolver.spec.ts
```diff
@@ -103,6 +104,36 @@ describe('InstructionResolver', function () {
   }
 });
 
+const setup = async (): Promise<{ au; container; host; router }> => {
+  const container = BasicConfiguration.createContainer();
+
+  const App = (CustomElementResource as any).define({ name: 'app', template: '<template><au-viewport name="left"></au-viewport><au-viewport name="right"></au-viewport></template>' });
```
fkleuver: `as any` cast shouldn't be necessary

packages/__tests__/router/instruction-resolver.spec.ts
```diff
@@ -103,6 +104,36 @@ describe('InstructionResolver', function () {
...
+  container.register(Router as any);
```
fkleuver: `as any` cast shouldn't be necessary

packages/__tests__/router/instruction-resolver.spec.ts
```diff
@@ -103,6 +104,36 @@ describe('InstructionResolver', function () {
   }
 });
 
+const setup = async (): Promise<{ au; container; host; router }> => {
```
fkleuver: Can we make this `async function setup()` please? That would keep it consistent with the rest of the project. Also, you can rely on type inference so you don't necessarily explicitly need to specify the return value

packages/__tests__/router/instruction-resolver.spec.ts
```diff
@@ -103,6 +104,36 @@ describe('InstructionResolver', function () {
...
+  const container = BasicConfiguration.createContainer();
+
+  const App = (CustomElementResource as any).define({ name: 'app', template: '<template><au-viewport name="left"></au-viewport><au-viewport name="right"></au-viewport></template>' });
+  container.register(Router as any);
+  container.register(ViewportCustomElement as any);
+
+  const host = document.createElement('div');
```
fkleuver: Don't use globals please. You can use the `TestContext` for this:

```ts
const ctx = TestContext.createHTMLTestContext();

const host = ctx.createElement('div');
ctx.doc.body.appendChild(host);
// ...
```

packages/__tests__/router/queue.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+      const animal = qAnimal as Animal;
+      console.log('Animal', animal);
+      await wait(100);
+      if (animal.name === 'dog') {
+        qAnimal.reject();
+      } else {
+        qAnimal.resolve();
+      }
+    });
+    q.enqueue(new Animal('dog', 'Pluto') as Animal);
```
fkleuver: I can't imagine this `as Animal` cast is necessary immediately after calling the constructor

packages/__tests__/router/e2e/doc-example/app/src/app.ts
```diff
@@ -45,6 +45,9 @@ export class App {
   public color: string = 'lightgreen';
   constructor(private readonly router: Router, authorsRepository: AuthorsRepository, private readonly state: State) {
     authorsRepository.authors(); // Only here to initialize repositories
+  }
+
+  public bound() {
```
fkleuver: Why did you need to move this to `bound`? Did you try `binding`? I know it's just an example app, but it's also e2e test in the making. Would like to try to keep things tight.

packages/__tests__/router/instruction-resolver.spec.ts
```diff
@@ -1,6 +1,7 @@
+import { BasicConfiguration } from '@aurelia/jit-html-browser';
```
fkleuver: Don't import anything from the `-browser` packages in tests please. Use the `TestContext` to get a configured container etc. Otherwise the tests will fail in node.

packages/router/src/browser-navigation.ts
fkleuver: Could call it `replaceNavigationState` then. Just a bit more specific. "verb + noun" is generally preferable over just either verb or just noun.

packages/router/src/navigator.ts
fkleuver: Well, if someone decided to take this object and put stuff on it or observe properties, it wouldn't stay in sync. It depends on what the intended purpose is. Just have it be a deliberate, justified decision over an arbitrary one

packages/router/src/navigator.ts
fkleuver: If it's a conversion then i'd prefix it with `to` instead of `get` yes, but still prefix it with a verb

packages/router/src/router.ts
fkleuver: But what happens if `navigatorCallback` is called twice in a row synchronously? Something is going to happen in parallel then..

packages/router/src/queue.ts
```diff
@@ -0,0 +1,105 @@
...
+    if (!this.isActive) {
+      throw new Error('Queue has not been activated');
+    }
+    this.lifecycle.dequeueRAF(this.dequeue, this);
+    this.tickLimit = null;
+    this.clear();
+    this.isActive = false;
+  }
+
+  public enqueue(itemOrItems: T | T[], costOrCosts?: number | number[]): Promise<void> | Promise<void>[] {
```
fkleuver: What is `cost` here?

Also, could you use method overloads here to clarify which combinations are possible? I'm assuming either all must be arrays or non-arrays?

So:

```ts
public enqueue(item: T, cost?: number): Promise<void;
public enqueue(items: T[], costs?: number[]): Promise<void>[];
public enqueue(itemOrItems: T | T[], costOrCosts?: number | number[]): Promise<void> | Promise<void>[] {
```

packages/router/src/router.ts
jwx: It can't be (not unless something else is critically wrong) since it's the callback from the Navigator and the Navigator's got a Queue to make sure one navigation instruction is processed at a time.

packages/router/src/queue.ts
jwx: (You can also set the same `cost` for each item in a list of items.)

packages/router/src/navigator.ts
jwx: Yeah, that behavior of not being in sync was intentional. Keep it as it is for now?

packages/__tests__/router/e2e/doc-example/app/src/app.ts
jwx: Moved to `binding`.

packages/router/src/queue.ts
jwx: It's now `allowedExecutionCostWithinTick`. Better?

packages/router/src/navigator.ts
fkleuver: Why was it intentional?

packages/router/src/queue.ts
fkleuver: What is `ExecutionCost`?

packages/router/src/router.ts
fkleuver: Then please add this clarification as a comment

packages/router/src/queue.ts
fkleuver: And what *is* the `cost`?

packages/router/src/router.ts
```diff
@@ -144,66 +149,66 @@ export class Router implements IRouter {
       const context = scope.scopeContext();
       href = this.instructionResolver.buildScopedLink(context, href);
     }
-    this.historyBrowser.setHash(href);
+    this.goto(href).catch(error => { throw error; });
```
fkleuver: What if you click another link before the previous is handled? Please add small justifying comment for this floating promise, e.g. what happens if it's called multiple times in parallel

packages/router/src/router.ts
fkleuver: Is there a guarantee that this method is never called before the previous has finished? If so, what guarantees it? The comment isn't clear enough

packages/router/src/router.ts
```diff
@@ -327,9 +331,8 @@ export class Router implements IRouter {
         this.addedViewports.push(new ViewportInstruction(componentOrInstruction, viewport));
       }
     } else if (this.lastNavigation) {
-      this.pendingNavigations.unshift({ path: '', fullStatePath: '', isRepeat: true });
+      this.navigator.navigate({ instruction: '', fullStateInstruction: '', repeating: true }).catch(error => { throw error; });
```
fkleuver: This is public API, so what if `addProcessingViewport` is called before `navigator.navigate` is done?

packages/router/src/navigator.ts
```diff
@@ -0,0 +1,249 @@
...
+      if (instruction.fromBrowser) {
+        await this.options.store.popNavigationState();
+      }
+      this.currentEntry.index--;
+      this.entries[this.currentEntry.index] = this.toStorableEntry(this.currentEntry);
+      await this.saveState();
+    } else if (this.currentEntry.replacing) {
+      this.entries[this.currentEntry.index] = this.toStorableEntry(this.currentEntry);
+      await this.saveState();
+    } else {
```
fkleuver: What is the implied state of `currentEntry` in this `else` branch?

packages/router/src/browser-navigation.ts
```diff
@@ -0,0 +1,227 @@
...
+    }
+
+    calls.push({
+      target: target,
+      methodName: methodName,
+      parameters: parameters,
+    });
+    costs.push(1);
+
+    promises.push(this.pendingCalls.enqueue(calls, costs)[0]);
```
fkleuver: Can you store the result of `this.pendingCalls.enqueue(calls, costs)` in a variable with a meaningful name, and then also do the same for the first item of that array? I'm not sure what the semantic meaning is of this promise that you're adding to `promises` here.

packages/router/src/browser-navigation.ts
```diff
@@ -0,0 +1,227 @@
...
+    return promises[0];
```
fkleuver: What is the meaning of this first promise? Can you clarify why the first promise, and not the last for example?

packages/router/src/browser-navigation.ts
```diff
@@ -0,0 +1,227 @@
...
+    resolve();
+  }
+
+  private forwardState(state: ForwardedState): void {
+    this.forwardedState = state;
+  }
+
+  // Everything that wants to await a browser event should pass suppressPopstate param
+  // Events NOT resulting in popstate events should NOT pass suppressPopstate param
+  private enqueue(target: object, methodName: string, parameters: unknown[], suppressPopstate?: boolean): Promise<void> {
```
fkleuver: What is `target`?

packages/router/src/browser-navigation.ts
```diff
@@ -0,0 +1,227 @@
...
+  private callback: (ev?: INavigationViewerEvent) => void;
+
+  private forwardedState: ForwardedState;
+
+  constructor(
+    lifecycle: ILifecycle,
+  ) {
+    this.lifecycle = lifecycle;
+
+    this.window = DOM.window;
```
fkleuver: Please inject `HTMLDOM` (interface symbol is `IDOM` from `@aurelia/runtime-html`) and use the `wnd` (or `window`, can't remember) property from that instead. Makes it easier to mock/stub stuff.

packages/__tests__/router/instruction-resolver.spec.ts
```diff
@@ -103,15 +100,28 @@ describe('InstructionResolver', function () {
...
+  const App = CustomElementResource.define({ name: 'app', template: '<template><au-viewport name="left"></au-viewport><au-viewport name="right"></au-viewport></template>' });
+  container.register(Router);
+  container.register(ViewportCustomElement);
+
+  const host = ctx.createElement('div');
+  ctx.doc.body.appendChild(host);
+
+  const au = window['au'] = new Aurelia(container)
+    .register(DebugConfiguration, RouterConfiguration)
+    .app({ host: host, component: App })
```
fkleuver: FYI the result of `au.start()` is actually an `ILifecycleTask`, not `Aurelia`. So you'd need to do this:

```ts
const au = window['au'] = new Aurelia(container)
    .register(DebugConfiguration, RouterConfiguration)
    .app({ host: host, component: App });

await au.start().wait();
```

Also, please get rid of the `window` global already lol

packages/router/src/router.ts
jwx: It's not really intended as a public API. But to answer the question: calling `addProcessingViewport` while a navigation is in process will add the viewport instruction to that navigation.

packages/router/src/navigator.ts
jwx: An "ordinary" new entry. I've added a short comment.

packages/router/src/browser-navigation.ts
jwx: The object in which the method is called. Suggestion for a better name?

packages/router/src/browser-navigation.ts
jwx: It kind of already is the semantic meaning. Storing in more variables would require more logic. I've added a comment. See if that's enough?

packages/router/src/browser-navigation.ts
jwx: Added a comment.

packages/router/src/navigator.ts
jwx: So that modifying the object wouldn't affect the stored one (shallowly).

packages/router/src/router.ts
jwx: Come to think of it, it could very well be used as a public API.

packages/router/src/queue.ts
jwx: I've added a description to the Queue class.

packages/router/src/browser-navigation.ts
fkleuver: The method is a lifecycle hook? `viewModel` maybe? I exposed an interface `IViewModel` for that btw

jwx: It can be anything, really. Currently it's methods in the History api and in BrowserNavigation itself.

fkleuver: So:

```ts
private enqueue(
  target: History | IBrowserNavigation,
  methodName: keyof History | keyof IBrowserNavigation,
  parameters: unknown[],
  suppressPopstate?: boolean,
): Promise<void> {
```

?

PR #488, opened 2019-06-17
portal

fkleuver:
# Pull Request


packages/runtime-html/src/resources/custom-attributes/portal.ts
```diff
@@ -0,0 +1,340 @@
...
+        if (oldValue !== newValue) {
+          $this._renderContext = newValue;
+          $this.targetChanged(newValue, oldValue, flags | $this.$controller.flags);
+        }
+      },
+    };
+  })();
+
+  // tslint:disable-next-line: member-ordering
+  public readonly $observers: InlineObserversLookup<IAccessor> = Object.freeze({
```
bigopon: This looks pretty short-circuited

packages/runtime-html/src/resources/custom-attributes/portal.ts
```diff
@@ -0,0 +1,340 @@
...
+  }
+
+  private activate(target: T, flags: LifecycleFlags): ILifecycleTask {
+    const { activating, activated, callbackContext, view } = this;
+    let task = this.task;
+
+    view.hold(target);
+
+    if ((this.$controller.state & State.isAttachedOrAttaching) === 0) {
+      return task;
```
bigopon: this means if it's attaching, returns the attaching task, else, task.done?

packages/runtime-html/src/resources/custom-attributes/portal.ts
```diff
@@ -0,0 +1,340 @@
...
+
+    if (this.dom.isNodeInstance(target)) {
+      return target;
+    }
+
+    if (target == null) {
+      if (this.strict) {
+        throw new Error('Render target not found');
+      } else {
+        target = this.dom.document as unknown as ResolvedTarget<T>;
```
bigopon:
```suggestion
        target = this.dom.document.body as unknown as ResolvedTarget<T>;
```

packages/runtime/src/index.ts
```diff
@@ -435,6 +435,7 @@ export {
   ISubscriberCollection,
   ObservedCollection,
   ObserversLookup,
+  InlineObserversLookup,
```
bigopon: 🤔

packages/runtime-html/src/resources/custom-attributes/portal.ts
fkleuver: It is. Doing the same thing in the other resources (check `repeat` and `if`). For bindings that cannot be / don't need to be two-way, the only contract that must be fulfilled for bindings to work is `setValue` / `getValue`. Doing this is actually a very significant perf boost. Normally AOT would do this though, so perhaps when AOT is far enough along we can clean this up again and turn them into simple `@bindable`s.
For now it's also a bit of an exercise to become more familiar with the desired AOT output

packages/runtime-html/src/resources/custom-attributes/portal.ts
fkleuver: This means "if it's NOT attaching or attached, abort right here". Simply return the existing task (which will most likely be a `task.done`). It's similar to the `binding` guard here https://github.com/aurelia/aurelia/pull/488/files#diff-0251d569108c58da0a1101d4101b2ae4R223

packages/runtime/src/index.ts
fkleuver: Yeah it's a bit rough around the edges, still WIP. Haven't quite yet figured out a good API for manually instantiating observers lookup in the various scenarios

fkleuver:
@bigopon @bigopon @bigopon

bigopon:
⌨️ ⌨️ ⌨️

bigopon:
@fkleuver ready to go, but there are 2 failings tests related to ref bindings, which I'm not sure has anything to do with portal or not. This error: `Code 0: Cannot add observers to object`

fkleuver:
@bigopon Can you just amend the title of that `feat(portal): wip` commit? "wip" is not a feature :P

bigopon:
@fkleuver done. Do you need a todo so that later you know to come back here once render location refactoring is done?

fkleuver:
@bigopon I'm thinking about pushing that back a little bit and getting AOT going first. Is it a blocker for you?

bigopon:
No noy ay all. Please gp ahead

AureliaEffect:
JS Framework Benchmark: https://24907-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://25514-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/runtime/src/templating/controller.ts
```diff
@@ -109,6 +112,15 @@ const defaultControllerHoldOptions: IControllerHoldOptions = {
...
 };
 
+function bindingSorter(b1: IBinding, b2: IBinding): number {
+  const b1Order = typeof b1.order === 'number' ? b1.order || 0 : 0;
+  const b2Order = typeof b2.order === 'number' ? b2.order || 0 : 0;
+  // the behavior of ECMA sort with 0 return value is undefined
+  // so treating 0 as -1, which should mean keeping the same order
+  // todo: should this be normalize to only 1 | -1 for better browser JIT optimization?
+  return b1Order !== 0 && b2Order !== 0 ? (b1Order - b2Order) || -1 : -1;
+}
```
fkleuver: Suggestion:

```ts
function bindingSorter(b1: IBinding, b2: IBinding): number {
  // Note 1: we treat undefined as 0 so negative orders are sorted before undefined ones
  // Note 2: we treat a sort return value of 0 as -1 to ensure the original order is kept if 2 values are the same
  // Note 3: b1 and b2 are often polymorphic, so we store the properties here to make
  // the rest of the function monomorphic. We should be able to tell that this works;
  // the two lines below should show up with significantly higher samples in a profiler report
  // than the rest, when there are many different binding types, due to lookups.
  const o1 = b1.order;
  const o2 = b2.order;
  if (o1 === void 0) {
    if (o2 === void 0) {
      return -1;
    }

    return o2 === 0 ? -1 : -o2;
  }

  if (o2 === void 0) {
    return o1 === 0 ? -1 : o1;
  }

  // We could start with this comparison to make the function a bit shorter,
  // but both values will always be numbers here which more optimizeable
  return o1 === o2 ? -1 : o1 - o2;
}
```

packages/runtime/src/templating/controller.ts
```diff
@@ -539,6 +551,13 @@ export class Controller<
     }
   }
 
+  public reorderBindings(): void {
+    if (this.bindings !== void 0) {
+      (this as any)['sorted'] = true;
```
fkleuver: Just define this property in the ctor and init with `false`, otherwise you're making the lexical `this` polymorphic.

bigopon: oopsie, debug code

packages/runtime/src/templating/controller.ts
bigopon: I know it's garbage in, garbage out, but it could potentially go wrong in ways we don't know so I was using `typeof` op to ensure correctness of the sort operation. If it's garbage in garbage out then i'm fine with this change

bigopon:
Not really keen on resolving conflicts of this one, creating new PR

PR #489, opened 2019-06-18
Fixed typos in router.spec.ts

thephoenixofthevoid:
# Pull Request

## 📖 Description
It seems two colons are missing in one of the test, making these tests always pass no matter what.
Should here be used the positive (equals to) checks instead of negative (not equals to) ones?

### 🎫 Issues
Two typos and two useless test assertions as the very result.

## 👩‍💻 Reviewer Notes
Probably, this should be fixed by changing assertions to checking what values are, not that they no longer have some value.

## 📑 Test Plan
Not Applicable 

## ⏭ Next Steps
Explained above

### DIFF
--- a/packages/router/test/unit/router.spec.ts
+++ b/packages/router/test/unit/router.spec.ts
@@ -57,8 +57,8 @@ describe('Router', function () {
     await goto('bar@right', router);
     expect(host.textContent).to.contain('Viewport: bar');
     await goto('-', router);
-    expect(host.textContent).to.not.contain('Viewport foo');
-    expect(host.textContent).to.not.contain('Viewport bar');
+    expect(host.textContent).to.not.contain('Viewport: foo');
+    expect(host.textContent).to.not.contain('Viewport: bar');
 
     await teardown(host, router, 1);
   });


fkleuver:
@thephoenixofthevoid Thanks, good catch! :) Could you just merge latest master into your branch? I mistakenly left CI disabled a day ago, it's enabled again. Then we'll be able to tell how this affects the tests.

thephoenixofthevoid:
Merged. You probably need to change the type of assertion. To be honest, I have never worked with the library and don't know what exact value happens to be there. I'd say it's more of an academic interest to learn different approaches by reading source codes.

thephoenixofthevoid:
There are few change suggestions that would make code more clear and reusable. If you are okey with the fact that I may completely miss the point, because of never actually using that code, I will share my thoughts.

fkleuver:
@thephoenixofthevoid Hey, master is not yet merged actually:
image

Make sure you merge it from the correct remote :)

>  I'd say it's more of an academic interest to learn different approaches by reading source codes.

> There are few change suggestions that would make code more clear and reusable. If you are okey with the fact that I may completely miss the point, because of never actually using that code, I will share my thoughts.

Sure! By all means go ahead. A slight mess has started to accumulate here and there over the last few months, so a fresh pair of eyes is certainly welcome.

Anything in particular you're looking to learn/accomplish?

thephoenixofthevoid:
Now I double checked. I am actually doing these commits on my smartphone through the github UI. What I like to propose is changes to Scope class in router package. I will craft commit on a laptop in a few hours or even tomorrow.

Basically, it boils down to two things:
1) By providing a modifier (private, public or maybe readonly) before a constructor argument, we are automatically assigning it to the instance under the key which is the name of the argument variable.
2) Then we might assign values to remaining state properties right where  they are defined, again not littering in the constructor.

fkleuver:
Hey, I just noticed this is actually in a file that was moved to `packages/__tests__/router/...`. Could you merge/rebase/reopen and do the change in the appropriate place please? Thanks!

thephoenixofthevoid:
Here: https://github.com/aurelia/aurelia/pull/491

fkleuver:
Closing this then as it's superseded by https://github.com/aurelia/aurelia/pull/491

PR #491, opened 2019-06-20
Fixed typos in router.spec.ts

thephoenixofthevoid:
# Pull Request

## 📖 Description
It seems two colons are missing in one of the test, making these tests always pass no matter what.
Probably, this should be fixed by changing assertions to checking what values are, not that they no longer have some value.

### 🎫 Issues
see also [https://github.com/aurelia/aurelia/pull/489](url)


fkleuver:
Thanks, I'll merge this so @jwx can look into fixing it

PR #496, opened 2019-06-25
fix(replaceable): make part scopes also work when not immediately bound from the wrapping replaceable

fkleuver:
# Pull Request

## 📖 Description

A more involved fix that addresses a deeper underlying issue with the "quick-n-dirty" part propagation logic I wrote earlier: the scope from the target `replace-part` was only bindable when every binding inside the `replace-part` was bound at the same time as the wrapping `replaceable`.

Now, by propagating the selected part name through the binding and AST (in a similar manner to how the flags flow through it), it will also work in a large number of edge cases that weren't accounted for earlier.

Most notably, when the content of a `replace-part` is wrapped inside an `if` which starts as `false`, when that `if` turns to `true` the scope of the inner binding would not have proper access to the part scope. This case (and probably many others, still need more tests) is now working.

## 👩‍💻 Reviewer Notes

A number of files were touched but the changes themselves are fairly repetitive in nature, so the code should be able to speak for itself.

## 📑 Test Plan

Added one test specifically for the `if` scenario mentioned above. This will need more combinations (possibly a few thousand actually).

## ⏭ Next Steps

Implement said tests.

@bigopon you may want to look at this and see if any of your commented-out tests work now

### DIFF
--- a/packages/__tests__/jit-html/replaceable.spec.ts
+++ b/packages/__tests__/jit-html/replaceable.spec.ts
@@ -1,4 +1,4 @@
-import { Aurelia, CustomElementResource } from '@aurelia/runtime';
+import { Aurelia, CustomElementResource, LifecycleFlags as LF } from '@aurelia/runtime';
 import { TestContext, assert } from '@aurelia/testing';
 
 describe('replaceable', function () {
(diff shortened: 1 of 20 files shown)


bigopon:
@fkleuver why 
> Most notably, when the content of a replace-part is wrapped inside an if which starts as false, when that if turns to true the scope of the inner binding would not have proper access to the part scope. This case (and probably many others, still need more tests) is now working.

was solved passing part name through out the system

fkleuver:
@bigopon 

If you recall my initial implementation, the part name was set on a static property on `BindingContext` during `binding` hook of `replaceable`.

It was bad and naive in so many ways it's hard to imagine I ever thought it was acceptable to commit that code. Well, it make *something* work, at least.. 😆 

Now, part names are stored in the bindings and on the controllers so that on change handlers they can be passed through the AST where the binding context can then use it to get the target scope.

PR #504, opened 2019-06-29
Renames

fkleuver:
# Pull Request

## 📖 Description

This PR contains 3 "types" of renames:

- The `Resource` suffix is removed from the resources; `CustomElementResource` becomes `CustomElement`, `BindingBehaviorResource` becomes `BindingBehavior`, etc.
- The AST classes which represent expressions (so not statements/bindings) have the `Expression` suffix added to them. Mainly to solve the naming conflict with `ValueConverter` and `BindingBehavior`, but also for consistency and just general naming. Some of these node class names were just too short and prone to naming conflict anyway, imo.
- Same for the `Binding` classes. `Binding` is now `PropertyBinding`, `Call` is `CallBinding`, etc.

Would like to get everyone's opinion on whether all of these changes are desired. They make sense to me, but perhaps not to others.

@EisenbergEffect @bigopon @jwx @davismj @BBosman


EisenbergEffect:
Should we post-fix the AST classes with `Node`? Would that be more expected? I don’t have a strong opinion on it myself. I like dropping the suffix for the resource classes and I like making the binding classes clearer as well.

bigopon:
`Binding` suffix shoulda been done properly in vCurrent. But they come with a cost for minification. For now is good but later we may need to revisit this I think

fkleuver:
@EisenbergEffect What about `ForOfStatement` or `BindingIdentifier` then? I only added `Expression` to the nodes that are, as per the spec, expressions. We could shorten it to `Expr` if the length is a concern though.

fkleuver:
@bigopon I think we can use AOT to properly shorten the class names for production builds.

EisenbergEffect:
I’m not a big fan of using abbreviated suffixes most of the time. So, a no vote from me on `Expr`. I like the other ideas though. @fkleuver Ultimate decision is up to you of course.

fkleuver:
@EisenbergEffect 
What I meant to say is that we have different kinds of AST nodes that reflect their respective counterparts in the ES spec. So:

- `BinaryExpression`
- `ForOfStatement`
- `BindingIdentifier`

Would be the "official" names.

Currently, we have:

- `Binary`
- `ForOfStatement`
- `BindingIdentifier`

So, appending `Node` to everything:

- `BinaryNode`
- `ForOfStatementNode`
- `BindingIdentifierNode`

Would be consistent in one respect (everything in the AST is a node, which is fair) but at the same time inconsistent (a `BinaryNode` is still a `Binary` node when technically it's a `BinaryExpression` node)

Likewise, I wouldn't do something like this either because then it becomes an incorrect name:

- `ForOfStatementExpression`
- `ForOfExpression`

I have seen another framework name their AST with the `Node` suffix (don't recall if it was Angular or something else) but I don't believe they had things that were not expressions. If we only had expressions, it could work.

So yeah, I think in terms of renaming the expressions, given that you don't like the abbreviation (I don't quite like it either), I think the current changeset has the only consistent enough representation.

fkleuver:
Alright, I'll go ahead and merge this then.

EisenbergEffect:
I very much like:

* BinaryExpression
* ForOfStatement
* BindingIdentifier

PR #517, opened 2019-07-11
chore(all): remove browser field in package.json

3cp:
Previous change on browser field to umd had small side-effect. I forgot webpack actually uses "browser" over "module" field, and it complains about static analysis bla bla, although it still can compile successfully. Removing browser field just makes all bundlers happier.


3cp:
@fkleuver any luck to make your develop branch head to pass CI?

3cp:
@fkleuver is going to remove "browser" and "jsnext:main" on master branch.

Issue #509, opened 2019-07-05
[RFC] Tooling: scaffolding, bundling, cli helper, support of Aurelia convention.


3cp:
# 💬 RFC

Tooling around Aurelia 2: scaffolding, bundling, cli helper commands, and support of Aurelia convention.

## 🔦 Context

At Aurelia v1's early time, complexity around scaffolding and bundling hurt our early adoption.

**Our target is to make sure Aurelia 2 provides smoothest possible tooling for our users.**

Current aurelia-cli is mainly designed for Aurelia v1, especially the built-in bundler. While aurelia-cli provides a solid user experience, the all-in-one approach made it hard to add feature, test, release, and slow to install. aurelia-cli is a huge three-in-one: scaffolding, a bundler, and convenient helper commands.

With Aurelia 2, we are going to modularize tooling parts, separate scaffolding, bundling, convenient helper commands. **But this is just internal design to us, our end users would not and should not see any difference.** The separation of concerns would allow us to improve and iterate each part faster, plus easier testing.

### 1. Scaffolding

Zero-install scaffolding. This will be enabled by a new GitHub repo aurelia/new, a skeleton-only repo for "makes".

```bash
npx makes aurelia
```

["makes"](https://makes.js.org) is a scaffolding tool I designed based on my experience on aurelia-cli. It enables zero-install scaffolding, speed, great flexibility and simplicity, plus easy try-out on feature branch. It's fully documented, read more if you are interested.

### 2. Bundling

Different from Aurelia v1, Aurelia 2 has no loader abstraction layer. This simplification has made Aurelia 2 naturally working with any bundler without any effort. We got few [examples](https://github.com/aurelia/aurelia/tree/master/examples) in Aurelia 2 repo.

The simplification comes with a price. With Aurelia 2, users would need to use explicit decorator to mark custom elements, custom attributes, value converters, and binding behaviours. This almost made Aurelia 2 code as verbose as Angular. But we have a plan (explained in 4. Aurelia convention) to bring Aurelia convention back through our tooling. In the end, end users will see Aurelia 2 uses convention to cut out boilerplate code just like Aurelia v1 did.

Let's keep Aurelia 2 setup clean. Let webpack be webpack, don't use gulp + webpack. Let browserify be browserify, keep it as simple one line command in npm scripts if possible.

The scaffolding and bundling will provide standard npm scripts.

* `npm start` to start dev server.
* `npm test` to run unit tests.
* `npm run build` to build production files.

[...]

Issue #116, opened 2018-08-30
[RFC] Require/Import in View Templates


3cp:
vNext dropped `<require>` tag for special syntax `@import`

```html
@import './hello.html';
<template>
  <!-- ... -->
</template>
```

I like the decision to move dependency declaration out of `<template>` tag. This will solve the annoying restriction that `<tr>` cannot have a child `<require>` element.

But the new syntax destroys my editor's syntax validation. Should we change the syntax to this? To keep code editor calm.

## Option1

```html
<require from="./hello.html"></require>

<template>
  <!-- ... -->
</template>
```

## Option2
Maybe even like this since we do not feed those header lines to browser native html parser.

```html
<require from="./hello.html">

<template>
  <!-- ... -->
</template>
```

## Option3
Or use attribute.

```html
<template require="./hello.html, ./hello.css">
  <!-- ... -->
</template>
```

## Cleaner variant of option3
```html
<template require="./hello.html" require="./hello.css">
  <!-- ... -->
</template>
```

EisenbergEffect:
Actually, option three is kind of nice. Not sure if it would get crowded though. I'm pretty open on syntax as long as it's something we can extract inside of a loader plugin without needing to include an entire html parser to do it. The trick is that this data has to be extracted before the template is sent to the compiler and we want to avoid double parsing the html.

huochunpeng:
auto-trace cli already has htmlparser2 installed to deal with `<require>` tag.
aurelia-webpack-plugin also has access to webpack's html parser for same purpose.

The parsing part is not a problem. The extra parsing is in compile time, I don't think we have double parsing in runtime.

huochunpeng:
I got your point now. I think you are talking about dynamic component loading at runtime (supported by jit app).

We can enforce **only support `require` attribute if it's the first attribute of `template` tag**. I think it's a good trade-off for both syntax and usability.

davismj:
Vote option 1, or option 2 as `<require from="./hello.html" />`.

huochunpeng:
Probably option1 fits better with Aurelia's spirit of standard compliant.

In HTML5 browser native parser, `<require from="./hello.html" />` is treated exactly same as `<require from="./hello.html">`, result is same as missing closing pair. But I like the compactness.

StrahilKazlachev:
If we keep any declarative form of dependencies declaration in the *view*, that *supports* css, we **need** to resolve the css order issue.
[...]

PR #521, opened 2019-07-16
feat: plugin conventions

3cp:
# Pull Request

## 📖 Description

This is ongoing work to implement Aurelia naming conventions at bundler level for Aurelia 2.
Aurelia v1 deals with conventions at runtime, while it's very flexible, it added up tooling complexity and runtime complexity.
Aurelia 2 runtime is very simple. We hook conventions (inject needed code) at bundler level even before the transpiler see the source code.

### 🎫 Issues

#116 Preprocess html view template
Part of #509 naming conventions.

## 👩‍💻 Reviewer Notes

This is an ongoing effort.
The first project is `@aurelia/plugin-conventions`, it's the base to be wrapped by all plugins (webpack, parcel, gulp...).
The plugin-conventions provides pure functions to preprocess js/ts resources and html view template. No fs API used. So it can run inside browser (if we ever need it in an online IDE).

Recommend to review the test code inside `packages/__test__/plugin-conventions`, to see if I covered enough edge cases.

Also, I am not a TS guy, the code didn't use much typing. But we want to get this working first, before improving the implementation.

### Note on html template

Since we preprocess it, we now can tolerate bad html from user.

```html
<import from="a"></import> <!-- this is correct html5 -->
<!-- following are incorrect html5, but we can accept them -->
<import from="a">
<import from="a" />
```

Both `<import>` and `<require>` tags are acceptable, both before or after `<template>`. In fact, we accept them everywhere in the html template, even after the last `</template>`.

Aurelia 2 finally avoided `<require>` + `<td>` issue, because `<require>`(or `<import>`) is not present in final html string, the deps were picked out by preprocessor.

## 📑 Test Plan

The html template preprocessor has option to write some typing for TS code, in order to avoid complain from following transpiling.

The js/ts preprocessor has no option for adding typing. We will see what's needed.

We need full app build to find out possible missed parts (typing is my biggest concern).

## ⏭ Next Steps

- [x] experimental webpack plugin


EisenbergEffect:
Very excited to see this getting started! 🎉

3cp:
@EisenbergEffect have a look of the test cases.

packages/__tests__/plugin-conventions/name-convention.spec.ts
```diff
@@ -0,0 +1,33 @@
...
+  });
+
+  it('gets custom attribute like resource', function() {
+    const result = nameConvention('FooBarCustomAttribute');
+    assert.deepEqual(result, {name: 'foo-bar', type: 'customAttribute'});
+  });
+
+  it('gets value converter like resource', function() {
+    const result = nameConvention('FooBarValueConverter');
+    assert.deepEqual(result, {name: 'foo-bar', type: 'valueConverter'});
```
EisenbergEffect: This should actually be `fooBar` due to the fact that we can use a `-` a binding expression without it being interpreted as a minus. Same for the binding behavior convention below (but not binding command).

packages/__tests__/plugin-conventions/preprocess-resource.spec.ts
```diff
@@ -0,0 +1,292 @@
...
+export class FooBarCustomAttribute {}
+`;
+    const result = preprocessResource('bar/FooBar.js', code, true);
+    assert.equal(result.code, expected);
+  });
+
+  it('injects valueConverter decorator', function () {
+    const code = `export class FooBarValueConverter {}\n`;
+    const expected = `import { valueConverter } from '@aurelia/runtime';
+@valueConverter('foo-bar')
```
EisenbergEffect: See note above about value converter and binding behavior naming.

packages/__tests__/plugin-conventions/name-convention.spec.ts
3cp: Good catch! My mistakes.

EisenbergEffect:
@3cp I left a couple comments about name conventions. Otherwise, looks pretty good :)

packages/__tests__/plugin-conventions/name-convention.spec.ts
```diff
@@ -0,0 +1,33 @@
+import { nameConvention } from '@aurelia/plugin-conventions';
+import { assert } from '@aurelia/testing';
+
+describe('nameConvention', function () {
+  it('complains about empty input', function() {
+    assert.throws(() => nameConvention(''));
```
bigopon: maybe throw on anything that is potentially invalid? Just to keep the spirit of our testing, started by @fkleuver

packages/__tests__/plugin-conventions/name-convention.spec.ts
```diff
@@ -0,0 +1,33 @@
...
+  });
+
+  it('gets custom element like resource', function() {
```
bigopon: would love to see some number tests as well, sometime i use them 😬

packages/__tests__/plugin-conventions/preprocess-html-template.spec.ts
```diff
@@ -0,0 +1,89 @@
+import { preprocessHtmlTemplate } from '@aurelia/plugin-conventions';
```
bigopon: ❤️ this file

packages/__tests__/plugin-conventions/strip-html-import.spec.ts
```diff
@@ -0,0 +1,91 @@
...
+</template>
+`;
+
+    assert.deepEqual(stripHtmlImport(html), {
+      html: expected,
+      deps: ['./a']
+    });
+  });
+
+  it('strips mixed import/require tags with wrong or missing close tag', function() {
```
bigopon: woohoo welcome to html town 😆

packages/plugin-conventions/src/strip-html-import.ts
```diff
@@ -0,0 +1,48 @@
+import { parseFragment, DefaultTreeElement, ElementLocation } from 'parse5';
```
bigopon: This file looks simple but exciting 🎉

packages/__tests__/plugin-conventions/name-convention.spec.ts
3cp: It's kind of internal API to translate ClassName. BTW, what kind of invalid input you has in mind?

Even `WhatEverName` will be a custom element based on convention.

packages/__tests__/plugin-conventions/name-convention.spec.ts
3cp: What's number tests?

bigopon: Example would be
MyEl1 -> my-el1 or myel-1?
Just need to have a few tests for that i think.

packages/__tests__/plugin-conventions/name-convention.spec.ts
bigopon: Undefined, false symbol null number

packages/__tests__/plugin-conventions/name-convention.spec.ts
3cp: I see, will add it.
The kebabCase and camelCase is from `@aurelia/kernel`, that's why I didn't bother to test many edge cases.

packages/__tests__/plugin-conventions/name-convention.spec.ts
3cp: I thought because this is TS project, you don't need to do that kind of tests anymore?

3cp: I think the compiler will reject that kind of test code to build.

3cp: @fkleuver I tested `nameConvention(null)`, the `npm run dev` didn't complain about the usage against the typing `nameConvention(className: string)` in the test code. Is there any missed typing setup in our test code compilation?

EisenbergEffect:
@3cp Note that I think we’ll want some APIs or method of configuration in the future that let’s people customize the conventions. We don’t need that for the first pass though. I’d love to get everything working and then come back to that in the 2nd or 3rd pass, once we feel confident about the general approach.

3cp:
I see, that probably will be options passed to AOT first, then AOT might call conventions internally. We will see how it goes.

The current implementation is tiny, easy to change in future.

3cp:
plugin-webpack is renamed to webpack-loader, because it's not a plugin, it's a loader just like ts-loader. I tested it in babel and ts webpack apps.

@EisenbergEffect @fkleuver let's merge this PR and publish these two packages, so we can test in webpack app more easily.

fkleuver:
@3cp The builds do need to pass first. It's complaining somehow about not being able to find `@babel/parser`? https://circleci.com/gh/aurelia/aurelia/21357?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

3cp:
I see, modify-code use babel parse to tokenize all latest js/ts. I will check why it complains.

3cp:
@fkleuver it looks like those browser test I never did locally. Questions:

1. I guess I need to add some deps to `packages/__test__/package.json`?
2. how do I add plugin-conventions to browser test, but skip webpack-loader on browser?

3cp:
I don't quite understand why the karma+webpack setup didn't trace and bundle the necessary packages. If this is the first time an `@aurelia` package requires 3rd party npm package, we might have a hole in karma+webpack config.

PR #530, opened 2019-07-20
feat(i18n): skeleton implementation

Sayan751:
# Pull Request

## 📖 Description

This PR adds a skeleton implementation of the i18n plugin. At this point, it only provides replacement of the `textContent`. There are both unit tests and e2e tests for that.

## 👩‍💻 Reviewer Notes

As said before, this PR does not provide complete implementation of i18 plugin. The main objective is to setup a skeleton plugin along with the tests so that the basic integration is established that will make future development of this plugin easier.
 

## 📑 Test Plan

For unit tests the i18next is spied and injected into the `I18N`. The implementation however, relies on a wrapped instance of i18next, which in turn makes the mocking, and DI easier.

For the e2e tests an app is setup under `__tests__/app` with the integration tests next to it. 

## ⏭ Next Steps

The very next step would be to complete the implementation of `I18N` service, which means supporting translations by attributes, such `html`, `append`, `prepend`, and regular attributes. Once that is done, further development with the valueConverters, and bindingBehaviors can be started.


Sayan751:
Closed in favor of #533

PR #533, opened 2019-07-21
feat(i18n): skeleton implementation

Sayan751:
# Pull Request

## 📖 Description

This PR adds a skeleton implementation of the i18n plugin. At this point, it only provides replacement of the `textContent`. There are both unit tests and e2e tests for that.

## 👩‍💻 Reviewer Notes

As said before, this PR does not provide complete implementation of i18 plugin. The main objective is to setup a skeleton plugin along with the tests so that the basic integration is established that will make future development of this plugin easier.
 

## 📑 Test Plan

For unit tests the i18next is spied and injected into the `I18N`. The implementation however, relies on a wrapped instance of i18next, which in turn makes the mocking, and DI easier.

For the e2e tests an app is setup under `__tests__/app` with the integration tests next to it. 

## ⏭ Next Steps

The very next step would be to complete the implementation of `I18N` service, which means supporting translations by attributes, such `html`, `append`, `prepend`, and regular attributes. Once that is done, further development with the valueConverters, and bindingBehaviors can be started.


zewa666:
As for the nature of this PR I can't really say if it's enough or not. It certainly does something so it could be used as a start for having the grunt work all setup. I just hope we don't miss parts when re-implementing the features which is what currently bothers me a bit with the i18n service as it's stripped down to a bare min and excluding features.

packages/__tests__/i18n/Spy.ts
```diff
@@ -0,0 +1,48 @@
+import { assert } from '@aurelia/testing';
```
zewa666: do we really need to implement this here? It feels like this is a more general feature which could be part of testing itself.

packages/i18n/package.json
```diff
@@ -0,0 +1,54 @@
...
+  },
+  "publishConfig": {
+    "access": "public"
+  },
+  "dependencies": {
+    "@aurelia/kernel": "^0.3.0",
+    "@aurelia/runtime": "^0.3.0",
+    "@aurelia/runtime-html": "^0.3.0",
+    "@aurelia/jit-html-browser": "^0.3.0",
+    "i18next": "^17.0.6"
```
zewa666: this is right away a major breaking change, which on the other hand would be perhaps ok though with the rewrite for vNext. We just should point this out. vCurrent is still based on the 14.x branch

packages/i18n/package.json
```diff
@@ -0,0 +1,54 @@
...
+  },
+  "devDependencies": {
+    "@types/node": "^10.12.27",
```
zewa666: don't we need the `@types/i18next` devDependency any longer or is it totally outdated?

packages/i18n/src/configuration.ts
```diff
@@ -0,0 +1,22 @@
...
+        I18nConfigurationOptions.withDefault(x => x.callback(this.optionsProvider)),
+      );
+    },
+    customize(cb?: I18NConfigOptionsProvider) {
+      return createI18nConfiguration(cb || optionsProvider);
+    },
+  };
+}
+
+export const I18nConfiguration = createI18nConfiguration(() => Object.create(null));
```
zewa666: why return an empty object here? or is this just a placeholder? If so in general it would be good to highlight those parts with comments aka Todos

.circleci/config.yml
```diff
@@ -429,6 +446,8 @@ workflows:
           <<: *filter_ignore_develop_release
       - js_framework_benchmark:
           <<: *filter_ignore_develop_release
+      - e2e_cypress:
+         <<: *filter_ignore_develop_release
```
zewa666: so does this mean E2E tests will only be executed when merging into master?

packages/__tests__/i18n/Spy.ts
Sayan751: I started writing the spy only focusing on `i18next`. I wrote the generic version, almost as an afterthought. I would like to keep it here for now, as that would allow me to have a non-generic implementation. If the API matured in the process, it can be made part of `@aurelia/testing`.

packages/i18n/package.json
Sayan751: Nope, now `i18next` comes with its [own type definition](https://github.com/i18next/i18next/blob/master/package.json#L7) 😃

packages/i18n/src/configuration.ts
Sayan751: This is the default `i18next` init options provider, which returns an empty object. If devs want to customize the init options, they can provide their own call back via `customize`. In both cases the lambda acts as callback resolver for `I18nConfigurationOptions`. The default provider just acts as fallback.

.circleci/config.yml
Sayan751: I am not really sure, need input from @fkleuver

Sayan751:
> I just hope we don't miss parts when re-implementing the features which is what currently bothers me a bit with the i18n service as it's stripped down to a bare min and excluding features.

@zewa666 I understand your concern. I would like to point out that the e2e tests written for vCurrent will come in handy to validate that all features are ported correctly :smiley:

.circleci/config.yml
fkleuver: @zewa666 @Sayan751 `<<: *filter_ignore_develop_release` means only ignore the `develop` and `release` branches, this is standard for all jobs. It means the E2E tests will be executed on every other branch (so topic, master, etc)

packages/i18n/src/t-custom-attribute.ts
```diff
@@ -0,0 +1,27 @@
...
+export class TCustomAttribute {
+  // tslint:disable-next-line: prefer-readonly
+  @bindable private value: string = (void 0)!;
+  constructor(
+    @INode private readonly node: Node,
+    @I18N private readonly i18n: I18nService) {
+
+  }
+
+  public static register(container: IContainer): void {
```
fkleuver: You can remove this call. It is added by the decorator.

packages/testing/src/html-test-context.ts
```diff
@@ -176,6 +176,10 @@ export class HTMLTestContext {
     return this.doc.createElement(name);
   }
 
+  public createTextNode(data: string): Text {
```
fkleuver: You can use `createElementFromMarkup` and pass anything into it, it will return the correct node. If you pass in plain text, you'll get a text node.

packages/i18n/src/t-custom-attribute.ts
```diff
@@ -0,0 +1,27 @@
...
+    container.register(Registration.transient('custom-attribute:t', this));
+    container.register(Registration.transient(this, this));
+  }
+
+  public attaching(flags: LifecycleFlags) {
+    this.i18n.updateValue(this.node, this.value);
+  }
+
```
fkleuver: This and a few other redundant blank lines, if you wouldn't mind can you please tidy those up :)

packages/i18n/src/i18next-wrapper.ts
```diff
@@ -0,0 +1,12 @@
+import { DI } from '@aurelia/kernel';
+import i18next from 'i18next';
+
+export const I18nWrapper = DI.createInterface<I18nextWrapper>('I18nextWrapper').withDefault(x => x.singleton(I18nextWrapper));
```
fkleuver: Please avoid using `.withDefault` in the core framework as it makes it impossible for end users to override the default registration. This is more of an end user api.

What we should do instead is this:

```ts
export const I18nWrapper = DI.createInterface<I18nextWrapper>('I18nextWrapper').noDefault();
```

And then in the config:

```ts
      return container.register(
        // ...
        Registration.singleton(I18nWrapper, I18nextWrapper),
        // ...
      );
```

packages/i18n/src/configuration.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export type I18NConfigOptionsProvider = () => I18nConfigurationOptions;
+
+function createI18nConfiguration(optionsProvider: I18NConfigOptionsProvider) {
+  return {
+    optionsProvider,
+    register(container: IContainer) {
+      return container.register(
+        TCustomAttribute,
+        I18nConfigurationOptions.withDefault(x => x.callback(this.optionsProvider)),
```
fkleuver: Should be:

```ts
Registration.callback(I18nConfigurationOptions, this.optionsProvider),
```

packages/i18n/src/i18n-configuration-options.ts
```diff
@@ -0,0 +1,14 @@
...
+import i18next from 'i18next';
+
+export const I18nConfigurationOptions = DI.createInterface<I18nConfigurationOptions>('I18nConfigurationOptions');
+export interface I18nConfigurationOptions extends i18next.InitOptions {
+  /**
+   * Collection of i18next plugins to use.
+   */
+  // TODO: Have some basic duck typings for the i18next plugins. These come in various shapes and i18next does not provide any type definition for these plugins.
+  // tslint:disable-next-line: no-any
+  plugins?: any[];
```
fkleuver: Are these used in any way yet by the i18n plugin? If so, we should create a strongly typed shape. If not, let's remove these properties for now as they have no use yet and would just lead to confusion

zewa666: Never seen that plugins part yet. Earlier you used to call .use(Yourplugin) so perhaps a new thing with v18?

packages/i18n/src/i18next-wrapper.ts
zewa666: Out of curio, what happens If you only keep noDefault around and the Key is resolved?

packages/i18n/src/t-custom-attribute.ts
zewa666: Normally the linter should pick this up with consecutive new lines rule. Perhaps we need to tweak the settings

packages/i18n/src/i18n.ts
```diff
@@ -0,0 +1,110 @@
...
+
+    while (i--) {
+      const { attr, key } = this.extractAttributesFromKey(node, keys[i]);
+      this.applyTranslations(node, attr.split(','), key, params);
+    }
+  }
+
+  private extractAttributesFromKey(node: Node, key: string) {
+    const re = /\[([a-z\-, ]*)\]/ig;
+    let m, attr = 'text';
```
fkleuver: One variable declaration per line please. I'd like for everything to be explicitly typed.

packages/i18n/src/t-custom-attribute.ts
zewa666: What is that init of the value bindable doing? Compiler optimization? Also if its a bindable it is referenced from the template so it should be public

packages/i18n/src/i18n.ts
zewa666: In his protection this is just old copy pasted code from vCurrent but yeah we should fix this. Also the while loop above is ugly

packages/i18n/src/i18n.ts
```diff
@@ -0,0 +1,110 @@
...
+  }
+
+  private applyTranslations(node: Node, attrs: string[], key: string, params?: i18nextCore.TOptions<object>) {
+    let j = attrs.length;
+    while (j--) {
+      // handle various attributes
+      // anything other than text,prepend,append or html will be added as an attribute on the element.
+      switch (attrs[j].trim()) {
+        case 'text':
+          this.replaceTextContent(node, key, params);
```
fkleuver: You actually don't need the `replaceTextContent` method. You can just do this:
```ts
node.textContent = this.tr(key, params);
```

That replaces the text content just the same, but more efficiently

packages/testing/src/html-test-context.ts
Sayan751: That's cool to know. However, I am injecting the test-context as a mock for `IDOM` in `I18NService`, where this specific function is needed. From that perspective it makes sense to have this mock implementation unless a mock `createTextNode` is added as a proxy for `createElementFromMarkup` in the test class itself. However, that seems counter productive.

packages/__tests__/i18n/i18n.spec.ts
```diff
@@ -0,0 +1,87 @@
...
+import { Spy } from './Spy';
+
+const translation = {
+  simple: {
+    text: 'simple text',
+    attr: 'simple attribute'
+  }
+};
+
+export function i18nTests() {
```
fkleuver: Just wrap it in a describe directly, and remove `inedx.ts`. Also a few other nitpicks:

```ts
describe('aurelia-i18n', function () { // use function () instead of arrow () => to retain `this` test context for mocha
  // Declare a normal function named "setup" for consistency with the other tests
  async function setup(options: I18nConfigurationOptions = {}) {
      const mockContext = new Spy();
      const ctx = TestContext.createHTMLTestContext();
      const sut = new I18nService({ i18next: mockContext.getMock(i18next) }, options, ctx as unknown as IDOM<Node>);
      await sut['task'].wait();
      return { mockContext, ctx, sut }; // Declare variables locally and return them from setup function, instead of using shared test variables
  }

    // consume appropriately
    it('initializes i18next with default options on instantiation', async function () {
      const { mockContext } = await setup();

      mockContext.methodCalledOnceWith('init', [{
        lng: 'en',
        fallbackLng: ['en'],
        debug: false,
        plugins: [],
        attributes: ['t', 'i18n'],
        skipTranslationOnMissingKey: false,
      }]);
    });
});
```

packages/__tests__/i18n/Spy.ts
```diff
@@ -0,0 +1,48 @@
...
+  }
+
+  public methodCalledOnceWith(methodName: string, expectedArgs: any[]) {
+    this.methodCalledTimes(methodName, 1);
+    this.methodCalledNthTimeWith(methodName, 1, expectedArgs);
+  }
+
+  public methodCalledNthTimeWith(methodName: string, n: number, expectedArgs: any[]) {
+    const calls = this.callRecords.get(methodName);
+    assert.equal(JSON.stringify(calls[n - 1]), JSON.stringify(expectedArgs));
```
fkleuver: You can use `assert.deepStrictEqual`

packages/__tests__/i18n/Spy.ts
```diff
@@ -0,0 +1,48 @@
...
+      record = [args];
+    }
+    this.callRecords.set(methodName, record);
+  }
+
+  public clearCallRecords() { this.callRecords.clear(); }
+
+  public methodCalledTimes(methodName: string, times: number) {
+    const calls = this.callRecords.get(methodName);
+    assert.equal(!!calls, true);
```
fkleuver:
```ts
assert.notEqual(calls, null);
```

packages/i18n/src/i18n.ts
```diff
@@ -0,0 +1,110 @@
...
+
+  public tr(key: string | string[], options?: i18nextCore.TOptions<object>) {
+    return this.i18next.t(key, options);
+  }
+
+  public updateValue(node: Node, value: string, params?: i18nextCore.TOptions<object>) {
+    if (this.task.done) {
+      this.updateValueCore(node, value, params);
+    } else {
+      this.task = new ContinuationTask(this.task, this.updateValueCore, this, node, value, params);
```
fkleuver: if you `return` this task then it can be awaited by the consumer

packages/i18n/src/t-custom-attribute.ts
```diff
@@ -0,0 +1,27 @@
...
+    @I18N private readonly i18n: I18nService) {
+
+  }
+
+  public static register(container: IContainer): void {
+    container.register(Registration.transient('custom-attribute:t', this));
+    container.register(Registration.transient(this, this));
+  }
+
+  public attaching(flags: LifecycleFlags) {
```
fkleuver: Should use `binding` here and return the value of `updateValue` so that if it returns a task, it is awaited by binding and the DOM would only need to update once

packages/i18n/package.json
```diff
@@ -0,0 +1,54 @@
...
+    "dev": "tsc -b -w"
+  },
+  "publishConfig": {
+    "access": "public"
+  },
+  "dependencies": {
+    "@aurelia/kernel": "^0.3.0",
+    "@aurelia/runtime": "^0.3.0",
+    "@aurelia/runtime-html": "^0.3.0",
+    "@aurelia/jit-html-browser": "^0.3.0",
```
fkleuver: No package should depend on `@aurelia/jit-html-browser`

packages/i18n/src/i18n-configuration-options.ts
Sayan751: @zewa666 The `i18next` plugin API still is the same. Only in this implementation, the `i18next` init options are provided in a single object, including the plugins. The `I18nService#initializeI18next` registers those plugins.

@fkleuver The type definition of the plugins should ideally come from `i18next`. Unfortunately, it provides none. Maybe I can write some interfaces based on this doc: https://www.i18next.com/misc/creating-own-plugins

packages/i18n/src/i18next-wrapper.ts
fkleuver: [this](https://github.com/aurelia/aurelia/blob/master/packages/kernel/src/di.ts#L155) :)

packages/i18n/src/i18n-configuration-options.ts
fkleuver: @Sayan751 If you have time for it, that would be great. It needs to be done at some point anyway

packages/i18n/src/i18n.ts
fkleuver: Yeah I know. It is not criticism. Just saying we should fix what we touch. I don't think any of the vCurrent plugin landscape is fit for plain copy-paste.

packages/i18n/src/t-custom-attribute.ts
fkleuver: I'm pretty sure it does. @Sayan751 do you have tslint installed in vs code?

packages/testing/src/html-test-context.ts
fkleuver: I don't think using test context as a mock value is very scalable or future-maintainer friendly.
Why not just pass an inline object?
Also why do you need that mocked in the first place? If you need to verify that it works, verify the text content imo

packages/i18n/src/t-custom-attribute.ts
```diff
@@ -0,0 +1,27 @@
...
+import { bindable, customAttribute, INode, LifecycleFlags } from '@aurelia/runtime';
+import { I18N, I18nService } from './i18n';
+/**
+ * `t` custom attribute to facilitate the translation via HTML
+ * @export
+ */
+@customAttribute({ name: 't' })
+export class TCustomAttribute {
+  // tslint:disable-next-line: prefer-readonly
+  @bindable private value: string = (void 0)!;
```
fkleuver: Bindables should be public as @zewa666 pointed out.
Also, if you make this public then `tslint` won't nag about `prefer-readonly` either.

Furthermore I would type this as `unknown` and remove the initializer. You have no guarantee whatsoever about what the user binds to this, so the typing should reflect that.

packages/__tests__/i18n/i18n.spec.ts
```diff
@@ -0,0 +1,86 @@
...
+        en: { translation }
+      }
+    };
+    const { sut, ctx } = await setup(customization);
+
+    const span = ctx.createElement('span');
+    sut.updateValue(span as any, 'simple.text');
+    await sut['task'].wait();
+
+    assert.equal(span.textContent, translation.simple.text);
```
fkleuver: Please use `strictEqual`

AureliaEffect:
JS Framework Benchmark: https://21697-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/__tests__/i18n/i18n.spec.ts
```diff
@@ -0,0 +1,86 @@
...
+  async function setup(options: I18nConfigurationOptions = {}) {
+    let sut: I18nService, mockContext: Spy, ctx: HTMLTestContext;
+    mockContext = new Spy();
+    ctx = TestContext.createHTMLTestContext();
+    sut = new I18nService({ i18next: mockContext.getMock(i18next) }, options, ctx as unknown as IDOM<Node>);
+    await sut['task'].wait();
+    return { mockContext, sut, ctx };
+  }
+
+  it('initializes i18next with default options on instantiation', async () => {
```
fkleuver:
```suggestion
  it('initializes i18next with default options on instantiation', async function () {
```

packages/__tests__/i18n/i18n.spec.ts
```diff
@@ -0,0 +1,86 @@
...
+      lng: 'en',
+      fallbackLng: ['en'],
+      debug: false,
+      plugins: [],
+      attributes: ['t', 'i18n'],
+      skipTranslationOnMissingKey: false,
+    }]);
+  });
+
+  it('respects user-defined config options', async () => {
```
fkleuver:
```suggestion
  it('respects user-defined config options', async function () {
```

packages/__tests__/i18n/i18n.spec.ts
```diff
@@ -0,0 +1,86 @@
...
+      lng: customization.lng,
+      fallbackLng: ['en'],
+      debug: false,
+      plugins: [],
+      attributes: customization.attributes,
+      skipTranslationOnMissingKey: false,
+    }]);
+  });
+
+  it('registers external plugins provided by user-defined options', async () => {
```
fkleuver:
```suggestion
  it('registers external plugins provided by user-defined options', async function () {
```

packages/__tests__/i18n/i18n.spec.ts
```diff
@@ -0,0 +1,86 @@
...
+        }
+      ]
+    };
+    const { mockContext } = await setup(customization);
+
+    mockContext.methodCalledNthTimeWith('use', 1, [customization.plugins[0]]);
+    mockContext.methodCalledNthTimeWith('use', 2, [customization.plugins[1]]);
+  });
+
+  it('can update textContent of an element given translations', async () => {
```
fkleuver:
```suggestion
  it('can update textContent of an element given translations', async function () {
```

packages/i18n/src/i18n.ts
```diff
@@ -0,0 +1,98 @@
...
+  private extractAttributesFromKey(node: Node, key: string) {
+    const re = /\[([a-z\-, ]*)\]/ig;
+    // set default attribute to src if this is an image node
+    let attr = node.nodeName === 'IMG' ? 'src' : 'text';
+
+    // check if a attribute was specified in the key
+    const matches = re.exec(key);
+    if (matches) {
+      key = key.replace(matches[0], '');
+      attr = matches[1];
```
fkleuver: This whole method kind of confuses me. What exactly is the point here? Feels like this could be a lot simpler and/or clearer.

packages/i18n/src/i18n.ts
zewa666: Thanks god we have the e2e tests now in place. I remember that exactly this didnt work for the vCurrent so I had to do some mumbo jumbo with recreating textnodes. Now we can see if everything still behaves fine after refactor

packages/i18n/src/t-custom-attribute.ts
zewa666: If the TS version doesnt matter unknown is certainly better

packages/i18n/src/i18n.ts
zewa666: I'd propose we get this in as it is with a refactor todo as it would extend the scope of this PR.

Sayan751: Actually this method will come in handy when the use case of attribute translation will be implemented. I would like to keep this function as-is for now. Once those use-cases are implemented, it might be clearer how this function can be optimized.

packages/i18n/src/t-custom-attribute.ts
Sayan751: I disagree with the usage of type `unknown` in this case. I think it should rather reflect from the type definition that a string value is expected by the custom attribute. In `binding` (and later in change handler), a simple type checking can be done to assert that the value is indeed of type string, or else throw an error. But the fact that a string is needed here should be reflected by the type definition. That way we can justify throwing an error if the value is not a string. On the other hand if the type is `unknown`, devs should expect any type of value to work and that certainly should not be the case.

packages/i18n/src/i18n-configuration-options.ts
Sayan751: I will put those types in next PR.

packages/i18n/src/t-custom-attribute.ts
fkleuver: Actually you make a very good point there. I was in the mindset of this being internal api because of the `private` keyword but yeah it should be public anyway.

packages/i18n/src/i18next-wrapper.ts
Sayan751: Done

packages/i18n/src/i18n.ts
Sayan751: done

packages/__tests__/i18n/Spy.ts
Sayan751: done

packages/__tests__/i18n/Spy.ts
Sayan751: done

packages/i18n/package.json
Sayan751: removed

packages/i18n/src/i18n-configuration-options.ts
Sayan751: Also I think it will make sense if those type definitions come from `i18next`. I will first try to create some interfaces, and if it works then I will try to create a PR for i18next repo with the types. There is no reason to keep those type definitions in Aurelia, if they accept the PR.

packages/testing/src/html-test-context.ts
Sayan751: replaced the mock with `DOM`

packages/i18n/src/t-custom-attribute.ts
Sayan751: Yup, I missed.

packages/i18n/src/configuration.ts
Sayan751: done

packages/__tests__/i18n/i18n.spec.ts
Sayan751: done

AureliaEffect:
JS Framework Benchmark: https://21823-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/i18n/src/t-custom-attribute.ts
zewa666: But in this case we should definitely make sure to add that changedhandler with throwing an error otherwise the type gives fault safety. Its super easy to mess up with passing wrong arguments via html

packages/i18n/src/i18n.ts
Sayan751: Done

packages/i18n/src/t-custom-attribute.ts
zewa666: Alternatively what https://github.com/aurelia-contrib/aurelia-typed-observable-plugin does would be an option

Sayan751: Next I am planning to write some unit tests for this attribute, I will put the changeHandler then.

AureliaEffect:
JS Framework Benchmark: https://21843-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/i18n/src/i18n-configuration-options.ts
Sayan751: Found the type defs :)

AureliaEffect:
JS Framework Benchmark: https://21887-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/i18n/src/i18n.ts
```diff
@@ -0,0 +1,98 @@
+import { DI } from '@aurelia/kernel';
+import { ContinuationTask, IDOM, ILifecycleTask, PromiseTask } from '@aurelia/runtime';
+import i18nextCore from 'i18next';
+import { I18nConfigurationOptions } from './i18n-configuration-options';
+import { I18nextWrapper, I18nWrapper } from './i18next-wrapper';
+
+export const I18N = DI.createInterface<I18nService>('I18nService').withDefault(x => x.singleton(I18nService));
```
fkleuver: @Sayan751 should be:
```ts
export const I18N = DI.createInterface<I18nService>('I18nService').noDefault();
```

Paired with the appropriate registration on the exported config object

AureliaEffect:
JS Framework Benchmark: https://21909-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

PR #537, opened 2019-07-23
Allow default value for attributes that are bound with no explicit value being set

brandonseydel:
# Pull Request

## 📖 Description

-This allows a a bindable to be set to true when nothing is passed in to the attribute.  This greatly reduces the boilerplate for times when you want to set a static true value. *see below
```html
<element disabled outlined ></element>
```
```typescript
class Element{
@bindable public disabled: boolean = false; //THIS WILL BE TRUE
@bindable public outlined: boolean = false; //THIS WILL BE TRUE
}
```


3cp:
I love this feature, but what about `<element something.bind="value">` where `value` is an empty string, I expect `something` is bound to empty string instead of `true`.

We need at least some unit tests to cover some edge cases.

fkleuver:
`SetPropertyRenderer` is exclusively used when there is no binding command on the attribute, so this does not affect normal bindings. If it did, a couple hundred tests would have failed ;)

3cp:
Good to know. But should not we add test coverage for this new feature too?

PR #538, opened 2019-07-24
perf(all): remove tracer/profiler from ts source

fkleuver:
# Pull Request

## 📖 Description

This PR removes all `Tracer` and `Profiler` calls from the source code, as a first step in cleaning things up a bit and improving the performance for JIT. Later, AOT will add these calls based on configuration flags.

### 🎫 Issues

N/A

## 👩‍💻 Reviewer Notes

Just a project-wide find/replace, not much to see.

## 📑 Test Plan

No new behavior introduced. Existing tests should still pass.

### DIFF
--- a/packages/jit-html-browser/src/index.ts
+++ b/packages/jit-html-browser/src/index.ts
@@ -26,7 +26,6 @@ export const BasicConfiguration = {
    * Apply this configuration to the provided container.
    */
   register(container: IContainer): IContainer {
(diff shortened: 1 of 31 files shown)


EisenbergEffect:
I like the idea of having AoT add this based on configuration. Nice call.

bigopon:
Hmm it's amazing that only 31 files got touched. Though the description of this PR makes me very curious. Cant wait to see it.

PR #539, opened 2019-07-24
refactor conventions implementation to avoid duplicated code

3cp:
Push down common logic to base package plugin-conventions.


AureliaEffect:
JS Framework Benchmark: https://21770-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

PR #540, opened 2019-07-24
feat(runtime): initial implementation for startup tasks

fkleuver:
# Pull Request

## 📖 Description

This is a somewhat quick-n-dirty startup task implementation, mostly to get some API feedback and get the ball rolling. I hope others can help me refine/improve/test this.

The idea is that you create a task from a fluent API and register the created object with DI. The activator then resolves all tasks and executes them at the right times during initial startup.

There are 4 "slots": `beforeCreate`, `beforeRender`, `beforeBind` and `beforeAttach`.
- `beforeCreate` will run "before app()" in a sense, that is, it will run before the root App is instantiated from DI. This is where you can, for example, manipulate container registrations while all static registrations are already available.
-`beforeRender` runs before the app root is rendered / controller created. So you could for example manipulate the app root before its compilation. Technically equivalent to the `constructor` of the app root.
- `beforeBind` runs after render and before the bind lifecycle starts on the root. Technically equivalent to the `created` hook in the app root.
- `beforeAttach` runs after the `binding` + `bound` lifecycles fully completed and before `attach` starts. Technically equivalent to the `bound` hook in the app root.

So while `beforeRender`, `beforeBind` and `beforeAttach` can, timing-wise, be accomplished from within the app root, these hooks actually allow plugin authors (who have no control over the app root) to plug in logic at these points in time.

Furthermore, there are 2 "methods": `from` and `with`
- `from` creates the task "from" a promise or existing task.
- `with` allows you to specify a registration key, and a callback that returns that resolved component where you can then perform logic (e.g. call a method on the component) that then returns a promise or task.

To just dive in with some examples:

```ts
StartTask.with(IRouter).beforeAttach().call(router => router.activate()) // initial idea for the router
StartTask.with(IComponent).beforeCreate().call(component => component.doStuff())
StartTask.with(IComponent).at(TaskSlot.beforeCreate).call(component => component.doStuff())
StartTask.from(async () => {
  await doStuff();
  doOtherStuff();
}).beforeBind();
```

## 👩‍💻 Reviewer Notes

- There's a new task type `ProviderTask` for the DI callback, [implementation here](https://github.com/aurelia/aurelia/compare/start-task?expand=1#diff-5aa3f5afe915c59e4045e33de04f8339R288)
- `StartTaskManager` is the facade that wraps all tasks for a slot into an aggregate task, [implementation here](https://github.com/aurelia/aurelia/compare/start-task?expand=1#diff-5aa3f5afe915c59e4045e33de04f8339R187)
- The placement of `beforeCreate` sits in [the CompositionRoot](https://github.com/aurelia/aurelia/compare/start-task?expand=1#diff-0bcdaa7de1d39a267573c522d86a0850R92)
- The placements of `beforeRender`, `beforeBind` and `beforeAttach` sit [in the activator](https://github.com/aurelia/aurelia/compare/start-task?expand=1#diff-d98c0115270b529c1d794c910c81f1d5R48)

## 📑 Test Plan

There are no tests at the moment. Any volunteers to help out would be greatly appreciated.

## ⏭ Next Steps

Test it :)

### DIFF
--- a/packages/kernel/src/index.ts
+++ b/packages/kernel/src/index.ts
@@ -19,7 +19,8 @@ export {
   transient,
   Injectable,
   InterfaceSymbol,
-  InstanceProvider
+  InstanceProvider,
+  Resolved,
 } from './di';
 export {
   Class,
--- a/packages/router/src/configuration.ts
+++ b/packages/router/src/configuration.ts
@@ -1,8 +1,9 @@
 import { DI, IContainer, IRegistry } from '@aurelia/kernel';
+import { StartTask } from '@aurelia/runtime';
 
 import { NavCustomElement } from './resources/nav';
 import { ViewportCustomElement } from './resources/viewport';
-import { IRouterOptions, Router } from './router';
+import { IRouterOptions, Router, IRouter } from './router';
 
 export const RouterRegistration = Router as unknown as IRegistry;
 
@@ -11,7 +12,7 @@ export const RouterRegistration = Router as unknown as IRegistry;
  * - `IRouter`
  */
 export const DefaultComponents = [
-  RouterRegistration
+  RouterRegistration,
 ];
 
 export {
@@ -41,7 +42,8 @@ const routerConfiguration = {
   register(container: IContainer): IContainer {
     return container.register(
       ...DefaultComponents,
-      ...DefaultResources
+      ...DefaultResources,
+      StartTask.with(IRouter).beforeAttach().call(router => router.activate())
     );
   },
   /**
--- a/packages/runtime/src/activator.ts
+++ b/packages/runtime/src/activator.ts
@@ -4,11 +4,12 @@ import {
   IResolver,
   IServiceLocator,
   Registration,
+  Key,
 } from '@aurelia/kernel';
 
 import { INode } from './dom';
 import { LifecycleFlags } from './flags';
-import { IViewModel, IController } from './lifecycle';
+import { IController, IViewModel } from './lifecycle';
 import {
   ContinuationTask,
   ILifecycleTask,
@@ -27,6 +28,8 @@ export const IActivator = DI.createInterface<IActivator>('IActivator').withDefau
 
 /** @internal */
 export class Activator implements IActivator {
+  public static readonly inject: readonly Key[] = [IStartTaskManager];
+
   constructor(
     private readonly taskManager: IStartTaskManager,
   ) {}
--- a/packages/runtime/src/activator.ts
+++ b/packages/runtime/src/activator.ts
@@ -8,10 +8,11 @@ import {
 
 import { INode } from './dom';
 import { LifecycleFlags } from './flags';
-import { IViewModel } from './lifecycle';
+import { IViewModel, IController } from './lifecycle';
 import {
   ContinuationTask,
   ILifecycleTask,
+  IStartTaskManager,
 } from './lifecycle-task';
 import { IScope } from './observation';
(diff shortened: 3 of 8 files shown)


EisenbergEffect:
I have to say, I like this. It’s a great way to use DI and reminds me of “startable” services that I think Castle Windsor used to have many years ago. I really like it that any plugin can simple shove these in the root container and hook into just about anything. The router example is a nice touch there. I’m all for this.

fkleuver:
@EisenbergEffect Great. Timing wise, would one of those hooks address the specific use case you were talking about earlier?

EisenbergEffect:
Yep, `beforeCreate` handles my specific scenarios :) Seems like `beforeAttach` matches well with routers and anything that might need to mess with the DOM. Seems like the bases are covered by the four hooks.

HamedFathi:
Long-time ago I asked something [here](https://discourse.aurelia.io/t/aurelia-2018-q2-update-and-vnext-progress/1501/7)

> add excludePlugin() beside plugin() method. Sometimes we need to add the whole project except some plugins instead of adding plugins one by one.

@EisenbergEffect said

> For plugins, we’ve opted to enhance the dependency injection container to handle this scenario. As pictured in the code sample above BasicConfiguration is a plugin that adds all the basic capabilities of the framework. However, the new mechanism and framework design allows you to add each piece individually, so you can add only what you want/need. When we enable AOT, we’re hopeful that we’ll be able to code-generate a plugin, specific for your app, that installs only the things you actually use, so you don’t have to do that by hand.

@fkleuver 

With this, Can I filter plugins registration by my rules? For example `excludePlugins`. 
`Aurelia-Toolbelt` has many components I want to register all except some.

bigopon:
Start up tasks require tasks ordering, dependency tasks declaration & deduplication capabilities.

In vCurrent, one of the problems for component plugin authors is that you can specify dependencies without worrying about duplication (A -> B, C -> B, app starts with both A & C will only get B once), but B will be run after A or C:

```ts
// a.ts
// configure fn for plugin A
export function configure(config: FrameworkConfiguration): void {
  config
    // queue plugin B to make sure it's booted during start up
    .plugin('B')
  // code here runs when app root starts, but before B runs
}
```

To work around this, you either have to (1) specify dependencies as B then (A & C), or (2) just invoke B during A/C startup

For example, (1):
```ts
// main.ts
export async function configure(aurelia: Aurelia) {
  aurelia
    .use
      .plugin('B')
      .plugin('A')
}
```

vCurrent UX is currently suggesting (1) to handle this, and ther is a PR to make it also friendly to (2). Though both (1) and (2) are just work around. I feel we need (3): a way to register a dependency task, that will be guaranteed to completed before other task, and it doesn't rely on the order of the registration.

jwx:
@fkleuver I like this and think it'll give me some useful options for the router.

fkleuver:
@HamedFathi 

> Can I filter plugins registration by my own rules? For example excludePlugins.
> Aurelia-Toolbelt has many components I want to register all except some.
> Is it possible to use your approach to achieve this?

You can do anything you like with the container registration API. It's completely transparent. Instead of passing in strings that point to module (like vCurrent), you pass in resolvers or registration objects. So you can put your own custom code in there to handle which plugins to load.

AureliaEffect:
JS Framework Benchmark: https://21952-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

PR #541, opened 2019-07-24
fix plugin-conventions tsconfig.json syntax, plus properly fail in various scripts

3cp:


scripts/publish.ts
```diff
@@ -13,13 +13,13 @@ function parseArgs(): {tag: string; suffix: string; registry: string} {
...
+(async function (): Promise<void> {
   const { tag, suffix, registry } = parseArgs();
   const { major, minor, patch } = getCurrentVersion();
   const newVersion = getNewVersion(major, minor, patch, tag, 'none', suffix);
-  lerna(['publish', newVersion, '--dist-tag', tag, '--no-git-tag-version', '--no-push', '--no-verify-registry', '--no-verify-access', '--registry', registry, '-y']);
-}
-
-run().then(() => {
-  log(`Done.`);
+  await lerna(['publish', newVersion, '--dist-tag', tag, '--no-git-tag-version', '--no-push', '--no-verify-registry', '--no-verify-access', '--registry', registry, '-y']);
```
3cp: @fkleuver I added await in front of lerna because there was no await in this async func, but I am not sure whether lerna() returns a promise.

fkleuver:
Excellent, thanks!

PR #542, opened 2019-07-24
WIP: chore: standardize on single quotes

BBosman:
# Pull Request

## 📖 Description

Standardize on single quotes

## 👩‍💻 Reviewer Notes

This is part of my effort to switch from `tslint` to `eslint`, but as it's a very noisy PR I decided to split it off for easy review.

Only changes are changing " and ` (where not explicitly needed) to '

## 📑 Test Plan

CircleCI


BBosman:
The drop in coverage appears to only be from lines where I switched from ` to '. 
I assume that's because it doesn't need to transpile the template strings.

fkleuver:
I think the switch from double to single quote is an obvious one, but I'm not so sure about replacing the backticks.
The main reason I (fairly consistently) use backticks for everything - even when not needed - is because they're generally the more friendly format for generator code that's prone to undergo large sets of changes from time to time, such as our unit tests.
I've been in dozens of situations where I wanted to change some string from static to parameterized and constantly having to change the quotes back and forth is then just a waste of time and energy. Gets me out of the flow.

In general, I simply don't agree with the linter's arguments for why single quotes should be used when possible. It neither makes the code clearer nor more maintainable for me. But perhaps I'm doing something wrong.

What in your opinion are the important benefits here?

BBosman:
After discussing this with @fkleuver on Discord we decided to switch to backticks instead of single quotes.

I'm running into some `typescript-eslint` bugs doing so though, so marking this one as WIP for now.
(Issues 756 and 757 on https://github.com/typescript-eslint/typescript-eslint for those interested)

BBosman:
Putting this on the back burner for now.

PR #545, opened 2019-07-25
fix(plugin-gulp): fix html pair checking in plugin-gulp

3cp:
It's caused by vinyl file.relative is not relative to current working dir.


AureliaEffect:
JS Framework Benchmark: https://22090-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

PR #546, opened 2019-07-25
Simplify code

fkleuver:
# Pull Request

## 📖 Description

This PR essentially reverses a few optimizations I've done a while back, and makes code clearer and easier to understand for other maintainers. Making the code base more contributor friendly hopefully results in more people helping out over time.

Now, in hindsight, some of my perf tweaks might not necessarily have provided that much perf benefit in the first place.

I have learned more about V8 since then, and while some of my optimizations would technically lead to less CPU work required, it also increased the number of megamorphic call sites which in effect might have worsened perf in realistic scenarios.

Furthermore, for a first iteration of AOT, my life would be easier if the core framework actually uses the framework's own APIs wherever possible. It could then apply the same optimizations to the JIT bundle as it could apply to a user's application.

So, we're doing a 180 degree turn here. The optimizations I've done will serve as a useful reference for the kind of code that AOT needs to output. Now we're going to actually let AOT output that and keep the code base nice, clean and idiomatic to how Aurelia should be used.

These particular "deoptimizations" in this PR should not affect performance per se, which is why I think it's acceptable to get them in before we actually have AOT optimizing the bundle.

### 🎫 Issues

N/A

## 👩‍💻 Reviewer Notes

Mostly consists of removing `define` calls and replacing them with decorators again.

Removes a few `static inject` properties, replacing them with parameter decorators.

Also gets rid of a few todo's here and there I stumbled upon.

Introduces a new perf tweak to the renderer (which I might remove again in a week or two though) to make the renderer call sites monomorphic.

## 📑 Test Plan

No behavior added/removed, so existing tests should pass and all should be good.

## ⏭ Next Steps

More of this needs to happen, but one thing at a time.


AureliaEffect:
JS Framework Benchmark: https://22201-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

fkleuver:
Merging now so I can use the changes for my local work. Let me know if there are any comments/questions @EisenbergEffect @bigopon @BBosman @jwx

EisenbergEffect:
Looks good to me 😄

PR #547, opened 2019-07-26
LifecycleFlag propagating template controllers

fkleuver:
# Pull Request

## 📖 Description

This PR is a fun one :)

It introduces 3 new template controllers:
- `infrequent-mutations`- causes all target observers to update the DOM synchronously. The benefit of doing that is, when you have a lot of elements that rarely ever update (such as a very large table), you just need the element creation to be as fast as possible. You won't benefit from scheduling updates, so that won't happen. noticeable in the js-framework-benchmark.
- `frequent-mutations` - the opposite. Eagerly enqueues all target observer in the RAF queue and polls them every frame for new changes. Benefit of doing this, is if you have a lot of bindings that update all the time (e.g. animations), you'll save the overhead of constantly queueing and dequeueing observers. *this was previously the default, but must now be explicitly turned on* - noticeable in the 1kcomponents and sierpinski triangle benchmarks
- `observe-shallow` - only observe "leaf" properties in bindings. Currently, if you have a binding expression like `foo.bind="bar.baz"`, both the `bar` property on the viewmodel as well as the `baz` property on the `bar` object will be observed, when often times you really only need `baz` to be observed. This will often be the case in repeaters, as reordering the items will physically reorder the nodes and they don't need to be re-bound. probably noticeable in about every benchmark and most apps, actually.

## 👩‍💻 Reviewer Notes

There's a base class `FlagsTemplateController` which is actually almost identical to `replaceable`. And, like `BindingModeBehavior`, there are 3 subclasses that only pass one out of 3 specific lifecycle flags into it.

## 📑 Test Plan

Only making sure existing tests pass at the moment. And verifying it through the js-framework-benchmark. I'll add some bulk tests again in a week or two.

## ⏭ Next Steps

Nothing specific apart from adding tests.


AureliaEffect:
JS Framework Benchmark: https://22260-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://22314-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://22324-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://22339-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://22347-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://22356-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://22368-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

Issue #331, opened 2018-12-04
@children decorator [RFC]


davismj:
# 💬 RFC

`@children` should be expanded in the following ways:

1. It should be able to access child view models at various depths and selectors.
2. It should be accessible outside of a decorator.

## 🔦 Context

In vCurrent, I have found `@children` to be the best and only way to solve complex component use cases. Furthermore, this is the only use I have found for it. 

Let's say we have a component `<pane>` that collapses and expands based on its content.

Let's also say we have a component `<accordion>` that has several `<pane>`s. When one `<pane>` opens, whether it is through a click or through a programmatic call to `PaneCustomElement.open()`, all the other `<pane>`s in the `<accordion>` close. It is `<accordion>`'s job to close the other `<pane>`s. To do this, `<accordion>` needs a reference to all of the child `<pane>`s that were added to its slot. 

Note that a `<pane>` can live outside of an `<accordion>`, in which case it is independent of any other `<pane>`.

```html
<accordion>
  <pane id="1"></pane>
  <div>
    <pane id="2"></pane>
  </div>
  <pane id="3">
    <pane id="4"></pane>
  </pane>
  <another-custom-element>
    <pane id="5"></pane>
  </another-custom-element>
</accordion>
```

Right now, the best way to solve this problem is through `@children('pane')`. However, this requires that all `<pane>` elements are direct children of `<accordion>`. In the example above, `<accordion>` can only find out about the view-model for pane#2 by manually looking through the DOM on `attached()` and accessing the private `element.au.controller.viewModel` API. This is hacky solution and in practice I've always let Aurelia force me into a particular DOM structure.

Since the above has been in my experience the most common use case, I think it would be best if `@children('pane')` in `AccordionCustomElement` would access pane#1, pane#2, pane#3. Since we may or may not want to access pane#4 and pane#5, we need some method of supporting those cases on request.

## Other Solutions

**Children.of**

Add a new DI resolver function `Children`, which hoists a newly instantiated dependency to its parent container, up to a configurable depth.

```
@inject(Children.of(PaneCustomElement))
@inject(Children.of(PaneCustomElement), depth = 1)
```

## 💻 Examples

```
<parent-element>
  <child-element id="1"></child-element>
  <div>
    <child-element id="2"></child-element>
  </div>
  <child-element id="3">
    <child-element id="4"></child-element>
  </child-element>
</parent element>
```

`@children('child-element')` on the parent custom element should get a reference to elements 1, 2, 3, and not 4.

[...]

PR #548, opened 2019-07-26
Child Observation

EisenbergEffect:
# Pull Request

## 📖 Description

The ShadowDOMProjector was observing and providing it’s shadow children instead of content children. This PR fixes that. It also introduces filter and select functions to the ChildrenObserver, so that consumers can decide just which children they want and whether they want the element, controller, or viewModel (or whatever else).

There’s still no `@children` decorator or runtime capability to set this up. That still needs to be done. I’d like to be able to have some simple defaults like `@children(TabItem)` or `@children(HTMLElement)` or even `@children(TabItem, ‘.some-selector’)` to get all child view models of a type, or all elements, or all child view models of a type where the element matches a selector, for example. Here’s something that’s typical:

```TypeScript
export class TabControl {
  @children(TabItem) items: TabItem[];
}
```

Advanced scenarios could allow passing the custom filter and select functions, which all of this would be built on.

## 👩‍💻 Reviewer Notes

The code changes are relatively small at this point. Mainly this was about fixing the bugs in the projector and the child observer, and laying a small foundation for the scenarios listed above.

## ⏭ Next Steps

I need some guidance from @fkleuver when he gets a chance on what the best way to handle the decorator and runtime is with the latest iteration. This PR could be merged as is and then we could come back and add the decorator bit, if that’s involved.

### DIFF
--- a/packages/runtime-html/src/projectors.ts
+++ b/packages/runtime-html/src/projectors.ts
@@ -76,9 +76,9 @@ export class ShadowDOMProjector implements IElementProjector<Node> {
     return this.host.childNodes;
   }
 
-  public subscribeToChildrenChange(callback: () => void): void {
+  public subscribeToChildrenChange(callback: () => void, options = childObserverOptions): void {
     // TODO: add a way to dispose/disconnect
-    this.dom.createNodeObserver!(this.host, callback, childObserverOptions);
+    this.dom.createNodeObserver!(this.host, callback, options);
   }
 
   public provideEncapsulationSource(): CustomElementHost<ShadowRoot> {
@@ -117,9 +117,7 @@ export class ContainerlessProjector implements IElementProjector<Node> {
   }
 
   public subscribeToChildrenChange(callback: () => void): void {
-    // TODO: add a way to dispose/disconnect
-    const observer = new MutationObserver(callback);
-    observer.observe(this.host, childObserverOptions);
+    // Containerless does not have a container node to observe children on.
   }
 
   public provideEncapsulationSource(): Node {
--- a/packages/runtime-html/src/projectors.ts
+++ b/packages/runtime-html/src/projectors.ts
@@ -73,12 +73,12 @@ export class ShadowDOMProjector implements IElementProjector<Node> {
   }
 
   public get children(): ArrayLike<CustomElementHost<Node>> {
-    return this.shadowRoot.childNodes;
+    return this.host.childNodes;
   }
 
   public subscribeToChildrenChange(callback: () => void): void {
     // TODO: add a way to dispose/disconnect
-    this.dom.createNodeObserver!(this.shadowRoot, callback, childObserverOptions);
+    this.dom.createNodeObserver!(this.host, callback, childObserverOptions);
   }
 
   public provideEncapsulationSource(): CustomElementHost<ShadowRoot> {
--- a/packages/runtime-html/src/projectors.ts
+++ b/packages/runtime-html/src/projectors.ts
@@ -117,6 +117,7 @@ export class ContainerlessProjector implements IElementProjector<Node> {
   }
 
   public subscribeToChildrenChange(callback: () => void): void {
+    // TODO: turn this into an error
     // Containerless does not have a container node to observe children on.
   }
 
--- a/packages/runtime/src/definitions.ts
+++ b/packages/runtime/src/definitions.ts
@@ -22,6 +22,8 @@ import {
   ensureValidStrategy
 } from './flags';
 import { Bindable } from './templating/bindable';
+import { INode } from './dom';
+import { IController } from './lifecycle';
 
 export type IElementHydrationOptions = { parts?: Record<string, TemplateDefinition> };
 
@@ -32,6 +34,13 @@ export interface IBindableDescription {
   callback?: string;
   attribute?: string;
   property?: string;
+}
+
+export interface IChildrenObserverDescription {
+  callback?: string;
+  property?: string;
+  filter?: (node: INode, controller?: IController, viewModel?: any) => boolean;
+  select?: (node: INode, controller?: IController, viewModel?: any) => any;
 }
 
 /**
@@ -70,6 +79,7 @@ export interface ITemplateDefinition extends IResourceDefinition {
   build?: IBuildInstruction;
   surrogates?: ITargetedInstruction[];
   bindables?: Record<string, IBindableDescription> | string[];
+  childrenObservers?: Record<string, IChildrenObserverDescription>
   containerless?: boolean;
   shadowOptions?: { mode: 'open' | 'closed' };
   hasSlots?: boolean;
--- a/packages/runtime/src/definitions.ts
+++ b/packages/runtime/src/definitions.ts
@@ -43,7 +43,7 @@ export interface IChildrenObserverDescription<TNode extends INode = INode> {
   options?: MutationObserverInit;
   query?: (projector: IElementProjector<TNode>) => ArrayLike<TNode>;
   filter?: (node: INode, controller?: IController<TNode>, viewModel?: IViewModel<TNode>) => boolean;
-  map?: <T>(node: INode, controller?: IController<TNode>, viewModel?: IViewModel<TNode>) => T;
+  map?: (node: INode, controller?: IController<TNode>, viewModel?: IViewModel<TNode>) => any;
 }
 
 export type ChildrenObserverSource = Omit<IChildrenObserverDescription, 'property'>;
--- a/packages/runtime/src/definitions.ts
+++ b/packages/runtime/src/definitions.ts
@@ -23,7 +23,7 @@ import {
 } from './flags';
 import { Bindable } from './templating/bindable';
 import { INode } from './dom';
-import { IController } from './lifecycle';
+import { IController, IViewModel } from './lifecycle';
 import { IElementProjector } from './resources/custom-element';
 
 export type IElementHydrationOptions = { parts?: Record<string, TemplateDefinition> };
@@ -37,14 +37,16 @@ export interface IBindableDescription {
   property?: string;
 }
 
-export interface IChildrenObserverDescription {
+export interface IChildrenObserverDescription<TNode extends INode = INode> {
   callback?: string;
   property?: string;
   options?: MutationObserverInit;
-  query?: (projector: IElementProjector) => ArrayLike<INode>;
-  filter?: (node: INode, controller?: IController, viewModel?: any) => boolean;
-  map?: (node: INode, controller?: IController, viewModel?: any) => any;
+  query?: (projector: IElementProjector<TNode>) => ArrayLike<TNode>;
+  filter?: (node: INode, controller?: IController<TNode>, viewModel?: IViewModel<TNode>) => boolean;
+  map?: <T>(node: INode, controller?: IController<TNode>, viewModel?: IViewModel<TNode>) => T;
 }
+
+export type ChildrenObserverSource = Omit<IChildrenObserverDescription, 'property'>;
 
 /**
  * TargetedInstructionType enum values become the property names for the associated renderers when they are injected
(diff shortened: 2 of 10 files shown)


packages/runtime/src/rendering-engine.ts
```diff
@@ -425,7 +425,6 @@ export class ChildrenObserver implements Partial<IChildrenObserver> {
   private select: typeof defaultChildSelect;
 
   constructor(
-    lifecycle: ILifecycle,
```
EisenbergEffect: @fkleuver This seemed redundant since the controller has a lifecycle instance. We can just use that, right?

EisenbergEffect:
For the decorator, we could do something really simple and just call define property on the prototype, creating a getter and setter that lazily instantiated the child observer using the instance’s controller. Any reason not do to that @fkleuver ?

packages/runtime-html/src/projectors.ts
```diff
@@ -73,12 +73,12 @@ export class ShadowDOMProjector implements IElementProjector<Node> {
   }
 
   public get children(): ArrayLike<CustomElementHost<Node>> {
-    return this.shadowRoot.childNodes;
+    return this.host.childNodes;
```
bigopon: This is where we could throw a more meaningful error if host is not an element (containerless)

EisenbergEffect: In this particular case, it will always be an element. Take a look at the ContainerlessProjector. That’s where we may want to make some changes.

AureliaEffect:
JS Framework Benchmark: https://22328-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

fkleuver:
@EisenbergEffect 

> For the decorator, we could do something really simple and just call define property on the prototype, creating a getter and setter 

Doing it on the prototype seems fine by me but you do need to make sure the `$observers` lookup exists on the instance somehow. If `@children` is the only decorator, nothing will initialize that lookup.

Also, we shouldn't directly mutate the instance or prototype directly from the decorator because native decorators (when they land) won't allow you to arbitrarily monkey patch / mixin stuff.
They will allow defining arbitrary metadata, so we should work via the `description` ideally and let some part of the rendering process take care of defining the getter/setter.
This will ensure smooth transition to native decorators when they land.

packages/runtime/src/rendering-engine.ts
fkleuver: Yeah that's fine

EisenbergEffect:
@fkleuver If you get a chance to review, my last commit attempts to add runtime support for child observation by fixing up the children observer, adding the getter/setter code, and hooking that into the controller’s code that creates observers. I haven’t created the decorator yet. I wanted you to take a look at this bit first. Thanks!

packages/runtime/src/templating/controller.ts
```diff
@@ -1037,6 +1039,32 @@ function createObservers(
...
+            childrenDescription.callback,
+            childrenDescription.filter,
+            childrenDescription.select
+          );
+        }
+      }
+    }
+  }
+
   if (!useProxy) {
```
fkleuver: We should add a condition here, e.g.
```ts
if (!useProxy || hasChildrenObserver) {
```

To make sure the lookup still gets stored in proxy mode, since this is the only way to observe children even in proxy mode, and the lookup does need to get created in that case.

fkleuver:
@EisenbergEffect It looks good. Since it's not a target observer, it shouldn't go through the RAF scheduler, so good call on removing that bit.

This is not very important, but perhaps we want to allow users to provide a normal list of strings for the `childrenObservers` property like with the bindables, where the defaults are used? Makes creating on the fly definitions (and certain unit tests) a bit cleaner/easier.

Speaking of tests, there are some old children observer tests commented out that we could add back in for this. They may need some small tweaks, but should largely be usable as they were:

https://github.com/aurelia/aurelia/blob/a9cf6b1aad757209746e8cae7ba0626ca6f149b0/packages/__tests__/jit-html/repeater-custom-element.spec.ts#L181

https://github.com/aurelia/aurelia/blob/a9cf6b1aad757209746e8cae7ba0626ca6f149b0/packages/__tests__/jit-html/repeater-custom-element.spec.ts#L405

davismj:
Any spec about how this works? I've found that the au1 `@children` decorator is almost always too limited for practical use cases and therefore always opt directly into a strategy that uses `MutationObserver`. Hopefully au2 cma resolve some of those issues?

fkleuver:
Maybe add an argument to the decorator that allows specifying the options for the mutation observer (e.g. `childTree` and/or query selector for fetching the elements?

fkleuver:
Keeping [your RFC](https://github.com/aurelia/aurelia/issues/331) in mind perhaps we should have an additional `@descendants` decorator for that scenario, and keep this one simple?

davismj:
I'm still a fan of `@descendants(what, depth)` with `@children` as an alias for depth 1.

fkleuver:
Yeah that makes sense. We might need to defer that though, depending on how much time @EisenbergEffect has for this

EisenbergEffect:
For now, I want to keep it to just children and get it working end-to-end. After that we can expand that with subtree support. We might want to have a different decorator, such as `descendants` or `subtree` though. I’m not sure it makes sense to build children on top of a general descendants implementation yet though. There are different requirements. For example, getting children is pretty simple and cheap, where as descendants would require some sort of query selector, for example. If you want depth support, that has to be calculated for each node that the query selector matches. That’s not built into mutation observers or the dom itself, so that’s more expensive. It’s still an open question for me as to whether or not we want depth support, as that seemed mostly needed in vCurrent because of the incomplete shadow dom polyfill, which didn’t patch up the mutation observer. However, native shadow dom or the shady dom polyfill would not suffer from that issue.

EisenbergEffect:
@fkleuver There’s a huge similarity between the bindable and child observer setup code in the controller. I’m wondering if we should generalize this. For example, we could make the bindable description take a constructor or a unified shape as part of its configuration and then pass the configuration and context data to that. Then, we could register the child observers in the same object as the bindables and open this up to customizations, such as creating a completely different implementation for subtrees, if needed. For example:

```TypeScript
  const { length } = observableNames;
  let name: string;
  for (let i = 0; i < length; ++i) {
    name = observableNames[i];
    const config = bindables[name];

    if (observers[name] == void 0) {
      observers[name] = new config.ctor(
        config,
        controller,
        flags,
        useProxy
      );
    }
  }
```

fkleuver:
@EisenbergEffect 

In my opinion, `@bindable` and `@children` are significantly different in nature and usage, and the similarity in their initialization is merely coincidental. I don't think generalizing their initialization into a single abstract operation conceptually makes sense, and has a fair chance of causing difficulties later down the road, when we add additional capabilities to `@bindable` and integrate with AOT, etc.

So, I would prefer keeping the two initializations as separate as possible, even if this leads to some duplicate code.

EisenbergEffect:
That’s fine, I was just thinking of ways that someone like @davismj could add what he wanted without it being required to be hard-coded into the framework.

fkleuver:
Why not just parameterize the children retrieval and mutation observer arguments then?

```ts
export function filterChildren(
  projector: IElementProjector, // <---
  retrieve: typeof defaultChildRetrieve, // <---
  filter: typeof defaultChildFilter,
  select: typeof defaultChildSelect
): any[] {
  const children = [];
  const nodes = retrieve(projector); // <--

  // ...

  return children;
}

function defaultChildRetrieve(projector: IElementProjector): INode[] {
  return projector.children; // <--
}

// ... in projectors.ts

  public subscribeToChildrenChange(callback: () => void, options = childObserverOptions): void {
    this.dom.createNodeObserver!(this.host, callback, options);  // ^
  }

// ...

export interface IChildrenObserverDescription {
  callback?: string;
  property?: string;
  options?: MutationObserverInit; // <--
  retrieve?: (projector: IElementProjector) => readonly INode[]; // <--
  filter?: (node: INode, controller?: IController, viewModel?: any) => boolean;
  select?: (node: INode, controller?: IController, viewModel?: any) => any;
}

// ...

  private tryStartObserving() {
    if (!this.observing) {
      this.observing = true;
      this.children = filterChildren(this.projector, this.retrieve, this.filter, this.select);
                                          // ^               ^
      this.projector.subscribeToChildrenChange(() => { this.onChildrenChanged(); }, this.options);
                                                                                       // ^
    }
  }
```

EisenbergEffect:
Yeah, I was thinking of something pretty similar to that. I can add it in my next commit.

packages/runtime-html/src/projectors.ts
```diff
@@ -117,9 +117,7 @@ export class ContainerlessProjector implements IElementProjector<Node> {
   }
 
   public subscribeToChildrenChange(callback: () => void): void {
-    // TODO: add a way to dispose/disconnect
-    const observer = new MutationObserver(callback);
-    observer.observe(this.host, childObserverOptions);
+    // Containerless does not have a container node to observe children on.
```
fkleuver: Shouldn't we just throw? If a user puts a `@children` decorator on a containerless, that is an error situation anyway. Instead of leaving them wondering why it's not working, throwing an error would immediately let them know what's wrong.

packages/runtime/src/definitions.ts
```diff
@@ -34,6 +37,15 @@ export interface IBindableDescription {
   property?: string;
 }
 
+export interface IChildrenObserverDescription {
+  callback?: string;
+  property?: string;
+  options?: MutationObserverInit;
+  query?: (projector: IElementProjector) => ArrayLike<INode>;
+  filter?: (node: INode, controller?: IController, viewModel?: any) => boolean;
+  map?: (node: INode, controller?: IController, viewModel?: any) => any;
```
fkleuver: Suggestion:

```ts
export interface IChildrenObserverDescription<TNode extends INode = INode> {
  callback?: string;
  property?: string;
  options?: MutationObserverInit;
  query?: (projector: IElementProjector<TNode>) => ArrayLike<TNode>;
  filter?: (node: INode, controller: IController<TNode>, viewModel: IViewModel<TNode>) => boolean;
  map?: <T>(node: INode, controller: IController<TNode>, viewModel: IViewModel<TNode>) => T;
}
```

Passing through the `TNode` type argument allows the end user to properly expose the real `lib.dom.d.ts` typings and get access to the DOM api's in a strongly typed manner.

As for making `controller` and `viewModel` non-optional: these are callbacks invoked by the framework, and these arguments are guaranteed to be provided, so them being optional would only force the user to perform a (redundant) nullcheck or use non-null assertion.
The user is never forced to supply a function that accepts all these arguments anyway.

fkleuver: See also: https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html#optional-parameters-in-callbacks

EisenbergEffect: Actually, there won’t necessarily be a controller or a view model. If they just want to select html elements that aren’t custom elements, those values will be undefined.

EisenbergEffect:
Ok, so first pass at integrating the decorator. Here’s what the basic usage would be:

```TypeScript
export class TabControl {
  @children tabItems;
}
```

That would get you view models for all child elements that are custom components. I’d probably want to only get TabItem instances, and so to be safe could do this:

```TypeScript
export class TabControl {
  @children({ filter: (node, ctrl, vm) => vm instanceof TabItem })
  tabItems;
}
```

@davismj could do what he wants like this:

```TypeScript
export class TabControl {
  @children({ 
    query: p => p.host.querySelectorAll(‘.some-class’)
    filter: node => calcDepth(node) < 2
  })
  tabItems;
}
```

AureliaEffect:
JS Framework Benchmark: https://22517-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://22736-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

EisenbergEffect:
@fkleuver Ok, I’ve got this implemented and added a few tests for the most common scenarios. It’s not exhaustive, but I’m hoping it’s enough to get this in and move forward. @davismj may want to add some more tests for custom mutation observer options, etc. in another PR, but the basic query, filter, map, change callback, and children property update are all tested.

AureliaEffect:
JS Framework Benchmark: https://22748-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

PR #552, opened 2019-07-29
feat(focus): add focus attribute

bigopon:
# Pull Request

## 📖 Description

Add focus attribute, as one of core attributes in vCurrent

### 🎫 Issues

N/A

## 👩‍💻 Reviewer Notes

N/A

## 📑 Test Plan

Need to re-enable multiple binding scenarios once the support is back:

```html
Working:
<input focus.two-way="hasFocus">
Not working:
<input focus="value.two-way: hasFocus">
```

## ⏭ Next Steps

Add more tests

### DIFF
--- a/packages/__tests__/jit-html/focus.spec.ts
+++ b/packages/__tests__/jit-html/focus.spec.ts
@@ -150,17 +150,10 @@ describe('focus.spec.ts', function() {
           const isFocusable = ceProp && (typeof ceProp.tabIndex !== undefined || ceProp.contentEditable);
           // tslint:disable-next-line:insecure-random
           const ceName = `ce-${Math.random().toString().slice(-6)}`;
+          const CustomEl = defineCustomElement(ceName, ceTemplate, { tabIndex: 1 }, shadowMode);
 
           it(`works with ${isFocusable ? 'focusable' : ''} custom element ${ceName}, #shadowRoot: ${shadowMode}`, function() {
             let callCount = 0;
-
-            const { au, component, ctx } = setup<IApp>(
-              `<template><${ceName} focus.bind=hasFocus></${ceName}></template>`,
-              class App {
-                public hasFocus = true;
-              }
-            );
-            const CustomEl = defineCustomElement(ctx, ceName, ceTemplate, { tabIndex: 1 }, shadowMode);
             // only track call, virtually no different without this layer
             CustomEl.prototype['focus'] = function focus(options?: FocusOptions): void {
               callCount++;
@@ -178,6 +171,13 @@ describe('focus.spec.ts', function() {
                 return HTMLElement.prototype.focus.call(this, options);
               }
             };
+
+            const { au, component, ctx } = setup<IApp>(
+              `<template><${ceName} focus.bind=hasFocus></${ceName}></template>`,
+              class App {
+                public hasFocus = true;
+              }
+            );
 
             const activeElement = ctx.doc.activeElement;
             const ceEl = ctx.doc.querySelector(`app ${ceName}`);
@@ -348,7 +348,7 @@ describe('focus.spec.ts', function() {
     return { ctx, container, lifecycle, host, au, component, observerLocator };
   }
 
-  function defineCustomElement(ctx: HTMLTestContext, name: string, template: string, props: Record<string, any> = null, mode: 'open' | 'closed' | null = 'open') {
+  function defineCustomElement(name: string, template: string, props: Record<string, any> = null, mode: 'open' | 'closed' | null = 'open') {
     class CustomEl extends HTMLElement {
       constructor() {
         super();
--- a/packages/__tests__/jit-html/focus.spec.ts
+++ b/packages/__tests__/jit-html/focus.spec.ts
@@ -1,9 +1,13 @@
-import { Constructable } from '@aurelia/kernel';
+import { Constructable, PLATFORM } from '@aurelia/kernel';
 import { Aurelia, CustomElement } from '@aurelia/runtime';
 import { Focus } from '@aurelia/runtime-html';
 import { assert, eachCartesianJoin, HTMLTestContext, TestContext } from '@aurelia/testing';
 
 describe('focus.spec.ts', function() {
+
+  if (!PLATFORM.isBrowserLike) {
+    return;
+  }
 
   interface IApp {
     hasFocus?: boolean;
--- a/packages/__tests__/jit-html/focus.spec.ts
+++ b/packages/__tests__/jit-html/focus.spec.ts
@@ -16,29 +16,11 @@ describe('focus.spec.ts', function() {
     selectedOption?: string;
   }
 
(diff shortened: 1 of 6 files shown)


AureliaEffect:
JS Framework Benchmark: https://22546-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://22552-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://22570-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://22607-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/runtime-html/src/resources/custom-attributes/focus.ts
```diff
@@ -0,0 +1,133 @@
...
+   */
+  private needsApply: boolean;
+
+  // This is set by the controller after this instance is constructed
+  // tslint:disable-next-line: prefer-readonly
+  private $controller!: IController;
+
+  constructor(
+    element: HTMLElement,
+    dom: HTMLDOM
```
fkleuver:
```ts
constructor(
  @INode private readonly element: HTMLElement,
  @IDOM private readonly dom: HTMLDOM,
)
```
?

packages/runtime-html/src/resources/custom-attributes/focus.ts
```diff
@@ -0,0 +1,133 @@
...
+   * Invoked everytime the bound value changes.
+   * @param newValue The new value.
+   */
+  public valueChanged(): void {
+    // In theory, we could/should react immediately
+    // but focus state of an element cannot be achieved
+    // while it's disconnected from the document
+    // thus, there neesd to be a check if it's currently connected or not
+    // before applying the value to the element
+    if (this.$controller.flags & State.isAttached) {
```
fkleuver:
```suggestion
    if (this.$controller.state & State.isAttached) {
```

`.flags` is `persistentFlags`, has nothing to do with state. Should probably in fact be hidden from public api

packages/runtime-html/src/resources/custom-attributes/focus.ts
```diff
@@ -0,0 +1,133 @@
...
+    element: HTMLElement,
+    dom: HTMLDOM
+  ) {
+    this.element = element;
+    this.dom = dom;
+    this.needsApply = false;
+  }
+
+  public binding(): void {
+    this.valueChanged();
```
fkleuver: I don't think this makes sense. `valueChanged()` will just set `this.needsApply = true` 100% of the time, because this hook is never invoked when the controller is attached. And if you're always setting `this.needsApply = true` then why bother with this hook at all, and not just always call `apply()` in attached()? Boils down to the same thing but with simpler code

packages/runtime-html/src/resources/custom-attributes/focus.ts
```diff
@@ -0,0 +1,133 @@
...
+    addListener(el, 'blur', this);
+  }
+
+  /**
+   * Invoked when the attribute is detached from the DOM.
+   */
+  public detached(): void {
+    // the following block is to help minification
+    // as addEventListener method name, together with element property
+    // adds quite a bit of munification unfriendly code
```
fkleuver: I'd rather do this in a unified and consistent manner, ideally with AOT or at least in a way that the whole code base does this the same way. Please let's not have loose functions scattered over the place. Nothing is minification friendly quite yet so it's ok for this to be the same. We can fix that later.

packages/__tests__/jit-html/focus.spec.ts
```diff
@@ -0,0 +1,379 @@
...
+
+  interface IApp {
+    hasFocus?: boolean;
+    isFocused?: boolean;
+    isFocused2?: boolean;
+    selectedOption?: string;
+  }
+
+  let $aurelia: Aurelia;
+  let $host: HTMLElement;
```
fkleuver: Can we use setup/teardown functions please instead of shared variables with hooks :P

packages/__tests__/jit-html/focus.spec.ts
```diff
@@ -0,0 +1,379 @@
...
+          this[p] = props[p];
+        }
+      }
+    }
+    customElements.define(name, CustomEl);
+    return CustomEl;
+  }
+
+  function waitForDelay(time = 0): Promise<void> {
+    return new Promise(r => setTimeout(r, time));
```
fkleuver: What about rAF (or 2 rAFs) instead of 50ms? There should be something deterministic possible here

bigopon: I copy-pasted this from pre-deterministic time long long ago, will fix

packages/runtime-html/src/resources/custom-attributes/focus.ts
bigopon: will make it consistent with the rest

packages/runtime-html/src/resources/custom-attributes/focus.ts
bigopon: Will there be scenario where an instance of this attribute is detached, but still bound?

packages/runtime-html/src/resources/custom-attributes/focus.ts
bigopon: Will there be scenario where an instance of this attribute is detached, but still bound?

AureliaEffect:
JS Framework Benchmark: https://22897-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://22903-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/runtime-html/src/resources/custom-attributes/focus.ts
fkleuver: @bigopon It is possible yes, in particular this situation happens when the repeater reorders items - it detaches them at their old location and re-attaches them at their new location. Though I'm really not sure if it's 1) going to stay that was and 2) how we want stuff to behave in this scenario in the first place

packages/runtime-html/src/resources/custom-attributes/focus.ts
fkleuver: Yes. Moving an item in the DOM (e.g. repeater reorder) can make it detach and then attach again, without unbinding and binding in between.

PR #554, opened 2019-07-29
feat(observer): Add the ability to bind an object to class

brandonseydel:
# Pull Request

## 📖 Description

- Add the ability to bind an object to class and check value of each property and if true then add
the property name to the class list.
- Fixed an issue when binding a class list with an empty string at the end by changing the regex to match instead of string split. (Minor perf improvement as well)

### 🎫 Issues

N/A

## 👩‍💻 Reviewer Notes

N/A

## 📑 Test Plan

Added additional tests to cover different scenarios when binding an object.

## ⏭ Next Steps

E2E tests need added.

### DIFF
--- a/packages/__tests__/runtime-html/target-observers.spec.ts
+++ b/packages/__tests__/runtime-html/target-observers.spec.ts
@@ -5,7 +5,7 @@ import {
   DataAttributeAccessor,
   StyleAttributeAccessor
 } from '@aurelia/runtime-html';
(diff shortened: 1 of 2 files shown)


AureliaEffect:
JS Framework Benchmark: https://22810-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

PR #558, opened 2019-08-01
 feat(observer): Add the ability to bind an array of objects and strings to a class

brandonseydel:
# Pull Request

## 📖 Description

This PR gives the ability to pass an array of strings or objects to class bind

```typescript
['class1', {class2: true, 'class3 class4': true, class5: false}, 'class6']
```
results in 
```html
<input class="class1 class2 class3 class4 class6" />
```

### 🎫 Issues

N/A
## 👩‍💻 Reviewer Notes

N/A

## 📑 Test Plan

Unit tests added to accomodate changes.

## ⏭ Next Steps

E2E needs added


AureliaEffect:
JS Framework Benchmark: https://23038-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/runtime-html/src/observation/class-attribute-accessor.ts
```diff
@@ -110,14 +106,28 @@ export class ClassAttributeAccessor implements IAccessor<unknown> {
     }
   }
 
-  private getClassesToAdd(object: Record<string, unknown>): string[] {
+  private splitClassString(classString: string): string[] {
+    return classString.match(/\S+/g) || [];
+  }
+
+  private getClassesToAdd(object: Record<string, unknown> | [] | string): string[] {
```
fkleuver: Perf optimization suggestion:

```ts
  private splitClassString(classString: string): string[] {
    const matches = classString.match(/\S+/g);
    if (matches === null) {
      return PLATFORM.emptyArray;
    }
    return matches;
  }

  private getClassesToAdd(object: Record<string, unknown> | [] | string): string[] {
    if (typeof object === 'string') {
      return this.splitClassString(object);
    }

    if (object instanceof Array) {
      const len = object.length;
      if (len > 0) {
        const classes: string[] = [];
        for (let i = 0; i < len; ++i) {
          classes.push(...this.getClassesToAdd(object[i]));
        }
        return classes;
      } else {
        return PLATFORM.emptyArray;
      }
    } else if (object instanceof Object) {
      const classes: string[] = [];
      for (const property in object) {
        // Let non typical values also evaluate true so disable bool check
        // tslint:disable-next-line: strict-boolean-expressions
        if (!!object[property]) {
          // We must do this in case object property has a space in the name which results in two classes
          if (property.indexOf(' ') >= 0) {
            classes.push(...this.splitClassString(property));
          } else {
            classes.push(property);
          }
        }
      }
      return classes;
    }
    return PLATFORM.emptyArray;
  }
```

3cp:
Something wrong with codecov, it generated and emailed out 4 copies of result.

fkleuver:
yeah it happens on occasion, probably an overzealous retry mechanism somewhere

AureliaEffect:
JS Framework Benchmark: https://23227-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

PR #563, opened 2019-08-06
feat(i18n): core translation service and related auxiliary features

Sayan751:
# Pull Request

## 📖 Description
This PR adds the core implementation of updating the content and attribute of the elements in view. The implementation is based on attribute pattern, binding, and render combination. The previous approach with custom attribute is removed in favor of abstracting the DOM manipulation via Binding, and ObserverLocator.

Moreover this PR is updated to include the auxiliary i18n features such as locale specific number, date, and relative time formatting.

This PR includes following.

- Various syntax for `t`, such as `t="string.key"`, `t="${interpolated.key}"`, `t="${multi}.${interpolation}"`, `t.bind="vm.prop"`.
- Manipulation of the translations using `t-params.bind`. This should encompass the `i18next` use cases of context-based translation, pluralization, key nesting etc.
- Support for aliases; i.e. translation attribute aliases can be registered while configuring the plugin
- Change of locale and updating the translations with that
- Following use cases in vCurrent
  - replacement of `textContent` by defualt
  - replacement of `src` of `img` element by default
  - attribute translation `t="[attr]key"`
  - multiple translation target `t="[attr1,attr2]key1;[attr3]key2"`
  - replacement of `innerHTML` with `t="[html]key"`
  - appending and prepending additional text `t="[append]key1;[prepend]key2"`
  - translation value converter and binding behavior (both available as `t`, same as vCurrent)
  - number formatting
    - `I18nService#nf` provides formatted number string (`i18n.nf(number, [options], [locale]): string`). Note that this is a _breaking change_ as in vCurrent the similar method returned an `Intl.NumberFormat` instance. Similar method is still available as `I18nService#createNumberFormat`
    - signalable value converter and binding behavior 
  - date formatting: Similar behavior as described above for `nf` but for date. This is also a _breaking change_.
  - relative time formatting: this is again a _breaking change_. Instead of a  completely separate class for solely relative time formatting function, in vNext this is just a method as in `I18nService#rt`. The implementation relies on new `Intl.RelativeTimeFormat` API, which is supported already in v8 and FF. For the other browsers, polyfills are available.

Note that all value converters are signalable, unlike vCurrent.

## 👩‍💻 Reviewer Notes

1. The renderer section at this point is bit messy. As the value of `t` attribute can come in different shapes, it is important to be able to parse a wide range of expressions. At this point there is a different renderer for `t-params`, which should be like that. However, there are 2 renderers to support the `t` and `t.bind`. The binding  command for `t` uses `CustomCommand` binding type, which causes parsing of the key expression altogether, which is intended as the i18n key expression parsing happen in binding. Therefore it is not a problem to pass on the string expression as is.  All 3 renderers instantiate and manipulate a single instance of binding that does the actual work of translation. Any idea for a more compact implementation would be great.
2. In the binding, there are still some direct DOM manipulations, due to the support for `[append]`, and `[prepend]`. @zewa666 if you think this is an underused feature, can this be dropped? In that case, the binding implementation would be a lot less complicated. Otherwise, if the current implementation is too messy, alternative approaches needs to be devised.
3. As the new relative time formatting relies on `Intl` API, there is a slight change in output as compared to vCurrent. It seems that for relative time formatting the [lowest time unit supported](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RelativeTimeFormat/format#Parameters) is `second`. However, in vCurrent those cases are formatted as a variation of "now". In vNext those cases are approximated to 1 second and formatted using the `Intl` API. This saves us the maintenance of custom translations. aye or nay?

## 📑 Test Plan

At present mostly there are integration and e2e tests. There are also few unit tests written, but as the underlying components are in nascent stage and still changing, more involved unit tests are postponed.

## ⏭ Next Steps

Docs, docs, and docs. And if there is any missing features detected.


AureliaEffect:
JS Framework Benchmark: https://23663-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://24063-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://24085-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://24184-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://24224-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://24232-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://24318-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://24628-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/__tests__/e2e/src/startup.ts
```diff
@@ -1,21 +1,41 @@
...
+          en: { translation: en['default'] },
+          de: { translation: de['default'] },
+        },
+        skipTranslationOnMissingKey: !!new URL(location.href).searchParams.get('skipkey')
+      };
+    })
   )
+  .register(...[
+    SutI18N, CustomMessage
+  ] as unknown as IRegistration[])
```
zewa666: why is that unknown cast first needed?

packages/__tests__/e2e/tests/integration/i18n.e2e.ts
```diff
@@ -1,26 +1,315 @@
...
+
 describe('i18n', () => {
   beforeEach(() => { cy.visit('/'); });
 
   const assertContent = (selector: string, expected: string, isHtmlContent: boolean | undefined = false) => {
     cy.get(selector).should(isHtmlContent ? 'have.html' : 'have.text', expected);
   };
 
-  const specs = [
+  const changeCurrentLocaleToDe = () => { cy.get('#locale-changer-de').click(); };
```
zewa666: this is super over-the-top and not related to this PR but I was wondering whether we should extract all id selectors into consts and assign the actual IDs via bindings. Over in the tests we then simply import the const and use it as selector. This way the connection between SUT and Test should be more stable.

packages/__tests__/e2e/tests/integration/i18n.e2e.ts
```diff
@@ -1,26 +1,315 @@
...
   ];
 
-  for (const { name, suts } of specs) {
-    it(name, () => {
-      for (const sut of suts) {
-        assertContent(sut.selector, sut.expected);
-      }
+  describe('translates via HTML that', () => {
+    for (const { name, suts, isHtmlContent } of [...enSpecificTests, ...specs]) {
+      it(name, function () {
```
zewa666: do you need a new scope or why the use of function vs fat-arrow here?

packages/__tests__/e2e/tests/integration/i18n.e2e.ts
```diff
@@ -1,26 +1,315 @@
...
+      },
+      {
+        name: 'interval',
+        suts: [{ selector: '#i18n-code-interval', expected: 'a lot of items' }]
+      },
+      {
+        name: 'nf',
+        suts: [{ selector: '#i18n-code-num', expected: '123,456,789' }]
+      },
+      ...['simple', 'locale', 'currency', 'text', 'minus'].map((part) => (
```
zewa666: super slick, like this approach

packages/__tests__/i18n/df/date-format-value-converter.spec.ts
```diff
@@ -0,0 +1,62 @@
+import { DateFormatValueConverter, I18nService } from '@aurelia/i18n';
+import { EventAggregator } from '@aurelia/kernel';
+import { assert, MockSignaler } from '@aurelia/testing';
+import i18next from 'i18next';
+
+describe('df', function () {
+
+  async function setup() {
+    const i18n = new I18nService({ i18next }, {}, new EventAggregator(), new MockSignaler());
+    await i18n['task'].wait();
```
zewa666: whats this?

packages/__tests__/i18n/i18n.spec.ts
```diff
@@ -62,24 +68,285 @@ describe('I18N', function () {
...
+
+  it('setLocale changes the active language of i18next', async function () {
+    const { sut, eaSpy, mockSignaler } = await setup();
+
+    await sut.setLocale('de');
+
+    eaSpy.methodCalledOnceWith('publish', [I18N_EA_CHANNEL, { newLocale: 'de', oldLocale: 'en' }]);
+    const dispatchCall = mockSignaler.calls.find((call) => call[0] === 'dispatchSignal');
+    assert.notEqual(dispatchCall, undefined);
+    const [, args] = dispatchCall;
```
zewa666: your code is full of some great little tricks

packages/__tests__/i18n/i18n.spec.ts
```diff
@@ -62,24 +68,285 @@ describe('I18N', function () {
...
+      const { sut } = await setup();
+
+      const nf = sut.createDateTimeFormat({ month: 'short', timeZoneName: 'long' });
+      assert.instanceOf(nf, Intl.DateTimeFormat);
+      const options = nf.resolvedOptions();
+      assert.equal(options.month, 'short');
+      assert.equal(options.timeZoneName, 'long');
+    });
+  });
+  describe('df', function () {
```
zewa666: OCD: missing new-line

packages/__tests__/i18n/i18n.spec.ts
```diff
@@ -62,24 +68,285 @@ describe('I18N', function () {
...
+      const { sut } = await setup();
+
+      const nf = sut.createNumberFormat({ currency: 'EUR', style: 'currency' });
+      assert.instanceOf(nf, Intl.NumberFormat);
+      const options = nf.resolvedOptions();
+      assert.equal(options.currency, 'EUR');
+      assert.equal(options.style, 'currency');
+    });
+  });
+  describe('nf', function () {
```
zewa666: OCD: newline

packages/__tests__/i18n/i18n.spec.ts
```diff
@@ -62,24 +68,285 @@ describe('I18N', function () {
...
 
-    const span = DOM.createElement('span');
-    sut.updateValue(span as any, 'simple.text');
-    await sut['task'].wait();
+    for (const multiplier of [1, -1]) {
+      for (const value of [7, 14]) {
+        it(`works for time difference in weeks - ${multiplier > 0 ? 'future' : 'past'} - ${value > 7 ? 'plural' : 'singular'}`, async function () {
+          const { sut } = await setup();
+          const input = new Date();
+          input.setDate(input.getDate() + multiplier * value);
```
zewa666: sweet

packages/__tests__/i18n/t/translation-integration.spec.ts
```diff
@@ -0,0 +1,939 @@
...
+        name: 'app', template: `<custom-message t="[message]simple.text"></custom-message>`
+      })
+      class App { }
+
+      const host = DOM.createElement('app');
+      const { de, container } = await setup(host, new App());
+      const i18n = container.get(I18N);
+      await i18n.setLocale('de');
+
+      await new Promise((resolve) => {
```
zewa666: can we use the aurelia-taskqueue with a queued macro task instead of this?

packages/__tests__/i18n/t/translation-parameters-render.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+  });
+
+  it('compiles the binding to a TranslationParametersBindingInstruction', function () {
+    const sut = setup();
+    const syntax: AttrSyntax = { command: 't-params.bind', rawName: 't-params.bind', rawValue: '{foo: "bar"}', target: '' };
+
+    // tslint:disable: no-object-literal-type-assertion
+    const actual = sut.compile({
+      command: new AttrBindingCommand(),
+      flags: (void 0)!,
```
zewa666: huch? whats that construct for? perf optimization?

packages/__tests__/i18n/t/translation-parameters-render.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+
+  function setup() {
+    const container = DI.createContainer();
+    container.register(RuntimeBasicConfiguration, I18nConfiguration);
+    return container;
+  }
+
+  it('instantiated with instruction type', function () {
+    const container = setup();
+    const sut: IInstructionRenderer = new TranslationParametersBindingRenderer(container.get(IExpressionParser), {} as unknown as IObserverLocator);
```
zewa666: does container.get accept a generic?

`container.get<IObserverLocator>`

packages/__tests__/i18n/t/translation-render.spec.ts
```diff
@@ -0,0 +1,207 @@
...
+    const value = 'simple.key';
+
+    const actual: AttrSyntax = sut[pattern](pattern, value, []);
+    assert.equal(actual.command, pattern);
+    assert.equal(actual.rawName, pattern);
+    assert.equal(actual.rawValue, value);
+    assert.equal(actual.target, '');
+  });
+
+  // it('creates attribute syntax with `to` part when `T.bind="expr"` is used', function () {
```
zewa666: leftover?

packages/__tests__/i18n/t/translation-render.spec.ts
```diff
@@ -0,0 +1,207 @@
...
+    const suts = setup(aliases);
+
+    assert.equal(suts.length, aliases.length);
+    assert.equal(
+      suts.every((sut) => sut instanceof TranslationBindingCommand),
+      true);
+  });
+
+  it('compiles the binding to a TranslationBindingInstruction', function () {
+    const sut = setup()[0];
```
zewa666: const [ sut ] = setup();

packages/i18n/src/Intl.d.ts
```diff
@@ -0,0 +1,23 @@
+/**
```
zewa666: I don't like that they're located inside the src folder.
Is there any chance we move that out into something like `packages/i18n/custom-defintions`?

packages/i18n/src/configuration.ts
```diff
@@ -23,4 +87,4 @@ function createI18nConfiguration(optionsProvider: I18NConfigOptionsProvider) {
   };
 }
 
-export const I18nConfiguration = createI18nConfiguration(() => Object.create(null));
+export const I18nConfiguration = createI18nConfiguration(() => {/* noop */ });
```
zewa666: OCD: missing space before /*

packages/i18n/src/i18n-configuration-options.ts
```diff
@@ -9,12 +9,17 @@ export type I18nModule =
   | i18next.I18nFormatModule
   | i18next.ThirdPartyModule;
 
-export const I18nConfigurationOptions = DI.createInterface<I18nConfigurationOptions>('I18nConfigurationOptions');
-export interface I18nConfigurationOptions extends i18next.InitOptions {
+export const I18nInitOptions = DI.createInterface<I18nInitOptions>('I18nInitOptions');
```
zewa666: why the rename? If kept around I'd also vote for renaming the file itself to stay in match

packages/i18n/src/i18n.ts
```diff
@@ -12,76 +20,177 @@ export const I18N = DI.createInterface<I18nService>('I18nService').noDefault();
...
+    @IEventAggregator private readonly ea: IEventAggregator,
+    @ISignaler private readonly signaler: ISignaler
+  ) {
     this.i18next = i18nextWrapper.i18next;
     this.task = new PromiseTask(this.initializeI18next(options), null, this);
+    this.intl = PLATFORM.global.Intl;
   }
 
-  public tr(key: string | string[], options?: i18nextCore.TOptions<object>) {
+  public evaluate(keyExpr: string, options?: i18nextCore.TOptions): I18nKeyEvaluationResult[] {
```
zewa666: This one could use a jsdoc comment. From its name its not immediately obvious what this function actually does.

zewa666:
as for 3. I'd definitely go with what saves us maintenance efforts.

packages/__tests__/e2e/package.json
```diff
@@ -25,7 +25,9 @@
     "@aurelia/kernel": "^0.3.0",
     "@aurelia/runtime": "^0.3.0",
     "@aurelia/runtime-html": "^0.3.0",
-    "@aurelia/runtime-html-browser": "^0.3.0"
+    "@aurelia/runtime-html-browser": "^0.3.0",
+    "i18next-intervalplural-postprocessor": "^1.0.0",
```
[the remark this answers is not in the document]
Sayan751: I added the i18next plugin to test if the plugin integration is working correctly. Moreover, I think that translation resources based on interval is not a very rare scenario.

packages/__tests__/e2e/src/plugins/sut-i18n.ts
```diff
@@ -0,0 +1,55 @@
...
+
+@customElement({ name: 'sut-i18n', template })
+export class SutI18N {
+  public obj = { key: 'simple.text', foo: 'bar' };
+  public dispatchedOn = new Date(2020, 1, 10, 5, 15);
+  public deliveredOn = new Date(2021, 1, 10, 5, 15);
+  public params = { context: 'delivered', date: this.deliveredOn };
+  public translations: { [key: string]: string | number };
+  private readonly myDate: Date;
+  private locale: string;
```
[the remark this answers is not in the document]
Sayan751: That's  nice use of TS. In practice though, for e2e tests that might be bit overkill.

packages/__tests__/e2e/tests/integration/i18n.e2e.ts
Sayan751: I think that would really be overkill. Plus it might make the debugging difficult as then you might have to look up the ids in a different file/place. Above all, it will be troubling because then the app needs to be changed in order to accommodate tests.

packages/__tests__/e2e/tests/integration/i18n.e2e.ts
Sayan751: my bad, copy-paste error. will be fixed

packages/__tests__/i18n/df/date-format-value-converter.spec.ts
Sayan751: waited for the i18next initialization.

packages/__tests__/e2e/src/plugins/sut-i18n.ts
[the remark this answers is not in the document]
zewa666: Perhaps but keep in mind that our tests will also act as a kind of docs so we can show best practices here as well

packages/__tests__/i18n/df/date-format-value-converter.spec.ts
zewa666: So is there a reason for the string index access of `task`? Is this a Symbol?

packages/__tests__/i18n/i18n.spec.ts
Sayan751: new goodies from Intl API 😃

packages/__tests__/i18n/t/translation-integration.spec.ts
Sayan751: the vNext thing would be to use ContinuationTask, but actually could not make this work with either Promise or Task. Therefore, skipped it for now.

@fkleuver is there anyway this can be done?

packages/__tests__/i18n/t/translation-parameters-render.spec.ts
Sayan751: Yup. if i am not mistaken then this is being preferred over `undefined` in vNext.

packages/__tests__/e2e/src/plugins/sut-i18n.ts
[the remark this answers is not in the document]
Sayan751: Agreed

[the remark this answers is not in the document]
Sayan751: Can make the change

packages/__tests__/i18n/df/date-format-value-converter.spec.ts
Sayan751: Nah, just a `private` :)

packages/__tests__/i18n/t/translation-parameters-render.spec.ts
zewa666: @fkleuver any idea why that is? I'm fine with doing so would just like to understand the construct

packages/__tests__/i18n/df/date-format-value-converter.spec.ts
zewa666: Well if we do need that in our tests chances are the user might do as well, so should we create a public method for this?

Sayan751: Agreed. I will change that `task` to public and will make few other changes in the binding. That way the task don't need to be waited for explicitly by devs, the binding should take care of it. If need be, devs can wait for the task explicitly.

packages/__tests__/i18n/t/translation-parameters-render.spec.ts
Sayan751: AFAIK, `undefined` is a global variable, which is resolved by well looking up in scope. That is not the case for `void 0`. So, it should be bit faster than the other, at least in theory.

On related note, historically `undefined` was also assignable (!), which is not the case anymore.

packages/__tests__/e2e/src/startup.ts
```diff
@@ -1,21 +1,41 @@
...
 import { Aurelia } from '@aurelia/runtime';
+import * as intervalPlural from 'i18next-intervalplural-postprocessor';
 import { App } from './app';
 import * as de from './locales/de/translations.json';
 import * as en from './locales/en/translations.json';
+import { CustomMessage } from './plugins/custom-message';
+import { SutI18N } from './plugins/sut-i18n';
+
+// Intl.RelativeTimeFormat polyfill is needed as Cypress uses electron and does not seems to work with puppeteer
+import RelativeTimeFormat from 'relative-time-format';
```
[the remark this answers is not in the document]
Sayan751: I would like to open another PR for the whole Docs thing. I will add it to docs then.

packages/__tests__/e2e/src/startup.ts
Sayan751: Refactored

packages/__tests__/e2e/tests/integration/i18n.e2e.ts
Sayan751: Done

packages/__tests__/i18n/df/date-format-value-converter.spec.ts
Sayan751: Done. Take a look at the changed `configuration.ts` and `I18nService`. On related note, as the formatting services are based on `Intl` API, the wait for that `task` was superfluous; removed it.

packages/__tests__/i18n/i18n.spec.ts
Sayan751: Done

packages/__tests__/i18n/i18n.spec.ts
Sayan751: done

packages/__tests__/i18n/t/translation-parameters-render.spec.ts
Sayan751: done

AureliaEffect:
JS Framework Benchmark: https://24776-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/__tests__/i18n/t/translation-parameters-render.spec.ts
fkleuver: Historically, `void 0` has been "safer" than `undefined` but this is not really the case anymore. There are situations where `void 0` performs a bit better on initial evaluation, but that's in the same order of scale as `++i` vs `i++`; hardly measurable. If anything, it's shorter, and it's better to stay consistent.

packages/__tests__/e2e/src/startup.ts
```diff
@@ -1,21 +1,44 @@
...
+            de: { translation: de['default'] },
+          },
+          skipTranslationOnMissingKey: !!new URL(location.href).searchParams.get('skipkey')
+        };
+      })
+    );
+  au.container.register(SutI18N, CustomMessage);
+  au.app({ host, component });
+
+  await au.start().wait();
```
zewa666: That looks interesting, where is this wait coming from?

packages/__tests__/i18n/t/translation-parameters-render.spec.ts
zewa666: I personally feel like that `(void 0)!` is ugly as hell to look at and I'd sacrifice that tiny tiny portion of optimization every day for the readability of undefined, especially when that is in unit tests where it really doesn't matter. The cognitive load added by above statement is wasting much more time than needed, especially for a case where `flags` could be left out at all and it just indicates the reader that it's not used.

Anyways, if you guys prefer to use it though, I'd go with @fkleuver's statement to stay consistent

packages/__tests__/i18n/t/translation-parameters-render.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+    const expressionParser = container.get(IExpressionParser);
+    const renderable = ({} as unknown as IController);
+
+    sut.render(
+      LifecycleFlags.none,
+      DOM,
+      container as unknown as IRenderContext,
+      renderable,
+      DOM.createElement('span'),
+      { from: expressionParser.parse('{foo: "bar"}', BindingType.BindCommand) } as unknown as ITargetedInstruction);
```
zewa666: OCD: my heart burns looking at this. Is this how our linter is setup by default to not break the closing `)` onto a new line?

packages/__tests__/i18n/t/translation-parameters-render.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+  it('#render instantiates TranslationBinding if there are none existing', function () {
+    const container = setup();
+    const sut: IInstructionRenderer = new TranslationParametersBindingRenderer(container.get(IExpressionParser), container.get(IObserverLocator));
+    const expressionParser = container.get(IExpressionParser);
+    const renderable = ({} as unknown as IController);
+
+    sut.render(
+      LifecycleFlags.none,
+      DOM,
+      container as unknown as IRenderContext,
```
zewa666: how about casting the return value from `setup` instead? Those `as unknown as xxx` are always feeling like a hack

packages/__tests__/i18n/t/translation-render.spec.ts
```diff
@@ -0,0 +1,207 @@
...
+      TranslationAttributePattern.prototype.$patternDefs,
+      aliases.reduce(
+        (acc, alias) => {
+          acc.push({ pattern: alias, symbols: '' });
+          return acc;
+        },
+        []));
+
+    aliases.forEach((alias) => {
+      assert.equal(typeof sut[alias], 'function');
```
zewa666: assertion could be moved inline of the reducer

packages/i18n/src/df/date-format-value-converter.ts
```diff
@@ -0,0 +1,30 @@
...
+
+    if ((!value && value !== 0) || (typeof value === 'string' && value.trim() === '')) {
+      return value;
+    }
+
+    // covert '0' to 01/01/1970 or ISO string to Date and return the original value if invalid date is constructed
+    if (typeof value === 'string') {
+      const numValue = Number(value);
+      const tempDate = new Date(Number.isInteger(numValue) ? numValue : value);
+      if (isNaN(tempDate.getTime())) {
```
zewa666: nice trick 👍

packages/i18n/src/i18n.ts
```diff
@@ -12,76 +20,177 @@ export const I18N = DI.createInterface<I18nService>('I18nService').noDefault();
...
+    // return real number
+    return Number(result);
+  }
+
+  /**
+   * Returns `Intl.RelativeTimeFormat` instance with given `[options]`, and `[locales]` which can be used to format a value with associated time unit.
+   * If the `locales` is skipped, then the `Intl.RelativeTimeFormat` instance is created using the currently active locale.
+   */
+  public createRelativeTimeFormat(options?: Intl.RelativeTimeFormatOptions, locales?: string | string[]): Intl.RelativeTimeFormat {
+    return new this.intl.RelativeTimeFormat(locales || this.getLocale(), options);
```
zewa666: should we check before using that and throw a proper error with a hint to use a polyfill?

packages/i18n/src/t/translation-binding.ts
```diff
@@ -0,0 +1,232 @@
...
+interface TranslationBindingCreationContext {
+  parser: IExpressionParser;
+  observerLocator: IObserverLocator;
+  context: IRenderContext;
+  renderable: IController;
+  target: HTMLElement;
+  instruction: ICallBindingInstruction;
+  isParameterContext?: boolean;
+}
+type ContentAttribute = 'textContent' | 'innerHTML' | 'prepend' | 'append';
```
zewa666: Remove unnecessary duplication of array values:

```typescript
const contentAttributes = ['textContent', 'innerHTML', 'prepend', 'append'] as const;
type ContentAttribute = typeof contentAttributes[number];


...

export class ... {
  private readonly contentAttributes = contentAttributes;
}
```

packages/i18n/src/t/translation-binding.ts
```diff
@@ -0,0 +1,232 @@
+import { IEventAggregator, IServiceLocator } from '@aurelia/kernel';
+import { connectable, CustomElement, CustomExpression, DOM, IBindingTargetAccessor, IConnectableBinding, Interpolation, IObserverLocator, IPartialConnectableBinding, IScope, IsExpression, LifecycleFlags, State, IExpressionParser, IRenderContext, IController, ICallBindingInstruction, ensureExpression, BindingType, addBinding } from '@aurelia/runtime';
```
zewa666: can we break this import onto multiple lines. Hard to scroll all the way to the right

zewa666: Would be cool to have a linter rule enforcing max line-width btw ;)

packages/i18n/src/t/translation-binding.ts
```diff
@@ -0,0 +1,232 @@
...
+      this.updateContent(content, flags);
+    }
+  }
+
+  private updateAttribute(attribute: string, value: string, flags: LifecycleFlags) {
+    const controller = CustomElement.behaviorFor(this.target);
+    const observer = controller && controller.viewModel
+      ? this.observerLocator.getAccessor(LifecycleFlags.none, controller.viewModel, attribute)
+      : this.observerLocator.getAccessor(LifecycleFlags.none, this.target, attribute);
+    observer.setValue(value, flags);
```
zewa666: didn't we have issues with this one on setting bindables of a custom attribute? I can't recall anymore if this got clarified. An e2e test for this could be useful.

packages/i18n/src/t/translation-binding.ts
```diff
@@ -0,0 +1,232 @@
...
+  private prepareTemplate(content: ContentValue, marker: string, fallBackContents: ChildNode[]) {
+    const template = DOM.createTemplate() as HTMLTemplateElement;
+
+    this.addTextContentToTemplate(template, content.prepend, marker);
+
+    // build content: prioritize [html], then textContent, and falls back to original content
+    if (content.innerHTML) {
+      const fragment = DOM.createDocumentFragment(content.innerHTML) as DocumentFragment;
+      for (const child of Array.from(fragment.childNodes)) {
+        Reflect.set(child, marker, true);
```
zewa666: great idea to use a marker with Reflection vs the original `_xxx` spamming

packages/i18n/src/t/translation-binding.ts
```diff
@@ -0,0 +1,232 @@
...
+      }
+    }
+
+    this.addTextContentToTemplate(template, content.append, marker);
+    return template;
+  }
+
+  private addTextContentToTemplate(template: HTMLTemplateElement, additionalText: string | undefined, marker: string) {
+    if (additionalText) {
+      const addendum: Node = DOM.createTextNode(additionalText) as Node;
```
zewa666: `const addendum = `

packages/i18n/src/t/translation-parameters-renderer.ts
```diff
@@ -0,0 +1,43 @@
+import { attributePattern, AttrSyntax, bindingCommand, BindingSymbol, getTarget, IBindingCommand, PlainAttributeSymbol } from '@aurelia/jit';
+import { BindingMode, BindingType, ICallBindingInstruction, IController, IDOM, IExpressionParser, IInstructionRenderer, instructionRenderer, IObserverLocator, IRenderContext, IsBindingBehavior, LifecycleFlags } from '@aurelia/runtime';
```
zewa666: same as above

packages/i18n/src/t/translation-parameters-renderer.ts
```diff
@@ -0,0 +1,43 @@
...
+import { TranslationBinding } from './translation-binding';
+
+export const TranslationParametersInstructionType = 'tpt';
+// `.bind` part is needed here only for vCurrent compliance
+const attribute = 't-params.bind';
+
+@attributePattern({ pattern: attribute, symbols: '' })
```
zewa666: this is soooo freaky cool

packages/i18n/src/i18n-configuration-options.ts
Sayan751: actually it is not renamed strictly per se. an interface with the original name is still there.

packages/__tests__/e2e/src/startup.ts
Sayan751: `start` returns a [`ILifeCycleTask` which can be awaited](https://github.com/aurelia/aurelia/blob/master/packages/runtime/src/lifecycle-task.ts#L228) :)

packages/i18n/src/i18n.ts
Sayan751: Then we also need to come back here to clean up later. I think it is better to prepare good documentation instead.

packages/i18n/src/t/translation-binding.ts
Sayan751: Do you mean `@bindable`s in custom _element_? I already have e2e tests for that.

In case you mean custom _attribute_ instead, never saw an use cae in conjunction with i18n.

packages/__tests__/i18n/t/translation-render.spec.ts
Sayan751: fixed

packages/__tests__/i18n/t/translation-render.spec.ts
Sayan751: fixed

packages/i18n/src/Intl.d.ts
Sayan751: fixed

packages/i18n/src/configuration.ts
Sayan751: fixed

packages/i18n/src/i18n.ts
Sayan751: fixed

packages/__tests__/i18n/t/translation-parameters-render.spec.ts
Sayan751: fixed

packages/__tests__/i18n/t/translation-parameters-render.spec.ts
Sayan751: fixed

packages/__tests__/i18n/t/translation-render.spec.ts
Sayan751: refactored

packages/i18n/src/t/translation-binding.ts
Sayan751: fixed

packages/i18n/src/t/translation-binding.ts
Sayan751: fixed

packages/i18n/src/t/translation-binding.ts
Sayan751: fixed

packages/i18n/src/t/translation-parameters-renderer.ts
Sayan751: fixed

Sayan751:
@zewa666 not sure which one you referred

> as for 3. I'd definitely go with what saves us maintenance efforts.

AureliaEffect:
JS Framework Benchmark: https://24996-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/__tests__/e2e/src/plugins/sut-i18n.ts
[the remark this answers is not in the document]
Sayan751: made the changes, though it is looking a bit stretched https://github.com/aurelia/aurelia/pull/563/commits/eabf0e31a3448d9d8ebd2ced8c9f4c8f70499390

AureliaEffect:
JS Framework Benchmark: https://25031-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

zewa666:
@Sayan751 3 was relating to your reviewer notes, point 3 (As the new relative time formatting relies on Intl API, there is a slight change...)

packages/i18n/src/t/translation-binding.ts
zewa666: yeah I was referring to the attribute (aka bindable) of a custom element. Sorry for the confusion

packages/i18n/src/i18n-configuration-options.ts
zewa666: oki doki

zewa666:
I'm super happy with the outcome of this one ... was quite a block but it looks definitely good. It's also such a great and reassuring feeling to see the e2e still covered form vCurrent. If @fkleuver you could just reply to the last open issue with regards to ContinuationTask I think we're fine to get this ready for merge

packages/__tests__/i18n/t/translation-render.spec.ts
fkleuver: You could add a new assertion `assert.typeOf(sut[alias], 'function')` to `@aurelia/testing` if you want, btw. You can also do this for any other asserts that might make sense. That's one of the points of having our own assertion lib :)

packages/i18n/src/t/translation-binding.ts
Sayan751: all good :)

packages/__tests__/i18n/t/translation-integration.spec.ts
Sayan751: Done

packages/__tests__/i18n/t/translation-render.spec.ts
Sayan751: done :)

AureliaEffect:
JS Framework Benchmark: https://25149-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/i18n/src/i18n.ts
```diff
@@ -1,9 +1,17 @@
...
+import { ILifecycleTask, ISignaler, PromiseTask } from '@aurelia/runtime';
 import i18nextCore from 'i18next';
-import { I18nConfigurationOptions } from './i18n-configuration-options';
+import { I18nInitOptions } from './i18n-configuration-options';
 import { I18nextWrapper, I18nWrapper } from './i18next-wrapper';
 
+export const I18N_EA_CHANNEL = 'i18n:locale:changed';
+export const I18N_SIGNAL = 'aurelia-translation-signal';
+
+export interface I18nKeyEvaluationResult {
```
fkleuver: I'd prefer this to be a class, so [here](https://github.com/aurelia/aurelia/pull/563/files#diff-75c517b4cc4e13c096a8449d4cf85d60R64) you can do `.push(new I18nKeyEvaluationResult(attributes, translation))`, etc. May want to consider making the properties readonly as well.

Could even take it a small step further and encapsulate the logic from `extractAttributesFromKey` into the class constructor?

packages/i18n/src/i18n.ts
```diff
@@ -1,9 +1,17 @@
...
-import { ContinuationTask, IDOM, ILifecycleTask, PromiseTask } from '@aurelia/runtime';
+import { DI, IEventAggregator, PLATFORM } from '@aurelia/kernel';
+import { ILifecycleTask, ISignaler, PromiseTask } from '@aurelia/runtime';
 import i18nextCore from 'i18next';
-import { I18nConfigurationOptions } from './i18n-configuration-options';
+import { I18nInitOptions } from './i18n-configuration-options';
 import { I18nextWrapper, I18nWrapper } from './i18next-wrapper';
 
+export const I18N_EA_CHANNEL = 'i18n:locale:changed';
+export const I18N_SIGNAL = 'aurelia-translation-signal';
```
fkleuver: What do you think about making something like this instead:

```ts
export const enum Constants {
  EA_Channel = 'i18n:locale:changed',
  Signal = 'aurelia-translation-signal',
}
```

packages/i18n/src/i18n.ts
```diff
@@ -12,76 +20,193 @@ export const I18N = DI.createInterface<I18nService>('I18nService').noDefault();
...
-      return;
+  /**
+   * Returns a relative time format of the given `input` date as per the given `[options]`, and `[locales]`.
+   * If the `locales` is skipped, then the currently active locale is used for formatting.
+   */
+  public rt(input: Date, options?: Intl.RelativeTimeFormatOptions, locales?: string | string[]): string {
+    let difference = input.getTime() - new Date().getTime();
+    const formatter = this.createRelativeTimeFormat(options, locales);
+
+    let value: number = difference / 31104000000;
```
fkleuver: Can we name these please, e.g.:

```ts
const enum TimeSpan {
  Second = 1000,
  Minute = Second * 60,
  Hour = Minute * 60,
  Day = Hour * 24,
  Week = Day * 7,
  Year = Day * 360, // 31104000000 - quirk?
}
```

packages/i18n/src/nf/number-format-value-converter.ts
```diff
@@ -0,0 +1,19 @@
+import { valueConverter } from '@aurelia/runtime';
+import { I18N, I18N_SIGNAL, I18nService } from '../i18n';
+
+export const numberFormatValueConverterName = 'nf';
```
fkleuver: Feels a little bit like overkill. But if you feel strongly about having single point of modification, maybe move to `Constants` enum proposed earlier?

packages/i18n/src/t/translation-binding.ts
```diff
@@ -0,0 +1,254 @@
...
+    }
+
+    const template = this.prepareTemplate(content, marker, fallBackContents);
+
+    // difficult to use the set property approach in this case, as most of the properties of Node is readonly
+    // const observer = this.observerLocator.getAccessor(LifecycleFlags.none, this.target, '??');
+    // observer.setValue(??, flags);
+
+    this.target.innerHTML = '';
+    for (const child of Array.from(template.content.childNodes)) {
```
fkleuver:
```suggestion
    for (const child of PLATFORM.toArray(template.content.childNodes)) {
```

Is about 10x faster than `Array.from` :)

packages/i18n/src/i18n.ts
Sayan751: That's nice. I would go for `enum Signals` though.

packages/i18n/src/i18n.ts
Sayan751: Done

packages/i18n/src/i18n.ts
Sayan751: Done

packages/i18n/src/nf/number-format-value-converter.ts
Sayan751: Moved to a different enum in `utils`

packages/i18n/src/t/translation-binding.ts
Sayan751: Done

packages/i18n/src/i18n.ts
Sayan751: Done

AureliaEffect:
JS Framework Benchmark: https://25179-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

PR #569, opened 2019-08-08
feat(attr-childlist): add childlist attribute

bigopon:
# Pull Request

## 📖 Description

Ability to bind a childlist of any arbitrary element with default binding mode is from view. Example usage:

```html
<div childlist.bind="children_of_the_div">
  <div></div>
  <button></button>
  <input/>
  <select></select>
</div>
```

### 🎫 Issues

N/A

## 👩‍💻 Reviewer Notes

Should see how feasible this is, in conjunction with `@children` decorator, potentially some consolidation if this is usable?
Timing is another issue, `attaching` is enough in normal case, but if this supports surrogate element, it needs to be done initially during `attached`

## 📑 Test Plan

N/A

## ⏭ Next Steps

N/A

### DIFF
--- a/packages/__tests__/jit-html/childlist.integration.spec.ts
+++ b/packages/__tests__/jit-html/childlist.integration.spec.ts
@@ -1,0 +1,105 @@
+import { Constructable } from '@aurelia/kernel';
+import { Aurelia, CustomElement } from '@aurelia/runtime';
+import { ChildList } from '@aurelia/runtime-html';
+import { assert, TestContext } from '@aurelia/testing';
+
+describe('childist.integration.spec.ts', function() {
+
+  it('gets the right childist', function() {
+    const { dispose, ctx, component } = setup(
+      `<div childlist.bind="childrenOfTheWhale">
+        <div></div>
+        <button></button>
+        <input/>
+        <select></select>
+      </div>`,
+      class App {
+        public childrenOfTheWhale: HTMLElement[];
+      }
+    );
+
+    assert.instanceOf(component.childrenOfTheWhale, Array, 'should have been array');
+    assert.equal(component.childrenOfTheWhale.length, 4, 'should have had 4 elements');
+
+    const gbl = ctx.wnd as unknown as typeof globalThis;
+    assertElements(
+      component.childrenOfTheWhale,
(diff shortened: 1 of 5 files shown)


EisenbergEffect:
@bigopon I wonder if a better approach would be to enhance the `children` decorator so that it can be used on any custom attribute, not just custom elements. Then, it would be really easy for anyone to take advantage of it as a more fundamental building block. An generic attribute like this could then be built with that if desired. Thoughts?

bigopon:
@EisenbergEffect why do you want to have a `@children` on an arbitrary attribute? Also, if we are to consolidate `ChildrenObserver` so that this can be built on top of that, probably will rework it a lot, since it will align with vCurrent more that way. Though we need @fkleuver to comment on this, since `ChildrenObserver` seems to be using a lot of internal APIs for now, and I'm not sure future plan for it (read integrated platform, where there's no mutation observer)

EisenbergEffect:
Someone may want to create a custom attribute that works with the children of the element that it's placed on. I believe I saw that request with vCurrent quite some time ago. As such, anything like you have here could be built on top of the foundation and I'd prefer to provide the building blocks and make custom elements and attributes share capabilities, rather than introduce special case attributes.

To make the children decorator work with custom attributes, I believe that all we need to do is provide a projector for custom attribute controllers. With that in place, it should "just work". The projector can be really simple, as it only needs to capture the element the attribute is being attached to (host), provide a way to get its children (very simple) and provide a way to subscribe to child changes (also simple). It will be very close to the existing shadow dom projector, but without implementations for provideEncapsulationSource, project, and take. What we could do is create a new interface that has only host, children, and subscribeToChildrenChange. The projector interface would derive from that to add on more capabilities, but the children observer would rely only on that, since that's all it needs. All in all, I think a very small set of additions would make this work, without a need to change hardly any existing code.

bigopon:
It sounds like you want to enhance custom attribute. I will have a look at that. But this PR serve completely different purpose that children decorator cannot serve: arbitrary mutation. Though I think the argument on the other PR still holds: we dont want to make the core larger and this can be a separated repo/doc

bigopon:
@EisenbergEffect I think existing infra would be insufficient for this attribute, as all of the query would be configurable, and has dynamic/bindable values. Maybe another example in the list instead?

bigopon:
should be done in #603

PR #570, opened 2019-08-09
feat(attr-rectsize): add rectsize attribute

bigopon:
# Pull Request

## 📖 Description

Add new attribute `rectsize` to provide automatic element sizing binding. Example:
```html
<form style="width: 100px">
  <div id=formField1 rectsize.bind="firstnameFieldSize" style="height: 60px">
    <label>First name</label>
    <input name=firstname />
  </div>
</form>
```
```ts
/**
 * Editor’s Draft, 7 June 2019 at https://drafts.csswg.org/resize-observer-1/#resizeobserversize
 */
firstnameFieldSize === { inlineSize: 100, blockSize: 60 }
```

### 🎫 Issues

N/A

## 👩‍💻 Reviewer Notes

N/A

## 📑 Test Plan

Test many combinations with template controllers
Test with element in Shadow DOM (probably hard without native support)
Test with containerless element

## ⏭ Next Steps

N/A

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -348,6 +348,12 @@
         }
       }
     },
+    "@juggle/resize-observer": {
+      "version": "2.3.0",
+      "resolved": "https://registry.npmjs.org/@juggle/resize-observer/-/resize-observer-2.3.0.tgz",
+      "integrity": "sha512-vCS2dz1HAG760+vADa7Fol/NhrmdyDa58Nvq39TpdwYjCqTEDFmIFCDGRiKGSw4GPoDRcx/+xwOYlJdMeNUKRw==",
+      "dev": true
+    },
     "@lerna/add": {
       "version": "3.16.2",
       "resolved": "https://registry.npmjs.org/@lerna/add/-/add-3.16.2.tgz",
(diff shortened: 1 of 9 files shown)


EisenbergEffect:
I'm not sure what this does. Does it observe the element's size and then update the properties on the object when the size changes? I can see size observation as something that's interesting, but I'm not sure we need a special attribute to bind sizes to properties right now. We're adding more and more special attributes to the core, and that concerns me because it makes the framework larger, incurs more maintenance cost, but most of all makes it harder to learn and use. I'd rather not bother with this particular item right now when we've still got tons of core work that we need to focus on. This seems like it could be much better as a recipe in our docs, rather than a builtin attribute.

bigopon:
Feel the feeling again, where we have to reject stuff to the core 😆
Ill keep it out of core

EisenbergEffect:
I think you should add the RectSize attribute to the docs. There's a section for examples, and this would be a fun one to add. Just take your implementation (no tests or anything else), and put it in a  markdown files under docs/user-docs/examples, and give it a nice name, maybe something like 'Bind to Element Size Changes" or something like that.

Issue #524, opened 2019-07-17
[RFC] Handling CSS


EisenbergEffect:
# 💬 RFC

Component and application developers need a way to provide style for what they build. While this can be handled independently of Aurelia, by simply including a stylesheet in the `index.html` page, this doesn’t always play well with componentized architectures and certainly doesn’t work well with Shadow DOM. To address these shortcomings, Aurelia’s component system should provide a mechanism for providing global/shared styles as well as per-component styles.

## Common Setups

For the purpose of this discussion, I want to present three typical ways that Aurelia would be used.

### No Shadow DOM

In this architecture, real Shadow DOM is not used in the application. Slotted composition, if required, is handled by the end user adding the [ShadyDOM Polyfill](https://github.com/webcomponents/polyfills/tree/master/packages/shadydom) with `ShadyDOM = {force: true}` to ensure consistent behavior regardless of browser.

This configuration is optimal for apps that leverage Bootstrap or other component CSS libraries which rely on a large set of global styles. However, developers may still want to build their app with per-component styles, following a naming convention. e.g. `my-component.js`, `my-component.html`, `my-component.css`. Typically, in this setup, CSS would still be handled by a CSS build step that concatenates all these styles, which the developer would then add to the head of the document. However, in a componentized world, there is often a desire to still keep individual app component styles isolated. For this, CSS Modules is the recommended solution.

#### Requirements

* Aurelia's runtime should detect slots and use shadow dom automatically, only for those scenarios. This should work with the Shady DOM polyfill.
* There should be a reasonable way to configure Webpack/Dumber/etc. for CSS Modules without breaking our AoT and conventions. 
  * The AoT/conventions system needs to know to leave the CSS files alone in this setup, as they should be handled by the CSS Modules plugin. 
  * The CSS Modules plugin needs to run earlier in the process, since it will alter the HTML to use the new class names generated through CSS Modules. 
  * CSS imports should be removed from source, if present, since the CSS Modules build will handle creating the final CSS output.
* Document the setup and tradeoffs of this configuration.

### Only Native Shadow DOM

[...]

PR #573, opened 2019-08-12
Runtime Support for Styles

EisenbergEffect:
# Pull Request

## 📖 Description

This PR implements the necessary infrastructure to support both CSS Modules Preprocessing and Shadow DOM CSS styles in the HTML runtime. The Shadow DOM styles implementation supports using adoptedStyleSheets if present, and falling back to style elements if not. When using adoptedStyleSheets, it can handle both strings and CSSStyleSheet instances, providing support for the upcoming CSS Modules spec as well.

### 🎫 Issues

* #524 

## 👩‍💻 Reviewer Notes

The basic idea is that "styles" are a dependency of the template. So, the infrastructure is based around plugging in different ways to handle CSS, based on application preference. Depending on configuration, when styles are added to a component dependency list, a different registry will be used to register the correct set of services within the component's render context. For CSS Modules, a custom class attribute is registered. For Shadow DOM CSS, a style manager is configured, which is later used by the ShadowDOMProjector. Two different style managers for Shadow DOM have been implemented, with the best implementation being chosen based on the runtime capabilities of the browser.

## ⏭ Next Steps

* Get conventions and template build-time support to use this.

### DIFF
--- a/packages/__tests__/runtime-html/styles.spec.ts
+++ b/packages/__tests__/runtime-html/styles.spec.ts
@@ -1,5 +1,5 @@
-import { DI, PLATFORM, Registration } from '@aurelia/kernel';
-import { Controller, CustomAttribute, CustomElement, INode, Aurelia } from '@aurelia/runtime';
+import { DI, Registration } from '@aurelia/kernel';
+import { Aurelia, Controller, CustomAttribute, CustomElement, INode } from '@aurelia/runtime';
 import {
   AdoptedStyleSheetsStyles,
   CSSModulesProcessorRegistry,
--- a/packages/__tests__/runtime-html/styles.spec.ts
+++ b/packages/__tests__/runtime-html/styles.spec.ts
@@ -1,5 +1,5 @@
 import { DI, PLATFORM, Registration } from '@aurelia/kernel';
-import { Controller, CustomAttribute, CustomElement, INode } from '@aurelia/runtime';
+import { Controller, CustomAttribute, CustomElement, INode, Aurelia } from '@aurelia/runtime';
 import {
   AdoptedStyleSheetsStyles,
   CSSModulesProcessorRegistry,
@@ -12,11 +12,23 @@ import {
 import { assert, TestContext } from '@aurelia/testing';
 
 describe('Styles', () => {
-  describe('CSS Modules Processor', () => {
-    it('config adds correct registry for css', () => {
-      const container = DI.createContainer();
+  async function startApp(configure: (au: Aurelia) => void) {
+    const ctx = TestContext.createHTMLTestContext();
+    const au = new Aurelia(ctx.container);
+    configure(au);
+    const host = ctx.createElement('div');
+    class App { }
+    const component = CustomElement.define({ name: 'app', template: ' ' }, App);
+    au.app({ host, component });
+    await au.start().wait();
+    return { au, ctx, host, container: au.container };
+  }
 
-      container.register(StyleConfiguration.cssModulesProcessor());
+  describe('CSS Modules Processor', () => {
+    it('config adds correct registry for css', async () => {
+      const { container } = await startApp(au => {
+        au.register(StyleConfiguration.cssModulesProcessor());
+      });
 
       const registry = container.get('.css');
       assert.instanceOf(registry, CSSModulesProcessorRegistry);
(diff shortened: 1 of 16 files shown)


AureliaEffect:
JS Framework Benchmark: https://23946-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://23965-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://23995-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://24012-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://24113-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://24128-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://24132-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://24147-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://24153-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://24241-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://24288-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://24297-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://24409-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/__tests__/runtime-html/styles.spec.ts
```diff
@@ -0,0 +1,262 @@
...
+      const container = DI.createContainer();
+      container.register(Registration.instance(INode, element));
+      const registry = new CSSModulesProcessorRegistry();
+      const cssModulesLookup = {};
+
+      registry.register(container, cssModulesLookup);
+
+      const attr = container.get(CustomAttribute.keyFrom('class'));
+
+      assert.equal(CustomAttribute.isType(attr.constructor), true);
```
fkleuver: We could add an assertion for this? `assert.isCustomAttributeType`, `assert.isCustomAttributeInstance`, etc?

They would go here: <https://github.com/aurelia/aurelia/blob/master/packages/testing/src/assert.ts#L413>

packages/runtime-html/src/styles/shadow-dom-registry.ts
```diff
@@ -0,0 +1,21 @@
...
+export type ShadowDOMStylesFactory =
+  (localStyles: (string | CSSStyleSheet)[], sharedStyles: IShadowDOMStyles | null) => IShadowDOMStyles;
+
+export class ShadowDOMRegistry {
+  constructor(
+    private sharedStyles: IShadowDOMStyles,
+    private createStyles: ShadowDOMStylesFactory
+  ) { }
+
+  public register(container: IContainer, ...params: any[]) {
```
fkleuver: We should use `unknown` for user-provided input if there are no constraints. But are there really no constraints though?

packages/__tests__/runtime-html/styles.spec.ts
```diff
@@ -0,0 +1,262 @@
...
+
+      assert.equal(root.adoptedStyleSheets.length, 1);
+      assert.strictEqual(root.adoptedStyleSheets[0], sheet);
+    });
+
+    it('adopted styles merge sheets from parent', () => {
+      const ctx = TestContext.createHTMLTestContext();
+      const sharedCSS = '.my-class { color: red }';
+      const localCSS = '.something-else { color: blue }';
+      const root = { adoptedStyleSheets: [] };
```
fkleuver: Nice, I like the minimalism. Is there any interface that describes this contract as well?
And this is also working in node now?

packages/kernel/src/di.ts
```diff
@@ -782,6 +782,14 @@ export const Registration = Object.freeze({
   alias<T>(originalKey: T, aliasKey: Key): IRegistration<Resolved<T>> {
     return new Resolver(aliasKey, ResolverStrategy.alias, originalKey);
   },
+  defer(key: Key, ...params: any[]): IRegistry {
+    return {
```
fkleuver: My initial thought here is it should be a class, but perhaps it's overkill. What do you think?
```ts
export class ParameterizedRegistry implements IRegistry {
  constructor(
    private readonly key: Key,
    private readonly params: unknown[]
  ) {}

  public register(container: IContainer): void {
    const registry = container.get<IRegistry>(this.key);
    registry.register(container, ...this.params);
  }
}

// ...

defer(key: Key, ...params: unknown[]): IRegistry {
  return new ParameterizedRegistry(key, params);
}
```

packages/__tests__/runtime-html/styles.spec.ts
EisenbergEffect: The `adoptedStyleSheets` property is present on instances of `ShadowRoot`. Unfortunately that type is missing from JSDOM. It's present on the TS type though.

packages/kernel/src/di.ts
EisenbergEffect: Hmm. I don't know, but I'm happy to make this change.

AureliaEffect:
JS Framework Benchmark: https://24563-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/__tests__/runtime-html/styles.spec.ts
fkleuver: `ShadowRoot` does exist in jsdom: <https://github.com/jsdom/jsdom/blob/0a39afae36ced79d2cb6636cf5fdf649520f2657/lib/jsdom/living/index.js#L77>
Maybe the types are just outdated. Try this in the initializer and see if it works?
```ts
(this.jsdom.window as unknown as { ShadowRoot: typeof ShadowRoot }).ShadowRoot
```

packages/kernel/src/di.ts
fkleuver: What makes you unsure?

EisenbergEffect: At first, I wasn't sure if it was worth it, but I went ahead and made the change. I do think that this will probably get called quite a bit in real apps, so there's probably an advantage to having it in this form.

packages/__tests__/runtime-html/styles.spec.ts
EisenbergEffect: Interesting. I'll change things over to this and see how it works. If just about anything is there with that name, it should be fine.

AureliaEffect:
JS Framework Benchmark: https://24616-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

PR #583, opened 2019-08-17
refactor(router): refactor interfaces and types

jwx:
# Pull Request

## 📖 Description

This PR refactors and restructures interfaces and types, including renaming `IRouteableCustomElement*` to `IRouteableComponent*` and using them where appropriate. It also fixes a small bug in `au-nav`s active route.

### 🎫 Issues

## 👩‍💻 Reviewer Notes

## 📑 Test Plan

## ⏭ Next Steps

- Implement scope traversal to Router's `goto` method
- Finish navigation skeleton
- Review the extension points

### DIFF
--- a/packages/router/src/guard.ts
+++ b/packages/router/src/guard.ts
@@ -1,12 +1,13 @@
+import { Constructable } from '@aurelia/kernel';
 import { GuardIdentity, GuardTypes, IGuardOptions, } from './guardian';
-import { GuardFunction, GuardTarget, IGuardTarget, INavigatorInstruction, IRouteableComponentType } from './interfaces';
+import { GuardFunction, GuardTarget, IGuardTarget, INavigatorInstruction, IRouteableComponentType, ComponentAppellationResolver, ViewportAppellationResolver, IComponentAndOrViewportOrNothing } from './interfaces';
 import { Viewport } from './viewport';
 import { ViewportInstruction } from './viewport-instruction';
 
 export class Guard {
   public type: GuardTypes;
-  public includeTargets: Target[];
-  public excludeTargets: Target[];
+  public includeTargets: Target<Constructable>[];
+  public excludeTargets: Target<Constructable>[];
   public guard: GuardFunction;
   public id: GuardIdentity;
 
@@ -40,23 +41,24 @@ export class Guard {
   }
 }
 
-class Target {
-  public component?: IRouteableComponentType;
+class Target<C extends Constructable> {
+  public component?: IRouteableComponentType<C>;
   public componentName?: string;
   public viewport?: Viewport;
   public viewportName?: string;
 
   constructor(target: GuardTarget) {
-    const { component, componentName, viewport, viewportName } = target as IGuardTarget;
     if (typeof target === 'string') {
       this.componentName = target;
-    } else if (component || componentName || viewport || viewportName) {
-      this.component = component;
-      this.componentName = componentName;
-      this.viewport = viewport;
-      this.viewportName = viewportName;
+    } else if (ComponentAppellationResolver.isType(target as IRouteableComponentType<C>)) {
+      this.component = target as IRouteableComponentType<C>;
+      this.componentName = ComponentAppellationResolver.getName(target as IRouteableComponentType<C>);
     } else {
-      this.component = target as IRouteableComponentType;
+      const cvTarget = target as IComponentAndOrViewportOrNothing;
+      this.component = ComponentAppellationResolver.isType(cvTarget.component) ? ComponentAppellationResolver.getType(cvTarget.component as IRouteableComponentType<C>) : null;
+      this.componentName = ComponentAppellationResolver.getName(cvTarget.component)
+      this.viewport = ViewportAppellationResolver.isInstance(cvTarget.viewport) ? cvTarget.viewport as Viewport : null;
+      this.viewportName = ViewportAppellationResolver.getName(cvTarget.viewport);
     }
   }
 
--- a/packages/router/src/guard.ts
+++ b/packages/router/src/guard.ts
@@ -1,7 +1,7 @@
 import { GuardIdentity, GuardTypes, IGuardOptions, } from './guardian';
+import { GuardFunction, GuardTarget, IGuardTarget, INavigatorInstruction, IRouteableComponentType } from './interfaces';
 import { Viewport } from './viewport';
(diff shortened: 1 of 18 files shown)


packages/router/src/guard.ts
```diff
@@ -43,7 +42,7 @@ export class Guard {
 }
 
 class Target {
-  public component?: Partial<ICustomElementType>;
+  public component?: IRouteableComponentType;
```
fkleuver: `component` would refer to instance, so this is confusing. Should be either named `type` or `componentType`. Just `type` is probably fine.

Also, should add a generic type arg to `Target` and pass it through to `IRouteableComponentType`

packages/router/src/guardian.ts
```diff
@@ -1,12 +1,11 @@
...
-
 import { Guard } from './guard';
 import { INavigatorInstruction } from './navigator';
 import { Viewport } from './viewport';
+import { IRouteableComponentType } from './viewport-content';
 import { ViewportInstruction } from './viewport-instruction';
 
 export interface IGuardTarget {
-  component?: Partial<ICustomElementType>;
+  component?: IRouteableComponentType;
```
fkleuver: Wait so is this `IGuardTarget` an interface with only optional properties? What kind of "thing" can this be?

packages/router/src/router.ts
```diff
@@ -29,16 +30,16 @@ export interface IRouterOptions extends INavigatorOptions, IRouteTransformer {
...
 
 export interface IViewportComponent {
-  component: string | Partial<ICustomElementType>;
+  component: string | IRouteableComponentType;
   viewport?: string | Viewport;
   parameters?: Record<string, unknown> | string; // TODO: Allow unknown[] for parameters
 }
 
-export type NavigationInstruction = string | Partial<ICustomElementType> | IViewportComponent | ViewportInstruction;
+export type NavigationInstruction = string | IRouteableComponentType | IViewportComponent | ViewportInstruction;
```
fkleuver: Should also have the generic type arg. Possibly needs to be passed through to all 3 even:

```ts
export type NavigationInstruction<C extends Constructable = Constructable> = (
  string |
  IRouteableComponentType<C> |
  IViewportComponent<C> |
  ViewportInstruction<C>
);
```

packages/router/src/viewport-content.ts
```diff
@@ -108,7 +108,7 @@ export class ViewportContent {
       return Promise.resolve(true);
     }
 
-    const contentType: IRouteableCustomElementType = this.component !== null ? this.component.constructor : this.content as IRouteableCustomElementType;
+    const contentType: IRouteableComponentType = this.component !== null ? this.component.constructor : this.content as IRouteableComponentType;
```
fkleuver: This is another example of `type` vs `instance` discrepancy. Conceptually we can't be renaming `content` to `contentType`. One of the two is wrongly named.

packages/router/src/viewport-instruction.ts
```diff
@@ -72,12 +72,12 @@ export class ViewportInstruction {
...
   }
 
-  public componentType(context: IRenderContext): IRouteableCustomElementType {
+  public componentType(context: IRenderContext): IRouteableComponentType {
     if (this.component !== null) {
       return this.component;
     }
     const container = context.get(IContainer);
-    const resolver = container.getResolver<Constructable & IRouteableCustomElementType>(CustomElement.keyFrom(this.componentName));
+    const resolver = container.getResolver<Constructable & IRouteableComponentType>(CustomElement.keyFrom(this.componentName));
```
fkleuver: This is an example of why the generic type arg should be passed through. Consider this method body:

(also note I added `| null` to the return type because it seems like `null` can be returned. Not sure why typescript isn't complaining about this..)

```ts
  public componentType<C extends Constructable = Constructable>(context: IRenderContext): IRouteableComponentType<C> | null {
    if (this.component !== null) {
      return this.component;
    }
    const container = context.get(IContainer);
    const resolver = container.getResolver<IRouteableComponentType<C>>(CustomElement.keyFrom(this.componentName));
    if (resolver !== null) {
      return resolver.getFactory(container).Type;
    }
    return null;
  }
```

Now, you can do: `instruction.componentType<Foo>(context);` and you'll have a properly typed thing with user properties included.

AureliaEffect:
JS Framework Benchmark: https://24898-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://24921-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://25069-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://25083-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/router/src/guard.ts
jwx: The name will change in an upcoming PR.
We tried the generic type arg and went back on it.

AureliaEffect:
JS Framework Benchmark: https://25103-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/router/src/guardian.ts
jwx: This will change in an upcoming PR based on our discussion.

packages/router/src/router.ts
jwx: We tried and backtracked.

packages/router/src/viewport-content.ts
jwx: This will be sorted in an upcoming PR.

packages/router/src/viewport-instruction.ts
jwx: We tried the change and went back again.

AureliaEffect:
JS Framework Benchmark: https://25110-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

Issue #375, opened 2019-01-12
JIT-Mode Usability


EisenbergEffect:
I've been working with vNext to start building an app. It's a bit rough for normal use due to little things here and there. I wouldn't recommend it to anyone outside of the core team now. We've fixed a few issues already along the way though and I'm getting pretty excited.

I wasn't sure how nice we'd be able to make JIT mode. I knew AOT would be great. So, I've been playing with some ideas for JIT mode and require.js+CLI bundler. With the help of @huochunpeng we got some pretty cool stuff working.

By using a require.js plugin I wrote, I can do something like this:

```javascript
import { component } from 'au!./auth-experience.html';
import { NewAccount } from './new-account';
import { UserLogin } from './user-login';

@component
export class AuthExperience {
  public static inject = [UserLogin, NewAccount];
  public currentScreen: UserLogin | NewAccount;

  constructor(private userLogin: UserLogin, private newAccount: NewAccount) { }
}
```

The plugin exports a `component` decorator for an HTML file. The decorator basically wraps our `customElement` decorator and binds it to a template definition based on the HTML. The component name is derived from the html file name. It also does some app-specific stuff for my app, such as always configuring shadow dom open mode, and always injecting shared styles into the shadow DOM. Furthermore, because it's a require.js plugin, it's able to parse the `require` elements from the template, load them, and provide the dependencies to the template definition, so the compiler can process everything properly. It also handles html-only elements when the `require` has an `.html` extension.

Notice that I'm importing `component` as a named import. That's because I made the plugin generate a module that exports other useful things. For example, you could  `import { definition } from 'au!...` to get the actual template definition object, which you can do various things with. In fact, internal to the plugin, this is how the HTML-only elements are handled. It looks at the dependency, grabs the definition directly, and creates the html-only element from that.

A third export that the module provides is another decorator called `associateView`. It's a simple decorator that allows you to take the template description and tag it on an arbitrary class constructor as a `$view` property. So, you can do something like this:

```javascript
import { associateView } from 'au!./new-account.html';

@associateView
export class NewAccount {
  public message = 'New account placeholder text.';
}
```

[...]

PR #584, opened 2019-08-19
WIP: View Locator

EisenbergEffect:
# Pull Request

## 📖 Description

This PR enables any class to have one or more associated views. With a view associated, the `au-compose` element can then render instances of the class by using the `view` value converter.

### 🎫 Issues

* #375 

## 👩‍💻 Reviewer Notes

The interesting part is the implementation of the ViewLocator, which creates (and caches) components on the fly to allow existing, non-component object instances to be rendered as components with their associated view.

## 📑 Test Plan

Tests coming in future commits to this PR...

## ⏭ Next Steps

We should add a new convention for associating views. Today, if you have a class in a file named `foo.js` and you have a view in `foo.html`, the convention is to create a component. We should add that if the html file has the "view" suffix, then the class should merely have the view associated rather than being made into a component. For example, `foo.js` and `foo-view.html` will not result in a component. The view will be compiled the same, but the `@view(template)` decorator will be added to the class in `foo.js` instead of the `@customElement(template)` decorator. This will enable robust, convention-based, dynamic composition of vanilla JS classes.

### DIFF
--- a/packages/runtime/src/templating/view.ts
+++ b/packages/runtime/src/templating/view.ts
@@ -1,4 +1,4 @@
-import { PLATFORM, Reporter } from '@aurelia/kernel';
+import { PLATFORM, Reporter, Constructable, DI, Writable } from '@aurelia/kernel';
 import { ITemplateDefinition, TemplatePartDefinitions } from '../definitions';
 import { INode } from '../dom';
 import { LifecycleFlags, State } from '../flags';
@@ -9,6 +9,9 @@ import {
 } from '../lifecycle';
 import { ITemplate } from '../rendering-engine';
 import { Controller } from './controller';
+import { CustomElement } from '../resources/custom-element';
+import { Scope } from '../observation/binding-context';
+import { IScope } from '../observation';
 
 export class ViewFactory<T extends INode = INode> implements IViewFactory<T> {
   public static maxCacheSize: number = 0xFFFF;
@@ -93,5 +96,155 @@ export class ViewFactory<T extends INode = INode> implements IViewFactory<T> {
     } else {
       Object.assign(this.parts, parts);
     }
+  }
+}
+
+type HasAssociatedViews = {
+  $views: ITemplateDefinition[];
+};
+
+export function view(v: ITemplateDefinition) {
+  return function<T extends Constructable>(target: T & Partial<HasAssociatedViews>) {
+    const views = target.$views || (target.$views = []);
+    views.push(v);
+  };
+}
+
+function hasAssociatedViews<T>(object: T): object is T & HasAssociatedViews {
+  return object && '$views' in object;
+}
+
+export const IViewLocator = DI.createInterface<IViewLocator>('IViewLocator')
+  .withDefault(x => x.singleton(ViewLocator));
+
+export interface IViewLocator {
+  getViewComponentForModelInstance(model: object, requestedViewName?: string): Constructable | null;
+}
+
+const lifecycleCallbacks = [
+  'binding',
+  'bound',
+  'attaching',
+  'attached',
+  'detaching',
+  'caching',
+  'detached',
+  'unbinding',
+  'unbound'
+];
+
+export class ViewLocator implements IViewLocator {
+  private modelInstanceToBoundComponent: WeakMap<object, Record<string, Constructable>> = new WeakMap();
+  private modelTypeToUnboundComponent: Map<object, Record<string, Constructable>> = new Map();
+
+  public getViewComponentForModelInstance(model: object, viewName?: string) {
+    if (model && hasAssociatedViews(model.constructor)) {
+      const availableViews = model.constructor.$views;
+      const resolvedViewName = this.getViewName(availableViews, viewName);
+
+      return this.getOrCreateBoundComponent(
+        model,
+        availableViews,
+        resolvedViewName
+      );
+    }
+
+    return null;
+  }
+
+  private getOrCreateBoundComponent(model: object, availableViews: ITemplateDefinition[], resolvedViewName: string) {
+    let lookup = this.modelInstanceToBoundComponent.get(model);
+    let BoundComponent;
+
+    if (!lookup) {
+      lookup = {};
+      this.modelInstanceToBoundComponent.set(model, lookup);
+    } else {
+      BoundComponent = lookup[resolvedViewName];
+    }
+
+    if (!BoundComponent) {
+      const UnboundComponent = this.getOrCreateUnboundComponent(
+        model,
+        availableViews,
+        resolvedViewName
+      );
+
+      BoundComponent = class extends UnboundComponent {
+        constructor() {
+          super(model);
+        }
+      };
+
+      lookup[resolvedViewName] = BoundComponent;
+    }
+
+    return BoundComponent;
+  }
+
+  private getOrCreateUnboundComponent(model: object, availableViews: ITemplateDefinition[], resolvedViewName: string) {
+    let lookup = this.modelTypeToUnboundComponent.get(model.constructor);
+    let UnboundComponent;
+
+    if (!lookup) {
+      lookup = {};
+      this.modelTypeToUnboundComponent.set(model.constructor, lookup);
+    } else {
+      UnboundComponent = lookup[resolvedViewName];
+    }
+
+    if (!UnboundComponent) {
+      UnboundComponent = CustomElement.define(
+        this.getView(availableViews, resolvedViewName),
+        class {
+          protected $scope!: IScope;
+
+          constructor(protected viewModel: any) {}
+
+          public created() {
+            this.$scope = Scope.fromParent(0, this.$scope, this.viewModel);
+
+            if ('created' in this.viewModel) {
+              this.viewModel.created();
+            }
+          }
+        }
+      );
+
+      const proto = UnboundComponent.prototype as any;
+
+      lifecycleCallbacks.forEach(x => {
+        if (x in model) {
+          proto[x] = function() { return this.viewModel[x](); };
+        }
+      });
+
+      lookup[resolvedViewName] = UnboundComponent;
+    }
+
+    return UnboundComponent;
+  }
+
+  private getViewName(views: ITemplateDefinition[], requestedName?: string) {
+    if (requestedName) {
+      return requestedName;
+    }
+
+    if (views.length === 1) {
+      return views[0].name;
+    }
+
+    return 'default';
+  }
+
+  private getView(views: ITemplateDefinition[], name: string): ITemplateDefinition {
+    const v = views.find(x => x.name === name);
+
+    if (!v) {
+      // TODO: user Reporter
+      throw new Error(`Could not find view: ${name}`);
+    }
+
+    return v;
   }
 }
--- a/packages/runtime/src/configuration.ts
+++ b/packages/runtime/src/configuration.ts
@@ -39,6 +39,7 @@ import { Repeat } from './resources/custom-attributes/repeat';
 import { Replaceable } from './resources/custom-attributes/replaceable';
 import { With } from './resources/custom-attributes/with';
 import { SanitizeValueConverter } from './resources/value-converters/sanitize';
+import { ViewValueConverter } from './resources/value-converters/view';
 
 export const IObserverLocatorRegistration = ObserverLocator as IRegistry;
 export const ILifecycleRegistration = Lifecycle as IRegistry;
@@ -67,6 +68,7 @@ export const RepeatRegistration = Repeat as IRegistry;
 export const ReplaceableRegistration = Replaceable as unknown as IRegistry;
 export const WithRegistration = With as IRegistry;
 export const SanitizeValueConverterRegistration = SanitizeValueConverter as unknown as IRegistry;
+export const ViewValueConverterRegistration = ViewValueConverter as unknown as IRegistry;
 export const DebounceBindingBehaviorRegistration = DebounceBindingBehavior as unknown as IRegistry;
 export const OneTimeBindingBehaviorRegistration = OneTimeBindingBehavior as unknown as IRegistry;
 export const ToViewBindingBehaviorRegistration = ToViewBindingBehavior as unknown as IRegistry;
@@ -92,6 +94,7 @@ export const DefaultResources = [
   ReplaceableRegistration,
   WithRegistration,
   SanitizeValueConverterRegistration,
+  ViewValueConverterRegistration,
   DebounceBindingBehaviorRegistration,
   OneTimeBindingBehaviorRegistration,
   ToViewBindingBehaviorRegistration,
--- a/packages/runtime/src/index.ts
+++ b/packages/runtime/src/index.ts
@@ -290,6 +290,9 @@ export {
 } from './templating/controller';
 export {
   ViewFactory,
+  IViewLocator,
+  ViewLocator,
+  view
 } from './templating/view';
 
 export {
--- a/packages/runtime/src/resources/value-converters/view.ts
+++ b/packages/runtime/src/resources/value-converters/view.ts
@@ -1,0 +1,14 @@
+import { IViewLocator } from '../../templating/view';
+import { valueConverter } from '../value-converter';
+
+@valueConverter('view')
+export class ViewValueConverter {
+  constructor(@IViewLocator private viewLocator: IViewLocator) {}
+
+  public toView(subject: object, viewName?: string) {
+    return this.viewLocator.getViewComponentForModelInstance(
+      subject,
+      viewName
+    );
+  }
+}


AureliaEffect:
JS Framework Benchmark: https://24696-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/runtime/src/templating/view.ts
```diff
@@ -95,3 +98,153 @@ export class ViewFactory<T extends INode = INode> implements IViewFactory<T> {
...
+  'unbinding',
+  'unbound'
+];
+
+export class ViewLocator implements IViewLocator {
+  private modelInstanceToBoundComponent: WeakMap<object, Record<string, Constructable>> = new WeakMap();
+  private modelTypeToUnboundComponent: Map<object, Record<string, Constructable>> = new Map();
+
+  public getViewComponentForModelInstance(model: object, viewName?: string) {
+    if (model && hasAssociatedViews(model.constructor)) {
```
fkleuver: It's a nit, but if `model` can be nil then the argument type should reflect that. `object` means "not primitive", so you can't pass null or undefined (or anything that's not an object) into it.

Perhaps `unknown` is the more appropriate type for `model` if we really can't be sure what it is?

packages/runtime/src/templating/view.ts
```diff
@@ -95,3 +98,153 @@ export class ViewFactory<T extends INode = INode> implements IViewFactory<T> {
...
+        resolvedViewName
+      );
+    }
+
+    return null;
+  }
+
+  private getOrCreateBoundComponent(model: object, availableViews: ITemplateDefinition[], resolvedViewName: string) {
+    let lookup = this.modelInstanceToBoundComponent.get(model);
+    let BoundComponent;
```
fkleuver: This is implicit `any`, what kind of type is this supposed to be?

packages/runtime/src/templating/view.ts
```diff
@@ -95,3 +98,153 @@ export class ViewFactory<T extends INode = INode> implements IViewFactory<T> {
...
+            this.$scope = Scope.fromParent(0, this.$scope, this.viewModel);
+
+            if ('created' in this.viewModel) {
+              this.viewModel.created();
+            }
+          }
+        }
+      );
+
+      const proto = UnboundComponent.prototype as any;
```
fkleuver: Could cast this to `IViewModel`. That has all the optional hooks properties on it. Then `lifecycleCallbacks` could be typed as `keyof IViewModel` and then the typings should be sound, even in copying the methods

packages/runtime/src/templating/view.ts
```diff
@@ -95,3 +98,153 @@ export class ViewFactory<T extends INode = INode> implements IViewFactory<T> {
...
+
+      lifecycleCallbacks.forEach(x => {
+        if (x in model) {
+          proto[x] = function() { return this.viewModel[x](); };
```
fkleuver: What about other methods, getters, etc?

packages/runtime/src/templating/view.ts
```diff
@@ -95,3 +98,153 @@ export class ViewFactory<T extends INode = INode> implements IViewFactory<T> {
...
+        this.getView(availableViews, resolvedViewName),
+        class {
+          protected $scope!: IScope;
+
+          constructor(protected viewModel: any) {}
+
+          public created() {
+            this.$scope = Scope.fromParent(0, this.$scope, this.viewModel);
+
+            if ('created' in this.viewModel) {
```
fkleuver: Is `this.viewModel` supposed to be a custom element or a non-compiled/rendered arbitrary class? If it's already rendered there are some `$controller` properties available to help with this

packages/runtime/src/templating/view.ts
EisenbergEffect: There shouldn't be a need to copy anything beyond what Aurelia internally might look for. Everything else is available via the scope fixup that points it at the model.

AureliaEffect:
JS Framework Benchmark: https://25010-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://25223-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://25237-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://25259-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://25394-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://25472-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/runtime/src/templating/view.ts
```diff
@@ -95,3 +99,160 @@ export class ViewFactory<T extends INode = INode> implements IViewFactory<T> {
...
+    views.push(v);
+  };
+}
+
+function hasAssociatedViews<T>(object: T): object is T & HasAssociatedViews {
+  return object && '$views' in object;
+}
+
+export const IViewLocator = DI.createInterface<IViewLocator>('IViewLocator')
+  .withDefault(x => x.singleton(ViewLocator));
```
fkleuver: I'd prefer if we use only `noDefault()` in the framework and explicitly register stuff in the exported `Configuration` for a few reasons:
- No explicit coupling between interface and implementation
- One place to find all registrations
- Easier to override default registrations without having to worry about temporal coupling (e.g. if you resolve an interface with a default then you can't override it anymore and it could leave users scratching their heads on why their implementation isn't resolved, whereas without a default they would get an error and know immediately that they need to register their implementation sooner)

packages/runtime/src/templating/view.ts
```diff
@@ -95,3 +99,160 @@ export class ViewFactory<T extends INode = INode> implements IViewFactory<T> {
...
+
+export interface IViewLocator {
+  getViewComponentForObject(object: ComposableObject | null | undefined, viewNameOrSelector?: string | ViewSelector): Constructable | null;
```
fkleuver:
```suggestion
  getViewComponentForObject<T extends ComposableObject>(
    object: T | null | undefined,
    viewNameOrSelector?: string | ViewSelector,
  ): ConstructableClass<T> | null;
```

This would retain the type information on the `prototype` and `constructor` (of the instance) properties of the type the user passed in.

bigopon:
How does this change worm with static view declaration? In vCurrent we allow template configuration via staic `$view` property as well

packages/runtime/src/templating/view.ts
```diff
@@ -95,3 +99,160 @@ export class ViewFactory<T extends INode = INode> implements IViewFactory<T> {
...
+      return views[0].name;
+    }
+
+    return 'default-view';
+  }
+
+  private getView(views: ITemplateDefinition[], name: string): ITemplateDefinition {
+    const v = views.find(x => x.name === name);
+
+    if (!v) {
```
fkleuver: Can only ever be `undefined` if there is no result, so this is slightly better for perf as it avoids coercion

```suggestion
    if (v === void 0) {
```

packages/runtime/src/templating/view.ts
```diff
@@ -95,3 +99,160 @@ export class ViewFactory<T extends INode = INode> implements IViewFactory<T> {
...
+    }
+
+    return null;
+  }
+
+  private getOrCreateBoundComponent(object: ComposableObject, availableViews: ITemplateDefinition[], resolvedViewName: string) {
+    let lookup = this.modelInstanceToBoundComponent.get(object);
+    let BoundComponent: Constructable | undefined;
+
+    if (!lookup) {
```
fkleuver: Ditto
```suggestion
    if (lookup === void 0) {
```

packages/runtime/src/templating/view.ts
```diff
@@ -95,3 +99,160 @@ export class ViewFactory<T extends INode = INode> implements IViewFactory<T> {
...
+  'attached',
+  'detaching',
+  'caching',
+  'detached',
+  'unbinding',
+  'unbound'
+] as (keyof Omit<ComposableObject, 'created'>)[];
+
+export class ViewLocator implements IViewLocator {
+  private modelInstanceToBoundComponent: WeakMap<object, Record<string, Constructable>> = new WeakMap();
```
fkleuver:
```suggestion
  private readonly modelInstanceToBoundComponent: WeakMap<object, Record<string, Constructable>> = new WeakMap();
```

packages/runtime/src/templating/view.ts
```diff
@@ -95,3 +99,160 @@ export class ViewFactory<T extends INode = INode> implements IViewFactory<T> {
...
+  private modelTypeToUnboundComponent: Map<object, Record<string, Constructable>> = new Map();
```
fkleuver:
```suggestion
  private readonly modelTypeToUnboundComponent: Map<object, Record<string, Constructable>> = new Map();
```

packages/runtime/src/templating/view.ts
```diff
@@ -95,3 +99,160 @@ export class ViewFactory<T extends INode = INode> implements IViewFactory<T> {
...
+    let BoundComponent: Constructable | undefined;
+
+    if (!lookup) {
+      lookup = {};
+      this.modelInstanceToBoundComponent.set(object, lookup);
+    } else {
+      BoundComponent = lookup[resolvedViewName];
+    }
+
+    if (!BoundComponent) {
```
fkleuver:
```suggestion
    if (BoundComponent === void 0) {
```

packages/runtime/src/templating/view.ts
```diff
@@ -95,3 +99,160 @@ export class ViewFactory<T extends INode = INode> implements IViewFactory<T> {
...
+            }
+          }
+        }
+      );
+
+      const proto = UnboundComponent.prototype;
+
+      lifecycleCallbacks.forEach(x => {
+        if (x in object) {
+          proto[x] = function() { return this.viewModel[x](); };
```
fkleuver:
```suggestion
          const fn = function () { return this.viewModel[x](); };
          Reflect.defineProperty(fn, 'name', { configurable: true, value: x });
          proto[x] = fn;
```

- Missing space between keyword and open paren (styling thing)
- Retain the original function name for better debugging / stack traces

packages/runtime/src/templating/view.ts
```diff
@@ -95,3 +99,160 @@ export class ViewFactory<T extends INode = INode> implements IViewFactory<T> {
...
+        : this.getViewName(availableViews, viewNameOrSelector);
+
+      return this.getOrCreateBoundComponent(
+        object,
+        availableViews,
+        resolvedViewName
+      );
+    }
+
+    return null;
```
fkleuver: A bit nitpicky, but general coding style suggestion for consistency (and imo better readability). 

There's also the control flow thing, it looks like an empty array of `availableViews` would in every realistic scenario result in an error being thrown anyway, so for the empty array scenario a specific error seems more helpful:

```ts
  public getViewComponentForObject(object: ComposableObject | null | undefined, viewNameOrSelector?: string | ViewSelector) {
    if (object == null) {
      return null;
    }

    if (!hasAssociatedViews(object.constructor)) {
      throw new Error(`The component has no associated views`);
    }

    const resolvedViewName = typeof viewNameOrSelector === 'function'
      ? viewNameOrSelector(object, object.constructor.$views)
      : this.getViewName(object.constructor.$views, viewNameOrSelector);

    return this.getOrCreateBoundComponent(
      object,
      object.constructor.$views,
      resolvedViewName
    );
  }
```

packages/__tests__/runtime/view-location.spec.ts
```diff
@@ -0,0 +1,222 @@
...
+    'binding',
+    'bound',
+    'attaching',
+    'attached',
+    'detaching',
+    'caching',
+    'detached',
+    'unbinding',
+    'unbound'
+  ].forEach(lifecycleHook => {
```
fkleuver: Maybe an idea to export `lifecycleCallbacks` from `lifecycle.ts` so it can be reused. Could be useful to users as well for the same reasons

packages/__tests__/runtime/view-location.spec.ts
```diff
@@ -0,0 +1,222 @@
...
+      const locator = new ViewLocator();
+      const model = new MyModel();
+
+      const Component = locator.getViewComponentForObject(model as any);
+      const instance: any = new Component();
+
+      instance.$scope = {};
+      instance[lifecycleHook]();
+
+      assert.equal(CustomElement.isType(Component), true);
```
fkleuver: I think we should add this as an assertion, e.g. (in `packages/testing/src/assert.ts`):

```ts

export function isCustomElementType(actual: any, message?: string): void {
  if (CustomElement.isType(actual)) {
    innerFail({
      actual: false,
      expected: true,
      message,
      operator: 'isCustomElementType' as any,
      stackStartFn: isCustomElementType
    });
  }
}

const assert = Object_freeze({
  // ...
  isCustomElementType,
  // ...
});
```

And then:

```ts
assert.isCustomElementType(Component);
```

This would make test error message more helpful

packages/__tests__/runtime/view-location.spec.ts
```diff
@@ -0,0 +1,222 @@
...
+      }
+
+      MyModel.prototype[lifecycleHook] = function(this: MyModel) {
+        this.invoked = true;
+      };
+
+      const locator = new ViewLocator();
+      const model = new MyModel();
+
+      const Component = locator.getViewComponentForObject(model as any);
```
fkleuver:
```suggestion
      const Component = locator.getViewComponentForObject(model);
```

fkleuver:
@bigopon By the looks of it, it's perfectly compatible. The decorator just adds def to static `$views` array if it exists or creates new array if it doesn't. If you create the static array directly yourself, it seems like the effect would be the same.
Perhaps a unit test to demonstrate that would be a nice addition tho?

bigopon:
@fkleuver Yes, I was confirming with @EisenbergEffect before asking for some test to demonstrate that behavior.

packages/runtime/src/templating/view.ts
EisenbergEffect: There is a reason to allow the empty array. I forgot that test case. I'll add it. Basically, this enables providing views dynamically through the selector function, if none is associated.

packages/__tests__/runtime/view-location.spec.ts
EisenbergEffect: This causes a compiler error.

packages/__tests__/runtime/view-location.spec.ts
EisenbergEffect: In my case, it wouldn't help because we don't want the created callback in this case. So, I'd have to apply a filter on that if I wanted to re-use the array. I'd rather wait on that for now to see if there other situations that would justify that change.

EisenbergEffect:
@bigopon I added a test to show statically defining the views instead of using the decorator. However, I don't think either this or the decorator are the main way it will be used. The beauty of this is how it works with conventions. Once conventions are in place, this enables you to simply create an HTML file to describe how any part of your model (domain model or vanilla view-models) can be rendered. This whole PR is about opening up to even more vanilla JS architectures. Using the converter and service in this PR, I've built an entire app with basically no components; just vanilla JS classes composed through DI and associated views.

AureliaEffect:
JS Framework Benchmark: https://25739-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/runtime/src/templating/view.ts
```diff
@@ -95,3 +99,160 @@ export class ViewFactory<T extends INode = INode> implements IViewFactory<T> {
...
+}
+
+export const IViewLocator = DI.createInterface<IViewLocator>('IViewLocator')
+  .withDefault(x => x.singleton(ViewLocator));
+
+export interface IViewLocator {
+  getViewComponentForObject(object: ComposableObject | null | undefined, viewNameOrSelector?: string | ViewSelector): Constructable | null;
+}
+
+export type ComposableObject = Omit<IViewModel, '$controller'>;
```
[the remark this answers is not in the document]
EisenbergEffect: Yes, that's correct. ComposableObject represents an arbitrary object that the developer provides. That's essentially the whole purpose of this feature set: to allow composition of truly vanilla JS objects. I had introduced the ComposableObject at your request from an earlier comment to base this on IViewModel. However, perhaps you missed the details of this scenario when you first made that recommendation.

I do think IViewModel should remain an interface. I think some TS users will want to implement that explicitly. The only real difference in terms of type signatures is that the ComposableObject doesn't have a `$controller` because it's not a component itself. It just supports the same lifecycle methods so that the actual component can delegate those through to the model.

I don't have strong opinions about how to handle this. I can make whatever changes you think are best. I think this is the last issue to resolve before merging, if you do want to make a change. @fkleuver

packages/__tests__/runtime/view-location.spec.ts
fkleuver: @EisenbergEffect What error exactly? I think it should be fixed at a typing level, so that end users don't need to perform this hacky cast when they are doing unit tests.

EisenbergEffect: It's an odd error, and one that seems like a TS bug to me. I've put together a TS playground that shows the error, so you can take a look easily: http://www.typescriptlang.org/play/index.html#code/JYOwLgpgTgZghgYwgAgJIDVgQO4FkD2AJhADbIDeAUMjcgEaiGgDmA-ABQCUAXMgG75ghANzVacMGEQALFhx79BIygF9KlBCTgBnbclwBPTDgLEyVWsgAOAVzolgCZNvwBbCAAUo+K9DAHeOnx8Egg4EGQAXmR4Em0IUTVKGBsQBDBgfAjCfABlNwgwWRBmdj4sPCJSXgwK01JOCnUkhCztMGRQdvCkKOQQHH0jOqqSLlEc-PcilnYuqTSITmEgA

fkleuver: Ah yes. It's not really a bug, more of a quirk in a principle of the type system which matches by properties/methods.

An interface with only optional properties will only match `{}` (an object with no properties, which it is by default equivalent to) or an object that has at least one of its properties. Otherwise the two types are simply unrelated as far as the type system is concerned.

This would be one way to solve that:

```ts
type ClassInstance<T> = T & {
  readonly constructor: Function;
}

getViewComponentForObject<T extends ClassInstance<ComposableObject>>(
    object: T | null | undefined,
    viewNameOrSelector?: string | ViewSelector,
  ): ComposableObjectComponentType<T> | null;
```

This works, because every class instance (and plain object) has a `constructor` property. So that's a property the types will then always have in common, if nothing else.

fkleuver:
@EisenbergEffect I do thing a proper integration test or two would be appropriate here, regardless of whether that's how people will actually use it or not. For documentation and for verification that e.g. the way you create the scope will actually work (because I have this vague feeling it doesn't, since the scope sits on `this.$controller.scope`)

I tried to come up with an integration test, don't know if it is supposed to work according to the way you designed the API, but perhaps you can use it as a starting point to make a few working examples?

You could add this to e.g. `packages/__tests__/jit-html/compose.spec.ts` since it's a `compose` integration if I'm not mistaken:

```ts
  describe('With the ViewLocator value converter', function () {
    it('can bind to a property from the parent scope', async function () {
      const { au, host } = setup();

      @view({ name: 'default-view', template: '{message}' })
      class MyModel {}

      @customElement({ name: 'app', template: '<au-compose subject.bind="model | view"></au-compose>' })
      class App {
        public model = new MyModel();
        public message = 'Hello world!';
      }

      au.app({ host, component: App });
      await au.start().wait();

      assert.visibleTextEqual(au.root, 'Hello world!');

      await au.stop().wait();
    });
  });
```

fkleuver:
@EisenbergEffect Awesome :) And that's why I only trust integration tests. For future reference, you can easily debug tests by changing the file filter in the webpack config and running `npm run test-chrome:debugger` and opening the debug window. I think that's a fair bit easier than console.log statements.
The problem with the line numbers is because the test process is still in an intermediate state where transpilation happens separately, so source maps aren't propagated properly. It's something I intend to fix with AOT

fkleuver:
Ah, I recall changing that `this` to `Type`. That was part of a slight overhaul where I changed static `this` references to direct class references. The reasoning for that is twofold:
1. Performance. It stores the type in the closure at declaration time and thus saves the browser from having to do a type lookup (polymorphism thing)
2. Flexibility. It allows users to store the static functions in variables and they will still work, even though they lost their correct `this` context.

I did not realize in the case of registrations that it could have this particular effect, so my apologies for breaking that. It goes to show that we need more integration tests with inheritance, to see if there happen to be any other places where it falls apart. In any case I'm glad it surfaced now :)

3cp:
Sure, I am aware of the new view convention, I will make another PR for it after we test and merge the new features on conventions.

3cp:
Question: given
```js
import template from './person-view.html';
import { view } from '@aurelia/runtime';

@view(template)
export class Person {
  firstName: string;
  lastName: string;
}
```

Following usage is fine
```html
<au-compose subject.bind="personInstanceProperty | view"></au-compose>
```

But how would webpack trace dependency in
```html
<au-compose subject.bind="personInstanceProperty | view: 'detail-view'"></au-compose>
```

fkleuver:
It needs to be registered in DI somewhere.

EisenbergEffect:
My thought on this was simply that if someone wants to use a named view (not the conventional default), then they just have to add the `@view` decorator themselves. So, the new convention will handle the most common case where there's a single default view. If the developer needs more than one view or a specially-named view, they are responsible for overidding the convention by adding explicit decorators.

Does that make sense?

3cp:
@EisenbergEffect something is missing here. Your `@view(string)` has no name. You probably need to change api to something like:

```js
import template from './person-view.html';
import template2 from './other-view.html';
import { view } from '@aurelia/runtime';

@view({ name: 'default', template })
@view({ name: 'other-view', template2 })
export class Person {}
```

Or
```js
import template from './person-view.html';
import * as t1 from './other-view.html';
import { view } from '@aurelia/runtime';

@view({ name: 'default', template })
@view(t1) // only uses t1.name and t1.template
export class Person {}
```

3cp:
NVM, I read your code, and realised your sample code is not correct. You mean to use
```js
import * as t0 from './person-view.html';
import * as t1 from './other-view.html';
import { view } from '@aurelia/runtime';

@view(t0)
@view(t1)
export class Person {}
```
The first associated view is the default by default. Am I correct?

3cp:
Is this a good idea to update `@view` API to accept `@view('./person-view.html')` which by default will raise a runtime error: `"this doesn't work at runtime, make sure you configured aurelia conventions or AoT correctly in your toolchain."`?

EisenbergEffect:
@3cp The `@view` decorator takes an `ITemplateDefinition` just like the `@customElement` decorator does. I think that my original sample code was incorrect. I didn't mean for it to only import the template string, I meant for it to import the full template definition and pass that along to the decorator.

Currently, if there is more than one view, then the default is the one with the name `default-view`. I'm not sold on that though. We could make it the first view. What do you think?

3cp:
> make default the first view

This sounds intuitive without special naming.

Anyway, this multi-views use case is out of the scope of plugin-conventions.

EisenbergEffect:
I'll make the behavior change. It shouldn't affect the conventions though, yes, since it's only for the single view scenario.

3cp:
For `au-compose`, the multi-views is the dominant use case. Otherwise, user will not use `au-compose`.

EisenbergEffect:
I have an entire app built with single, default views. No components except for the root component. The reason for this is because I need to control the instantiation of all the models through DI, sometimes using child containers, and other techniques. Also, some of what I am rendering is instances of object that come from a web service, for which I've built classes. So, I can just associate a view with any class to render it. So, there are lots of other use cases for the single view, but it's more of an architecture choice in the way I'm using it.

3cp:
That sounds very flexible!

Issue #585, opened 2019-08-19
[RFC] more metadata in html template


3cp:
# 💬 RFC

Aurelia 2 has new import/require tags above the opening `<template>` tag. It looks like we can extend this practice to cover more metadata.

## 🔦 Context

This RFC is related to #116, #465. While keeping the old v1 syntax, we can have a new recommended syntax to put all metadata above the opening `<template>` tag for better readability and deliver a visual clue that these metadata are preprocessed in Aurelia 2, comparing to runtime processing in Aurelia v1.

## 💻 Examples

In v1, we have:

```html
<template use-shadow-dom="open" bindable="firstName, lastName" containerless>
  <require from="./foo.css"></require>
  <require from="./bar"></require>
</template>
```

In v2, **the above continues to work** (but preprocessed). This RFC suggests a new recommended syntax.

```html
<use-shadow-dom mode="open">
<containerless>
<bindable name="firstName" mode="one-way"> <!-- TBD mode="oneWay" or mode="one-way"  -->
<bindable name="lastName">
<import from="./foo.css">
<import from="./bar">

<template>
</template>
```

Notes:
1. The metadata nodes need no closing tag (like `</import>`).
  * They are above the opening `<template>` tag, they are all preprocessed before browser native html parser sees html template. This is very different from Aurelia v1 where all html string are processed at runtime by browser native html parser.
2. `<import>` and `<require>` tag are already implemented.
3. `use-shadow-dom` is for #465.
4. `bindable` I am not quite sure. Is it still processed at runtime in Aurelia 2? Should we move it to preprocessing?

eriklieben:
The above would also open up the possibility to enable you to set the other options for ```bindable``` right, which currently isn't possible in v1. Not sure if that is useful/makes sense for all the options, but seems like a logical step.

```typescript
@bindable({
    name:'myProperty',
    attribute:'my-property',
    changeHandler:'myPropertyChanged',
    defaultBindingMode: bindingMode.oneWay,
    defaultValue: undefined
  }) 
```
If this is how we would use our html only component:
```html
<custom-welcome firstname.bind="firstName"></custom-welcome>
```
Then we could potentially allow you to do:
```html
<bindable name="firstname" default-binding-mode="one-time" default-value="you">
<template>
  <p>Hey, ${firstname}!</p>
</template>
```
Looks like it's easier to read than:
```html
<template bindable="firstname">
[...]

Issue #578, opened 2019-08-14
Shadow DOM by Convention


EisenbergEffect:
# 🙋 Feature Request

When using the new conventions plugin, we'd like to be able to specify for it to use a particular shadow DOM configuration for elements.

## 🤔 Expected Behavior

Turning on shadow dom in the convention settings results in shadow dom config being generated properly for each compiled view.

## 😯 Current Behavior

There's no way to make shadow dom the default today.

## 💁 Possible Solution

Enhance the conventions module so that it can take a configuration object that sets up the conventions for the project it is being used to build. The first convention could be `shadowOptions`. Here's how that is defined int he framework:

```
shadowOptions?: { mode: 'open' | 'closed' } | null,
```

EisenbergEffect:
@3cp I created this based on our conversation the other day. I've got a new styles PR to add support for shadow dom styles. I'd like to try to get this feature in close in timing to the styles support for the feature I'm working on.

If you've got questions, ask away! 😄

3cp:
Question, to use shadow dom, is it a simple option on tooling? or do we require user to turn it on in `src/main.js` too?

Ideally, we should give user a single place, instead of two, to turn this feature on.

EisenbergEffect:
Currently, there's no framework-level setting to turn it on everywhere. I thought about that, but I felt that it could be a bad idea and negatively affect library authors. For example, if you are using a library with elements in it that don't use shadow dom and then you turn it on in your app, you just broke their elements. So, I'm thinking it should be a build configuration option and only affect the HTML in your app.

For libraries, I'm thinking they will have a build that applies their own conventions, which happens before they distribute.

Does that make sense?

3cp:
That works for me.

For clarify, can you review/fix the expected results for 2) shadow dom open and 3) shadow dom closed?

`foo.html`
```html
<import from="./foo.css">
<import from="./bar">
<template>
  <bar></bar>
</template>
```
1. no shadow dom.
```js
import './foo.css';
import * as d1 from './bar';
export const name = 'foo';
export const template = '<template>\n  <bar></bar>\n</template>\n';
export default template;
export const dependencies = [d1];
let _e;
export function getHTMLOnlyElement() {
  if (!_e) {
    _e = CustomElement.define({ name, template, dependencies });
  }
  return _e;
}
```
2. shadow dom open
```js
// first, show me where to import Registration

// for webpack
import d0 from 'raw-loader!./foo.css';
// or for dumber
[...]

Issue #465, opened 2019-03-10
Enable Shadow DOM through HTML


EisenbergEffect:
# 🙋 Feature Request

HTML Only elements should be able to declare that they use shadow dom.

## 🤔 Expected Behavior

Developers should be able to declare shadow dom use through HTML, not just JavaScript.

## 😯 Current Behavior

Shadow DOM can only be enabled through JavaScript.

## 💁 Possible Solution

See here for vCurrent: See here for the vCurrent behavior: https://github.com/aurelia/templating-resources/pull/368

EisenbergEffect:
@3cp this might be an interesting one to pick up while you are doing the shadow dom convention.

3cp:
Sounds good. Following per-file attribute will overwrite the option on bundler plugin.

```html
<template use-shadow-dom>
<template use-shadow-dom="open">
<template use-shadow-dom="closed">
```

Different from vcurrent, the `use-shadow-dom` attribute will be stripped off from the html text during preprocessing. Unless you have a reason to retain it for runtime usage.

3cp:
Since Aurelia 2 recommends to move import/require to above template tag, while supporting existing `use-shadow-dom` attribute, should we also recommend following syntax?

```html
<use-shadow-dom mode="closed">
<import from="./foo.css">

<template></template>
```

The `mode="closed"` reads better than `use-shadow-dom="closed"`.

EisenbergEffect:
I do like the fact that mode is an explicit attribute, which enables us to support other shadow dom properties in the future. I'd like to hear some other opinions on this though, since historically we've only had the require/import as a special case element and things like this have been put on the template.

PR #586, opened 2019-08-20
Various enhancement of conventions

3cp:
# Pull Request

Put in WIP mode.
- [x] Waiting for #584.
- [x] Add config for ShadowDOM and CSSModule
- [x] To implement #585 on bindable and containerless
- [x] To implement conventional css pair: auto import foo.css in foo.html.
- [x] Add config for file extensions for html file pair (e.g. `.html`, `.md`, `.pug`, `.haml`) and css file pair (e.g. `.css`, `.scss`, `.sass`, `.less`, `.styl`). The default config covered all popular file extensions.

## 📖 Description

Add defaultShadowOptions to webpack-loader and plugin-gulp.
Support per-html-file `use-shadow-dom` attribute and tag.
Warn user for one-word element with ShadowDOM at both compile time and runtime.

Add containerless, bindable.
Add support of css pair.
Add config for file extensions for html pair and css pair.

### 🎫 Issues

This PR implemented #465 and #578, more details in the two issues.
It also fixed a small issue in build script that left unintended change in `packages/__tests__/tsconfig.json`.

#585 containerless, bindable, omit top `<template>` tag.

## 👩‍💻 Reviewer Notes

I probably needs a better abstracted logger for the warning for that one-word element issue. We can improve logging in future if this is not too bad now.

## 📑 Test Plan

Tested against apps with webpack and dumber.

## ⏭ Next Steps

To turn it on in [aurelia/new](https://github.com/aurelia/new) after dev release.


3cp:
Again, codecov reported some files not touched here.

fkleuver:
@3cp This means your branch is based on an outdated version of master. Blur attr was merged yesterday and increased code coverage, now your branch does not include that and that's what the difference comes from. rebase :)

3cp:
Thx. Ready for review.

AureliaEffect:
JS Framework Benchmark: https://25050-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/plugin-conventions/src/preprocess-html-template.ts
```diff
@@ -2,44 +2,77 @@ import { kebabCase } from '@aurelia/kernel';
...
+    statements.push(`console.warn(${JSON.stringify(error)});\n`);
+  }
 
   deps.forEach((d, i) => {
     const ext = path.extname(d);
-    let importStatement: string;
 
     if (isCss(ext)) {
-      importStatement = `import ${s(d)};\n`;
+      if (shadowMode) {
```
EisenbergEffect: It should also be possible to configure the build so that it uses the defer registration for non-shadow mode scenarios. For example, if you are using CSS Modules Preprocessor, importing the css actually results in an object being imported that has key/values that map the friendly class names to the obfuscated names. We need that information at runtime, and it gets passes to the style system via the defer registration.

packages/plugin-conventions/src/preprocess.ts
```diff
@@ -7,16 +7,19 @@ export function preprocess(
   // The filePath is used in sourceMap.
   filePath: string,
   contents: string,
-  ts: boolean = false,
   // The base file path that filePath is related to. Used for checking existence of html pair.
+  // We separated filePath and basePath because filePath will be written into source map.
   basePath: string = '',
+  defaultShadowOptions: { mode: 'open' | 'closed' } | null = null,
```
EisenbergEffect: I wonder if we should turn this into an options object so that we can handle more settings in the future.

packages/plugin-gulp/README.md
```diff
@@ -19,3 +19,40 @@ For our nightly builds:
 ```bash
 npm install --save @aurelia/plugin-gulp@dev
 ```
+
+## Usage
```
EisenbergEffect: It makes me really happy that you always document things!

packages/plugin-conventions/src/preprocess-html-template.ts
3cp: Code sample pls.

packages/plugin-conventions/src/preprocess.ts
3cp: That's why I asked about the repeated shadowOptions typing in `@aurelia/runtime`. This should share the same typing. We'd better refactor all the typing for shadowOptions. I will defer this until someone cleanup `@aurelia/runtime` typings.

packages/plugin-conventions/src/preprocess-html-template.ts
EisenbergEffect: The developer writes the same thing but the CSS gets handled differently. With the styles features, they can do this in their app setup.

```JavaScript
aurelia.register(StyleConfiguration.cssModulesProcessor());
```

This will register a different defer handlers for `.css`, which will expect the result of the CSS import to be the CSS Modules object. For an explanation of CSS Module Preprocessing, see here: https://github.com/css-modules/css-modules There's a section at the bottom on how the css-loader works in Webpack as well.

So, in this particular case, we want to output the exact same thing as you are for shadow dom, except it won't be in shadow dom mode. Basically, any non JS/HTML import should always get turned into a `defer` call with the file extension as the first parameter, so that the runtime can then turn that type of import into something the templating engine can understand. 

Does that make sense?

AureliaEffect:
JS Framework Benchmark: https://25199-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://25208-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

3cp:
@EisenbergEffect now all unknown resources are wrapped in `Registration.defer`.

**TODO: first, you need to implement default behaviour of `Registration.defer('.ext_without_handler', foo)` that simply returns `foo`.**

The new behaviour for css resource:

1. No shadowdom, when CSSModule is on or off
```
import d0 from "./foo.css";
Registration.defer('.css', d0);
```
2. With shadowdom, when CSSModule is off
```
import d0 from "raw-loader!./foo.css";
Registration.defer('.css', d0);
```
3. (Doesn't make sense, but the code didn't care) With shadowdom, when CSSModule is on
```
import d0 from "./foo.css";
Registration.defer('.css', d0);
```

Note in situation No.1, `import d0 from "./foo.css";` will go through standard bundler setup. By default, bundler will inject the style onto html head. This is also the standard behaviour when CSSModule is in use.

Note we got the behaviour that will return `d0` (which is a string when CSSModule is off) from `Registration.defer('.css', d0);`. **TODO: you have to make sure DI will ignore a pure string.** because bundler has always done everything needed in `import d0 from "./foo.css";`.

Note the conventions implementation will call `Registration.defer('.css', d0);` for all known css extensions (css/scss/sass/less/styl). You don't need to do any alias for style handle in DI.

The two mentioned TODOs need to be implemented before merging this PR.

EisenbergEffect:
I'll fix up `defer` so that it handles this absent handler and string scenarios. Thanks for working on this @3cp !

EisenbergEffect:
@3cp Here's the solution I came up with:

```TypeScript
export const Registration = Object.freeze({
  defer(key: Key, ...params: unknown[]): IRegistry {
    return new ParameterizedRegistry(key, params);
  }
});

export class ParameterizedRegistry implements IRegistry {
  constructor(
    private readonly key: Key,
    private readonly params: unknown[]
  ) {}

  public register(container: IContainer): void {
    if (container.has(this.key, true)) {
      const registry = container.get<IRegistry>(this.key);
      registry.register(container, ...this.params);
    } else  {
      this.params.forEach(x => container.register(x));
    }
  }
}
```

So, if the container has a handler for the key, it will use it. If not, it will go through each items that was passed and try to register it directly. If any of those items can be registered, they will. If they can (e.g. string), then nothing will be placed into the container.

Sound good?

3cp:
That looks right.

jwx:
@3cp Will conventions also add a css to matching html/?s to "complete" a component?

EisenbergEffect:
@3cp How hard do you think it would be to automatically add an import for css files with the same name as the html?

3cp:
That's good idea. Should be easy.

3cp:
Put in WIP mode.
1. Waiting for #584.
2. To implement #585 on bindable and containerless

AureliaEffect:
JS Framework Benchmark: https://25493-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://25517-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://25530-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

3cp:
Tested with webpack and dumber bundler with babel and TS.

AureliaEffect:
JS Framework Benchmark: https://26186-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

EisenbergEffect:
An amazing achievement @3cp ! Congratulations on this :) It's a lot of code. I made pretty quick pass, mostly reading over the test cases. They look solid. I'm also very happy how thin the webpack and gulp integrations are. If you are ready, I'm happy to merge this. Let me know.

3cp:
Merge pls 😄 
It is quite thin, considering the covered cases.

repo | loc
--- | ---
plugin-conventions | 531
plugin-gulp | 36
webpack-loader | 46

AureliaEffect:
JS Framework Benchmark: https://26200-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://26214-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://26217-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

EisenbergEffect:
All merged @3cp 😄 You up for adding that associated view convention next by any chance?

3cp:
Yes, if that's urgent, I can do it first before updating aurelia/new skeletons.

EisenbergEffect:
How much time do you think it will take to update the skeletons? I think we should keep them up to date, but want to get an idea for timelines.

3cp:
I can update and test the skeletons with only shadowdom and css module in 2/3 days, and leave planed sass/less/stylus/postcss skeletons to another update.

PR #587, opened 2019-08-20
chore: partially fix renovate

BBosman:
# Pull Request

## 📖 Description

This should (partially) get Renovate back in line.

### 🎫 Issues

* #551 

## 👩‍💻 Reviewer Notes

This PR fixes two separate issues related to dependency updates:
* There is a newer version of `chromedriver`. The problem is that that package is directly tied to the version of Chrome installed on the machine. In this case the version of Chrome installed on the Docker image used on CircleCI. My solution is to pin `chromedriver` to the currently used version `^74.0.0` across the board. Note that this is a temporary solution, because when CircleCI decides to update its version of Chrome stuff will start breaking again. This would happen with or without `renovate` though...
* `test/js-framework-benchmark/webdriver-ts-results` is using version `3.3.3333` of `typescript`. Once you upgrade it to the current `3.5.3` (that the rest of the repo is already using) the build breaks due to casting errors. This PR introduces the necessary casts to have it build again.

## 📑 Test Plan

Trust CircleCI

## ⏭ Next Steps

Fix the remaining issues (if any) that block #551.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -33,6 +33,10 @@
       "allowedVersions": "^3.3.6"
     },
     {
+      "packageNames": ["chromedriver"],
+      "allowedVersions": "^74.0.0"
+    },
+    {
       "packageNames": ["karma"],
       "allowedVersions": "~4.1.0"
     },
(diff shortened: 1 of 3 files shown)


BBosman:
@fkleuver This is the follow up to your comment at https://github.com/aurelia/aurelia/pull/551#issuecomment-515811286. This time it isn't really Renovate's fault though. 😉 

Could you review/merge if you have time?

fkleuver:
Thanks a lot @BBosman looks good :)

PR #588, opened 2019-08-20
feat(default-replaceable): allow replace conv

brandonseydel:
# Pull Request

## 📖 Description

This allows projection of elements without defining default replacements when a containerless element is being used.

See below for usage examples.
```ts
@customElement({
    name: 'cool-button', containerless: true, template: `
<button>
<template replaceable />
</button
`})
export class CoolButton {
message = 'I am inside the coolest button on earth!';
}
```

```html
<template>

<!-- Straight content rendered inside button -->
<cool-button> I am still pretty cool since I live in some unknown realm
owned by the coolest button on earth </cool-button>
<!--
RENDERS
<button>I am still pretty cool since I live in some unknown realm
owned by the coolest button on earth</button>
-->

<!-- Render message from current controller-->
<cool-button>${message} </cool-button>
<!--
RENDERS
<button>I am the parent of the coolest button on earth!</button>
-->


<!-- Render message from within the cool button -->
<cool-button> <template>${message} </template> </cool-button>

<!--
RENDERS
<button>I am inside the coolest button on earth!</button>
-->

</template>
```

```ts

export class App{
message= 'I am the parent of the coolest button on earth!';
}


```

### 🎫 Issues

N/A
## 📑 Test Plan

Tests writen.
## ⏭ Next Steps

E2E tests.

### DIFF
--- a/packages/__tests__/jit-html/replaceable.spec.ts
+++ b/packages/__tests__/jit-html/replaceable.spec.ts
@@ -28,8 +28,8 @@ describe('replaceable', function () {
   ]) {
     it(`replaceable - ${title}`, function () {
 
-      const App = CustomElement.define({ name: 'app', template: `<template><foo>${appMarkup}</foo></template>` }, class {});
-      const Foo = CustomElement.define({ name: 'foo', template: `<template>${ceMarkup}</template>` }, class {});
+      const App = CustomElement.define({ name: 'app', template: `<template><foo>${appMarkup}</foo></template>` }, class { });
+      const Foo = CustomElement.define({ name: 'foo', template: `<template>${ceMarkup}</template>` }, class { });
 
       const ctx = TestContext.createHTMLTestContext();
       ctx.container.register(Foo);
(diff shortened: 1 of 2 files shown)


EisenbergEffect:
I see what this is trying to accomplish, but it's so similar to `slot` that it feels like duplicated functionality. Could we rather detect when there's a `slot` inside a containerless element and just make it work like this?

AureliaEffect:
JS Framework Benchmark: https://25066-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

bigopon:
I proposed something close to this, a legacy mode for the compiler to compile slot into replaceable. I think its the better default for folks like me who prefer the traditional dom

brandonseydel:
Using the slot idea definitely won't take a whole lot of effort.  Just have to force containerless before shadow and rename the au-replace.  Just didn't want the same element serving two purposes?  As long as documentation is on point or something should be ok?

brandonseydel:
This essentially allows the "transparent element" that users have been asking for quite some time without having all of the replace parts everywhere.

fkleuver:
@EisenbergEffect It's a feature gap. Either we need this, or slot emulation (and not in the form of shadydom polyfill). I think this is better.

EisenbergEffect:
I'd just prefer to use the element name `slot` since this is exactly emulating the default slot behavior but for a containerless element. Having two elements that do the same thing but with different names, and you have to remember which to use in which type of component, is confusing. The more special things we add, the harder things get to learn and remember. This could be expanded out to be a simple slot emulation for regular elements as well, but we'll want to limit it like this is so it's concerned only with default slots (no named slots, re-projection of slots, dynamic slots, etc).

fkleuver:
Oh, I was actually just talking about the default name for `replaceable` template controller. We could do without the custom element, no? @brandonseydel

brandonseydel:
Yes the element is just syntatic sugar.

brandonseydel:
Other variations that work the same
```html
<template replaceable /> works the same as <replace />
<template part replaceable /> &
<template part='default' replaceable ></template>
```

fkleuver:
Personally I agree with @EisenbergEffect in the sense that as a custom element it's too similar to `<slot>`, and it's also introducing 2 practically identical ways to do the same thing. I think we should be conservative with that, or at least consistent: if we add a custom element for `replaceable`, we should also have a custom element for `if`, `else`, `repeat`, etc.
If we go down that road, we'd better do that a bit later when things have stabilized and if there is indeed certain demand for it.

EisenbergEffect:
Introducing elements for those things would be such a massive breaking change. I'd prefer not to do that. I'm not sure there's a real benefit.

fkleuver:
I don't see the breaking change as they would be an addition instead of a replacement, and I suppose the benefit sits in the semantic exterior. Template controllers stand out more on a long list of attributes when they're elements that wrap the thing, instead of just another attribute.
But, that same benefit can be accomplished by just placing them on standalone `<template>` elements. So I agree that it's probably better not to

EisenbergEffect:
Even if it's additive, it makes a second way to do the same thing, which causes a problem with learnability and people who move across projects that use one or the other. So, I'd favor just keeping one way if we can.

fkleuver:
Yeah. @brandonseydel let's just remove the custom element?

brandonseydel:
Will do.

AureliaEffect:
JS Framework Benchmark: https://25158-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

PR #596, opened 2019-08-26
feat(router): configure to not use url fragment hash

jwx:
# Pull Request

## 📖 Description

This PR makes it possible to configure the router to not use url fragment hash (to "use push state").

```typescript
au.register(
    RouterConfiguration.customize({ useUrlFragmentHash: false }),
  )
```

Note that the web server need to be configured to handle "redirects to the right application page" for deep links and other browser url manipulations to work. 

### 🎫 Issues

## 👩‍💻 Reviewer Notes

## 📑 Test Plan

- Added two new tests that tests the configuration difference.

## ⏭ Next Steps

- Implement scope traversal to Router's `goto` method
- Finish navigation skeleton
- Review the extension points

### DIFF
--- a/packages/__tests__/router/browser-navigator.spec.ts
+++ b/packages/__tests__/router/browser-navigator.spec.ts
@@ -69,7 +69,7 @@ describe('BrowserNavigator', function () {
   it('can be activated', async function () {
     const { sut, tearDown, callback } = setup();
 
-    await sut.activate(callback);
+    await sut.activate({ callback });
 
     assert.strictEqual(sut['isActive'], true, `sut.isActive`);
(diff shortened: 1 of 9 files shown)


3cp:
Isn't `pushState: true` more intuitive and inline with v1?

jwx:
No, not really. `pushState` describes which History API is used and it has no actual connection to which part of the URL is used for navigation. The question is whether the URL fragment is used for navigation data or not. And since not using it (and using the URL path instead) requires configuration on the web server, using the URL fragment needs to be the default.

It's a breaking change from v1, sure, but since it's so small, and the configuration in itself is a breaking change, I don't think it'll be an issue.

PR #600, opened 2019-08-28
fix(di): defer should not register primitives

EisenbergEffect:
# Pull Request

## 📖 Description

A basic fix to the `Registration.defer` implementation that prevents primitive data values from being passed to the container when there's no handler. This is specifically for CSS view build scenarios where no special CSS handling is needed at runtime.


fkleuver:
@EisenbergEffect Shouldn't you throw instead of just filter? People need some kind of notification as to why stuff is not working

fkleuver:
I see, makes sense :)

PR #602, opened 2019-08-31
fix(ref): fix ref usage

bigopon:
# Pull Request

## 📖 Description

Currently, ref is specially handled, making it only capable on reference the element. Adjust binding command, compiler, and renderer to bring back feature parity with vCurrent, via following example:
```html
<div ref="div">
<div element.ref="div">
<custom-element view-model.ref="ceVm">
<div custom-attr.bind="..." custom-attr.ref="caViewModel">
<custom-element controller.ref="controller">
``` 

New in v2: Now ref can be placed before the referenced target:
```
element.ref === ref.element
view-model.ref === ref.view-model
```

### 🎫 Issues

N/A

## 👩‍💻 Reviewer Notes

N/A

## 📑 Test Plan

N/A

## ⏭ Next Steps

N/A


AureliaEffect:
JS Framework Benchmark: https://26274-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://26281-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://26289-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://26303-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://26313-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/jit-html/src/template-binder.ts
```diff
@@ -400,8 +414,18 @@ export class TemplateBinder {
...
       symbol.bindings.push(new BindingSymbol(command, attrInfo.bindable, expr, attrSyntax.rawValue, attrSyntax.target));
     }
-    this.manifest!.attributes.push(symbol);
-    this.manifest!.isTarget = true;
+    const manifest = this.manifest!;
+    const attributes = manifest.attributes;
+    let firstCustomAttrIndex = 0;
+    for (let i = 0, ii = attributes.length; ii > i; ++i) {
+      if ((attributes[i].flags & SymbolFlags.isCustomAttribute) > 0) {
+        firstCustomAttrIndex = i;
```
fkleuver: Shouldn't you break here?

packages/jit-html/src/template-binder.ts
```diff
@@ -400,8 +414,18 @@ export class TemplateBinder {
...
+      } else {
+        break;
+      }
+    }
+    attributes.splice(firstCustomAttrIndex, 0, symbol);
```
fkleuver: Maybe explain with a comment what's going on here? It's a bit confusing to me

packages/jit-html/src/template-binder.ts
```diff
@@ -114,14 +116,26 @@ export class TemplateBinder {
...
+        } else if (attrInfo.isTemplateController) {
+          throw new Error('Cannot have template controller on surrogate element.');
+          // TODO: use reporter
+        } else {
+          this.bindCustomAttribute(attrSyntax, attrInfo);
+        }
       } else {
-        this.bindCustomAttribute(attrSyntax, attrInfo);
+        // map special html attributes to their corresponding properties
+        attrSyntaxTransformer.transform(node, attrSyntax);
```
fkleuver: Isn't this coming from a previous pr? did you rebase?

bigopon: This is to fix surrogate compation, which i missed in previous PR

packages/jit-html/src/template-binder.ts
bigopon: We need to loop to find the last index of custom attr. Bad naming, will fix

bigopon: Should we have a separate array for custom attribute?

packages/jit-html/src/template-compiler.ts
```diff
@@ -312,8 +312,8 @@ export class TemplateCompiler implements ITemplateCompiler {
   }
 
   private compileAttribute(symbol: IAttributeSymbol): HTMLAttributeInstruction {
-    if (symbol.syntax.target === 'ref') {
-      return new RefBindingInstruction(symbol.syntax.rawValue);
+    if (symbol.syntax.command === 'ref') {
```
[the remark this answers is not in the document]
bigopon: Yes. I forgor this. Was wondering why compile method of ref command was not hit in the coverage

packages/runtime/src/lifecycle.ts
```diff
@@ -125,6 +125,10 @@ export interface IController<
 
 export const IController = DI.createInterface<IController>('IController').noDefault();
 
+export type ComponentHost<T extends INode = INode> = IRenderLocation<T> & T & {
```
fkleuver: I don't think this should be `IRenderLocation` because if it's not a containerless then the properties of it won't exist

packages/runtime/src/renderer.ts
```diff
@@ -172,6 +173,48 @@ export function getTarget(potentialTarget: object): object {
...
+  referenceName: string
+): void {
+  let $auRefs = host.$au;
+  if ($auRefs === void 0) {
+    $auRefs = host.$au = new ControllersLookup() as Record<string, IController<T>>;
+  }
+  $auRefs[referenceName] = controller;
+}
+
+class ControllersLookup {}
```
fkleuver:
```suggestion
class ControllersLookup<T extends INode> {
  public [key: string]: IController<T>;
}
```

Then you don't need the `as Record<string, IController<T>>` cast

packages/runtime/src/renderer.ts
```diff
@@ -197,6 +240,8 @@ export class CustomElementRenderer implements IInstructionRenderer {
       instruction as IElementHydrationOptions,
     );
 
+    setControllerReference(controller, controller.host as ComponentHost, instruction.res);
```
fkleuver: Wouldn't it be cleaner to do this inside the controller?

packages/jit-html/src/template-binder.ts
fkleuver: I'm not necessarily against that. Could you show me what you mean?

bigopon: At the moment, im trying to reorder a custom attribute to the start of the attributes, but it needs to respect the order of other custom attributes declared before it on this same element. The ordering process feels funny, and i think can simply be solved by having a separate array on the manifest for custom attributes instead. So when compiling a manifest, you only need to process that mew array first.

packages/jit/src/attribute-patterns.ts
```diff
@@ -17,15 +17,23 @@ export class DotSeparatedAttributePattern {
...
+  { pattern: 'PART.ref', symbols: '.' }
 )
 export class RefAttributePattern {
+  // v1 syntax
   public ['ref'](rawName: string, rawValue: string, parts: string[]): AttrSyntax {
-    return new AttrSyntax(rawName, rawValue, 'ref', null);
+    return new AttrSyntax(rawName, rawValue, 'element', 'ref');
   }
 
+  // v2 addition syntax
```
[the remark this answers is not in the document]
bigopon: 💔

packages/runtime/src/lifecycle.ts
bigopon: You can see in template controller renderer change that i added, it uses render location. So i think this typing is necessary. Maybe a union?

packages/runtime/src/renderer.ts
bigopon: I did, intially, though i ran to cases where attributes do not have host, because its not always passed in.
The look of static `Controller.forCustomAttribute` suggests we can create ca controller without a host, i think it could have potential usage, thus i moved it here

packages/__tests__/jit-html/template-compiler.ref.spec.ts
```diff
@@ -0,0 +1,516 @@
...
+        },
+        {
+          title: `basic WRONG ref usage with [ref.${refTarget}]`,
+          testWillThrow: true,
+          template: `<div ref.${refTarget}=hello>`
+        },
+      ] as IRefIntegrationTestCase[];
+    }),
+    {
+      title: `basic WRONG ref usage with [repeat.ref] as cannot reference template controller`,
```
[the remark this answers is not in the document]
bigopon: Its hard to cover all scenarios actually, i felt. Template controller are basically on a different element.

AureliaEffect:
JS Framework Benchmark: https://26469-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://26480-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://26502-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

PR #612, opened 2019-09-05
fix(plugin-conventions): upgrade modify-code to fix a preprocess bug

3cp:
@EisenbergEffect pls merge this fix.


AureliaEffect:
JS Framework Benchmark: https://26843-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

PR #614, opened 2019-09-06
refactor(plugin-conventions): simplify usage of html only element

3cp:
# Pull Request

## 📖 Description

As Rob suggested, simplified HTML only element by implementing `register(container)` method in html module.

## 📑 Test Plan

Passed manually tested in app.

## ⏭ Next Steps

Need to update makes skeleton after this release.

### DIFF
--- a/packages/__tests__/plugin-conventions/preprocess-html-template.spec.ts
+++ b/packages/__tests__/plugin-conventions/preprocess-html-template.spec.ts
@@ -11,11 +11,11 @@ export const template = "<template></template>";
 export default template;
 export const dependencies = [  ];
 let _e;
-export function getHTMLOnlyElement() {
+export function register(container) {
   if (!_e) {
     _e = CustomElement.define({ name, template, dependencies });
   }
-  return _e;
+  container.register(_e);
 }
 `;
     const result = preprocessHtmlTemplate({ path: path.join('lo', 'foo-bar.html'), contents: html }, preprocessOptions());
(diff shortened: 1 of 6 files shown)


AureliaEffect:
JS Framework Benchmark: https://26969-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

3cp:
@EisenbergEffect question, do we want to support IE10 or below? If so, I need to replace `const` and `let` with `var` in compiled html code.

In all our default bundlers setup, html only goes through conventions plugin, not ts/babel. All bundlers understand esm code (we generated for html), so there is no problem.

IE11 and all other browsers support `const` and `let`.

brandonseydel:
@3cp I would vote not.  Throughout all my consulting hours through various major companies they all either have IE11 or the ability to go to it and haven't.  Let's see what others think.

EisenbergEffect:
@3cp I think we can leave it as is. If we get someone who makes the request later on, we could even add an option for that.

EisenbergEffect:
This looks good. I'm going to merge. Glad it was a pretty easy to make the change. Not sure why I didn't think of this the first time.

EisenbergEffect:
Note I may hold to merge this until later today so I don't break someone who may be in the process of installing deps this morning. I don't want them to get my old code with the new webpack plugin (they are on dev releases). But, I will get this in shortly.

PR #616, opened 2019-09-07
fix(kernel): cover more edge cases in camel/kebabCase

3cp:
Update camel/kebabCase to cover more edge cases.
This new camel/kebabCase ensures we can map UAFoo.html and class UAFoo to `<ua-foo>`.

### DIFF
--- a/packages/__tests__/kernel/platform.spec.ts
+++ b/packages/__tests__/kernel/platform.spec.ts
@@ -5,6 +5,39 @@ import { _, assert } from '@aurelia/testing';
 
 const toString = Object.prototype.toString;
 
+const stringCases = [
+  ['FooBarBaz', 'fooBarBaz', 'foo-bar-baz'],
+  ['fooBar1Baz23', 'fooBar1Baz23', 'foo-bar1-baz23'],
+  ['foobarbaz', 'foobarbaz', 'foobarbaz'],
+  ['FOOBARBAZ', 'foobarbaz', 'foobarbaz'],
+  ['FOOBarBAZ', 'fooBarBaz', 'foo-bar-baz'],
+  ['FOOBarBaz', 'fooBarBaz', 'foo-bar-baz'],
+  ['fOObARbAZ', 'fOObARbAz', 'f-o-ob-a-rb-az'],
+  ['foo-bar-baz', 'fooBarBaz', 'foo-bar-baz'],
+  ['foo-123-baz', 'foo123Baz', 'foo-123-baz'],
+  ['Foo-Bar-Baz', 'fooBarBaz', 'foo-bar-baz'],
+  ['FOO-BAR-BAZ', 'fooBarBaz', 'foo-bar-baz'],
+  ['foo_bar_baz', 'fooBarBaz', 'foo-bar-baz'],
+  ['Foo_Bar_Baz', 'fooBarBaz', 'foo-bar-baz'],
+  ['FOO_BAR_BAZ', 'fooBarBaz', 'foo-bar-baz'],
+  ['foo.bar.baz', 'fooBarBaz', 'foo-bar-baz'],
+  ['Foo.Bar.Baz', 'fooBarBaz', 'foo-bar-baz'],
+  ['FOO.BAR.BAZ', 'fooBarBaz', 'foo-bar-baz'],
+  ['foo bar baz', 'fooBarBaz', 'foo-bar-baz'],
+  ['Foo Bar Baz', 'fooBarBaz', 'foo-bar-baz'],
+  ['FOO BAR BAZ', 'fooBarBaz', 'foo-bar-baz'],
+  ['_foo_bar_', 'fooBar', 'foo-bar'],
+  ['__foo__bar__', 'fooBar', 'foo-bar'],
+  ['-foo-bar-', 'fooBar', 'foo-bar'],
+  ['--foo--bar--', 'fooBar', 'foo-bar'],
+  ['.foo.bar.', 'fooBar', 'foo-bar'],
+  ['..foo..bar..', 'fooBar', 'foo-bar'],
+  ['*foo*bar*', 'fooBar', 'foo-bar'],
+  ['**foo**bar**', 'fooBar', 'foo-bar'],
+  [' foo bar ', 'fooBar', 'foo-bar'],
+  ['\t\tfoo\t\tbar\t\t', 'fooBar', 'foo-bar']
+];
+
 describe(`The PLATFORM object`, function () {
   if (typeof global !== 'undefined') {
     it(`global references global`, function () {
@@ -65,42 +98,17 @@ describe(`The PLATFORM object`, function () {
   });
 
   describe(`camelCase()`, function () {
-    for (const sep of ['.', '_', '-']) {
-      for (const count of [1, 2]) {
-        for (const prepend of [true, false]) {
-          const f = prepend ? 'F' : 'f';
-          for (const append of [true, false]) {
-            for (const [[foo, bar, baz], expected] of [
-              [['foo', 'bar', 'baz'], `${f}ooBarBaz`],
-              [['Foo', 'Bar', 'Baz'], `${f}ooBarBaz`],
-              [['FOO', 'BAR', 'BAZ'], `${f}OOBARBAZ`],
-              [['fOO', 'bAR', 'bAZ'], `${f}OOBARBAZ`],
-              [['foo', 'bar42', '42baz'], `${f}ooBar4242baz`]
-            ]) {
-              const actualSep = count === 1 ? sep : sep + sep;
-              let input = [foo, bar, baz].join(actualSep);
-              if (prepend) input = actualSep + input;
-              if (append) input += actualSep;
-              it(`${input} -> ${expected}`, function () {
-                const actual = camelCase(input);
-                assert.strictEqual(actual, expected, `actual`);
-                assert.strictEqual(camelCase(input), actual, `camelCase(input)`); // verify via code coverage report that cache is being hit
-              });
-            }
-          }
-        }
-      }
+    for (const [input, expected] of stringCases) {
+      it(`${input} -> ${expected}`, function () {
+        const actual = camelCase(input);
+        assert.strictEqual(actual, expected, `actual`);
+        assert.strictEqual(camelCase(input), actual, `camelCase(input)`); // verify via code coverage report that cache is being hit
+      });
     }
   });
 
   describe(`kebabCase()`, function () {
-    for (const [input, expected] of [
-      ['FooBarBaz', 'foo-bar-baz'],
-      ['fooBarBaz', 'foo-bar-baz'],
-      ['foobarbaz', 'foobarbaz'],
-      ['FOOBARBAZ', 'f-o-o-b-a-r-b-a-z'],
-      ['fOObARbAZ', 'f-o-ob-a-rb-a-z']
-    ]) {
+    for (const [input, _ignore, expected] of stringCases) {
       it(`${input} -> ${expected}`, function () {
         const actual = kebabCase(input);
         assert.strictEqual(actual, expected, `actual`);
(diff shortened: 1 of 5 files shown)


fkleuver:
The use case you mentioned could easily be fixed in the kernel with a small modification:

```ts
export function kebabCase(input: string): string {
  let value = kebabCaseLookup[input];
  if (value !== void 0) {
    return value;
  }
  value = '';
  let prevWasUpper = false;
  let char: string;
  let lower: string;
  const len = input.length;
  for (let i = 0; i < len; ++i) {
    char = input.charAt(i);
    lower = char.toLowerCase();
    if (char  === lower) {
      if (prevWasUpper) {
        value += '-' + char;
        prevWasUpper = false;
      } else {
        value += char;
      }
    } else {
      value += lower;
      prevWasUpper = true;
    }
  }
  return kebabCaseLookup[input] = value;
}
```

I think that's an overall improvement with negligible perf impact, and I'd much rather do that instead of dragging in the monstrous overkill and unstable dependency that is lodash 😆

3cp:
@fkleuver our kebabCase assumes the input is camel case string, while generic kebabCase should cover more. For example, string including separators like `foo_bar` `foo bar`. Our camelCase has less assumption on input.
The generic implementation like lodash covers much more edge cases. Let me add bit more (but not all) coverage to our version. There is some different behaviour with our version and generic version. For example, `kebabCase('foo1bar23')`, our version says `foo1-bar23`, generic version says `foo-1-bar-23`. I will keep our behaviour to avoid breaking change.

3cp:
@fkleuver question, `camelCase('_foo_bar')`, our version says `FooBar`, generic version says `fooBar`. Do we rely on this behaviour somewhere? If not, I would like to change it to follow generic behaviour.

fkleuver:
@3cp The way our `camelCase` and `kebabCase` behave is more or less the same as in v1. I might have removed some edge cases that didn't seem to make much sense to me.

Anyway, the tests are [here](https://github.com/aurelia/aurelia/blob/master/packages/__tests__/kernel/platform.spec.ts#L67) and [here](https://github.com/aurelia/aurelia/blob/master/packages/__tests__/kernel/platform.spec.ts#L96)

I'd say let's start with adding the cases we need to support, failing tests. Add some combinations or just loose cases if you prefer, that's up to you.

Here's the [lodash](https://github.com/lodash/lodash/blob/master/test/camelCase.test.js) tests. They seem a bit limited tbh

3cp:
@fkleuver I believe we don't rely on `kebabCase('_foo_bar') => 'FooBar'`. Indeed, I started with test cases I want to cover. I have committed those test cases, please have a look.

For the behaviour, I am going to treat all non-alphanumeric chars as separators (not just `.`, `-` and `_`).

AureliaEffect:
JS Framework Benchmark: https://27048-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://27054-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://27067-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

3cp:
The new implementation did a preflight on all chars, so it can support looking forward and backward (to determine words boundary).

codeclimate thinks my base method is too complex. I can break it into 2 methods, but I don't feel it's necessary.

AureliaEffect:
JS Framework Benchmark: https://27080-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

fkleuver:
I like the idea of generalizing with `baseCase`, the only problem with this flags approach is it won't work for non-ascii characters. The `toLowerCase` comparison with `charAt` I did was intentional for precisely that reason, otherwise we'd have to build a big lookup with the full Latin1 set like I did in the expression parser: https://github.com/aurelia/aurelia/blob/master/packages/jit/src/expression-parser.ts#L978.

Or you need to do a big juicy (slow) RegExp like lodash does.

If we don't, we might get some unhappy users from languages with accents or other letters, who want to use those special letters in their component names.

We can keep the general approach though, just need to use `toUpperCase()` and `toLowerCase()` to determine whether something is a non-separator.

There are also some tweaks:

- By using 3 variables that shift place, we can avoid creating the array (as well as avoid the extra loop to create it)
- Reuse a single function declaration instead of passing in a new arrow each time
- Use `const enum` instead of regular `enum` so the values are inlined
- Can avoid additive operation in kebabCase when not a separator

There are some other small tweaks I've wanted to make for a while. So I guess now is as good a time as any. I wanted to use an IIFE to encapsulate the lookups etc, to keep module scope a bit cleaner. So putting it all together would make it something like this:

```ts
const baseCase = (function () {
  const enum CharFlags {
    none  = 0,
    digit = 1,
    upper = 2,
    lower = 3,
  }

  const isDigit = Object.freeze(Object.assign(Object.create(null) as {}, {
    '0': true,
    '1': true,
    '2': true,
    '3': true,
    '4': true,
    '5': true,
    '6': true,
    '7': true,
    '8': true,
    '9': true,
  } as Record<string, true | undefined>));

  function charToFlags(char: string): CharFlags {
    const upper = char.toUpperCase();
    if (char !== upper) {
      return CharFlags.lower;
    }

    const lower = char.toLowerCase();
    if (char !== lower) {
      return CharFlags.upper;
    }

    if (isDigit[char] === true) {
      return CharFlags.digit;
    }

    return CharFlags.none;
  }

  return function (input: string, cb: (char: string, sep: boolean) => string): string {
    const len = input.length;
    if (len === 0) {
      return input;
    }
    
    let sep = false;
    let output = '';

    let prevFlags = CharFlags.none;

    let curChar = '';
    let curFlags = CharFlags.none;

    let nextChar = input.charAt(0);
    let nextFlags = charToFlags(nextChar);

    for (let i = 0; i < len; ++i) {
      prevFlags = curFlags;

      curChar = nextChar;
      curFlags = nextFlags;

      nextChar = input.charAt(i + 1);
      nextFlags = charToFlags(nextChar);

      if (curFlags === CharFlags.none) {
        if (output.length > 0) {
          sep = true;
        }
      } else {
        if (!sep && output.length > 0 && curFlags === CharFlags.upper) {
          sep = prevFlags === CharFlags.lower || nextFlags === CharFlags.lower;
        }

        output += cb(curChar, sep);
        sep = false;
      }
    }

    return output;
  };
})();

export const kebabCase = (function () {
  const lookup = Object.create(null) as Record<string, string | undefined>;

  function callback(char: string, sep: boolean): string {
    return sep ? '-' + char.toLowerCase() : char.toLowerCase();
  }

  return function (input: string): string {
    let output = lookup[input];
    if (output === void 0) {
      output = lookup[input] = baseCase(input, callback);
    }

    return output;
  };
})();

export const camelCase = (function () {
  const lookup = Object.create(null) as Record<string, string | undefined>;

  function callback(char: string, sep: boolean): string {
    return sep ? char.toUpperCase() : char.toLowerCase();
  }

  return function (input: string): string {
    let output = lookup[input];
    if (output === void 0) {
      output = lookup[input] = baseCase(input, callback);
    }

    return output;
  };
})();
```

The above is untested, so I can't guarantee the tests will still pass. We also need some extra tests for unicode characters and other strange edge cases like empty names, only-separators, etc. Can you work with this?

3cp:
@fkleuver thx. I certainly need more to cover non-ascii string. We need unit tests coverage even without new implementation.

But let's re-evaluate the situation. Because this is exactly why I decided not to touch internal camel/kebab in the very first version of this PR.

Our existing camelCase only treats `.` `-` `_` as separators, retains all other chars. Our existing kebabCase assumes input is a camel string. **These limitations are perfectly fine because internally we only use them to deal with js variable name and html attribute name.**

We can however only do minimum enhancement as you suggested before to just cover `UXFoo` use case. But it would not change the fact that this is still not generic enough for `plugin-conventions` who needs to deal with file names.

So it all comes back to my first question: why make internal camel/kebab generic but bigger and slower for something irrelevant for aurelia runtime consumption? To me, it's an unnecessary enhancement for resulting actually runtime penalty with a tiny benefit to share small piece of code to toolchain.

3cp:
Regarding your code, I think it's solid. I will test the new code. The question is whether we need to pay the price.

fkleuver:
> this is exactly why I decided not to touch internal camel/kebab in the very first version of this PR.

This is why I decided to help you out :)

> Our existing kebabCase assumes input is a camel string. These limitations are perfectly fine because internally we only use them to deal with js variable name and html attribute name.

The `kebabCase` and `camelCase` functions in the kernel were originally meant just for resource name handling, but just like `path` and a few other non-DI-specific concerns with a wide variety of uses, they ended up in the kernel because they're useful features even outside of the context of Aurelia as an SPA framework, just like DI is.

> why make internal camel/kebab generic but bigger and slower for something irrelevant for aurelia runtime consumption? To me, it's an unnecessary enhancement for resulting actually runtime penalty with a tiny benefit to share small piece of code to toolchain.

On one hand, you're right.

On the other hand, why keep its scope narrow and its use cases limited to runtime internals, when with a bit more code they can be useful general-purpose functions that plugin authors (like yourself :)) can utilize as well? They would now have fairly well-defined behavior.

In any case, I'd like to avoid dragging in lodash as a dependency. Lodash has a very old-fashioned (for backwards compatibility to who-knows-what) and thus bloated code base that deals with a gazillion edge cases. Even the simplest 1-liner manages to become 5 layers of helper functions.

If lodash solves a problem that's too time-consuming to solve yourself then by all means copy-paste the function and perhaps clean it up a little bit.
This helps keep our code base resilient against ecosystem changes, and easier to debug.

3cp:
Unfortunately, not that easy to support unicode :(

```js
    if (char !== char.toUpperCase()) {
      return CharFlags.lower;
    }

    if (char !== char.toLowerCase()) {
      return CharFlags.upper;
    }
```

This does not work for unicode like 变, the code thinks this is neither upper nor lower case. We only needs to treat those `%` `$` `@` as neither upper nor lower case, but treat 变 as an acceptable char (lower case maybe). It looks inevitable to have a white list or black list.

3cp:
Or I can use regex to find acceptable char, it should be more efficient than manual white/black list.

3cp:
Damn, I just learnt JS `\w` only matches ascii alphanumeric, unlike Perl `\w` which can match a "word" character (Perl default behaviour is ASCII-safe mode off).

fkleuver:
I know :)
With regex you run into the same problem. You'd need to have a fully-defined set of unicode ranges.
Lodash is doing this with a [monster regex](https://github.com/lodash/lodash/blob/master/.internal/unicodeWords.js)

Either way, the lookup has to happen somewhere. Be it in your code or internally in the browser. No shortcuts in unicode support :)

Hence the decompress trick in the [jit expression parser](https://github.com/aurelia/aurelia/blob/master/packages/jit/src/expression-parser.ts#L933). This is more than twice as fast as a monster regex.

But it depends on your design decision. If you feel that support for `变` in filenames is a must then I'd just copy paste lodash logic into your plugin (with attribution) - if they indeed support this as well - and be done with it.

3cp:
No, I don't think we need to support full unicode in conventions. I don't think users will complain either.

I was just thinking about generic design. Anyway your code passed existing tests. Let's use this version as a good enough implementation.

AureliaEffect:
JS Framework Benchmark: https://27119-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

3cp:
BTW, should we rename flags to flag?

fkleuver:
Sure. Or `CharKind`. Doesn't really matter much it's only internal

3cp:
Done.

AureliaEffect:
JS Framework Benchmark: https://27149-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

fkleuver:
@3cp is this ready?

3cp:
Yes

PR #617, opened 2019-09-07
fix(custom-attr): define parsing behavior clearer

bigopon:
# Pull Request

## 📖 Description

Make the compiler always default to parse as multiple bindings, except when it's clearly not multiple bindings. Cases where it's clearly not multiple bindings:
  * has binding command, ie: `<div my-attr.bind="...">`. In this scenario, the value of the custom attributes is required to be a valid expression
  * has no colon: ie: `<div my-attr="abcd">` In this scenario, it's simply invalid syntax. Consider style attribute rule-value pair: `<div style="rule: ruleValue">`

The following example are consider single binding:
```html
<div square="red">
<div square="5">
<div square.bind="diameter">
<div square.bind="diameter | toNumber:2">
<div square.bind="diameter & aa:bb">
<div square="${diameter}">
<div square="${diameter | aa}">
```
Note: for interpolation, it will be considered multiple bindings when there is a colon, either in string parts, or in binding behavior/value converter arguments. Examples
```html
<div square="${color | toColor:'rgb'}">
```

Which means if one wants to pass colon as literal to a custom attribute, which could be the case in URL value scenarios, then it's required to use a binding command/literal combo:
```html
<div click-logger.bind="'http://my.api'"> // it's `'http://my.api'` inside `""`
```

When it is a single binding, the template compiler should create an instruction against the custom attribute primary bindable property, or the first bindable property, or `'value'` property if there's no bindable declared.

The following are considered multiple bindings:
```html
<div square="color: red; diamter: 5">
<div square="color.bind: color; diamter: 5">
<div square="color: ${color}; diamter.bind: diameter">
<div square="color: ${color | toColor:'rbg'}; diamter.bind: diameter & numberBinder">
```

### `hasDynamicOptions`

One option is always treat custom attribute as dynamic by default, as if all the bindings need not to have matching bindables. And use `hasDynamicOptions: false` explicitly to turn this behavior off.

Another option is remove `hasDynamicOptions`

## 👩‍💻 Reviewer Notes

N/A

## 📑 Test Plan

TBU

## ⏭ Next Steps

Complete the parser to make the following work

```ts
  type IHasMultipleBindingParserTestCase = [/* attr value */string, /* has multiple bindings? */boolean];

  const testCases: IHasMultipleBindingParserTestCase[] = [
    ['color', false],
    ['${c}', false],
    ['http\\://abc.def', false],
    ['1111', false],
    ['1.11', false],
    ['${a | b:{c:b}}', false],
    ['${a & b:{c:b}}', false],
    ['${a & b:{c:b}} ${a & b:{c:b}}', false],
    ['a:b', true],
    ['a:a;b: b', true],
    ['a:1;b: 2', true],
    ['a.bind:1;b: 2', true],
    ['a:1; b.bind: 2', true],
    ['a:1 | c:d; b.bind: 2', true],
    ['a.bind:1 | c:d; b.bind: 2', true],
    ['a: ${a | c:d} abcd; b.bind: 2', true],
    ['a: http\\:/ahbc.def; b.bind: 2', true],
    ['a-a.bind: bcd; b-b5.bind: 2', true],
  ];
```


EisenbergEffect:
Two quick thoughts:

* If there's no real need for `hasDynamicOptions`, then lets just get rid of that.
* I want to make sure that `<div my-attr="abcd">` works if `my-attr` has multiple bindables. In this case, the primary bindable would be set to "abcd".

bigopon:
Should we get rid of `hasDynamicOptions`? It could be used to prevent dynamic binding to undeclared bindables

EisenbergEffect:
Let's just get rid of it. It can be added back if there's ever a real need.

AureliaEffect:
JS Framework Benchmark: https://27446-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

bigopon:
@fkleuver @EisenbergEffect ready for review. Basically just a removal of `hasDynamicOptions` with a todo for parsing the multiple binding usages.

bigopon:
@fkleuver please help rerun the firefox unit tests

AureliaEffect:
JS Framework Benchmark: https://27474-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

EisenbergEffect:
I may have missed it, but there should be a test that asserts that `square="red"` sets the `color` property of the square custom attribute to "red".

bigopon:
@EisenbergEffect It's here https://github.com/aurelia/aurelia/blob/01bfdf280ac290db7d50ec97dc5e7be8bd951b6a/packages/__tests__/jit-html/template-compiler.primary-bindable.spec.ts#L85-L110

EisenbergEffect:
Excellent. Thank you @bigopon !

AureliaEffect:
JS Framework Benchmark: https://28126-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://28135-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://29796-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://29816-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://29829-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/jit-html/src/template-binder.ts
```diff
@@ -507,6 +571,23 @@ export class TemplateBinder {
...
+  const len = rawValue.length;
+  let ch = 0;
+  for (let i = 0; i < len; ++i) {
+    ch = rawValue.charCodeAt(i);
+    if (ch === Char.Backslash) {
+      ++i;
+      // Ignore whatever comes next because it's escaped
+    } else if (ch === Char.Colon) {
+      return true;
+    } else if (ch === Char.Dollar && rawValue.charCodeAt(i + 1) === Char.OpenBrace) {
```
bigopon: This seems it will fail the following:
```html
<div square="color: ${color};">
```
But the tests are passing, I'm not sure why

fkleuver: @bigopon Because the colon came first

packages/__tests__/jit-html/has-multi-bindings.unit.spec.ts
```diff
@@ -0,0 +1,336 @@
...
+        {
+          target: 'a',
+          rawValue: '${a | c:d} abcd',
+        },
+        {
+          target: 'b',
+          rawValue: '2',
+        },
+      ],
+    },
```
bigopon: I think this test alone should give us enough confidence to go ahead, as it should cover 95 - 99 out of 100 cases at least.
Nicely done @fkleuver

EisenbergEffect:
Exciting to see this because I think this gets things working that didn't work correctly in au1. I'll pull the latest into my project on Monday and make sure all my stuff is working. Based on the tests, I think things should be fine though. Thanks for working together on this and making it happen. One step closer to alpha 🎆

AureliaEffect:
JS Framework Benchmark: https://29853-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

Issue #622, opened 2019-09-13
Local Resource Renaming


EisenbergEffect:
# 🙋 Feature Request

We need to support `<import as="foo" from="bar">` in order to support renaming for local disambiguation.

## 🤔 Expected Behavior

A resourced named bar should be registered as foo only for the view where it's imported as foo.

## 😯 Current Behavior

it is not possible to rename a resource upon import.

## 💁 Possible Solution

First we need to change resource registration from this:

```TypeScript
  Type.register = function register(container: IContainer): void {
      const key = CustomElement.keyFrom(description.name);
      Registration.transient(key, this).register(container);
      Registration.alias(key, this).register(container);
    };
```

To this:

```TypeScript
  Type.register = function register(container: IContainer, name?: string): void {
      const key = CustomElement.keyFrom(name || description.name);
      Registration.transient(key, this).register(container);
      Registration.alias(key, this).register(container);
    };
```

Then we need a helper function similar to this:

```TypeScript
export function renameResource(resourceModule, name) {
  return {
    register(container) {
      for (const key in resourceModule) {
        const value = resourceModule[key];
        if (value.register) {
          resource.register(container, name);
          return;
        }
      }
    }
  };
}
```

Note: Helper function above needs some modification to also handle if there's a register method directly on the module, as is the case with html only elements.

Then when you have this:

```HTML
<import as="bar" from="./another/foo">
```

You generate this:

```TypeScript
import * as d0 from './another/foo';
export const dependencies = [ renameResource(d0, 'bar'), /* ... */ ];
```

fkleuver:
I would actually propose updating `RuntimeCompilationResources.find` instead, so that if the passed-in name does not match the name on the description, it clones and renames the description before returning it. Since the `name` property of the description is used in quite a few places related to view factories and replaceable as well, doing it this way would ensure several obscure edge cases will work correctly as well. And there would be no need to update the `register` methods.

Something like this may do the trick:

```ts
  public find<TDef, TProto>(kind: IResourceKind<TDef, TProto>, name: string): ResourceDescription<TDef> | null {
    const key = kind.keyFrom(name);
    const resourceLookup = (this.context as unknown as { resourceLookup: Record<string, IResolver | undefined | null> }).resourceLookup;
    let resolver = resourceLookup[key];
    if (resolver === void 0) {
      resolver = resourceLookup[key] = this.context.getResolver(key, false);
    }

    if (resolver != null && resolver.getFactory) {
[...]

PR #625, opened 2019-09-18
WIP: runtime-renaming

brandonseydel:
# Pull Request

## 📖 Description

This allows runtime renaming to occur without conflict if two elements with the same name are used.

### DIFF
--- a/packages/kernel/src/resource.ts
+++ b/packages/kernel/src/resource.ts
@@ -26,7 +26,7 @@ export interface IResourceType<TDef, TProto, TClass extends ConstructableClass<T
 }
 
 export interface IResourceDescriptions {
-  find<TDef, TProto>(kind: IResourceKind<TDef, TProto>, name: string): ResourceDescription<TDef> | null;
+  find<TDef extends IResourceDefinition, TProto>(kind: IResourceKind<TDef, TProto>, name: string): ResourceDescription<TDef> | null;
   create<TDef, TProto>(kind: IResourceKind<TDef, TProto>, name: string): TProto | null;
 }
 
@@ -37,7 +37,28 @@ export class RuntimeCompilationResources implements IResourceDescriptions {
     this.context = context;
   }
 
-  public find<TDef, TProto>(kind: IResourceKind<TDef, TProto>, name: string): ResourceDescription<TDef> | null {
+  private cloneAndRenameTemplateDefinition<TDef extends IResourceDefinition>(definition: TDef, name: string): TDef {
+    const definitionClone: Partial<TDef> = {};
+    for (const key in definition) {
+      const value = definition[key];
+      if (Array.isArray(value)) {
+        definitionClone[key] = [...value] as any;
+        continue;
+      }
+
+      if (value instanceof Object) {
+        definitionClone[key] = { ...value };
+        continue;
+      }
+      
+      definitionClone[key] = value;
+    }
+
+    definitionClone.name = name;
+    return definitionClone as TDef;
+  }
+
+  public find<TDef extends IResourceDefinition, TProto>(kind: IResourceKind<TDef, TProto>, name: string): ResourceDescription<TDef> | null {
     const key = kind.keyFrom(name);
     const resourceLookup = (this.context as unknown as { resourceLookup: Record<string, IResolver | undefined | null> }).resourceLookup;
     let resolver = resourceLookup[key];
@@ -49,8 +70,17 @@ export class RuntimeCompilationResources implements IResourceDescriptions {
       const factory = resolver.getFactory(this.context);
 
       if (factory != null) {
-        const description = (factory.Type as IResourceType<TDef, TProto>).description;
-        return description === undefined ? null : description;
+
+        const { aliases, description } = (factory.Type as IResourceType<TDef, TProto> & IResourceDefinition);
+        if (description === void 0) {
+          return null;
+        }
+        if (description.name !== name &&
+          (aliases === void 0 || !aliases.includes(name)) &&
+          (description.aliases === void 0 || !description.aliases.includes(name))) {
+          return this.cloneAndRenameTemplateDefinition(description, name);
+        }
+        return description;
       }
     }
 
--- a/packages/plugin-conventions/src/preprocess-html-template.ts
+++ b/packages/plugin-conventions/src/preprocess-html-template.ts
@@ -1,4 +1,4 @@
-import { kebabCase } from '@aurelia/kernel';
+import { kebabCase, Registration } from '@aurelia/kernel';
 import modifyCode, { ModifyCodeResult } from 'modify-code';
 import * as path from 'path';
 import { IFileUnit, IPreprocessOptions } from './options';
@@ -31,10 +31,8 @@ export function preprocessHtmlTemplate(unit: IFileUnit, options: IPreprocessOpti
 
   const viewDeps: string[] = [];
   const statements: string[] = [];
-  let registrationImported = false;
+  let printRegisterImport = false;
   let importAsImported = false;
-  const registrationStatement = `import { Registration } from '@aurelia/kernel';\n`;
-
   // Turn off ShadowDOM for invalid element
   if (!name.includes('-') && shadowMode) {
     shadowMode = null;
@@ -48,19 +46,16 @@ export function preprocessHtmlTemplate(unit: IFileUnit, options: IPreprocessOpti
     const ext = path.extname(d.from);
     if (ext && ext !== '.js' && ext !== '.ts' && !options.templateExtensions.includes(ext)) {
       // Wrap all other unknown resources (including .css, .scss) in defer.
-      if (!registrationImported) {
-        statements.push(registrationStatement);
-        registrationImported = true;
-      }
       const isCssResource = options.cssExtensions.indexOf(ext) !== -1;
       let stringModuleId = d.from;
-
+      printRegisterImport = true;
       if (isCssResource && shadowMode && options.stringModuleWrap) {
         stringModuleId = options.stringModuleWrap(d.from);
       }
       statements.push(`import d${i} from ${s(stringModuleId)};\n`);
       viewDeps.push(`Registration.defer('${isCssResource ? '.css' : ext}', d${i})`);
     } else {
+
       if (d.resourceName == null) {
         statements.push(`import * as d${i} from ${s(d.from)};\n`);
       }
@@ -70,23 +65,16 @@ export function preprocessHtmlTemplate(unit: IFileUnit, options: IPreprocessOpti
 
       if (d.as == null) {
         viewDeps.push(`d${i}`);
+        return;
       }
-      else {
-        if (registrationImported && !importAsImported) {
-          const statementIndex = statements.indexOf(registrationStatement);
-          statements[statementIndex] = statements[statementIndex].replace('import { Registration }', 'import { Registration, importAs }');
-          registrationImported = true;
-        }
-
-        if (!registrationImported) {
-          statements.unshift(`import { Registration, importAs } from '@aurelia/kernel';\n`);
-          registrationImported = true;
-        }
-        importAsImported = true;
-        viewDeps.push(`importAs(\'${d.as}\',d${i})`);
-      }
+      importAsImported = true;
+      viewDeps.push(`importAs(\'${d.as}\',d${i})`);
     }
   });
+
+  if (printRegisterImport || importAsImported) {
+    statements.unshift(`import { Registration${importAsImported ? ', importAs' : ''} } from '@aurelia/kernel';\n`);
+  }
 
   const m = modifyCode('', unit.path);
   m.append(`import { CustomElement } from '@aurelia/runtime';\n`);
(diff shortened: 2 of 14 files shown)


AureliaEffect:
JS Framework Benchmark: https://27913-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

brandonseydel:
#622

AureliaEffect:
JS Framework Benchmark: https://27936-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/kernel/src/resource.ts
```diff
@@ -37,7 +37,28 @@ export class RuntimeCompilationResources implements IResourceDescriptions {
     this.context = context;
   }
 
-  public find<TDef, TProto>(kind: IResourceKind<TDef, TProto>, name: string): ResourceDescription<TDef> | null {
+  private cloneAndRenameTemplateDefinition<TDef extends IResourceDefinition>(definition: TDef, name: string): TDef {
+    const definitionClone: Partial<TDef> = {};
+    for (const key in definition) {
+      const value = definition[key];
+      if (Array.isArray(value)) {
+        definitionClone[key] = [...value] as any;
```
brandonseydel: May need deeper clone

AureliaEffect:
JS Framework Benchmark: https://28005-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/plugin-conventions/src/preprocess-html-template.ts
```diff
@@ -44,25 +45,46 @@ export function preprocessHtmlTemplate(unit: IFileUnit, options: IPreprocessOpti
...
       statements.push(`import d${i} from ${s(stringModuleId)};\n`);
       viewDeps.push(`Registration.defer('${isCssResource ? '.css' : ext}', d${i})`);
     } else {
-      statements.push(`import * as d${i} from ${s(d)};\n`);
-      viewDeps.push(`d${i}`);
+      if (d.resourceName == null) {
+        statements.push(`import * as d${i} from ${s(d.from)};\n`);
+      }
+      else {
+        statements.push(`import {${d.resourceName} as d${i}} from ${s(d.from)};\n`);
```
3cp: It was not desided yet, but I was proposing using resource name, not export name. But resource name is more relevant to end users. It means we still need star export here, then use a helper to find out which export is the resource with that resource name.

packages/plugin-conventions/src/preprocess-html-template.ts
```diff
@@ -44,25 +45,46 @@ export function preprocessHtmlTemplate(unit: IFileUnit, options: IPreprocessOpti
...
+      }
+
+      if (d.as == null) {
+        viewDeps.push(`d${i}`);
+      }
+      else {
+        if (registrationImported && !importAsImported) {
+          const statementIndex = statements.indexOf(registrationStatement);
+          statements[statementIndex] = statements[statementIndex].replace('import { Registration }', 'import { Registration, importAs }');
```
3cp: We might need to cleanup how statements was generated.

The other lazy thing we can do is always importing `importAs`. Those minifiers in production build step can remove unused code.

packages/plugin-conventions/src/preprocess-html-template.ts
```diff
@@ -44,25 +45,46 @@ export function preprocessHtmlTemplate(unit: IFileUnit, options: IPreprocessOpti
...
+          registrationImported = true;
+        }
+
+        if (!registrationImported) {
+          statements.unshift(`import { Registration, importAs } from '@aurelia/kernel';\n`);
+          registrationImported = true;
+        }
+        importAsImported = true;
+        viewDeps.push(`importAs(\'${d.as}\',d${i})`);
```
3cp: When there is an "as" but no resource name, `d${i}` is from a star import. I don't think your `importAs` can handle that. We need another helper to find the first resource to feed to importAs.

The other case is when `d${i}` is a from a HTML only element. that `d${i}` contains no resource but a `.register()` method. Need to find a way how to use `importAs` on that. (BTW, @EisenbergEffect's original idea on `.register()` is easier to cover this case)

packages/plugin-conventions/src/strip-meta-data.ts
```diff
@@ -103,9 +108,13 @@ function stripAttribute(node: DefaultTreeElement, tagName: string, attributeName
...
 // <import from="./foo"></import>
 // <require from="./foo"></require>
-function stripImport(node: DefaultTreeElement, cb: (dep: string | undefined, ranges: [number, number][]) => void) {
+function stripImport(node: DefaultTreeElement, cb: (dep: IDependency | undefined, ranges: [number, number][]) => void) {
   return stripTag(node, ['import', 'require'], (attrs, ranges) => {
-    cb(attrs.from, ranges);
+    cb({
+      from: attrs.from,
+      as: attrs.as,
+      resourceName: attrs['resource-name']
```
3cp: Maybe we should make some decision on this attribute name? How about `name` or `resource` which is shorter than `resource-name` :)

packages/plugin-conventions/src/preprocess-html-template.ts
bigopon: I second what @3cp said. In v1, because of loader, i guess it makes sense to use export name, and thats why we could only build app via module. Later the static convention uses resource names only, and I think its better and consistent

packages/plugin-conventions/src/preprocess-html-template.ts
3cp: @EisenbergEffect any idea on the issue around HTML only element? ☝️

3cp:
Maybe we should use term `rename` not `alias` on this feature? In order to avoid confusion with the `alias` feature on customElement.

packages/plugin-conventions/src/preprocess-html-template.ts
EisenbergEffect: Agree on using the resource name.

packages/plugin-conventions/src/strip-meta-data.ts
EisenbergEffect: `name` seems good to me.

packages/plugin-conventions/src/preprocess-html-template.ts
EisenbergEffect: Perhaps, if `resource` is not defined, we assume that it's a custom element resource and use that to create the key from the name. If we do that, we'll have to move this helper out of kernel.

3cp: If there is a helper can hide all details, that will be great for conventions.

```js
smartImportAs(
  'new-resource-name',
  resource_or_namespace,
  'optional-original-resource-name'
)
```

The 2nd argument can be
* a resource
* or a namespace containing one/more resources
* or a namespace/object containing `.register()` method. (not sure how you can handle this, because the real resource is inside `.register()` call)

The 3rd argument is only needed for namespace with multiple resources.

3cp: > if resource is not defined, we assume that it's a custom element resource and use that to create the key from the name

Technically, Aurelia 2 doesn't require any similarity between resource name and export name. You cannot figure out export name simply from resource name.

packages/plugin-conventions/src/preprocess-html-template.ts
brandonseydel: I am not sure on the resource name.

1. Intellisense/refactor 
2. Generated code will make no sense logically what happened.
3. Manually coding it you would import the element by export which this is a replacement for.
4. Two elements with the same resource name can exist.  How to know which one?

packages/plugin-conventions/src/preprocess-html-template.ts
brandonseydel: I’ll just store the index rather than bool when added.  Then I will replace with full text rather than replace.

packages/plugin-conventions/src/preprocess-html-template.ts
3cp: > Two elements with the same resource name can exist. How to know which one?

I don't understand how can you have the situation.

Don't forget we are designed a meta language in html template here.
User doesn't care about the generated js code. In html template, users think about resource name (like `<foo-bar>` element`), not js export name.

My argument was we cannot force users to know how to import this 3rd party resource which is perfectly legit in Aurelia 2:
```
@customElement({name: 'foo-bar', ...})
export class Lorem {}
```

brandonseydel: Lorem2 in the same file with foo-bar name?

3cp: I mean it's hard for users to know to write

```html
<import name="Lorem" as="foo-bar2" from './some-file'>`
```

Users are more likely to understand
```html
<import name="foo-bar" as="foo-bar2" from './some-file'>`
```

brandonseydel: the File name alone will be difficult enough if your that far you cannot just look at the file?  I think latest pr would already goto if you f12 that file.

3cp: Sure this is not an issue when the imported file is written by the user him/herself.

When the imported file is from a 3rd party lib, even the source file might be a minified file. 3rb party lib is more likely having documentation on resource names rather than export names.

Again, this issue is only for file with multiple exports. Using this renaming feature is rare enough, using renaming on file with multiple exports is even rarer.

brandonseydel: I don’t think this will be true.  Authors will show how to use by registering globally or in the dependency array on the custom element.  There is just absolutely no way this can work correctly with resource name.  A barrel export will become an absolute disaster.

3cp: > There is just absolutely no way this can work correctly with resource name.

That was what I wondered. If there is no easy API to inspect resource name, it's hard to do.

3cp: BTW, currently all imports in our HTML module are star (barrel) import. It relies on DI to screen all exports of one namespace to find registrable resources.

brandonseydel: Let’s see what others think.  Contextually it makes sense because we are in html, but we cannot guarantee anything with resource name.

packages/plugin-conventions/src/preprocess-html-template.ts
EisenbergEffect: I'm not sure what I was thinking. Yeah, HTML-only elements are a problem. I'm not sure how to handle that without adding the optional key param to the register method. Was there really an issue with doing that?

3cp: FP will hurt your brain :-)

```js
import * as d0 from './foo.html';

export const dependencies = [ renameHtmlOnlyELement(d0, 'bar') ];

function renameHtmlOnlyELement(d0, newName) {
  return {
    register(container) {
      return d0.register({
        register(resource) {
          return importAs(newName, resource);
        }
      });
    }
  }
}
```

EisenbergEffect: 🤣

3cp: @EisenbergEffect see if this helper could work?

```js
function smartImportAs(d, newName, optionalOldName) {
  return {
    register(container) {

      if (d && typeof d.register === 'function') {
        // optionalOldName is irrelevant here.
        return d.register({
          register(resource) {
            return importAs(newName, d);
          }
        });
      }
      
      for (let key in d) {
        const obj = d[key];

        if (obj && typeof obj.register === 'function') {
          // not sure can we assume obj is a resource here.
          if (!optionalOldName || optionalOldName === someAPItoGetResourceName(obj)) {
            // this line is probably not right.
            return container.register(importAs(newName, d));
          }
        }
      }

    }
  }
}
```

3cp: Fixed few problems in the above code.

3cp: Probably we can find out whether `d` (and `obj`) is resource or not by testing `d.kind` and `obj.kind` is one of `CustomElement`, `BindingBehavior`, `ValueConverter`, `CustomAttribute` and `BindingCommand`. That means this helper should be in Aurelia code base because of too many references.

EisenbergEffect: Yes, I think we'd check for kind. We can put this in Aurelia, just not in kernal. I think it would need to go in runtime.

brandonseydel: obj.kind.keyFrom should suffice

AureliaEffect:
JS Framework Benchmark: https://28243-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

3cp:
Should we check in importAs first? We can design the wrapper and implement conventions later.

EisenbergEffect:
@3cp Yes, I think we need that in the runtime first.

fkleuver:
@3cp @brandonseydel What is the status on this / is this still relevant?

Issue #623, opened 2019-09-13
Do not jit register primitives in the container


fkleuver:
# 🐛 Bug Report

## 🤔 Expected Behavior

When accidentally registering a resource name instead of its type, or in some other variant when registering something that can be assumed to be a key (e.g. a parameter decorator), the container should throw instead of passing it through to jit register

## 😯 Current Behavior

If you register a string instance by accident, the string will be resolved. This is far more often than not, undesired behavior.

## 💁 Possible Solution

Add type guards in `jitRegister` that throws when a primitive or parameter decorator is passed on

## 🔦 Context

This happens primarily when you forget to register a resource and try to resolve it by string. The string is then registered as a type, which never makes sense.

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Aurelia          |
| Language         |
| Browser          |
| Bundler          |
| Operating System |
| NPM/Node/Yarn    |

Issue #615, opened 2019-09-06
Container/Component Hierarchy


EisenbergEffect:
# 🐛 Bug Report

The DI container hierarchy isn't properly matching the component hierarchy.

## 🤔 Expected Behavior

When traversing through component container parents, the parent container should match the component parent all the way up the chain.

## 😯 Current Behavior

The parent container is the container for the component where the child component is first used, not necessarily the actual parent.

## 💁 Possible Solution

Give container a unique id, so that they can be cached and associated with components.
When creating a child container, it can append its id to its parent's, thus creating an id that's unique for the hierarchy.

EisenbergEffect:
Notes on what we think is the better solution:

1. We need to fix up the hierarchy of containers when contexts are created. Here's a snippet from @fkleuver showing something similar to what we need to do in RenderingEngine.getViewFactory

Change this

```
let factory = this.viewFactoryLookup.get(definition);
    if (!factory) {
```

to this

```
let factory = this.viewFactoryLookup.get(definition);
    if (!factory || parentContext != null && parentContext.id !== factory.parentContextId) {
```

2. We need to only pass resources along from the root container

```
 public createChild(): IContainer {
    const config = this.configuration;
    const childConfig = { factories: config.factories, resourceLookup: Object.assign(Object.create(null), this.root.config.resourceLookup) };
    const child = new Container(childConfig);
    child.parent = this;
    return child;
  }

3. We need to set up root when createContainer is called and pass root along when createChild is called.
```

PR #627, opened 2019-09-19
fix(kernel): only propagate globally registered resources to child render contexts

fkleuver:
# Pull Request

## 📖 Description

Fixes https://github.com/aurelia/aurelia/issues/615 - see that issue for a description.

### 🎫 Issues

Fixes https://github.com/aurelia/aurelia/issues/615
Fixes https://github.com/aurelia/aurelia/issues/623
Needed for https://github.com/aurelia/aurelia/issues/622

## 👩‍💻 Reviewer Notes

- Shuffled a few things around in the container properties. Removed `config`. Constructor looks into its parent for what to copy, instead of `createChild` passing the info down, because no one is supposed to call the constructor directly anyway.
- Added `root` and `id` properties.
- Refactored `createRenderContext` into a class
- Added the necessary guards/logic to solve aforementioned issues

## 📑 Test Plan

There is just one big test right now that verifies the following:
- Globally registered resources are compiled+rendered anywhere in the tree
- Locally registered resources are only compiled+rendered in the declaring component, not in its children or siblings.
- Resources can still be resolved directly from the container from any level

## ⏭ Next Steps

- Might need a few more tests / edge cases specifically for the scenarios @EisenbergEffect ran into. @bigopon maybe you have some thoughts on this as well?
- Verify / integration test with https://github.com/aurelia/aurelia/issues/622


AureliaEffect:
JS Framework Benchmark: https://27988-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

EisenbergEffect:
Code looks correct based on our conversations. I'll push an integration test for the CSS modules bug.

EisenbergEffect:
@fkleuver I pushed a simple test addition to my styles code that asserts the css modules inheritance is working as expected.

AureliaEffect:
JS Framework Benchmark: https://28444-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://28461-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

fkleuver:
@EisenbergEffect What's missing is the integration with the rendering process, which is a pretty crucial moving part in the entirety of this functionality.
That test you added is useful in its own right, but could you please also add a full integration test that actually applies CSS modules somewhat in the way an end user would, and add an assertion with `getCalculatedStyle` that the correct style is applied? This full end-to-end scenario is important for preventing regressions.

AureliaEffect:
JS Framework Benchmark: https://28488-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

fkleuver:
@EisenbergEffect I'll go ahead and merge. Could you keep that integration test as described above, as a (not very urgent but preferably within a couple weeks) todo?

EisenbergEffect:
Definitely. Where is the best place to put it and do you have a "template" test I can use to grab the boilerplate from?

PR #628, opened 2019-09-20
fix(array-observer): fix splice edge case

fkleuver:
# Pull Request

## 📖 Description

Fixes cases where 0 or 1 arguments are passed into `splice` would not be observed properly.


AureliaEffect:
JS Framework Benchmark: https://28032-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

PR #632, opened 2019-09-24
fix(compose): define an infra for passing through bindings

### LINKED ISSUE #319: [Feature] Transparent Wrapper Components in Aurelia
# 🙋 Feature Request

You can use this like

```html
```

```html
```

image

## 🤔 Expected Behavior

## 😯 Current Behavior

## 💁 Possible Solution

image

## 💻 Examples

### LINKED ISSUE #422: [Bug] Compose Should Transfer Attributes to Composed Elements
# 🐛 Bug Report

## 🤔 Expected Behavior

Given the following HTML:

```XML
</template>
```

```xml
```

## 😯 Current Behavior

```xml
<my-space></my-space>
```

## 💁 Possible Solution

## 🔦 Context

### LINKED ISSUE #497: au-compose should pass through bindings and content
# 🐛 Bug Report

## 🤔 Expected Behavior

## 😯 Current Behavior

## 🔦 Context

## 💻 Code Sample

```HTML
  <div>Some content...</div>
</au-compose>
```

```HTML
  <div>Some content...</div>
</my-element>
```

bigopon:
# Pull Request

## 📖 Description

Define new config for custom element to be able to relay any non-bindable bindings put on its host for later use within its template.

### 🎫 Issues

Closes #422 
Closes #497
Closes #319 

Resolves https://discourse.aurelia.io/t/assign-all-new-attributes-to-specific-element-transparent-wrapper-components/1949

## 👩‍💻 Reviewer Notes

TBU

## 📑 Test Plan

TBU

## ⏭ Next Steps

Work out children/slot issues with `<compose/>`


AureliaEffect:
JS Framework Benchmark: https://28270-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

EisenbergEffect:
Thanks for kicking this off @bigopon . I'm thinking that we may need two ways to use this:

* Designate an element within the template to specifically relay commands to.
* Capture the commands and dynamically apply them as needed (au-compose)

Do you think we can accomplish this with the same mechanism?

bigopon:
@EisenbergEffect yes, those two ways of using it you mentioned is the reason why i went with this approach.

The first way is achievable via a custom attribute:
```ts
@customAttribute('relay-target')
export class RelayTarget {
  constructor(
    @IHydrateElementInstruction private readonly elementInstruction
    @INode private readonly element,
    @IRenderingEngine private readonly renderingEngine
  ) {
    renderingEngine.render(elementInstruction.instructions, element)
  }
}
```
Usage:
```html
<template>
  <input relay-target >
</template>
```

The second usage is the `<compose/>` element

AureliaEffect:
JS Framework Benchmark: https://28468-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://28548-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

fkleuver:
@bigopon Is this the same issue as https://github.com/aurelia/aurelia/pull/668 or is this different?

bigopon:
Yes, its the same, but older. We dont need to keep this

PR #641, opened 2019-09-30
chore: upgrade to typescript 3.6.3

BBosman:
# Pull Request

## 📖 Description

Upgrade `typescript` to `3.6.3` and fix resulting typing errors. This should unblock #604.

### 🎫 Issues

* #604 

## 👩‍💻 Reviewer Notes

I deduced the typings from nearby code and/or the given error messages, but a spot check would be nice.

## 📑 Test Plan

CircleCI.

## ⏭ Next Steps

n/a

### DIFF
--- a/packages/runtime/src/binding/ast.ts
+++ b/packages/runtime/src/binding/ast.ts
@@ -1164,11 +1164,11 @@ export class ForOfStatement implements IForOfStatement {
   }
 
   public count(flags: LifecycleFlags, result: ObservedCollection | number | null | undefined): number {
-    return CountForOfStatement[toStringTag.call(result)](result as any);
+    return CountForOfStatement[toStringTag.call(result)](result as never);
   }
 
   public iterate(flags: LifecycleFlags, result: ObservedCollection | number | null | undefined, func: (arr: Collection, index: number, item: unknown) => void): void {
-    IterateForOfStatement[toStringTag.call(result)](flags | LifecycleFlags.isOriginalArray, result as any, func);
+    IterateForOfStatement[toStringTag.call(result)](flags | LifecycleFlags.isOriginalArray, result as never, func);
   }
 
   public connect(flags: LifecycleFlags, scope: IScope, binding: IConnectableBinding, part?: string): void {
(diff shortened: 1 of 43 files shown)


packages/runtime/src/binding/ast.ts
```diff
@@ -1166,11 +1166,11 @@ export class ForOfStatement implements IForOfStatement {
   }
 
   public count(flags: LifecycleFlags, result: ObservedCollection | number | null | undefined): number {
-    return CountForOfStatement[toStringTag.call(result)](result as any);
+    return CountForOfStatement[toStringTag.call(result)](result as never);
```
fkleuver: The intersection between the parameter types of the different `CountForOfStatement` signatures results in `never`, but this is not actually the type we want here.

What we should do here is turn this into a `switch` statement and peel off the methods from `CountForOfStatement` into module-scoped functions (and just delete that object). Then the types stay sound and polymorphism is reduced, so it should perform better too. Does that make sense?

AureliaEffect:
JS Framework Benchmark: https://28865-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/runtime/src/binding/ast.ts
BBosman: @fkleuver Something like this?

AureliaEffect:
JS Framework Benchmark: https://28919-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/runtime/src/binding/ast.ts
fkleuver: @BBosman Yeah basically, just without the object to save the member expression. Since there's some reserved names in there, maybe put a dollar in front? e.g. `function $array(...)`, etc

AureliaEffect:
JS Framework Benchmark: https://28975-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

fkleuver:
@BBosman I meant this 😄 

```ts
 case '[object Array]': return $array(// ...
//                                          ^ not wrapped in object

// ...

function $array(// ...
// ^ not wrapped in object
```

AureliaEffect:
JS Framework Benchmark: https://29043-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

PR #642, opened 2019-09-30
chore: linting fixes

BBosman:
# Pull Request

## 📖 Description

Some `linting` fixes. All done through `--fix` with a manual spot check afterwards.

### 🎫 Issues

n/a

## 👩‍💻 Reviewer Notes

- Adding/removing some newlines.
- Adding/removing some semicolons.
- Removing superfluous `async` statements.
- Consistent casting style (use `x as type` instead of `as x<type>`), mostly related to Array and ReadOnlyArray
- Add `readonly` where appropriate.

## 📑 Test Plan

CircleCI

## ⏭ Next Steps

n/a

### DIFF
--- a/packages/testing/src/startup.ts
+++ b/packages/testing/src/startup.ts
@@ -1,7 +1,6 @@
 import { Constructable } from '@aurelia/kernel';
-import { TestContext } from '.';
 import { CustomElement, Aurelia } from '@aurelia/runtime';
-import { HTMLTestContext } from './html-test-context';
+import { HTMLTestContext, TestContext } from './html-test-context';
 
 export function setup<T>(template: string | Node,
     $class?: Constructable<T>,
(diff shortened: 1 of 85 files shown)


packages/testing/src/startup.ts
```diff
@@ -1,7 +1,7 @@
 import { Constructable } from '@aurelia/kernel';
-import { TestContext } from '.';
 import { CustomElement, Aurelia } from '@aurelia/runtime';
 import { HTMLTestContext } from './html-test-context';
+import { TestContext } from '.';
```
fkleuver: This was not intentional. I'd rather avoid single '.' dot paths like this

BBosman: @fkleuver Should be fixed.

AureliaEffect:
JS Framework Benchmark: https://28927-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

PR #645, opened 2019-10-01
WIP: feat(safe-mode): Empty instead of undefined

brandonseydel:
# Pull Request

## 📖 Description
This makes a default "safe" mode for aurelia that will print empty strings instead of undefined.
```ts
class {test}
```
```html
<div>${'1' + test}</div>
<!--strict mode = '1undefined' -->
<!--safe mode = '1' -->
```

## 📑 Test Plan
Need to create more tests for switching the flag back and forth from safe to strict.


3cp:
FYI, vcurrent expression did some special treatment to number +/- too. I noticed when I copy vcurrent parser to bcx-expression-evaluator project. Not sure if Aurelia 2 carried the same behaviour.

https://www.npmjs.com/package/bcx-expression-evaluator#default-result-for---operators

brandonseydel:
@3cp I was thinking about turning them to zero as well.  Do we want to port this over for sure.  @fkleuver @bigopon

bigopon:
I don't think it's an option. It's gonna get folks angry if not

brandonseydel:
@fkleuver 
  // note: autoConvertAdd (and the null check) is removed because the default spec behavior is already largely similar
  // and where it isn't, you kind of want it to behave like the spec anyway (e.g. return NaN when adding a number to undefined)
  // this makes bugs in user code easier to track down for end users
  // also, skipping these checks and leaving it to the runtime is a nice little perf boost and simplifies our code

brandonseydel:
This is OK?

${message + message2} === 0 in au1 when those could have been implied strings?

```ts
export class App {
  message:string = null;
  message2:string = null;
}
```

3cp:
I tested vcurrent aurelia-binding v2, it still carried the old aurelia-binding v1 behaviour.
Then I tested Aurelia 2 `@aurelia/binding`, it didn't carry this behaviour, `undefined + 1` now yields NaN which is same as JavaScript. This must be a conscious decision from @fkleuver when he moved from aurelia-binding v2 to `@aurelia/binding`.

@bigopon I think there is not much to worry about angry users. This is an edge case rarely depended by user apps. It might have some convenient use cases, but my brain could not find out any right now.

I was only wondering if Aurelia 2 behaves the same.

bigopon:
Yes thanks. Maybe its a reasonable breakage that we need

fkleuver:
Let's put it like this. Safe mode equals "behaves like v1". We could also call it legacy mode then, to make this clearer?
So that would include the number conversion thing.

brandonseydel:
Moved to #692 - Rebase went haywire.

PR #647, opened 2019-10-02
docs: update doc to use latest "aurelia" package

3cp:
Also added two re-exports `ISignaler` and `HTMLDOM` in "aurelia" package because i18n sample code uses them.


AureliaEffect:
JS Framework Benchmark: https://29263-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

PR #648, opened 2019-10-02
Remove incorrect migration step

rluba:
Injecting `I18nService` causes multiple instances to be created. (See Discord discussion in #i18n from 2019-10-02)


EisenbergEffect:
Pinging @Sayan751 for review.

Sayan751:
@rluba May I ask how you are currently injecting `I18N`? Also as you said in your description about the multiple instances, if I am not mistaken, you missed registering `I18nConfiguration`.

rluba:
Hi @Sayan751,
> in your description about the multiple instances, if I am not mistaken, you missed registering `I18nConfiguration`.

No, that was a different issue. When I forgot to register `I18nConfiguration`, it simply failed completely with an error 16 (which currently has the wrong error text).

> how you are currently injecting `I18N`?

At first I tried:
```TypeScript
import {I18nService} from '@aurelia/i18n';

export class SomeState {
	constructor(public i18n: I18nService) {
	}
}
```
which caused two instances of `I18nService` to be created – and therefore `i18next` to be initialized twice.

When I switched back to…
```TypeScript
import {I18N} from '@aurelia/i18n';

export class SomeState {
	constructor(public i18n: I18N) {
	}
}
```
… the problem disappeared.

rluba:
I just noticed that I get a TS error in SublimeText on `constructor(public i18n: I18N)`, that I didn’t get yesterday, for whatever reason: `I18N refers to a value but is used as a type here.`

But it works nonetheless. 🤔
I’ve seen that `TranslationValueConverter` uses `constructor(@I18N i18n: I18nService)` but that seems more complex than it should be.

Why should we require users to reference both `I18N` and `I18nService`?

Sayan751:
@rluba That error is indeed expected. As of now there is no `@autoinject`. The other alternative is to use the `@inject`, which I am not a fan of for obvious reason. Therefore to avoid confusion, I would like to keep that line as-is. I think the current recommendation is there in the docs.

Sayan751:
@rluba I have updated the i18n package and now there is no need to explicitly refer `I18NService`. Check this PR: https://github.com/aurelia/aurelia/pull/708. Closing this in favor of that.

PR #649, opened 2019-10-02
Fix/i18n interpolation changes

rluba:
## 📖 Description

i18n doesn’t handle updates to the `t-params` binding correctly. I’ve found and fixed two issues and written tests for them.

## 👩‍💻 Reviewer Notes

~This PR already contains a fix for `works for interpolation when the interpolation changes`, but the second test (`works for interpolation when a string changes`) is still broken.

The problem is that `TranslationBinding` cannot distinguish whether its `expr` or its `parametersExpr` has changed. It uses the same change handler for both and believes that it can distinguish the events based on the `newValue` type (which is incorrect).

The solution would be to [pass a different change handler than `this` to `parametersExpr.connect(…)`](https://github.com/aurelia/aurelia/blob/ad629adbf442841821345417da00874960049cd8/packages/i18n/src/t/translation-binding.ts#L101). But I don’t know what’s the best/cleanest way to create such a second change handler.~

I’ve applied @fkleuver’s (via Discord) suggested fix for the second issue: We’ve added a new persistent flag to distinguish the two observer cases.

There’s still the pending issue with the missing `part` parameter to the `evaluate` call (see ToDo).


Sayan751:
@rluba The changes looks fine to me. 

@fkleuver I have couple of questions:
1. In which condition the `part` in `$bind()` is truthy? I think for some specific bindings like multi-interpolation (and subsequently interpolation), it is truthy, but in general when I should expect that to be truthy? It seems for `translation-binding` it is always undefined. From that perspective, the usages of `part` can be removed from `$bind()`. What is your take on this?
2. As of now the "Subscriber"s are notified only via `handleChange`, and it seems that registering a custom handler for a specific expression might be quite a big change. Which might be an overkill if multi-expression connectables are not common.
3. However, it seems that the change notification has very limited context. For example, for an expression (let's call this root), composed of multiple expressions, the subscriber gets notified for the delta. Which is fine, but then it also need to evaluate the root expression once again to get the complete info, because only from the `newValue` it is difficult to determine which part of the expression changed. Example: multi-interpolation expression, or object literal expression. It would be nice to either put more context in the change notification. Or it might be easier to provide the reevaluated value for multi-expression collection scenario. Then the root expression can register itself as subscriber for the child expressions, and notify the original subscriber after accumulating the changes, and/or reevaluating the expression.

fkleuver:
@Sayan751 

1. When there is a `replaceable` below
2. Agree, though the current solution is not very pretty either
3. This is a general issue in Aurelia's binding, that you can't easily know which part of an expression the change is coming from, and also causes some inefficiencies. I intend to solve this problem in the future when I have time for it, by completely changing the flags system

fkleuver:
ty for discovering and helping out on this @rluba

PR #656, opened 2019-10-03
chore: resolve e2e nesting issues

BBosman:
# Pull Request

## 📖 Description

As discussed on Discord (with @fkleuver) I'm seeing some issues (with `eslint`) because we have a nested `lerna` package. Inside `__tests__` we have `e2e`. Although we need to think about how we want to handle  `cypress` in the future (the reason `e2e` isn't part of `__tests_`) for now we move `__tests__/e2e` to it's own top level `__e2e__` directory, to resolve the nesting issues.

I also noticed that the recently merged `aurelia` meta-package was still using an older version of `typescript` and also still using `tslint`, breaking `lerna` hoisting, so I took the liberty to fix that as well.

And lastly, the `scripts` in `package.json` of `__tests__` had a duplicate `test-chrome:verbose` entry, so I removed one.

### 🎫 Issues

n/a

## 📑 Test Plan

CircleCI


AureliaEffect:
JS Framework Benchmark: https://29555-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://29607-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

PR #658, opened 2019-10-04
feat(router): add au-href

jwx:
# Pull Request

## 📖 Description

This PR adds the `au-href` attribute as an alternative to `href` and `route-href`. The presence of an `au-href` attribute on an anchor will override any `href` attribute. If there's no `au-href` on the anchor, the router option `useHref` decides whether the router should use the `href` attribute instead.

```html
<a au-href="foo(bar)">foo with bar</a>
<a au-href.bind="'foo(bar)'">foo with bar</a>
<a au-href.bind="{ component: 'foo', parameters: 'bar' }">foo with bar</a>
<a au-href.bind="{ component: Foo, parameters: 'bar' }">foo with bar</a>
```
The last example currently requires that the `Foo constructor` is known within the view model. For example
```typescript
import { Foo } from './foo';

export class MyComponent {
  private readonly Foo = Foo;
}
```

### 🎫 Issues

## 👩‍💻 Reviewer Notes

## 📑 Test Plan

- Add new tests for `au-href` vs `href`
- Add new tests for proper navigations with `au-href`

## ⏭ Next Steps

- Add first version of configuration
- Review parameters in viewport instructions
- Fix `realworld`
- Fix proper sync/async for lifecycle hooks
- Finish navigation skeleton
- Review the extension points


AureliaEffect:
JS Framework Benchmark: https://29680-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://29969-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://29979-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://30046-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://30085-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://30142-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://30157-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://30181-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://30210-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/router/src/link-handler.ts
```diff
@@ -62,19 +84,16 @@ export class LinkHandler {
...
+    if ((auHref === null || auHref.length === 0) && (href === null || href.length === 0)) {
       return info;
     }
 
     info.anchor = target;
-    info.href = href;
+    info.instruction = auHref || href;
 
-    const leftButtonClicked = (event as MouseEvent).which === 1;
+    const leftButtonClicked: boolean = (event as MouseEvent).which === 1;
```
fkleuver: Couple things here. `MouseEvent.which` is not standardized (see https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/which) and so `MouseEvent.button` (see https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/button) would be preferred.

Also, this cast to `MouseEvent` shouldn't be necessary. Just make the parameter type `MouseEvent` instead of `Event` because on a `"click"` listener you're guaranteed to get `MouseEvent`. I would change this signature down below:

```ts
private readonly handler: EventListener = (e) => {
```

To this:

```ts
private readonly handler: (e: MouseEvent) => void = (e) => {
```

Then the event type is correct from the start.

As for `which`, better do this instead:

```ts
const leftButtonClicked = event.button === 0;
```

AureliaEffect:
JS Framework Benchmark: https://30247-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/router/src/link-handler.ts
jwx: Fixed. (The change from `which` also made node tests pass again.)

AureliaEffect:
JS Framework Benchmark: https://30253-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

PR #659, opened 2019-10-04
Add several failing tests for attributes with multiple bindable properties

rluba:
# Pull Request

@fkleuver, as discussed in Discord: Aurelia doesn't handle attributes with multiple `@bindable` properties correctly.

## 📖 Description

This PR adds several tests that currently fail, but shouldn’t.


EisenbergEffect:
Pinging @bigopon because I believe he's got a PR open where he has addressed this or is in the process of addressing it.

bigopon:
This should be fixed by #617

bigopon:
@rluba we have fixed the core issue related to multiple bindings and custom attribute, but your tests will still fail because you are expecting v1 behavior of reactivity by default (without `bind` lifecycle hook, change handler will be called during initialization).

We have a RFC here for this change #670 , please chime in.

rluba:
@bigopon Thank you. I’ve updated the tests.

PR #668, opened 2019-10-06
feat(capture-attrs): basic runtime infra for capturing attrs

### LINKED ISSUE #319: [Feature] Transparent Wrapper Components in Aurelia

### LINKED ISSUE #422: [Bug] Compose Should Transfer Attributes to Composed Elements

### LINKED ISSUE #497: au-compose should pass through bindings and content

bigopon:
# Pull Request

## 📖 Description

Define new config for custom element to be able to relay any non-bindable bindings put on its host for later use within its template.

### 🎫 Issues

Closes #422 
Closes #497
Closes #319 

Resolves https://discourse.aurelia.io/t/assign-all-new-attributes-to-specific-element-transparent-wrapper-components/1949

## 👩‍💻 Reviewer Notes

TBU

## 📑 Test Plan

TBU

## ⏭ Next Steps

Work out children/slot issues with `<compose/>`


AureliaEffect:
JS Framework Benchmark: https://29897-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://30191-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

bigopon:
For the following example:
```ts
@customElement({
  name: 'foo'
  captureAttrs: true,
})
export class Foo{}
```
```html
<foo class="hello" id="au" value.bind="barz"></foo>
```
should we capture `class`, `id` as well? or by default, only capture the attribute with bindings?

fkleuver:
What about:

```ts
@customElement({
  name: 'foo'
  captureAttrs: AttrFilter.all, // capture `class`, `id`, etc
})
export class Foo{}
```

```ts
@customElement({
  name: 'foo'
  captureAttrs: AttrFilter.bindingCommands, // only capture bindings
})
export class Foo{}
```

```ts
@customElement({
  name: 'foo'
  captureAttrs: AttrFilter.plain, // only capture non-bindings
})
export class Foo{}
```

bigopon:
That is nice looking config 👍

bigopon:
@fkleuver So it will be like this:

```ts
@customElement({
  name: 'foo'
  captureAttrs: AttrFilter | string[]
})
export class Foo{}
```

fkleuver:
@bigopon I meant this actually:

```ts
@customElement({
  name: 'foo'
  captureAttrs: AttrFilter | RegExp
})
export class Foo{}
```

It seems a bit simpler and more flexible to use a regexp for an api like this (one of the few exceptions I guess) but I don't feel strongly about it.

bigopon:
I think string is more intuitive as a config, and Regexp isn't always welcomed. Let's go with `string[]`, we can always add Regexp later

AureliaEffect:
JS Framework Benchmark: https://30382-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

brandonseydel:
I like the capture attrs.  I wonder though if this should include an object for include/exclude explicit.

If i wanted every attribute except style we would need to exclude it.

```ts

type CaptureAttrDef = {
include: AttrFilter | string | string[],
exclude: AttrFilter | string | string[],
}

captureAttrs: {
include: AttrFilter.All,
exclude: 'style'
}
```

bigopon:
I think exclude only has some meaning when it's used together with `AttrFilter.all` or a `RegExp`, for now, we don't have Regexp (yet) and `all`, well I'm not sure. Maybe it makes more sense than I thought. Could just do:
```ts
captureAttrs: {
  include: AttrFilter.All,
  exclude: ['style', 'class', 'id']
}
```

fkleuver:
This is starting to feel a bit like bike shedding though. I'm inclined to say let's stick to just the enum until someone asks for more flexibility.

EisenbergEffect:
As a quick FYI, for au-compose, we need the ability to transfer all attributes, both bound and unbound, to the composed element, with the exception of whatever bindables that au-compose defines for itself. e.g. subject

fkleuver:
> until someone asks for more flexibility.

->

> with the exception of whatever bindables that au-compose defines

I suppose someone has asked for more flexibility 😄

bigopon:
> Quick real use case: for au-compose, we need the ability to transfer all attributes, both bound and unbound, to the composed element, with the exception of whatever bindables that au-compose defines for itself. e.g. subject

Yes this was default behavior I wanted it to be, and I think it would be pretty confusing if not

AureliaEffect:
JS Framework Benchmark: https://30563-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://30885-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

PR #688, opened 2019-10-08
refactor(jit-html): cleanup template-binder and improve semantic-model types

fkleuver:
# Pull Request

## 📖 Description

- Simplified the `TemplateBinder` by passing the `manifest`, `manifestRoot`, etc via method parameters rather than storing&restoring them as class instance properties. Data flow should be easier to understand now. Also added some comments.
- Removed the symbol interfaces from the semantic model, using type unions of the classes instead. Also introduced node generic type arguments to make the types more accurate and require a little bit less casting.
- Cleaned up the resource model classes.
- Fixed a number of linting issues along the way.

### 🎫 Issues

Fixes https://github.com/aurelia/aurelia/issues/677

## 👩‍💻 Reviewer Notes

The semantic-model in `jit-html` looks a bit hacky, but this structure is needed to provide html-specific defaults for the generic type arguments. We might opt to get rid of those generics entirely and simply move the semantic model to jit-html at some point.

## 📑 Test Plan

No functional changes. Existing tests should still pass.

## ⏭ Next Steps

We could look into merging `TemplateBinder` and `TemplateCompiler` (@bigopon ran into some difficulties with `@processContent` caused by the separation of these concerns). No concrete plans for that yet, though. Perhaps it's not needed.

### DIFF
--- a/packages/jit-html/src/semantic-model.ts
+++ b/packages/jit-html/src/semantic-model.ts
@@ -1,0 +1,56 @@
+import {
+  AnySymbol as $AnySymbol,
+  BindingSymbol,
+  CustomAttributeSymbol,
+  CustomElementSymbol as $CustomElementSymbol,
+  ElementSymbol as $ElementSymbol,
+  LetElementSymbol as $LetElementSymbol,
+  NodeSymbol as $NodeSymbol,
+  ParentNodeSymbol as $ParentNodeSymbol,
+  PlainAttributeSymbol,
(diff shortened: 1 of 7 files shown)


AureliaEffect:
JS Framework Benchmark: https://30101-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

Issue #657, opened 2019-10-04
templateController on class ending in Something


rluba:
# 🐛 Bug Report

@3cp, as discussed in Discord:

This throws an (undefined) Error 50:
```TypeScript
@templateController('something')
export class SomethingCustomAttribute<T extends INode = INode> {
    constructor(@IViewFactory public viewFactory: IViewFactory<T>) {
    }
}
```

… while this works: 
```TypeScript
@templateController('something')
export class Something<T extends INode = INode> {
    constructor(@IViewFactory public viewFactory: IViewFactory<T>) {
    }
}
```

3cp:
@rluba pls test and see whether it works as expected.

Issue #630, opened 2019-09-23
[RFC] single npm package "aurelia" for simplified imports


3cp:
# 💬 RFC

This is a record of the decision made on discord few months ago, it will be implemented before alpha release.

## 🔦 Context

To provide single npm package `aurelia` (currently an empty npm package owned by @EisenbergEffect) which just re-exports all internals from `@aurelia/kernel`, `@aurelia/runtime`, `@aurelia/jit`, ...

```js
import { AnyThing } from 'aurelia';
```

Update: a new way to start app, to reduce boilerplate.

```ts
import Aurelia, { StyleConfiguration, RouterConfiguration } from 'aurelia';
import { MyRootComponent } from './my-root-component';
// By default host to element name (<my-root-component> for MyRootComponent),
// or <body> if <my-root-component> is absent.
Aurelia.app(MyRootComponent).start();

// Or load additional aurelia features
Aurelia
  .register(
    StyleConfiguration.shadowDOM(),
    RouterConfiguration.customize({ useUrlFragmentHash: false })
  )
  .app(MyRootComponent)
  .start();

// Or host to <my-start-tag>
Aurelia
  .register(
    StyleConfiguration.shadowDOM(),
    RouterConfiguration.customize({ useUrlFragmentHash: false })
  )
  .app({
    component: MyRootComponent,
    host: document.querySelector('my-start-tag')
  })
  .start();
```

Note the `au` is not a fixed name, it's using default export from `aurelia` package.

## 💻 Examples

So that users don't have to know where to import certain features, they don't have to write 
```js
import {...} from '@aurelia/runtime';
import {...} from '@aurelia/kernel';
```

It will be just `import {...} from 'aurelia'';`. This will simplify user code, and reduce the learning curve.

The downside is that we need to rely on tree-shaking (from various bundlers) to remove unused Aurelia 2 features.

We also need to cleanup existing export names to avoid naming conflicts. For example, there are multiple `BasicConfiguration` exported from few Aurelia 2 core packages. This is a small breaking change.

3cp:
The list of troublesome duplicated exports for review.

```js
export {
  Tracer as DebugTracer // duplicated in @aurelia/kernel
} from '@aurelia/debug';
export { 
  Scope as RouterScope,  // duplicated in @aurelia/runtime
  State as RouterState // duplicated in @aurelia/runtime
} from '@aurelia/router';
export {
  DOM as RuntimeHtmlDOM // duplicated in @aurelia/runtime
[...]

PR #694, opened 2019-10-09
Support templateController in conventions.

3cp:
# Pull Request

## 📖 Description

1. Add support of `@templateController` in conventions.
    * If you already decorated class with `@templateController`, conventions will skip the resource (even if there is `CustomAttribute` suffix in class name).
    * If you name resource as `export class FooBarTemplateController` without decorator, conventions will inject `@templateController('foo-bar')`.

2. Add check on existing import statement on new "aurelia" package to avoid double import.

### 🎫 Issues

#657
#630 

## 📑 Test Plan

Manually tested in a webpack app.

## ⏭ Next Steps

To be tested in user apps.

### DIFF
--- a/packages/__tests__/plugin-conventions/name-convention.spec.ts
+++ b/packages/__tests__/plugin-conventions/name-convention.spec.ts
@@ -60,4 +60,16 @@ describe('nameConvention', function () {
       type: 'bindingCommand'
     });
   });
+
+  it('gets template controller like resource', function() {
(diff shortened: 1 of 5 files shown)


AureliaEffect:
JS Framework Benchmark: https://30345-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

PR #695, opened 2019-10-09
doc(user-docs/README): replace makes with cli

jwx:
# Pull Request

## 📖 Description

user-docs/README: replace makes with CLI

### DIFF
--- a/docs/user-docs/README.md
+++ b/docs/user-docs/README.md
@@ -28,19 +28,19 @@ Should you find anything to be incorrect or missing as you read our documentatio
 
 ## Creating Your First Aurelia App
 
-There are various ways that you can setup an Aurelia project, including everything from adding a simple script tag to your HTML page to creating a custom [Webpack](https://github.com/aurelia/aurelia/tree/master/packages/webpack-loader) configuration. We think one of the easiest and most powerful ways to get started is by using [the `makes` tool.](https://github.com/aurelia/new)
+There are various ways that you can setup an Aurelia project, including everything from adding a simple script tag to your HTML page to creating a custom [Webpack](https://github.com/aurelia/aurelia/tree/master/packages/webpack-loader) configuration. We think one of the easiest and most powerful ways to get started is by using the Aurelia CLI.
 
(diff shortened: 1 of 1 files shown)


fkleuver:
Okay so if I take a step back and look at this with fresh perspective, I can see the value of `npx au2 new` as an alias to `npx makes aurelia`, because:
- `npx au2 new` feels more intuitive
- `npx makes aurelia` is kind of aurelia-specific anyway
- It's scaffolding, which in itself is kind of an aurelia concern (if we need something more intelligent for Aurelia projects in the future, such as modifying an existing scaffold, that won't fit into `makes`)

However, I have problems with `npx au2 start` as a shortcut for `npm start` because npm commands that exist in the package.json are themselves already shortcuts and almost everyone with at least a few days of experience with npm knows about this. No matter how I look at it, I see this as extra complexity, not as a simplifying thing.

EisenbergEffect:
I really don't want "2" in the command. Is there something else we can do? Also, I agree that we don't want to simply shortcut the npm commands. I don't see any value to that and actually see it as creating more confusion.

jwx:
@EisenbergEffect Did you hear anything from the holder of the `au` package? I still think we should have a CLI that can do `npx au new` (or even `npx au2 new` like we have now if we can't get the `au`). As stated by @LarsKemmann in https://github.com/aurelia/aurelia/issues/733#issue-515454929: "Every little bit of non-hurdle helps."

fkleuver:
Regardless of where we're going with this, if we're going to have a CLI I think it should be here in the repo. We've discussed `@aurelia/cli` package before and decided to hold off on it several months ago. But now we actually have something ti put in it. So let's create the package then @EisenbergEffect ?

3cp:
We can have the package. But I suggest to focus our energy on what's important (like router, aot) right now, and defer the discussion of Aurelia 2 CLI design to much later stage.

fkleuver:
CLI will also be an interface to AOT, so I kind of just want to get it over with. CLI has been this nagging little voice for a few months now. @jwx already created something and I helped him tidy it up a bit so it's practically copy pastable.

EisenbergEffect:
@fkleuver Feel free to set up the cli package and start putting things there as needed. I'll work on the process to get the `au` package name from npm.

fkleuver:
Closing this as we now have the `au` package and the cli is located in the `packages/au` folder

PR #696, opened 2019-10-09
docs(README): replace makes with cli

jwx:
# Pull Request

## 📖 Description

README: replace makes with CLI

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,11 +61,11 @@ export class App {
 
 This example shows you some of the powerful features of the aurelia binding syntax. To learn further, please [see our documentation](docs/user-docs).
 
-Feeling excited? Check out how to how to use `makes` to get started in the next section.
+Feeling excited? Check out how to how to use the Aurelia CLI to get started in the next section.
 
 ## Getting Started
 
-To get started with a new Aurelia 2 project, with Node.js installed, simply run `npx makes aurelia`. You'll then be guided through a series of choices on how you'd like to setup your project. Once complete, you'll have a new Aurelia 2 project ready to run. For more information on Aurelia's use of `makes`, see [here](https://github.com/aurelia/new). If you aren't interested in taking our preferred approach to generating a project, you can also see [the examples folder in this repo](examples) for pure JIT setups (no conventions) with various loaders and bundlers.
+To get started with a new Aurelia 2 project, with Node.js installed, simply run `npx au2 new`. You'll then be guided through a series of choices on how you'd like to setup your project. Once complete, you'll have a new Aurelia 2 project ready to run. Aurelia uses the `makes` tool for project scaffolding, see [here](https://github.com/aurelia/new) for more information. If you aren't interested in taking our preferred approach to generating a project, you can also see [the examples folder in this repo](examples) for pure JIT setups (no conventions) with various loaders and bundlers.
 
 ## Documentation
 


3cp:
Seems unnecessary to have another wrapper.

jwx:
I think it's definitely necessary to provide a CLI with a single "entry point", `au2`, that helps users to create, build, run and test their applications. And in the long run, we'd probably like to be able to add and remove different tools for different tasks without having to update all our documentation and force users to re-learn which commands to use.

AureliaEffect:
JS Framework Benchmark: https://30375-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

3cp:
I disagree on re-learn or need to update doc. For me, there is nothing wrong to tell users to use almost universal commands in nodejs world:

```
npm start
npm run build
npm test
```

Those commands are what you intend to normalize, aren't they?

It could be weird for user to find out `npx au2 new` only wraps `npx makes aurelia`. Just my opinion.

jwx:
Those would be
```
npx au2 new
npx au2 start
npx au2 build
npx au2 test
```
or, after a `npx au2 globalize` or `npm i -g au2`, this
```
au2 new
au2 start
au2 build
au2 test
```

I don't think there's anything weird with the CLI being a wrapper over different tools and commands. On the contrary, I think it's a strength since it keeps it small. It also allows users to achieve the same or customized results without needing the CLI if (when) they've got the desire and necessary skills.

Are you proposing that we don't have a CLI at all?

3cp:
You (or anyone knows me) would not be surprised to know I prefer no CLI at all 😄 

I think I need to elaborate my point. This is just my opinion, not speaking for the whole team. Some part is also out of the scope of this PR.

To me, we might need something to do `au2 generate`, and that's all. But those generators might be better to be implemented in editor plugin (e.g. our vscode plugin). (To be discussed, too early for this pre-alpha stage.)

## 1. all-in-one offering doesn't mean better user experience.

@EisenbergEffect designed aurelia-cli to be the tool for everything for Aurelia v1. While it served the purpose really well, especially in the difficult time of Aurelia tooling during 2015-2016. I love Rob, but it doesn't mean I agree on every engineering decisions he made.

From engineering point of view, aurelia-cli did too many things in one product: a scaffolding tool, a special AMD bundler, a task runner (simply wraps gulp). Over the years, it suffers low test coverage, deep learning for contributors, and legacy features we need to keep while moving forward.

For example, also a point for you to think, to normalize build task for CLI built-in bundler and webpack, `au build` wraps webpack build command behind another gulp task. This created many confusion for webpack users. They need to understand there are three files affects `au build`, the `aurelia.json` config file, the `tasks/build` task file, and the `webpack.config.js` itself. And users unsurprisingly asked why we complicated webpack so much.

@Sayan751 helped to remove aurelia-cli from our webpack setup. Now for backward compatibility, `au build` calls `npm run build:dev` which only uses `webpack.config.js` to control webpack build process. So you see, we are actually undoing `aurelia-cli` all-in-one offering.

Another point, aurelia-cli also implemented few features unnecessary. For example, CLI built-in bundler minification can be left for other gulp plugin to handle, built-in bundler file writing can be left for `gulp.dest` to handle.

aurelia-cli also tries to give user a centralised `aurelia.json` for config everything. But the truth is the config is out of context (not in their related task or related code), hard for user to understand/remember where/how to change. The json config format is also a small mistake, for example, aurelia-cli has lengthy code to support conditional config `{"cache": "dev & stage"}`, which can be replaced by `{cache: process.env.NODE_ENV === 'development' || process.env.NODE_ENV === 'stage'}` without any additional code to support it if chose js as config format.

My opinion against all-in-one solution is not limited to aurelia-cli. My stand is same for other all-in-one solution. Because not just for advanced users, but for every users with non-trivial production apps, it's inevitable for them to customise build/test tasks. aurelia-cli semi-hides the tasks behind local gulp task files, parcel v1 hides them as much as possible (I have not looked deep into parcel v2 yet). Both approach troubles the users who needs to customise them.

The stand `npm` scripts commands at least expose the details in `package.json` where users can understand what's going on.

## 2. small tools working together, doesn't mean bad user experience

Every unix shell users will tell how expressive it is to pipe small commands together. Because every commands only do a small part, they are easy to learn, and can be reused in all sorts of situations.

You can see what I did based on my experience learnt from working on aurelia-cli.
* I did [`makes`](https://github.com/makesjs/makes) which is based on refactored scaffolding part of aurelia-cli.
* I did [`dumber`](https://github.com/dumberjs/dumber) bundler which is a redesigned aurelia-cli built-in bundler. (The only product I have not find time to write full-length doc yet.)

Furthermore, I did not quite agree on the design of aurelia-validation, and didn't find anything on the market satisfying. So I did [`bcx-validation`](https://github.com/buttonwoodcx/bcx-validation) which only does validating js object.

There is also [`browser-do`](https://github.com/3cp/browser-do) because I don't enjoy karma.

All of them are generic, small tools, with very good test coverage. All are reusable out of Aurelia.

I don't think about whether they will be popular in future. I am not very social, not obsessed with popularity at all. Popularity only means more work to me :-) I designed my tools not to please others, but simply to please myself, because I have this psychological need to express myself. My tools might please few others in the process, but I knew I am not majority.

BTW, I am bit sad that the software industry (not only js) built up this trend on all-in-one offering. What's wrong to be small, nimble, flexible, and humble?

fkleuver:
In general I agree with @3cp in the sense that I also prefer to be close to the metal with my tooling. 

My problem with `au2` cli is that it's just like conventions, it makes things a bit cleaner and nicer for beginners or people who just like to save keystrokes. But in the framework where we have conventions we also have an easy switch back to more control.

With a cli this is trickier because the more you wrap and unify, the more you start to lose important details, you become prone to causing users to miss out on new stuff on the market because you'll lag behind.

`aurelia-cli` became a messy collection of hacks and band-aids to patch things up *just* enough to make it work with webpack. We all know how that ended, @3cp already illustrated it clearly.

Our vision (or at least my vision, but I hope the core team shares this) for Aurelia 2 in general: stick to the specs, stick to the standards, minimize special build dependencies, so that Aurelia can work with everything and everything can work with Aurelia.

I would almost like to take a sort of pride in, and show off, that you can do Aurelia without *any* kind of aurelia-specific CLI. Just plain vanilla webpack, plain vanilla gulp, or whatever it is that you like to use. Not aurelia-cli and its opinionated pre-defined (limited) set of options.

I vote against a unifying wrapper for reasons above. I understand and appreciate its merits, but it doesn't fit with our new direction in my opinion.

However, I'm open to have my mind changed. This is just what I think right now.

jwx:
I think there's some kind of misunderstanding here. (And if not, I simply don't get it at all.) I'm not talking about a big CLI that implements a lot of commands the way the v1 CLI does. I'm talking about a "small, nimble, flexible, and humble" tool that's got one job: to let the devs use command aliases in the form of 
```
au2 new
au2 start
au2 build
au2 test
au2 (help) // lists available commands (with description where available)
```
to execute the actual commands
```
npx makes aurelia
npm start
npm run build
npm test
(no help available!)
```

It's a small tool that helps devs to not _have to_ know the implementation details/tool of the different commands. But it doesn't perform anything by itself apart from that. It's completely optional. You and other devs (who are not really the target audience) can still use your direct commands, go straight to the metal. The actual work would still be performed by small tools in a standard way (including your excellent ones, @3cp). Your pride would still be intact, @fkleuver, because everything about this that you're proud of would still be true. (And you can even be _more_ proud since we achieved everything you want _and also_ gave (new) devs an easier and quicker path into our command line world.)

Why would we not offer devs that want or need it this extra help?

fkleuver:
> Your pride would still be intact, @fkleuver, because everything about this that you're proud of would still be true. (And you can even be more proud since we achieved everything you want and also gave (new) devs an easier and quicker path into our command line world.)

I did not mean to imply that my pride is of any concern here. I don't care about pride. I just want a good end result.

image

This doesn't add anything for me, it just hides information, namely what the real underlying tools are and how those underlying tools can be used outside of the context of aurelia.

> It's a small tool that helps devs to not have to know the implementation details/tool of the different commands.

Here is the fundamental disagreement: I don't believe in this. My many years with npm and its landscape have taught me that there are no shortcuts or magic tools. If you want to be productive, you need to get familiar with your tools.

So I would argue the oppsite: you *have* to know the tools you're working with, or you're in for a very very bad time. I don't want to create the appearance of something being simpler than it really is.

> Why would we not offer devs that want or need it this extra help?

Because I don't believe it helps at all.

EisenbergEffect:
**tl;dr;** The most important thing is that there's a seamless experience, not necessarily a single tool. Also, in keeping with Aurelia's values, we should try to stay as close to "vanilla" as possible and as modular as possible. I think vanilla would be basic npm scripts and modular would mean small, single-purpose tools.

The original CLI was born in a different time and under different constraints. Aurelia 1 really required a lot of custom setup at that time, and tools were a much bigger mess and harder to configure.

However, with Aurelia 2, we don't really need anything special to build a fully JIT project with no conventions. For conventions, we just have a simple plugin to whatever build system you are using. I like that. I like that I can use a `makes` to generate my project, and then it's just npm commands. Those npm commands can use a set of tools, but it's nice just being able to look inside the package json to see what's going on, and not needing any global tools or to even know that a special tool was used to create the project.

For generators, we can just have a small single-purpose tool. `makes` will generate npm scripts that leverage that tool.

So, I think the process can be super seamless. We just need the scaffolder (which we have) to set up a project with npm commands for common tasks, each of those commands using whatever tool is most appropriate.

jwx:
I've never suggested a big single tool that does all the work. Quite the contrary. And it wouldn't be replacing or hiding the tools that does the actual work either (the `au2` command outputs the underlying command so full transparency). It would be just an additional set of commands like conventions on top of `dependencies` or a configured route on top of a set of viewport instructions.

I can't really understand why we wouldn't offer a CLI to those who want it. (Which is never going to be any of you with your years of javascript experience and extensive knowledge about the tooling.) Not having it will just be yet another fight when trying to explain to devs why they should pick Aurelia over something else.

(And no help command? Ouch, why not?)

EisenbergEffect:
I'm not seeing what the cli would give that the other approach would not. @jwx Can you give me a concrete list of bullet points?

3cp:
@jwx it does not justify "CLI to those who want it" by wrapping `npm run build` behind `au2 build`.

`npm run build` is already as simple as it could be. A wrapper is not simpler, only stranger.

Nobody needs "years of javascript experience and extensive knowledge about the tooling" to run `npm run build`. It's basic stuff.

jwx:
> Nobody needs "years of javascript experience and extensive knowledge about the tooling" to run `npm run build`.

I've never said they do. Please don't put words in my mouth.

3cp:
@jwx I am sorry making you feel offended.

brandonseydel:
I don’t think we need a cli for the basic commands to start/lint/etc. see commands below


  npx aurelia new
 npx aurelia generate element/attr/etc
npx aurelia add router/logger/http (plugins)
npx aurelia update 
npx aurelia router add/remove/update

 Or just install globally.

EisenbergEffect:
I'm going to close this out for now because in the very least, we need further discussion.

Here's a link to the original RFC: https://github.com/aurelia/aurelia/issues/509

elmt1:
The first way I find out how things like that work is searching for "dotnet command line" or "sqlpackage command line". I think not having a specific cli tool would be OK, but the one thing that helps me after I have learned what each option does is to use -help like dotnet -help to remember the options and syntax.

If au -help isn't going to be there then it is helpful if there is an intuitive way to find out what the options are because so guessing it might be npx aurelia help or just npx aurelia?

jwx:
> npx aurelia new
> npx aurelia generate element/attr/etc
> npx aurelia add router/logger/http (plugins)
> npx aurelia update
> npx aurelia router add/remove/update

All of the above would be Aurelia CLI commands, so none of them would exist without a CLI.

jwx:
@elmt1 
> it might be npx aurelia help or just npx aurelia?

Anything that starts with `npx aurelia` (or just `aurelia`) would be Aurelia CLI commands, so none of those two would exist without a CLI.

Issue #686, opened 2019-10-07
[RFC] metadata api / native decorators integration (placeholder)


fkleuver:
# 💬 RFC

Issue #660, opened 2019-10-04
Injecting a parent’s custom attribute doesn’t work


rluba:
# 🐛 Bug Report

@fkleuver, as discussed in Discord: I converted two custom attributes from Aurelia 1 to Aurelia 2. One must be nested inside the other.

Usage looks like this:
```HTML
<li au-dropdown>
	<a au-dropdown-toggle>…</a>
	…
</li>
```
The inner attribute in V1 (works):
```JavaScript
@inject(DOM.Element, AuDropdownCustomAttribute)
export class AuDropdownToggleCustomAttribute {
```

When I try to inject the outer attribute into the inner one, Aurelia v2 injects a _new_ instance of the outer attribute (and doesn’t initialize the bindings at all):
```TypeScript
export class AuDropdownToggleCustomAttribute {
    constructor(public element: HTMLElement, public dropdown: AuDropdownCustomAttribute) {
```

## 🤔 Expected Behavior

The constructor should get the attribute instance from its parent element, as it did in Aurelia 1.

EisenbergEffect:
I think this is an important use case, but I'm not sure we should enable it in the same was that we did in v1. Here are a few quick thoughts. @rluba I'd love to hear what you think. FYI for @fkleuver as well.

In v1, we created a dedicated container instance for every DOM node instance that was a custom element or had a custom attribute on it. This meant there was a 1:1 mapping between container instances and Aurelia behaviors. Anything created out of one of these containers was a singleton. This enabled the pattern above, but at the cost of creating many, many DI containers, which all needed to be kept in memory. So, there were both perf and memory concerns there. It also made me feel pretty uncomfortable, since the one unique scenario it seemed to enable was the one above. It's a lot of cost on the entire system for just this one thing.

For v2 we're doing some different things with containers that enable us to have fewer of them. I think that's a good thing if we can keep it that way. So, how to address this use case?

Well, we've got a `@children` decorator. Why not add a `@parent` decorator? Since this is pretty DOM-centric, does it really need to be done with DI? My thought is that this is better handled through our component system and connected with lifecycle hooks.

[...]

Issue #636, opened 2019-09-25
[RFC] au-host, au-bind, au-transfer-bindings, au-transfer-attrs, etc


fkleuver:
Just some quick notes based on discussion on Discord, so we don't lose the conclusions/todos again.

- `au-host`
Placed on any element *inside* the template of a containerless custom element.
This element is then registered as the `INode`, `Node`, `Element` and `HTMLElement` to be injected into custom attributes that are placed on the containerless element on the outside.

Example:
**app.html**
```html
<template>
  <my-element my-custom-attr></my-element>
</template>
```

**my-element.html**
```html
<template>
  <div au-host>
    <!-- stuff -->
  </div>
</template>
```

**my-custom-attr.ts**
```ts
export class MyCustomAttr {
  constructor(el: Element) {
    // el is the HTMLDivElement that au-host was placed on
  }
}
```

- `au-bind` see https://github.com/aurelia/aurelia/issues/319

brandonseydel:
Some proposals for transfer options/naming convention possibilities the au-host is pretty much on point @fkleuver ;)
```html

<template>
<input au-accept="style" />
<input au-accept="class" />
<input au-accept />

</template>
```

```html
<!-- Transfer All attributes and binding -->
<parent-element au-transfer style.bind="style" class="test"/> 
<!-- Transfer ONLY the style binding -->
<parent-element au-transfer="style" style.bind="style" class="test"/>
<!-- Transfer All attributes and binding with specific  -->
<parent-element au-transfer style.bind="style" class="test"/> 
```

shahabganji:
Just my opinion around the idea:

```html
<template>
    <my-element my-attr></my-element>
</template>
```

```html
<template>
    <input au-host.accept="attr1, attr2" />
    <button au-host.only="attr3">Button</button>
</template>
```

maybe drop the `au-` prefix, just like `bind`, we could use `host`:

```html
<template>
    <input host.accept="attr1, attr2" />
    <button host.only="attr3">Button</button>
</template>
```

See [here](https://github.com/aurelia/aurelia/issues/319#issuecomment-457518284) in #319

EisenbergEffect:
I agree with that. If we're going to need to use a `.bind` with the attribute for normal use anyway, then we can drop the `au-` prefix since the attribute, in practice, has extremely low likelyhood of conflicting with any future web standard attribute or that of another library or framework.

bigopon:
One question we probably need to answer early is do we want to support dynamic values.

[...]

PR #706, opened 2019-10-12
Metadata

fkleuver:
# Pull Request

## BREAKING CHANGES

- Aurelia no longer writes to static class properties. So, while you can still *place* certain static properties (e.g. `bindables`, `aliases`, `template`, etc), the framework considers them read-only. The `Metadata.getOwn` and `CustomElement.getDefinition` etc methods are now how you retrieve the definitions
- The `build` property of the `customElement` decorator's object literal is gone, and is now instead a boolean property `needsCompile`

## 📖 Description

Goal of this PR is to make our decorators forward compatible with the official decorators proposal: https://github.com/tc39/proposal-decorators

The proposal allows for certain observation and even mixin-like behavior, but to be on the safe side I have limited Aurelia more or less to `@register`, a callback that is invoked when a class is defined.

We use this to associate metadata with classes via a spec-compliant `Reflect.defineMetadata` polyfill that uses a module-scoped WeakMap as its backing store, see:
https://rbuckton.github.io/reflect-metadata

All decorators in framework and user code can be replaced with a series of  `map.set` calls - that's all they boil down to now. The end result could easily be statically analyzed by tools/bundlers/the browser, as well as by our AOT. 

## Affected APIs

Only the decorators that are integral to DI, compilation and rendering were taken into account. Internal helpers/mixins like `@subscriberCollection` could easily be replaced with class/function composition or class inheritance, and were therefore left as-is for now.

Without further ado, the decorators that were overhauled:

### Resource definitions
- `@customElement`
- `@customAttribute`
- `@templateController` (wraps `@customAttribute`)
- `@valueConverter`
- `@bindingBehavior`
- `@bindingCommand`

### General-purpose class annotations
- `@view`
- `@alias`
- `@useShadowDOM`
- `@containerless`
- `@strict`
- `@inject`
- `@transient`
- `@singleton`

### Property annotations
- `@children`
- `@bindable`

### Parameter annotations
- `@all`
- `@InterfaceSymbol` (parameter decorators created by `DI.createInterface`)

### Resource definitions

The resource definitions were the bulk of the work due to them being heavily interconnected with almost all aspects of the framework. I made an attempt to simplify the types and generally improve consistency and readability across the different resources types.

Resource definitions are stored with their own keys, so it is possible, for instance, to **let the same class be a CustomElement as well as a CustomAttribute** or, perhaps more practical, a CustomElement as well as a RouteableComponent (so that the router can finally have its own resource model to work with).

The APIs are still the same, just their internals are quite different now.

### Protocol.resource / Protocol.annotation

For starters, the naming scheme (encapsulated via `Protocol.resource` and `Protocol.annotation` - naming improvement suggestions are welcome):

- Resource-specific metadata is prefixed with `au:resource`, so a `CustomElementDefinition` (initiated via `@customElement` or `CustomElement.define`) would be stored on `au:resource:custom-element`
- Resource-agnostic metadata is prefixed with `au:annotation` where DI annotations are further namespaced as `au:annotation:di`, such that a resource's resource-agnostic `dependencies` are stored on `au:annotation:dependencies` whereas a class constructor's parameter types (dependencies) are stored on `au:annotation:di:dependencies`

The `Protocol` api is also used to keep track of which metadata keys are associated with a type. This is just for performance reasons, to avoid having to iterate through the internal map on every invocation.

### Resource definition property resolution order

For each property of a resource definition there are 3 possible sources:
1. Per-property decorators. While not all of them are implemented yet, the wiring is there. 
2. Static properties on the class.
3. Object literal properties in the resource decorator.

They are also resolved in that order (this is not set in stone and very much open for discussion of course - I just had to pick something to start with), so in the following example:
```ts
@customElement({ name: 'foo', containerless: false })
@containerless
export class Foo {
  static containerless = false;
}
```
`containerless` will be `true` in the definition, because per-property decorators override static properties and resource decorator object literal properties.

In this example:
```ts
@customElement({ name: 'foo', containerless: false })
export class Foo {
  static containerless = true;
}
```
`containerless` will be `true` in the definition, because static properties override resource decorator object literal properties.

In the case of arrays, the properties are merged, so in the following example:
```ts
@customElement({ name: 'foo', aliases :['bar'] })
@alias('baz')
export class Foo {
  static aliases = ['qux'];
}
```
The node names `foo`, `bar`, `baz` and `qux` all resolve to the custom element `Foo`.

Same goes for records/array combinations, so `bindables` and `children` will also be merged from all 3 sources.

### Inheritance

Specifically `bindables` and `children` will also be merged from prototype annotations (but not definition or static properties). So in this example:

```ts
@customElement({ name: 'foo', bindables: ['fooProp1'] })
export class Foo {
  static bindables = ['fooProp2'];
  @bindable fooProp3;
}

@customElement({ name: 'bar', bindables: ['barProp1'] })
export class Bar extends Foo {
  static bindables = ['barProp2'];
  @bindable barProp3;
}
```

The `Bar` type will have the bindables `fooProp3`, `barProp1`, `barProp2` and `barProp3`.

We *could* merge everything, but my reasoning to only merge property decorators is because they are invoked on the prototype whereas the other pieces are associated with the class constructor itself. So the idea is to follow prototypical inheritance but not constructor inheritance - it gives users some ability to choose the level of encapsulation.

## The data model

To help better understand things, let's look at a practical example and then inspect the metadata model it translates to.

```ts
@customElement({
  name: 'name-tag',
  template: '${firstName} ${lastName}',
})
@alias('au-name-tag')
class NameTag {
  @bindable firstName;
  @bindable lastName;
}

@customElement({
  name: 'app',
  template: '<name-tag first-name.bind last-name.bind></name-tag>',
})
@dependencies(NameTag)
export class App {
  constructor(@ILifecycle lifecycle: ILifecycle) {}
}
```

The metadata (represented in abbreviated psuedo json form for readability) after DI registration but before compilation:
```json
[
  [
    "class NameTag#1",
    [
      [
        null,
        {
          "au:resource": [
            "au:resource:custom-element"
          ],
          "au:resource:custom-element": {
            "name": "name-tag",
            "key": "au:resource:custom-element:name-tag",
            "template": "${firstName} ${lastName}"
          },
          "au:annotation": [
            "au:annotation:aliases",
            "au:annotation:di:dependencies",
            "au:annotation:di:factory",
            "au:annotation:bindable:firstName",
            "au:annotation:bindable:lastName"
          ],
          "au:annotation:aliases": [
            "au-name-tag"
          ],
          "au:annotation:di:dependencies": [],
          "au:annotation:di:factory": "class Factory#1(0)"
        }
      ],
      [
        "firstName",
        {
          "au:annotation:bindable": {
            "property": "firstName",
            "attribute": "first-name",
            "mode": 2
          }
        }
      ],
      [
        "lastName",
        {
          "au:annotation:bindable": {
            "property": "lastName",
            "attribute": "last-name",
            "mode": 2
          }
        }
      ]
    ]
  ],
  [
    "class App#1",
    [
      [
        null,
        {
          "au:resource": [
            "au:resource:custom-element"
          ],
          "au:resource:custom-element": {
            "name": "app",
            "key": "au:resource:custom-element:app",
            "template": "<name-tag first-name.bind last-name.bind></name-tag>"
          },
          "au:annotation": [
            "au:annotation:dependencies",
            "au:annotation:di:dependencies",
            "au:annotation:di:factory"
          ],
          "au:annotation:dependencies": [
            "class NameTag#1"
          ],
          "au:annotation:di:dependencies": [
            "interface ILifecycle#1"
          ],
          "au:annotation:di:factory": "class Factory#1(1)"
        }
      ]
    ]
  ]
]
```

Which, by the way, would achieve the same thing as the functional API invocation as below (the difference being that the functional API, lacking decorators, doesn't support resource-agnostic annotations):
```ts
export const App = CustomElement.define({
  name: 'app',
  template: '<name-tag first-name.bind last-name.bind></name-tag>',
  dependencies: [
    CustomElement.define({
      name: 'name-tag',
      template: '${firstName} ${lastName}',
      bindables: ['firstName', 'lastName'],
      aliases: ['au-name-tag']
    })
  ]
}, class {
  static get inject() { return [ILifecycle]; }
  constructor(lifecycle: ILifecycle) {}
});
```

### 🎫 Issues

- Addresses: https://github.com/aurelia/aurelia/issues/686 but the RFC still needs to be written (this remains a TODO for later API refinement)
- Fixes: https://github.com/aurelia/aurelia/issues/678
- Facilitates: https://github.com/aurelia/aurelia/issues/660
- Facilitates: https://github.com/aurelia/aurelia/issues/636
- Facilitates: https://github.com/aurelia/aurelia/issues/622
- Facilitates: https://github.com/aurelia/aurelia/issues/497
- Facilitates: https://github.com/aurelia/aurelia/issues/424

Bugs fixed / shortcoming addressed / etc:

- If anything is a great candidate for proper FP-style immutability, it's decorators and their metadata. So, `CustomElementDefinition` (formerly `TemplateDefinition) is essentially immutable: the compiler creates a clone with the new compiled properties.
- The compiled template is registered, again via Metadata, as a scoped property of the non-compiled template. The property key is `au:resource:custom-element:${path}:compiled-template` where `path` is `renderContext.path`, and `renderContext.path` is a dot-separated string of context id's all the way up to the root.
  - This enables `transfer-attrs` as well as a number of other context-sensitive concerns, such as router scopes.
  - `ViewportCustomElement.render` no longer needed (it's removed now)
- Simplified/improved template compiler flow
- Simplified/improved template binder flow wrt replaceables
- Fixed some false positive replaceable tests
- Reworked all resource definitions and most annotations from the ground up to be more consistent, predictable and interoperable

## 👩‍💻 Reviewer Notes

Noteable files:
[packages/runtime/src/resources/custom-element.ts ](https://github.com/aurelia/aurelia/pull/706/files#diff-17f91605f3e2acbada1065929b44d1cf)
[packages/runtime/src/resources/value-converter.ts](https://github.com/aurelia/aurelia/pull/706/files#diff-fa59d16d85efa767a7e564e4a3262256)
[packages/runtime/src/templating/bindable.ts](https://github.com/aurelia/aurelia/pull/706/files#diff-a1bab50a2184d8f4f00aa8f28d83c456)
[packages/runtime/src/templating/children.ts](https://github.com/aurelia/aurelia/pull/706/files#diff-e4740d1de214600ae9f1c23fb75f62a4)
[packages/runtime/src/templating/controller.ts](https://github.com/aurelia/aurelia/pull/706/files#diff-43b0ef53a89721f28df850420b5883d0)
[packages/runtime/src/templating/view.ts](https://github.com/aurelia/aurelia/pull/706/files#diff-8ddee4cc38e27efb7b9268bccbcb3eaf)
[packages/runtime/src/resources/custom-attribute.ts](https://github.com/aurelia/aurelia/pull/706/files#diff-c1a3eceec58ce6e4badbfff248770446)
[packages/runtime/src/resources/binding-behavior.ts](https://github.com/aurelia/aurelia/pull/706/files#diff-d73dae43181a0c7b684bef294a9770b7)
[packages/runtime/src/render-context.ts](https://github.com/aurelia/aurelia/pull/706/files#diff-b12a454f7d84dd07c94592e2fb3b0a30)
[packages/runtime/src/rendering-engine.ts](https://github.com/aurelia/aurelia/pull/706/files#diff-7303a3a5dd9e044c24f021aff825f430)
[packages/kernel/src/di.ts ](https://github.com/aurelia/aurelia/pull/706/files#diff-a60e26e3e4fd00a5f2b949a9fabd1195)
[packages/kernel/src/resource.ts](https://github.com/aurelia/aurelia/pull/706/files#diff-96b1ac35f9c06f35423a70c12b6166c1)


AureliaEffect:
JS Framework Benchmark: https://31054-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://31224-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://31287-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://32013-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://32034-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://32043-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://32046-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://32058-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://32183-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://32216-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://32251-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://32260-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://32314-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://32317-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://32455-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://32467-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://32474-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://32498-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://32511-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://32519-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://32544-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/__tests__/jit-html/resources.spec.ts
```diff
@@ -0,0 +1,182 @@
...
+    );
+
+    assert.deepStrictEqual(
+      $proto,
+      {},
+      `$proto`,
+    );
+  });
+
+  xit('works with alias decorator before customAttribute decorator', function () {
```
bigopon: this doesn't work?

bigopon:
> Specifically bindables and children will also be merged from prototype annotations (but not definition or static properties)

Static properties are inherited by default, and from the code, we are simply doing `Type.--the_property_name--`, does that mean we also merge static properties by default or the lookup needs some tweaks?

packages/__tests__/jit-html/resources.spec.ts
fkleuver: Indeed. Decorator application order is bottom-up. There are ways to make it work regardless (e.g. by treating the resource definitions as mutable, retrieving them, etc) but I'd say let's not add that complexity just yet.
Decorator order matters, that's how it would behave natively as well, so I think it is reasonable to expect users to be at least somewhat aware of this: put annotations at the bottom, and resource definitions on top

bigopon:
I think this is an awesome cleanup, if the tests all pass then let's go. Can't wait for the next wave of improvements.

fkleuver:
@bigopon 

> we are simply doing Type.--the_property_name--, does that mean we also merge static properties by default or the lookup needs some tweaks?

No, I'm specifically excluding inherited property actually:
```ts
value = (Type as Constructable & TDef)[name] as TDef[K] | undefined;
if (value === void 0 || !hasOwn.call(Type, name)) {
  return getDefault();
}
```

So they are neither inherited nor merged from the prototype. I figured it's going to have to be neither or both, but not half/half.

bigopon:
@fkleuver this is not how a normal JS class user would expect it to work, so probably we will need a FAQ section for this

AureliaEffect:
JS Framework Benchmark: https://32578-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

fkleuver:
@bigopon 

As clarified before, my reasoning to exclude *static* properties is for consistency:
- class-level (static properties, class decorators) things are not inherited
- only prototype-level things (property decorators) are inherited

However, the point of consistency here is to be easy to explain/memorize, and at least somewhat intuitive to most people. I think perfect intuitiveness is not achievable in what is already kind of inconsistent natively.

However, if it would indeed be surprising behavior to everyone, then we shouldn't have a FAQ section, rather we should just fix it to work as users expect it to.

Where do we draw the line though? There are many more questions to answer:
- Should mergeable properties (objects, arrays) be overwritten (like they would be in normal JS) or merged?
- Should definitions then also be merged?
  - And individual mergeable properties from definitions?
  - Does that include instructions?

bigopon:
I see, somehow it didn't make as much sense to me before those last questions. It's a simple rule, resources definition will not be inherited, I think we can easily explain that

fkleuver:
To be more precise: only property decorators are inherited

PR #707, opened 2019-10-13
fix(ref): always notify during bind

bigopon:
# Pull Request

## 📖 Description

At the moment, ref binding update source during `$bind`, results in the change notification got swallowed in `SetterObserver`. Add a flag to bypass it.

### 🎫 Issues

N/A

## 👩‍💻 Reviewer Notes

N/A

## 📑 Test Plan

N/A

## ⏭ Next Steps

Finish up Portal, this is the last bit that blocks Portal


AureliaEffect:
JS Framework Benchmark: https://31146-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/runtime/src/observation/setter-observer.ts
```diff
@@ -51,7 +51,7 @@ export class SetterObserver {
       const currentValue = this.currentValue;
       this.currentValue = newValue;
       if (this.lifecycle.batch.depth === 0) {
-        if ((flags & LifecycleFlags.fromBind) === 0) {
+        if ((flags & LifecycleFlags.fromBind) === 0 || (flags & LifecycleFlags.alwaysNotify) > 0) {
```
fkleuver: So as already discussed on discord, problem with this new flag is we're already at 31 bits so we don't have room for another one right now.

Anyway, I don't think this is the right way to go. I'd consider removing the check `(flags & LifecycleFlags.fromBind) === 0` instead, actually, because that's an optimization that v1 doesn't have.

AureliaEffect:
JS Framework Benchmark: https://31187-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://31416-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://31430-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

PR #708, opened 2019-10-13
fix(i18n): type definition

Sayan751:
# Pull Request

## 📖 Description

Exported `I18N` interface so that devs don't have to explicitly refer to `I18NService`.


AureliaEffect:
JS Framework Benchmark: https://31296-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

PR #717, opened 2019-10-20
Fix multi binding compilation

bigopon:
# Pull Request

## 📖 Description

At the moment, multi binding attributes usage does not camel-case-ize the target attribute-property, making the following not working:
```html
<div square="border-radius: 5"></div>
```
But the following work;
```html
<div square="borderRadius: 5"></div>
```
Fix that by always convert target attribute-property to camel case

### 🎫 Issues

N/A

## 👩‍💻 Reviewer Notes

Should there be a warning in case an upper case character is detected in attribute name?

## 📑 Test Plan

N/A

## ⏭ Next Steps

N/A

### DIFF
--- a/packages/__tests__/jit-html/portal.spec.tsx
+++ b/packages/__tests__/jit-html/portal.spec.tsx
@@ -281,47 +281,100 @@ describe('portal.spec.tsx 🚪-🔁-🚪', function () {
           );
         }
       },
-      // todo: re-enable the test after fixing the parser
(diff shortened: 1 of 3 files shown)


AureliaEffect:
JS Framework Benchmark: https://32353-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

PR #718, opened 2019-10-20
chore: some more linting

BBosman:
# Pull Request

## 📖 Description

Another batch of linting tweaks

### 🎫 Issues

n/a

## 👩‍💻 Reviewer Notes

It's best to review this PR on a per commit basis, to see the rule changes in context.

## 📑 Test Plan

CircleCI

## ⏭ Next Steps

More linting


AureliaEffect:
JS Framework Benchmark: https://32430-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

.eslintrc.js
```diff
@@ -80,6 +80,7 @@ module.exports = {
     '@typescript-eslint/no-unnecessary-qualifier': 'error',
     '@typescript-eslint/no-unnecessary-type-arguments': 'error',
     '@typescript-eslint/no-useless-constructor': 'error',
+    '@typescript-eslint/prefer-for-of': 'error',
```
fkleuver: I'd rather not have this rule enabled as `for..of` is slower than `for` and we usually do this in perf critical places.

See: http://jsben.ch/1Nc4Z

test/kitchen-sink/src/startup.ts
```diff
@@ -140,7 +140,7 @@ const container = DI.createContainer().register(
   // custom inline registrations
   ValueConverter.define('pascal', class {
     public toView(input: string): string {
-      const camel = camelCase(input);
+      const camel: string = camelCase(input);
```
fkleuver: Why is this necessary? `camelCase` returns `string` right?

.eslintrc.js
BBosman: @fkleuver No problem. Do you want me to also revert the related changes? (I tried to do them only in non perf critical places)

test/kitchen-sink/src/startup.ts
BBosman: Yes, but `eslint` does not know that.

Because it's outside of `packages` and we don't do an `npm install` within `test/kitchen-sink` all dependencies are missing, so `eslint` defaults to `any` as it can't determine the actual type.

packages/testing/src/startup.ts
```diff
@@ -13,6 +13,7 @@ export function setup<T>(template: string | Node,
   const root = ctx.doc.body.appendChild(ctx.doc.createElement('div'));
   const host = root.appendChild(ctx.createElement('app'));
   const au = new Aurelia(container);
+  // eslint-disable-next-line @typescript-eslint/no-extraneous-class
```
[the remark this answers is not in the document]
BBosman: Usually empty classes in production code (but not in tests) are a red flag. The remaining two violations appear valid exceptions to that rule.

I agree the benefit for this rule is marginal, so if you want it out I'll take it out.

test/kitchen-sink/src/startup.ts
fkleuver: So this is not really solving anything then. I'd rather not introduce all kinds of type assertions just to work around what is technically a bootstrapping problem.
I don't have anything against installing all test projects as part of the lint job. Shall we do that instead?

.eslintrc.js
fkleuver: All of them except for the js framework benchmark are pretty perf sensitive / deliberately optimized pieces of code. Even if not hot paths, they are easy wins of little bits of perf improvements.
I also noticed you used `Array.from` on nodes once. That method is notoriously slow. We have `toArray` for that exported from the kernel. But I think it's still better to just leave that one be a for loop.

test/kitchen-sink/src/startup.ts
BBosman: But it also applies when running `npm run lint` locally. Or do you want to extend the `npm run bootstrap` task to `npm install` the test projects as well, as that would fix it for both local and CircleCI.

HamedFathi:
@fkleuver @BBosman 

`"@typescript-eslint/interface-name-prefix": "off"`
Is it valid for you to have `I` prefix for interfaces?

packages/testing/src/startup.ts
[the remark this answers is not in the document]
BBosman: Dropped it.

.eslintrc.js
BBosman: Dropped them.

BBosman:
> Is it acceptable for you to have `I` prefix for interfaces?

I'll let @fkleuver answer this one. He was doing some "magic" with exporting interfaces and classes with the same name to have the typescript compiler somehow merge those types, but I'm a bit hazy on the details as it was some time ago that we discussed it.

(For reference: There are currently 78 interfaces in the codebase that don't have the I prefix)

fkleuver:
Eventually we can try to move to it but it's not something I want to be a priority right now. Basically some of the non-prefixed interfaces need to stay interfaces whereas others (e.g. the merge scenario mentioned by @BBosman) should become types instead, but I need to take a look at this myself when the time feels right.

test/kitchen-sink/src/startup.ts
BBosman: Reverted this, as it's no longer needed as we brought those folders under control of `lerna`.

AureliaEffect:
JS Framework Benchmark: https://35507-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

BBosman:
@fkleuver I think I addressed all your comments.

AureliaEffect:
JS Framework Benchmark: https://35567-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://35571-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

PR #720, opened 2019-10-22
fix(metadata): add metadata and decorate function polyfills

fkleuver:
# Pull Request


EisenbergEffect:
After we fix the current issue we're working on, can this be closed? @fkleuver

PR #722, opened 2019-10-26
Get rid of PLATFORM timers

fkleuver:
# Pull Request

## 📖 Description

The idea of `PLATFORM.setTimeout`, `PLATFORM.requestAnimationFrame`, etc, was to normalize these timers in a runtime-agnostic manner. With the new scheduler this normalization is no longer useful (and may in fact be counter-productive) so the idea is now to simply use (and await) the scheduler for everything.

The real native timers are of course still accessible via `DOM.window.setTimeout`, `DOM.window.requestAnimationFrame` etc for SSR-compatibility, if needed by end-users for one reason or another, but in the framework we should avoid this to make things like `yield()` actually useful.

This PR addresses (most of) the topic by removing the timers from `PLATFORM` and refactoring to use the scheduler.

It also contains a slight rework of `Throttle` and `Debounce` binding behavior, utilizing the scheduler to make things a bit cleaner and improve consistency (you'll find that the two binding behaviors are pretty much identical apart from 2-3 LOC), as well as using `postRender` instead of `macroTask` to make things more timeslicing-friendly.

Also added `scheduler.yieldAll()` api as a way to await all tasks from all task queues; awaiting this would put you in a timing situation where all non-persistent tasks queued by the framework or by users are guaranteed to be completed.


AureliaEffect:
JS Framework Benchmark: https://32959-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

PR #723, opened 2019-10-26
Integration tests focusing runtime and runtime-html and bug fixes

Sayan751:
# Pull Request

## 📖 Description

This PR adds more integration tests focusing on following areas:

- `@aurelia/runtime`
  - Binding
    - `call-binding`
    - `let-binding`
    - `ast`
       - `BinaryExpression`  (partial)
       - `UnaryExpression`  (partial)
       - `ValueConverterExpression` (partial)
  - Observers
    - `dirty-checker`
    - `array-observer`
    - `map-observer`
- `@aurelia/runtime-html`
  - Observers
    - `checked-observer`
    - `setter-observer`
    - `select-value-observer`

Bug fixes:
- Order of `model`/`value` and `matcher` attribute needed to precede the `checked` attribute (`input[type=radio|checkbox]`) in order to work correctly. Fixed it by sorting the attributes in the correct order during compilation. Refer: https://github.com/aurelia/aurelia/commit/d8693cc66fc6206ee3d0982725bb51238ee9a19b#diff-60e172f5d67f09fb4a94996f7a44c3f8, https://github.com/aurelia/aurelia/commit/374ce9be4f362352fd56f455d667603fc1fde264#diff-60e172f5d67f09fb4a94996f7a44c3f8
- Subscribers were not notified if the value of an existing key in a map is changed. Refer: https://github.com/aurelia/aurelia/commit/55c9fdfd0baa1e17f24ab8839b09ee3c4e052650#diff-de43fb4b1cd41d6bc5ca7985e59a962c
- The `connect` for binary expression was pre-maturely short-circuiting, which resulted in incorrect change notification (prominently seen in `let-binding`). Refer: https://github.com/aurelia/aurelia/commit/039f4f2ef9e5f342201efa3e4a071ba4c23fa968#diff-596e652716d84f1a9398b642928c418b


packages/__tests__/integration/integration.spec.ts
```diff
@@ -20,34 +22,33 @@ describe('app', function () {
...
+  function assertCalls(calls: Call[], fromIndex: number, instance: any, expectedCalls: string[], unexpectedCalls?: string[], message?: string) {
+    const recentCalls = new Set(calls.slice(fromIndex).map(c => Object.is(c.instance, instance) && c.method));
     for (const expectedCall of expectedCalls) {
       assert.equal(recentCalls.has(expectedCall), true, `${message || ''} expected ${expectedCall}`);
     }
     for (const expectedCall of unexpectedCalls) {
       assert.equal(recentCalls.has(expectedCall), false, `${message || ''} not expected ${expectedCall}`);
     }
   }
+  function wait(ms = 300) {
```
fkleuver: With the new scheduler this shouldn't be needed anymore. You can get the scheduler from `ctx.scheduler` and then simply `await scheduler.yieldAll()`

packages/__tests__/integration/app/app.ts
```diff
@@ -19,16 +26,67 @@ export class App {
...
+  public user: User = new User('John', 'Doe', 0.1, 'Role1', 'Org1', 'City1', 'Country1');
+
+  // #region checked, map, repeat
+  public contacts1: Map<number, string> = new Map<number, string>([[123456790, 'mobile'], [9087654321, 'work'], [1122334455, 'home']]);
+  public contacts2: Map<number, string> = new Map<number, string>(Array.from(this.contacts1));
+  public chosenContact1: number = 9087654321;
+  public chosenContact2: number = 9087654321;
+  public contacts3: Contact[] = Array.from(this.contacts1).map(([number, type]) => ({ number, type }));
+  public chosenContact3: Contact = this.contacts3[0];
+  public contacts4: Contact[] = this.contacts3.slice(0);
```
fkleuver: Small nit, but `.slice()` without arguments does the same thing

packages/__tests__/integration/integration.spec.ts
```diff
@@ -1,12 +1,14 @@
...
+import { App, Product } from './app/app';
 import { startup, TestExecutionContext } from './app/startup';
+import { LetDemo } from './app/molecules/let-demo/let-demo';
 
-describe('app', function () {
-
+describe('app', function() {
   function createTestFunction(testFunction: (ctx: TestExecutionContext) => Promise<void> | void) {
-    return async function () {
+    return async function() {
```
fkleuver: We do have the convention of space between function keyword and opening paren btw. There's just no eslint rule for it atm

packages/__tests__/integration/integration.spec.ts
Sayan751: Fixed

packages/__tests__/integration/app/app.ts
Sayan751: Done

packages/__tests__/integration/integration.spec.ts
Sayan751: done

PR #729, opened 2019-10-28
chore(all): --save is removed for npm install

3cp:
Also cleanup --save-dev to shorter -D. Fixed few wrong install commands, for example webpack-loader should be a dev dep.


fkleuver:
If we're going the shortcut route why not abbreviate `install` to `i` as well? I always type `npm i -D something`, `npm i --save something`, `npm i -g something`, etc

3cp:
Yes, why not.

3cp:
I don't understand the failure on unit_test_node. Something wrong with current master head?

fkleuver:
@3cp Looks to me like it's all working. `js_framework_benchmark` is just broken atm for unrelated reasons. Is this ready to merge?

3cp:
Yes. Ready for me. But have a look first, no much changed.

fkleuver:
Haha that's a ton of completely unrelated changes in various build scripts too :P
Can't do any harm though I guess. It works universally. So I'll merge. Thanks!

PR #730, opened 2019-10-28
feat(replace): Nested replaceables didn't render

brandonseydel:
# Pull Request

## 📖 Description

There was an issue with the containerless option when nesting deep.  Need to check and make sure the parent element of the only child is a custom element.

## 📑 Test Plan

Added a failing test that now passes.


packages/jit-html/src/template-binder.ts
```diff
@@ -400,14 +400,14 @@ export class TemplateBinder {
...
-      if(type === 'checkbox' || type === 'radio') {
+      if (type === 'checkbox' || type === 'radio') {
         this.ensureAttributeOrder(manifest);
       }
     }
     processTemplateControllers(this.dom, manifestProxy, manifest);
 
     let replace = node.getAttribute('replace');
-    if (replace === '' || replace === null && manifestRoot !== null && manifestRoot.isContainerless) {
+    if (replace === '' || (replace === null && manifestRoot !== null && manifestRoot.isContainerless && ((parentManifest.flags & 16 /* isCustomElement */) > 0))) {
```
fkleuver: Can you use the actual flags please. I think it's `SymbolFlags`?

```suggestion
    if (replace === '' || (replace === null && manifestRoot !== null && manifestRoot.isContainerless && ((parentManifest.flags & SymbolFlags.isCustomElement) > 0))) {
```

AureliaEffect:
JS Framework Benchmark: https://34638-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

Issue #736, opened 2019-11-01
Component Registration Doc is inaccurate


eiswind:
https://github.com/aurelia/aurelia/blob/master/docs/user-docs/getting-started/components.md

says

```
Aurelia
  .register(
    NameTag // Here it is!
  )
  .app(App)
  .start();
```

Which does not compile because of register()'s type annotations.

Using

```
Aurelia.app(MyApp).register(NameTag).start();
```
does compile but does not work for me. If NameTag does have to implement IRegistry this could be documented.

fkleuver:
> If NameTag does have to implement IRegistry this could be documented.

It doesn't. `CustomElement.define` or the `@customElement` decorator do this indirectly for you, so that's a little typing issue left over from earlier refactors that I'll address shortly.

The docs currently only address the convention-based approach for which the webpack loader is required. Is your project generated from the cli or did you create it by hand? Could you share the project via a repo?

3cp:
@fkleuver tsc does not know the "correct" type, because the editor does the type check with original code.

Even with the error in editor, the webpack build should work without issue, because tsc will see modified code at compile time.

I will double check my guessing. We might need to relax the type in Aurelia subclass defined in "aurelia" package.

3cp:
I just checked. This is annoying.

It is where conventions failed. It is from a TS project conventions.

When webpack starts bundling, it starts with `src/main.ts` where NameTag is checked. The tsc will check the type by reading `src/name-tag.ts` because main.ts asks for it.

But at this point of time, `src/name-tag.ts` is not modified because webpack has not started processing it. This is all tsc doing extra reading for type check.

I can think of two possible solutions, both are not very idea.
1. pass an option to tsc in webpack.config.js to turn off cross-file type checking. This requires we add another npm script "lint" to run real type check before "build".
2. relax the typing of Aurelia subclass defined in "aurelia" package.

fkleuver:
@3cp as I said:

> that's a little typing issue left over from earlier refactors that I'll address shortly.

`.register` shouldn't have any restrictions on it, because things don't need to implement `IRegistry` to work in DI. Plain classes are jit registered as singletons, and resources are associated via metadata definitions which have a `register` method. So in both those valid cases (valid at build time and valid at runtime) they wouldn't conform to the api, meaning the api signature is wrong.

So the typing is an easy fix. What this is about is why his app does not work at runtime.

> Using

> Aurelia.app(MyApp).register(NameTag).start();

[...]

PR #739, opened 2019-11-02
fix(aurelia): update static register method signature to match parent class signature

3cp:
Partially addresses #736


EisenbergEffect:
@fkleuver Should this use `any` or `unknown`?

3cp:
Replaced by #740

fkleuver:
Oh sorry didn't notice this PR. But in general `unknown` is better for public api *inputs* because it allows the user to pass in anything (just like with `any`) but it forces us to perform some type checking and/or casting. So, it's the same for the user, but better for us in terms of type safety.

PR #746, opened 2019-11-05
route decorator WIP/poc

fkleuver:
# Pull Request

### DIFF
--- a/packages/router/src/routeable-component.ts
+++ b/packages/router/src/routeable-component.ts
@@ -1,0 +1,169 @@
+import { IIndexable, Constructable, ResourceType, IResourceKind, Protocol, Metadata, ResourceDefinition, mergeArrays, IContainer, Registration, PartialResourceDefinition } from '@aurelia/kernel';
+import { IViewModel, registerAliases, CustomElement } from '@aurelia/runtime';
+
+export interface IRouteConfig {
+  readonly name?: string;
+  readonly id?: string;
+  readonly path?: string | readonly string[];
+  readonly viewport?: string | null;
+  readonly nav?: boolean;
+  readonly meta?: IIndexable;
+}
+
+export type PartialRouteableComponentDefinition = Partial<PartialResourceDefinition<IRouteConfig & {
+  readonly routes?: readonly IRouteConfig[];
+}>>;
+
+export type RouteableComponentType<T extends Constructable = Constructable> = ResourceType<T, IViewModel & (T extends Constructable<infer P> ? P : {}), PartialRouteableComponentDefinition>;
+export type RouteableComponentKind = IResourceKind<RouteableComponentType, RouteableComponentDefinition> & {
+  isType<T>(value: T): value is (T extends Constructable ? RouteableComponentType<T> : never);
+  define<T extends Constructable>(def: PartialRouteableComponentDefinition, Type: T): RouteableComponentType<T>;
+  getDefinition<T extends Constructable>(Type: T): RouteableComponentDefinition<T>;
+  annotate<K extends keyof PartialRouteableComponentDefinition>(Type: Constructable, prop: K, value: PartialRouteableComponentDefinition[K]): void;
+  getAnnotation<K extends keyof PartialRouteableComponentDefinition>(Type: Constructable, prop: K): PartialRouteableComponentDefinition[K];
+};
+
+export type RouteableComponentDecorator = <T extends Constructable>(Type: T) => RouteableComponentType<T>;
+
+/**
+ * Decorator: Indicates that the decorated class is a routeable component.
+ *
+ * It can be navigated *to*
+ */
+export function routeableComponent(def: PartialRouteableComponentDefinition): RouteableComponentDecorator {
+  return function (target) {
+    return RouteableComponent.define(def, target);
+  };
+}
+
+/**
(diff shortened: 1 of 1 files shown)


packages/router/src/routeable-component.ts
```diff
@@ -0,0 +1,169 @@
+import { IIndexable, Constructable, ResourceType, IResourceKind, Protocol, Metadata, ResourceDefinition, mergeArrays, IContainer, Registration, PartialResourceDefinition } from '@aurelia/kernel';
+import { IViewModel, registerAliases, CustomElement } from '@aurelia/runtime';
+
+export interface IRouteConfig {
+  readonly name?: string;
+  readonly id?: string;
```
davismj: is id different from name somehow

**edit**
oh is this the component?

packages/router/src/routeable-component.ts
```diff
@@ -0,0 +1,169 @@
...
+ * It can be navigated *to*
+ */
+export function routeableComponent(def: PartialRouteableComponentDefinition): RouteableComponentDecorator {
+  return function (target) {
+    return RouteableComponent.define(def, target);
+  };
+}
+
+/**
+ * Decorator: Indicates that the decorated class is has the specified config(s) as child route(s).
```
davismj: is has is introducing confusion. 

This is how I would expect it to work:

```
@route({ path: ['', 'home'] })
export class HomeViewModel
```

And when the user navigates to '' or 'home', HomeViewModel is loaded in the default viewport. I would not expect that this would be interpreted that HomeViewModel has a child route.

packages/router/src/routeable-component.ts
```diff
@@ -0,0 +1,169 @@
...
+  if (Array.isArray(input)) {
+    return input.slice();
+  } else if (input === void 0) {
+    return [];
+  } else {
+    return [input as T];
+  }
+}
+
+export class RouteableComponentDefinition<T extends Constructable = Constructable> implements ResourceDefinition<T, IViewModel, PartialRouteableComponentDefinition> {
```
davismj: redirect may or may not belong in here. what do you think?

packages/router/src/routeable-component.ts
fkleuver: Yeah essentially the `name` would be the component name (perhaps rename to `componentName`?). The `id` is the route config id, the thing you would use in `route-href`

packages/router/src/routeable-component.ts
fkleuver: Oops. Is meant to say:

```ts
/**
 * Decorator: Indicates that the decorated class has the specified config(s) as child route(s).
 */
```

Anyway, so according to your expectation perhaps the two should be turned around.

In v1 we don't really have api equivalent of `@route`. You configure those at the component that is a parent relative to the configured route. And you do this in `configureRouter`.

Now we can't just blatantly replace the `configureRouter` idea with `@route` because that leaves a hole in the API that v1 used to fill: configuring child routes.

So to illustrate with the proposed decorators:

```ts
@route({ name: 'home' }, { name: 'settings' })
export class App {}

@routeableComponent({ name: 'home', path: ['', 'home'] })
export class HomePage {}

@routeableComponent({ name: 'settings' })
@route({ name: 'profile' }, { name: 'security' })
export class SettingsPage {}

@routeableComponent({ name: 'security' })
export class ProfileSettingsPage {}

@routeableComponent({ name: 'profile' })
export class SecuritySettingsPage {}
```

So, turning that around, would be e.g.:


```ts
@childRoute({ name: 'home' }, { name: 'settings' })
export class App {}

@route({ name: 'home', path: ['', 'home'] })
export class HomePage {}

@route({ name: 'settings' })
@childRoute({ name: 'profile' }, { name: 'security' })
export class SettingsPage {}

@route({ name: 'security' })
export class ProfileSettingsPage {}

@route({ name: 'profile' })
export class SecuritySettingsPage {}
```

Either way, we need something for both, wouldn't you agree?

packages/router/src/routeable-component.ts
fkleuver: Yeah, it does. Either I missed it or it was missing in one of your RFCs. But I'll make sure to add it

packages/router/src/routeable-component.ts
jwx: A component is a `@route` or `@routeable` (either works fine for me). A component can also define `@routes`. I think that's simple enough.

davismj: Turn it around. 👍
Need both. 😕

The `configureRouter` method does more than configure routes. So to replace it with basically a static property of route configurations means we lose some stuff that we will have to relocate. I can't remember if I relocated them or not. 

Either way, the `configureRouter` method usually has a `map` call with anywhere from 10 to 100 lines of routes depending on the app size. That's not a structure that lends itself well to a decorator.

Decorator is OK, but I don't think most will get as much use out of it as a static property alternative. Either way I think we may need a functional router configuration approach.

Fair?

@jwx I agree, both work really well. `@routes` is a nice name, if we do decide to go down that path, although just a touch close `@route`.

jwx:
```typescript
this.router.addRoutes([ { route }, { route }, ... ]);
```
instead of `map` within `configureRouter`.

davismj: Right. Decorator doesn't feel good for that.

PR #752, opened 2019-11-07
Bindable observer

### LINKED ISSUE #368: Type awareness and reflection for @bindable and @observable
# 💬 RFC

The following is a summary of how I'd like to see the `@bindable` and `@observable` evolve.
Specifically, it covers:

- The need for type awareness and coercion.
- The need for reflection of the value back to the DOM.
- The need for special handling of `boolean` attributes.

**Edit:** Important clarification based on the discussion below:
> The coercion I'm talking about here, is only relevant in terms of parsing an initial, hardcoded attribute value - and reflecting the current property value back to the attribute in the DOM.
I'm not suggesting any changes to how the actual value is exchanged between two properties bound to each other.

## 🔦 Context - and proposed solution

Let's say you have this component today:

```ts
export class Foo
{
    @bindable
    public number: number;
}
```
```html
<template>${value + 2}</template>
```

And you use it like this:

```html
<foo value="2"></foo>
```

You would expect this to show the number `4`, because `2 + 2 = 4`, right?
You would be wrong.

It shows the string `"22"`, because even though the property is declared as a `number`, `@bindable` is unaware of that, and therefore assigns the value as a `string`. A similar situation arises if the property was declared as `boolean`, where a value of `false` would end up as a string - which would actually be truthy.

This lack of type awareness can lead to some really nasty bugs, which due to automatic type conversion can easily go unnoticed - that `"22"` can easily end up as a number somewhere else, causing incorrect data or behavior, with no exceptions thrown to draw your attention to it.

This can be worked around by writing e.g `value.bind="2"` instead, but that's verbose and easy to forget.
Also, when the value is bound this way, we no longer have an attribute named `value` on our element - which would be a problem if you wanted to style the element differently based on the presence or value of that attribute. Another example of such a situation is the `disabled` attribute - more about that below.

#### Why the need for change?

Because literally everyone expects this to just work, and forgetting to convert the type has by far been our largest source of bugs - it's just way too easy to forget, and the bugs can be extremely nasty.
It's also super annoying, as the code to convert the values pollute our otherwise simple controllers.

This stuff is so tightly coupled to the binding internals, that I'm personally not comfortable taking a dependency on a third party plugin, which may or may not be updated to work with the latest version of the framework. This needs to be part of Aurelia itself.

I've seen other suggestions that Aurelia should just be smarter about it, and know how to deal with e.g. the `disabled` attribute, similar to how it has special logic for dealing with `aria-` attributes. I think this would be the wrong way to go though - it introduces too much magic and special cases, and it would not work for custom elements, as they might need non-standard attributes. This needs to be more configurable.

#### Support custom coercers

The `@bindable` decorator should support an option to specify which converter to use, which may be a named coercer registered somewhere, or a specific coercer class or instance - which supports coercion to (and from - more about that below) the attribute value.

```ts
@bindable({ coerce: 'custom' })
public foo: Custom;

@bindable({ coerce: CustomCoercer })
public foo: Custom;

@bindable({ coerce: customCoercer })
public foo: Custom;
```

**Edit:** From the discussion below, a note about build in coercers and an example:

Aurelia should, of course, come with coercers for the simple types `boolean` and `number` - and maybe a few more, such as `date`. So if that's all you need, there will be no need to create any coercers yourself. And if you're using TypeScript, it will use those automatically, as discussed next.

But if you have, say, a `Point` class with an `x` and `y` coordinate, and you'd like to be able to write `<foo position="0 42">` in your view template to get `{ x: 0, y: 42}`, then you'd create your own `PointValueCoercer` and maybe even register it globally with the name `point`. Then you can just annotate your property with `@bindable({ coerce: 'point' })`, and things will just work.

####  Leverage TypeScript metadata

The `@bindable` decorator should leverage TypeScript metadata, if available, to automatically determine which coercer to use - at least for simple types like `string`, `number` and `boolean`.
That way, this will just work as expected:

```ts
@bindable
public foo: number;
```

This can only be done for primitive types, but that probably covers 95% of the use cases for this.
For the rest, a custom coercer would need to be specified.

#### Support reflection back to the DOM

The `@bindable` decorator should have an option to reflect the value back to the DOM, for styling pourposes. That way, this:

```ts
@bindable({ reflect: true })
public theme: string = "success";

public message: string = "This is a success message";
```
```html
<message theme.bind="theme">${message}</message>
```
Would result in this DOM:
```html
<message theme="success">This is a success message</message>
```
Where we can use this CSS:
```css
[theme="success"] { color: green; }
[theme="error"] { color: red; }
```

#### Support reflection of boolean attributes

Reflection back to the DOM should support the special needs of boolean attributes
(Note: This is copied from https://github.com/aurelia/binding/issues/347#issuecomment-242106920)

Let's take the `disabled` attribute on input elements as an example - that is a `boolean` attribute with _no value_ - it's the fact that it is _present_ on the element that makes it `true`.

From the spec:

> A number of attributes are boolean attributes. The presence of a boolean attribute on an element represents the true value, and the absence of the attribute represents the false value.
> 
> If the attribute is present, its value must either be the empty string or a value that is an ASCII case-insensitive match for the attribute's canonical name, with no leading or trailing whitespace.
> 
> NOTE: The values "true" and "false" are not allowed on boolean attributes. To represent a false value, the attribute has to be omitted altogether.
> 
> https://www.w3.org/TR/html5/infrastructure.html#boolean-attributes

This  add/remove behavior could be achieved using a custom binding behavior, but I'd say it should be `@bindable` that controls whether this behavior is used. Think of a custom input element, which comes with styles for when the element is disabled - in this case, the CSS distributed with that element might apply different styles depending on the presence of the `disabled` attribute. It would be highly error prone if we had to remember to use a binding behavior every time we bind a `disabled` attribute, in order to make that styling work.

So, I'd say this property, which we know is `boolean`

``` typescript
class Foo
{
    @bindable
    public disabled: boolean;
}
```

Should be `true` for:

``` html
<foo disabled></foo>
<foo disabled=""></foo>
<foo disabled="disabled"></foo>
<foo disabled.bind="true"></foo>
```

And `false` for:

``` html
<foo></foo>
<foo disabled.bind="false"></foo>
```

And if `reflect` is enabled for that `@bindable`, I'd say it should actually default to the add/remove behavior for `boolean` properties.

So, while the element in my markup might be written as:

``` html
<foo disabled.bind="true">I'm disabled</foo>
<foo disabled.bind="false">I'm enabled</foo>
```

The element in the DOM should be:

``` html
<foo disabled>I'm disabled</foo>
<foo>I'm enabled</foo>
```

That way CSS like this will always work as expected:

``` css
foo[disabled] { color: gray }
```

Note that boolean `aria-` attributes are special in their own way - they require a `yes` or `no` value, which is why coercion might need to be at least a little bit configurable. I think the add/remove behavior should be the default though.

#### How a custom coercer might look

**Edit:** I've edited this part in response to the discussion below.
Here's some initial thoughts on how the interface of a coercer might look.

```ts
interface IValueCoercer<TValue>
{
    /** Converts an attribute value to a model value.
     * @param value The string value from the attribute in the DOM,
     * or undefined if the attribute is not specified.
     * @returns The value to assign to the property in the model.
     */
    fromAttribute?(value: string | undefined): TValue;

    /** Converts a model value to an attribute value.
     * @param value The value assigned to the property in the model.
     * @returns The string to use as the attribute value in the DOM,
     * or undefined to remove the attribute.
     */
    toAttribute?(value: TValue): string | undefined;
}
```

While this might *appear* very similar to the existing Value Converter concept, it really is different. It may have the same two methods, but their *behavior* is different - here it is specific to attributes:

* For example, a `DateValueConverter` would probably convert to/from a human readable string, while a `DateValueCoercer` would probably convert to/from an ISO date string.

* For example, a `BooleanValueConverter` might convert to/from a human readable string, such as `"True"` or `"False"`, while a `BooleanValueCoercer` would return either `""` or `undefined`, thus toggling the *presence* of the attribute on the element.

Mixing those two concepts would be very wrong.

Thoughts on all this?

bigopon:
# Pull Request

## 📖 Description

Provide the ability to config bindable with getter/setter interceptors. Note that those config are purely interceptors, and expected to be pure functions. Usage example would be:

Defining a setter for altering incoming value:
```ts
export class MyEl {
  @bindable({ set: Number })
  public value: number

  @bindable({ set: val => Number(val) || 0 })
  public credit: number
}
```
Then in template, the scenarios like following will work as expected:
```html
<input type="number" value.bind="value">
<input type="number" value.bind="credit">
```
Or in code:
```ts
myEl.value = '5';
// myEl.value === 5

myEl.value = 'a';
// myEl.value === NaN

myEl.credit = '1';
// myEl.credit === 1;

myEl.credit = 'a';
// myEl.credit === NaN
```

Defining a getter:
```ts
export class MyEl {

  @bindable({ get: val => Number(val) || 0 })
  public age: number
}
```

### 🎫 Issues

Closes #368 

## 👩‍💻 Reviewer Notes

Is getter necessary?

## 📑 Test Plan

TBU

## ⏭ Next Steps

N/A

### DIFF
--- a/packages/__tests__/2-runtime/property-observation.spec.ts
+++ b/packages/__tests__/2-runtime/property-observation.spec.ts
@@ -2,7 +2,7 @@ import { PLATFORM, Primitive, IIndexable } from '@aurelia/kernel';
 import {
   LifecycleFlags as LF,
   PrimitiveObserver,
-  SelfObserver,
+  BindableObserver,
   SetterObserver
 } from '@aurelia/runtime';
 import { SpySubscriber, assert, TestContext, ChangeSet } from '@aurelia/testing';
@@ -180,11 +180,11 @@ describe('SetterObserver', function () {
   });
 });
 
-describe('SelfObserver', function () {
+describe('BindableObserver', function () {
   function setup(flags: LF, obj: IIndexable, key: string) {
     const ctx = TestContext.createHTMLTestContext();
     const lifecycle = ctx.lifecycle;
-    const sut = new SelfObserver(lifecycle, flags, obj, key, `${key ? key.toString() : `${key}`}Changed`);
+    const sut = new BindableObserver(lifecycle, flags, obj, key, `${key ? key.toString() : `${key}`}Changed`);
 
     return { sut };
   }
--- a/packages/__tests__/2-runtime/property-observation.spec.ts
+++ b/packages/__tests__/2-runtime/property-observation.spec.ts
@@ -184,7 +184,7 @@ describe('BindableObserver', function () {
   function setup(flags: LF, obj: IIndexable, key: string) {
     const ctx = TestContext.createHTMLTestContext();
     const lifecycle = ctx.lifecycle;
-    const sut = new BindableObserver(lifecycle, flags, obj, key, `${key ? key.toString() : `${key}`}Changed`);
+    const sut = new BindableObserver(lifecycle, flags, obj, key, `${key ? key.toString() : `${key}`}Changed`, PLATFORM.noop, PLATFORM.noop);
 
     return { sut };
   }
--- a/packages/__tests__/2-runtime/property-observation.spec.ts
+++ b/packages/__tests__/2-runtime/property-observation.spec.ts
@@ -184,7 +184,7 @@ describe('BindableObserver', function () {
   function setup(flags: LF, obj: IIndexable, key: string) {
     const ctx = TestContext.createHTMLTestContext();
     const lifecycle = ctx.lifecycle;
-    const sut = new BindableObserver(lifecycle, flags, obj, key, `${key ? key.toString() : `${key}`}Changed`, PLATFORM.noop, PLATFORM.noop);
+    const sut = new BindableObserver(lifecycle, flags, obj, key, `${key ? key.toString() : `${key}`}Changed`, PLATFORM.noop);
 
     return { sut };
   }
--- a/packages/__tests__/5-jit-html/custom-attributes.spec.ts
+++ b/packages/__tests__/5-jit-html/custom-attributes.spec.ts
@@ -255,9 +255,9 @@ describe('custom-attributes', function () {
 
     it('does not invoke chane handler when starts with two way binding', function () {
       const { fooVm, tearDown } = setupChangeHandlerTest(`<div foo.two-way="prop"></foo>`);
-      assert.strictEqual(fooVm.propChangedCallCount, 0);
+      assert.strictEqual(fooVm.propChangedCallCount, 0, '#1 should have had 0 calls at start');
       fooVm.prop = '5';
-      assert.strictEqual(fooVm.propChangedCallCount, 1);
+      assert.strictEqual(fooVm.propChangedCallCount, 1, '#2 shoulda had 1 call after mutation');
       tearDown();
     });
 
@@ -313,10 +313,6 @@ describe('custom-attributes', function () {
       public propChanged(): void {
         this.propChangedCallCount++;
       }
-      public propertyChanged(...args: unknown[]): void {
-        this.propertyChangedCallCount++;
-        this.propertyChangedCallArguments.push(args);
-      }
     }
 
     @customAttribute('foo3')
@@ -326,9 +322,6 @@ describe('custom-attributes', function () {
       public propChangedCallCount: number = 0;
       public propertyChangedCallCount: number = 0;
       public propertyChangedCallArguments: unknown[][] = [];
-      public propChanged(): void {
-        this.propChangedCallCount++;
-      }
       public propertyChanged(...args: unknown[]): void {
         this.propertyChangedCallCount++;
         this.propertyChangedCallArguments.push(args);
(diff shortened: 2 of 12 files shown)


AureliaEffect:
JS Framework Benchmark: https://34567-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://35840-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://35843-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://35877-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://35889-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

PR #755, opened 2019-11-08
Emit decorator metadata

fkleuver:
# Pull Request

## 📖 Description

Previously we threw an error when the container gets `Object` as a key, but that turned out to have a few unintended side effect, such as not being able to do partial dependency resolution.

So, now we just report a warning instead, and we also do this on other natives (e.g. `Array`, `HTMLElement`, etc)

There is also a new exported helper function `isNativeFunction`, used by the container for the above, that will be useful in other parts of the framework and possibly also for end users.

Furthermore, there are a few issues with `emitDecoratorMetadata` in nodejs that I'm addressing now, because they're making things difficult in AOT.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -243,7 +243,9 @@ module.exports = {
       'Promise',
       'Map',
       'Set',
-      'Object.assign'
+      'Object.assign',
+      'WeakMap',
+      'WeakSet'
     ]
   }
 };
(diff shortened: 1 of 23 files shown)


AureliaEffect:
JS Framework Benchmark: https://34603-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

PR #764, opened 2019-11-14
Ci verbose logging for unit tests

fkleuver:
# Pull Request

## 📖 Description

Enables verbose logging for unit tests on circle ci. This makes it easier to see where and when a test run fails.
Also prevents circleci from timing out when a job runs for more than 10 minutes without logging anything (which may be a false negative)


AureliaEffect:
JS Framework Benchmark: https://35342-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

PR #765, opened 2019-11-14
perf(collection-observation): store observers in weakmaps

fkleuver:
# Pull Request

## 📖 Description

Stores collection observers in module-scoped weakmaps instead of as a property on the collections themselves.
This should help the garbage collector out a bit and reduce the chances of accidental memory leaks


AureliaEffect:
JS Framework Benchmark: https://35335-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

PR #766, opened 2019-11-14
switch to weakmap-based observer lookups

fkleuver:
# Pull Request

## 📖 Description

This PR:
- Is a **BREAKING CHANGE**: the `$observers` property no longer exists on binding contexts (or any object with observed properties for that matter)
- Uses weakmap lookups instead of directly creating an `$observers` property on each object
- Exposes new APIs for managing cached observers: `getObserver`, `setObserver`, `hasObserver`
- Should improve starup performance in the majority cases due to fewer total objects created: a global map keeps a weakmap per property name, rather than per object. Also helps the garbage collector out and reduces polymorphism
- Cleans up a bunch of messy types that were related to the `ObserversLookup`


AureliaEffect:
JS Framework Benchmark: https://35352-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://35421-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

fkleuver:
Abandoning this for now because it's not an urgent concern and probably needs more work and tests. Memory leak discovered recently was already fixed anyway.

AureliaEffect:
JS Framework Benchmark: https://35686-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

PR #770, opened 2019-11-14
chore: update and unpin http-server

BBosman:
# Pull Request

## 📖 Description

Update and unpin `http-server`.

### 🎫 Issues

n/a

## 👩‍💻 Reviewer Notes

<https://www.npmjs.com/advisories/830/versions> was fixed and `http-server` updated to the fixed version, so no more need for pinning.

## 📑 Test Plan

CircleCI.

## ⏭ Next Steps

n/a

### DIFF
--- a/examples/jit-browserify-ts/package.json
+++ b/examples/jit-browserify-ts/package.json
@@ -20,7 +20,7 @@
     "browserify": "latest",
     "fancy-log": "^1.3.3",
     "gulp": "^4.0.2",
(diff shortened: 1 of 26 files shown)


fkleuver:
As discussed, for `http-server` we need to stay on `0.9.0`. The latest version is still broken. Not really worth the time to ever look into upgrading imho. We might just toss it at some point in favor of our own simple version.

PR #771, opened 2019-11-14
chore: migrate from validate-commit-msg to commitlint

BBosman:
# Pull Request

## 📖 Description

The `validate-commit-msg` package has been deprecated, so switch to `commitlint`.

### 🎫 Issues

n/a

## 👩‍💻 Reviewer Notes

Configuration should be identical, so it should be a drop in replacement.

Type is mandatory and must be from a predefined list.
Scope is optional and can be anything
Description is mandatory.

Per request I made the config explicit instead of using `@commitlint/config-conventional`.

## 📑 Test Plan

CircleCI

I also tried to commit some local commits with invalid commit messages and they were rejected as before.

## ⏭ Next Steps

n/a

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -201,19 +201,342 @@
         }
       }
     },
+    "@commitlint/cli": {
+      "version": "8.2.0",
+      "resolved": "https://registry.npmjs.org/@commitlint/cli/-/cli-8.2.0.tgz",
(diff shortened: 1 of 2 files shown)


BBosman:
Done.

AureliaEffect:
JS Framework Benchmark: https://35623-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

PR #778, opened 2019-11-20
feat(runtime): add CustomElement.createInjectable api

fkleuver:
# Pull Request

## 📖 Description

Adds a new api that allows you to create an injection token for custom element instances so that you can inject the parent instance into a child, even recursively into itself.

This works by telling the rendering engine to not cache anything when the `injectable` property is specified, which may incur a large performance penalty on large components that require compilation because it will cause it to be re-compiled on every render. So, use with care.

However, for cheap-to-compile components (or those that don't have a template at all, such as `au-viewport` and `au-compose`), this should be fine.

Usage:

```ts
export const IMyElement = CustomElement.createInjectable();

@customElement({
  name: 'my-element',
  template: '<my-element-child></my-element-child>'
  injectable: IMyElement
})
export class MyElement {}

@customElement({ name: 'my-element-child' })
export class MyElementChild {
  constructor(@IMyElement private myElement: MyElement) {}
}
```

Also see the newly added integration test. Works recursively, non-recursively, with other elements in-between, with template controllers in-between, etc.

### DIFF
--- a/packages/__tests__/5-jit-html/injectable.spec.ts
+++ b/packages/__tests__/5-jit-html/injectable.spec.ts
@@ -1,0 +1,99 @@
+import { RouterConfiguration } from '@aurelia/router';
+import { CustomElement, Aurelia, customElement } from '@aurelia/runtime';
+import { TestContext, assert } from '@aurelia/testing';
+
+describe('CustomElement.createInjectable', function () {
+  it('properly links parent-child', async function () {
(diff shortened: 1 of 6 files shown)


AureliaEffect:
JS Framework Benchmark: https://36426-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

PR #784, opened 2019-12-08
fix(container): ignore primitive values in register

fkleuver:
# Pull Request

## 📖 Description

When primitive values are passed into `register()` this may cause unpredictable results, such as `Metadata` throwing an error, even when the end user has no control over that (such as when passing in a synthetically created module object by a bundler).

This PR causes the DI container to simply ignore those invalid values. I chose to do this in the top-level registration loop (as opposed to only in the inner `Object.keys` loop) so as to also make the container a bit more forgiving when passing in an array of registrations where one or more values is invalid. Being the most front-facing surface api, it should probably be forgiving in these scenarios.

We may want to let the reporter log a warning in this case in the future for good measure.

## 📑 Test Plan

Added tests for all primitive values shapes in the different ways they can be passed in

### DIFF
--- a/packages/kernel/src/di.ts
+++ b/packages/kernel/src/di.ts
@@ -735,6 +735,9 @@ export class Container implements IContainer {
     let jj: number;
     for (let i = 0, ii = params.length; i < ii; ++i) {
       current = params[i];
+      if (typeof current !== 'object' && typeof current !== 'function' || current === null) {
+        continue;
+      }
       if (isRegistry(current)) {
         current.register(this);
       } else if (Protocol.resource.has(current)) {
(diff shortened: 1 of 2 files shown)


packages/kernel/src/di.ts
```diff
@@ -756,6 +759,9 @@ export class Container implements IContainer {
         jj = keys.length;
         for (; j < jj; ++j) {
           value = current[keys[j]];
+          if (typeof value !== 'object' && typeof value !== 'function' || value === null) {
```
EisenbergEffect: Should we factor this out into a reusable function to remove the duplication?

fkleuver: Yeah good point, I wanted to avoid the fn call overhead but a small fn like `isObject` is probably quickly inlined anyway :)

AureliaEffect:
JS Framework Benchmark: https://37469-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

PR #802, opened 2020-01-05
feat(custom-attr): add no multi bindings cfg

bigopon:
# Pull Request

## 📖 Description

Add a new config for custom attribute to have ability to always parse their value as single binding. An example for this would be `goto` attribute:
```html
<a goto="http://outside.com"></a>
```

### 🎫 Issues

N/A

## 👩‍💻 Reviewer Notes

N/A

## 📑 Test Plan

N/A

## ⏭ Next Steps

Plan to spec dynamic options properly to complete custom attribute features

### DIFF
--- a/packages/__tests__/5-jit-html/custom-attributes.spec.ts
+++ b/packages/__tests__/5-jit-html/custom-attributes.spec.ts
@@ -8,6 +8,7 @@ import {
 import { assert, eachCartesianJoin, createFixture } from '@aurelia/testing';
 
 describe('custom-attributes', function () {
+  this.afterEach(assert.isSchedulerEmpty);
   // custom elements
   describe('01. Aliases', function () {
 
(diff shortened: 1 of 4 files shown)


AureliaEffect:
JS Framework Benchmark: https://39571-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

EisenbergEffect:
@bigopon Does the `goto` custom attribute have more than one bindable?

* If no, then there's no need for a config option. If there's only one bindable, it shouldn't support the multi-binding syntax.
* If yes, then how will it support setting those other bindables without multi-binding syntax?

bigopon:
@EisenbergEffect `goto` has only 1 bindable, though I think this is about general consistency: even with only one bindable, multi binding syntax is still valid, and should be inferred correctly from the usage.

> If yes, then how will it support setting those other bindables without multi-binding syntax?

It can still be set by getting reference manually and do it, not sure if anyone would do it though.

EisenbergEffect:
I'm not convinced this should go in. I don't see a reason to add a config setting to ignore multi-bindings when an attribute with a single bindable will *never* have multi-bindings. I don't even think the syntax should be supported. There's no reason anyone would ever write that.

Can we re-evaluate this please?

PR #804, opened 2020-01-12
refactor(styles): a more explicit api for shadow styles and css modules

EisenbergEffect:
# Pull Request

## 📖 Description

This PR adds new `cssModules` and `shadowCSS` functions that return DI registries capable of adding the appropriate resources to a custom element for style handling. This moves us away from the weak string-based, runtime determined css strategy to an explicit model either encoded by the conventions system or provided manually. This ensures that 3rd parties built with a certain CSS strategy will not break at runtime based on app-level configuration. It also simplifies the code needed to provide style registration.

## 📑 Test Plan

* Existing tests have been fixed up to use the new implementation.

## ⏭ Next Steps

@3cp Sorry this took so long, but better late than never. I'll need you to fix up your conventions code to use the `cssModules` function and the `shadowCSS` function instead of the previous functions used to register styles. As part of this change over, we should also fix the issue we have with multiple css file imports.

Example generated code should be something like:

```TypeScript
import sharedCSS from './shared.css';
import css from './my-element.css';``
import { shadowCSS } from '@aurelia/runtime-html';`

export const dependencies = [
  shadowCSS(sharedCSS, css)
];
```

@3cp You'll need to make these changes on this branch so that it all gets merged together, otherwise things will break :) Let me know if you have any questions.


3cp:
If user uses neither shadowCSS nor cssModule, is it:

```js
export const dependencies = [
  sharedCSS, css
];
```

Or
```js
export const dependencies = [
  Registration.defer('.css', sharedCSS), Registration.defer('.css', css)
];
```

3cp:
I guess following startup code should also be removed.
```js
.register(StyleConfiguration.shadowDOM())
.register(StyleConfiguration.cssModulesProcessor())
```

3cp:
I read your test code, maybe only for shadowCSS, we need a place holder.
```js
.register(StyleConfiguration.shadowDOM({
  sharedStyles: [
    // add your rootStyles
  ]
}));
```

Note, for following to work, webpack config has to skip style-loader for every css resources (even when they are loaded from js file).
```js
import css from './foo.scss';
//...
.register(StyleConfiguration.shadowDOM({
  sharedStyles: [ css ]
}));
```

EisenbergEffect:
@3cp For now, if they don't use css modules or shadow css, I guess we should generate:

```TypeScript
export const dependencies = [
  Registration.defer('.css', sharedCSS, css)
];
```

Though I'm a bit torn on that vs. only generating the import. Technically, defer is a noop if there's no handler registered. But, it seems like wasted processing. I'm not sure what scenario would warrant it.

The old startup code can be removed for sure, and we can add the placeholder for shadow css you recommend above.

packages/__tests__/plugin-conventions/preprocess-html-template.spec.ts
```diff
@@ -62,6 +62,28 @@ export function register(container) {
     assert.equal(result.code, expected);
   });
 
+  it('processes template with multiple css', function () {
```
3cp: @EisenbergEffect please view the changes in the tests for intended behaviour.

3cp:
Question: is that multi-defer call `Registration.defer('.ext', a, b, c)` functional right now?

EisenbergEffect:
@3cp That should be functional, it's just that nothing uses the defer stuff with the new styles system.

packages/__tests__/plugin-conventions/preprocess-html-template.spec.ts
EisenbergEffect: Looks good!

AureliaEffect:
JS Framework Benchmark: https://40082-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/runtime-html/src/styles/shadow-dom-registry.ts
```diff
@@ -1,21 +1,74 @@
...
   }
+
+  public static createStyleFactory(container: IContainer): IShadowDOMStyleFactory {
+    const dom = container.get(HTMLDOM);
+
+    if (AdoptedStyleSheetsStyles.supported(dom)) {
+      const styleSheetCache = new Map();
+
+      return {
+        createStyles(localStyles, sharedStyles) {
```
fkleuver: Let's use classes + DI here? May be more verbose but I feel it would still be beneficial. e.g.

```ts

class AdoptedStyleSheetsStylesFactory {
  private readonly cache = new Map();

  public constructor(
    @IDOM private readonly dom: HTMLDOM,
  ) {}

  public createStyles(
    localStyles: CSSStyleSheet,
    sharedStyles: IShadowDOMStyles | null,
  ): IShadowDOMStyle {
    return new AdoptedStyleSheetsStyles(
      this.dom,
      localStyles,
      this.cache,
      sharedStyles,
    );
  }
}

class StyleElementStylesFactory {
  public constructor(
    @IDOM private readonly dom: HTMLDOM,
  ) {}

  public createStyles(
    localStyles: string[],
    sharedStyles: IShadowDOMStyles | null,
  ): IShadowDOMStyle {
    if (localStyles.some(x => typeof x !== 'string')) {
      throw new Error('Shadow DOM CSS must be a string.');

    }
    return new StyleElementStyles(
      this.dom,
      localStyles,
      sharedStyles,
    );
  }
}

  public static createStyleFactory(container: IContainer): IShadowDOMStyleFactory {
    if (AdoptedStyleSheetsStyles.supported(container.get(HTMLDOM))) {
      return container.get(AdoptedStyleSheetsStylesFactory);
    }
    return container.get(StyleElementStylesFactory);
  }
```

Actually, perhaps the caching in the callback registration won't be necessary then either because the instance is cached on the container already?

packages/__tests__/3-runtime-html/styles.spec.ts
```diff
@@ -197,35 +181,41 @@ describe('Styles', function () {
...
+
+      const component = new FooBar();
+      const controller = Controller.forCustomElement(
+        component as ICustomElementViewModel<HTMLElement>,
+        ctx.container.get(ILifecycle),
+        host,
+        ctx.container, void 0
+      );
+
+      const seq = { appendTo() { return; } };
```
fkleuver: Can just pass `NodeSequence.empty` directly into `project`. Would behave the same way :)

packages/__tests__/3-runtime-html/styles.spec.ts
```diff
@@ -197,35 +181,41 @@ describe('Styles', function () {
...
+        Registration.instance(
+          IShadowDOMStyles,
+          new StyleElementStyles(ctx.dom, [css], null)
+        )
+      );
+
+      const component = new FooBar();
+      const controller = Controller.forCustomElement(
+        component as ICustomElementViewModel<HTMLElement>,
+        ctx.container.get(ILifecycle),
```
fkleuver: Convenience shorthand already exists

```suggestion
        ctx.lifecycle,
```

packages/__tests__/3-runtime-html/styles.spec.ts
```diff
@@ -197,35 +181,41 @@ describe('Styles', function () {
...
+        host,
+        ctx.container, void 0
+      );
+
+      const seq = { appendTo() { return; } };
+      const projector = controller.projector as any;
```
fkleuver:
```suggestion
      const projector = controller.projector as ShadowDOMProjector;
```

EisenbergEffect:
Thanks for the feedback @fkleuver I'll get the changes in shortly.

AureliaEffect:
JS Framework Benchmark: https://40176-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

3cp:
To reduce merge conflicts, I will wait for this to be merged before working on `@customElement('name')` conventions, as that also touches the same file.

packages/__tests__/3-runtime-html/styles.spec.ts
```diff
@@ -197,35 +181,41 @@ describe('Styles', function () {
...
+      const ctx = TestContext.createHTMLTestContext();
+      const host = ctx.createElement('foo-bar');
+      const FooBar = CustomElement.define(
+        {
+          name: 'foo-bar',
+          shadowOptions: { mode: 'open' }
+        }
+      );
+      const css = '.my-class { color: red }';
+      const context = ctx.container.createChild();
```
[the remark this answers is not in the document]
EisenbergEffect: I'm leaving this as is for now because the test is broken. I think it's probably needed once the test is fixed.

packages/runtime-html/src/styles/shadow-dom-registry.ts
```diff
@@ -1,21 +1,74 @@
...
+            styleSheetCache,
+            sharedStyles
+          );
+        }
+      }
+    }
+
+    return {
+      createStyles(localStyles, sharedStyles) {
+        if (localStyles.find(x => typeof x !== 'string')) {
```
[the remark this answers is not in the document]
EisenbergEffect: Fixed.

packages/runtime-html/src/styles/shadow-dom-registry.ts
```diff
@@ -1,21 +1,74 @@
...
+  return new ShadowDOMRegistry(css);
+}
+
+export interface IShadowDOMStyleFactory {
+  createStyles(localStyles: (string | CSSStyleSheet)[], sharedStyles: IShadowDOMStyles | null): IShadowDOMStyles;
+}
+
+export const IShadowDOMStyleFactory
+  = DI.createInterface<IShadowDOMStyleFactory>('IShadowDOMStyleFactory')
+    .withDefault(x => x.callback((handler, requestor, resolver: any) => {
```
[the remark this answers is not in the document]
EisenbergEffect: Fixed.

packages/runtime-html/src/styles/shadow-dom-registry.ts
EisenbergEffect: Switched to factory methods. We can't get rid of the caching though because callback is called every time the key is requested.

packages/__tests__/3-runtime-html/styles.spec.ts
EisenbergEffect: Fixed.

packages/__tests__/3-runtime-html/styles.spec.ts
EisenbergEffect: Fixed.

packages/__tests__/3-runtime-html/styles.spec.ts
EisenbergEffect: Fixed.

AureliaEffect:
JS Framework Benchmark: https://41085-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

EisenbergEffect:
@fkleuver I think this is ready now. I tried to run the tests in node and there was an issue not coming from my code. I'm not sure if that's expected or not. How should we proceed?

AureliaEffect:
JS Framework Benchmark: https://42801-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://43269-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://43334-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://43353-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

PR #806, opened 2020-01-13
Update docs

fkleuver:
# Pull Request

## 📖 Description

As per @EisenbergEffect 's request, this PR updates the docs to reflect the renamed lifecycle hooks.

I also added preliminary descriptions of the newly added hooks `create`, `beforeCompile`, `afterCompile` and `afterCompileChildren`. Preliminary, because they are probably a bit hard to digest and only meaningful to people already familiar with some of the lower level api's. So, it's more of a brain dump than anything else.

@EisenbergEffect @bigopon Could you please look at these new descriptions and let me know if the hooks make sense after reading those (and any other feedback on the descriptions is also welcome)


bigopon:
I love the the imaginable possibilities from reading the doc, but probably we will need more concrete examples to illustrate everything for users who are less familiar.

AureliaEffect:
JS Framework Benchmark: https://40129-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

fkleuver:
@bigopon I agree we need some examples. As a curiosity, a good example that shows the power of the `create` hook is the router's `au-viewport-scope` (this use case was also the reason why I (re)defined that hook in the first place)

This is an element that is effectively turned into a template controller by using `create` to return the `default` part as a new containerless custom element definition. So this completely overrides the element's own definition, dynamically returning something else: https://github.com/aurelia/aurelia/blob/master/packages/router/src/resources/viewport-scope.ts#L52

I can see myself using `afterCompile` for building a component tree structure with a recursive component by registering the instance with a specific key, which is then injected in the closest child instance.

`beforeCompile` is probably closest to v1's `processContent`, with the caveat that the `template` is not necessarily parsed yet (perhaps we should make it so?)

Beyond `processContent`-like scenario's I can't really think of anything concrete that you would specifically do here rather than in `afterCompile`, because besides mutating the template the only thing that you can do here, but not in `afterCompile`, is change the registrations. But it is a useful enough capability in itself that it seems silly not to have it, considering how cheap it is to have it.
You might do stuff like conditionally register an alternative binding syntax or override whatever global resource registrations (though not sure why you would do that here rather than in the `define` api).

`afterCompileChildren`, timing-wise, is basically what was previously `created`. I personally don't know what I would use it for, but it does fill a temporal gap in the lifecycles.

EisenbergEffect:
Looks good @fkleuver Thank you for updating this! Feel free to merge at will 😄

PR #811, opened 2020-01-22
feat: validation

Sayan751:
# Pull Request

## 📖 Description

I am submitting the phase1 of the work done for validation plugin. This can be considered as merely the port of the v1 aurelia-validation plugin. The API is more or less the same as v1, with some breaking changes (see below). Following are some example of how this can be used.

**Registration**
```typescript
au.register(ValidationConfiguration)

// or 

au.register(ValidationConfiguration.customize((options) => {/* manipulate the options here */}))
```
There are 3 customization options available at this point: validator implementation, custom messages, and default trigger (for validate binding behavior)

**Define validation rules and instantiate validation controller**
```typescript
class MyVm {
  constructor(
    @IValidationRules private validationRules: IValidationRules,
    // this will be injected to child VMs
    @newInstanceForScope(IValidationController) private controller: IValidationController,
   ){}

  async someMethod() {
    const person = new Person(name, age);
    validationRules
      .on(person)
      .ensure('name')
      .required()
      .ensure((p) => p.age)
      .min(42);
    const result = await controller.validate(new ValidateInstruction(person));
  }
}
```

**Using it in HTML**
```html
<div id="div1" validation-errors.bind="nameErrors">
   <input id="target1" type="text" value.two-way="person.name & validate">
   <span class="error" repeat.for="errorInfo of nameErrors">
     ${errorInfo.result.message}
   </span>
 </div>
```

Some of the major differences include following.
- **Rules are implemented in class** instead of lambdas in fluent rule applicator. Every rule (required, regex pattern matching etc.) gets its own class. Same goes also for custom rules; instead of registering the custom rule with the (previous) `ValidationRules` with a string name, one can just pass on an instance of the rule object in `.satisfiesRule(new CustomRule(...))`. The easy way of defining a lambda as a rule via `.satisifes()` stil exists, but it is internally normalized to an instance of class.
- **Enhanced usage of tags**: In v2 along with properties, objects can also be tagged. This helps defining multiple rulesets for a single object, which can later be used for selective validation.
- **Subscribers over validation renderers**:  In v2 there are no validation renderer. There is only subscribers. A component can register itself as a subscriber to a particular instance of validation controller. When the controller performs validation, it notifies all the registered subscribers. Therefore, instead of render instruction the subscribers gets notified with an instance of `ValidationEvent`. There are 3 out-of-the-box subscribers:
  - custom attribute
    ```html
    <div validation-errors.bind="nameErrors">
      <input type="text" value.two-way="person.name & validate">
      <span class="error" repeat.for="errorInfo of nameErrors">
        ${errorInfo.result.message}
      </span>
    </div>
    ```
  - custom attribute
    ```html
    <validation-container>
      <input type="text" value.two-way="person.name & validate">
    </validation-container>
    ``` 
  - presenter service: standalone DOM manipulator service that adds and removes validation messages, when registered as subscriber to the validation controller.
- **No additional event notification via EA**: In v1 the validation controller dispatches event via EA. In v2 it is removed in favor of the subscribers as mentioned previously.
- **One `validate` binding behavior**: Instead of many variants of the binding behavior for many kind of triggers, there is only one in v2. The trgiggers can be specified as the 1st argument to the binding behavior.
- **Single `validate` method in `IValidator` contract**: In v1 the validator has `validateObject` and `validateProperty` method. In v2 it is all consolidated in `validate`+`ValidateInstruction`.
- **Localization support**: Localizing the error messages and property display name is possible by using the `@aurelia/validation-i18n` plugin. The plugin combines both validation, and i18n plugin.
- **Model-based rule hydration support**: A JSON object defining the rules for object properties can be hydrated as validation rules, using `ValidationRules#applyModelBasedRules`. A sample rule set is shown below.
  ```json
    {
    "tag": "foo",
    "ruleset": {
      "prop": {
        "displayName": "Foo",
        "rules": [
          {
            "required": { "tag": "x", "messageKey": "req" },
            "regex": { "messageKey": "email", "pattern": { "source": "foo", "flags": "gi" }}
          }
        ]
      },
      "nav": {
        "subProp": {
          "subSubProp": {
            "displayName": "Bar",
            "rules": [
              {
                "maxLength": { "length": 42 }
              },
              {
                "expensiveRule": { "messageKey": "unique" }
              }
            ]
          }
        }
      }
    }
  }
  ```
  In case hydration support is needed for any custom rule, the hydrator can be extended and the plugin registration can be customized

## 👩‍💻 Reviewer Notes

- Parsing the property expression specified in `.ensure()`, and unparsing it from the binding expression via the BB, can be improved. Refer `parsePropertyName` is `rule-provider`, and `ValidationController#getPropertyInfo`. Completely open to suggestions.
- The validate binding behavior is implemented using `BindingInterceptor`. IMO the strategy can be improved, as incorrect usage/chaining might lead to unexpected results. Refer `ValidateBindingBehavior#updateSource`. Again, would love hear your suggestions.

## 📑 Test Plan

At this point there is nearly 100% test coverage for the plugin.


packages/validation/src/validation-customization-options.ts
```diff
@@ -0,0 +1,13 @@
+import { ICustomMessage } from './rule-provider';
+import { IValidator } from './validator';
+import { Class } from '@aurelia/kernel';
+import { ValidationTrigger } from './validate-binding-behavior';
+
+/**
+ * Customization options for the plugin.
+ */
+export interface ValidationCustomizationOptions {
+  validator: Class<IValidator>;
```
fkleuver: I'm inclined to think we should rename this to e.g. `ValidatorType` to make it clearer that a type is expected here. That's generally how we do it in the runtime too (including the PascalCase convention). I don't feel strongly about that though, but we do need consistency, so we could change it in the runtime to camelCase as well. What do you think?

packages/validation/src/validator.ts
```diff
@@ -0,0 +1,58 @@
...
+    const rules = instruction.rules ?? validationRulesRegistrar.get(object, instruction.objectTag) ?? [];
+
+    const result = await Promise.all(rules.reduce((acc: Promise<ValidationResult[]>[], rule) => {
+      const { name, expression } = rule.property;
+      if (validateAllProperties || name === propertyName) {
+        let value: unknown;
+        if (expression === void 0) {
+          value = object;
+        } else {
+          const scope = { bindingContext: object, parentScope: null, scopeParts: [], overrideContext: (void 0)! };
```
fkleuver: I think it's something like this:

```suggestion
          const scope = Scope.create(object);
```

Might also be another overload, but there should be an api for this. `scope` really should ideally be the `Scope` class from the runtime though.

packages/validation/src/validator.ts
```diff
@@ -0,0 +1,58 @@
...
+          value = expression.evaluate(LifecycleFlags.none, scope, null!);
```
fkleuver: Maybe add `flags` to the instruction so there is possibility of passing them along?

packages/validation/src/validation-controller.ts
```diff
@@ -0,0 +1,619 @@
...
+  create(validator?: IValidator): IValidationController;
+
+  /**
+   * Creates a new controller and registers it in the current element's container so that it's available to the validate binding behavior and subscribers.
+   */
+  createForCurrentScope(validator?: IValidator): IValidationController;
+}
+export const IValidationControllerFactory = DI.createInterface<IValidationControllerFactory>("IValidationControllerFactory").noDefault();
+
+export class ValidationControllerFactory implements IValidationControllerFactory {
```
fkleuver: Would it be possible/sensible to let this conform to the `IResolver` interface? Then it can also be used for lazy injection and other forms of deep DI integration. See for example the runtime's `ViewFactoryProvider` which has more stuff on it than just the resolver contract.

packages/validation/src/validation-controller.ts
```diff
@@ -0,0 +1,619 @@
...
+    const eventData: ValidationEvent = new ValidationEvent(kind, [], []);
+    // create a shallow copy of newResults so we can mutate it without causing side-effects.
+    newResults = newResults.slice(0);
+
+    const elements = this.elements;
+    for (const oldResult of oldResults) {
+      const removalTargets = elements.get(oldResult)!;
+      elements.delete(oldResult);
+
+      eventData.removedResults.push({ result: oldResult, targets: removalTargets });
```
fkleuver: Can we make this a class instead of obj literal?

packages/validation/src/validation-controller.ts
```diff
@@ -0,0 +1,619 @@
...
+        }
+        default:
+          throw new Error(`Unknown expression of type ${expression?.constructor.name}`); // TODO use reporter/logger
+      }
+      const separator = propertyName.startsWith('[') ? '' : '.';
+      propertyName = propertyName.length === 0 ? memberName : `${memberName}${separator}${propertyName}`;
+      expression = expression.object as ValidatableExpression;
+    }
+    if (expression === void 0) {
+      throw new Error('Unable to parse binding expression'); // TODO use reporter/logger
```
fkleuver:
```suggestion
      throw new Error(`Unable to parse binding expression: ${expression}`); // TODO use reporter/logger
```

packages/validation/src/validation-controller.ts
```diff
@@ -0,0 +1,619 @@
...
+
+    const scope = info.scope;
+    let expression = binding.sourceExpression.expression as ValidatableExpression;
+    const locator = binding.locator;
+    let toCachePropertyName = true;
+    let propertyName: string = "";
+    while (expression !== void 0 && !(expression instanceof AccessScopeExpression)) {
+      let memberName: string;
+      switch (true) {
+        case expression instanceof BindingBehaviorExpression || expression instanceof ValueConverterExpression:
```
fkleuver: We can use `switch (expression.$kind)` here. That would perform a bit better and imo also be a bit cleaner. I added that property specifically for this kind of purpose :)

packages/validation/src/validate-binding-behavior.ts
```diff
@@ -0,0 +1,250 @@
...
+      const controller = (target as CustomElementHost)?.$controller;
+      if (controller === void 0) {
+        throw new Error('Invalid binding target'); // TODO use reporter
+      }
+      this.target = controller.host as HTMLElement;
+    }
+  }
+}
+
+class Delta {
```
fkleuver: What kind of delta?

packages/__tests__/validation/rule-provider.spec.ts
```diff
@@ -0,0 +1,1089 @@
...
+      .minLength(2)
+      .maxLength(42)
+      .matches(/foo/)
+      .then()
+      .email()
+      .when(() => 'foo' as any === 'bar')
+      .equals('foo@bar.com')
+      .rules;
+
+    assert.equal(rules.length, 1);
```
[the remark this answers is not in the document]
Sayan751: In case of complete deep equality it might be bit tricky during starting. Typically, I like to focus my assertions on one specific aspect. Comparing all object properties all the time makes it bit unclear what the focus is. Moreover, if due to refactoring, the tests are failing, disconnected assertions makes the reason of failure bit prominent as compared to other. So it brings more atomicity I would say.

AureliaEffect:
JS Framework Benchmark: https://40585-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/validation/src/validation-customization-options.ts
Sayan751: Having consistent naming would be good. I can change that. Probably it is also a good idea to apply the linting rules. Not sure about the rule specifics though.

Sayan751: I am assuming this is just for the interface, right?

packages/validation/src/validator.ts
Sayan751: That's nice! I have not noticed that. Thank you for pointing it out :)

packages/validation/src/validator.ts
Sayan751: That's not a bad idea so long it is passed along. The reason I am telling this, is validation in general is not very sensitive to the lifecycle flags. However, there might be use cases to lifecycle-sensitive expression evaluation.

packages/__tests__/validation/rule-provider.spec.ts
```diff
@@ -0,0 +1,1089 @@
...
+  function setup() {
+    const container = DI.createContainer();
+    container.register(ValidationConfiguration);
+    return { sut: container.get(IValidationRules), container };
+  }
+
+  it('is transient', function () {
+    const { sut: instance1, container } = setup();
+    const instance2 = container.get(IValidationRules);
+    assert.equal(Object.is(instance1, instance2), false);
```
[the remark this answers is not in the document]
Sayan751: Done :)

packages/__tests__/validation/rule-provider.spec.ts
```diff
@@ -0,0 +1,1089 @@
...
+    { property: function (o: any) { "use strict"; return o.obj.prop['a']; },        expected: 'obj.prop[\'a\']' },
+    { property: function (o: any) { "use strict"; return o.obj.prop['a'].prop2; },  expected: 'obj.prop[\'a\'].prop2' },
+    { property: function (o: any) { "use strict"; return o.prop["a"]; },            expected: 'prop["a"]' },
+    { property: function (o: any) { "use strict"; return o.prop["a"].prop2; },      expected: 'prop["a"].prop2' },
+    { property: function (o: any) { "use strict"; return o.obj.prop["a"]; },        expected: 'obj.prop["a"]' },
+    { property: function (o: any) { "use strict"; return o.obj.prop["a"].prop2; },  expected: 'obj.prop["a"].prop2' },
+  ].map(({ property, expected }) =>
+    it(`parses ${property.toString()} to ${expected}`, function () {
+      const { parser } = setup();
+      assert.deepEqual(parsePropertyName(property, parser), [expected, parser.parse(expected, BindingType.None)]);
```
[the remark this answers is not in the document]
Sayan751: Done :)

packages/__tests__/validation/rules.spec.ts
```diff
@@ -0,0 +1,138 @@
...
+    { value: 41,        range: { min: 39,         max: 42        }, isInclusive: false, isValid: true,  key: 'between' },
+    { value: 42,        range: { min: 39,         max: 42        }, isInclusive: false, isValid: false, key: 'between' },
+    { value: 43,        range: { min: 39,         max: 42        }, isInclusive: false, isValid: false, key: 'between' },
+    { value: 38,        range: { min: 39,         max: 42        }, isInclusive: true,  isValid: false, key: 'range'   },
+    { value: 39,        range: { min: 39,         max: 42        }, isInclusive: true,  isValid: true,  key: 'range'   },
+    { value: 40,        range: { min: 39,         max: 42        }, isInclusive: true,  isValid: true,  key: 'range'   },
+    { value: 41,        range: { min: 39,         max: 42        }, isInclusive: true,  isValid: true,  key: 'range'   },
+    { value: 42,        range: { min: 39,         max: 42        }, isInclusive: true,  isValid: true,  key: 'range'   },
+    { value: 43,        range: { min: 39,         max: 42        }, isInclusive: true,  isValid: false, key: 'range'   },
+  ].map(({ value, range, isInclusive, isValid, key }) =>
```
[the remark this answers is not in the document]
Sayan751: I find this data driven approach to be bit more dumb and cleaner :) Made the changes with `for...of` loop.

packages/validation/src/validation-customization-options.ts
Sayan751: Changed to PascalCase.

packages/validation/src/validator.ts
Sayan751: Done :)

packages/validation/src/validation-controller.ts
Sayan751: Not sure why that stayed like that as I did already change the interface to class. Anyway, made the changes. Thanks for pointing it out.

packages/validation/src/validation-controller.ts
Sayan751: Done :)

packages/validation/src/validation-controller.ts
Sayan751: Done :)

packages/validation/src/validate-binding-behavior.ts
Sayan751: Changed the name to `ValidateArgumentsDelta`. Hope that clears things up.

AureliaEffect:
JS Framework Benchmark: https://40609-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/validation/src/validator.ts
Sayan751: Done :)

AureliaEffect:
JS Framework Benchmark: https://40612-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

bigopon:
> One validate binding behavior: Instead of many variants of the binding behavior for many kind of triggers, there is only one in v2. The triggers can be specified as the 1st argument to the binding behavior.

Among trigger, validation controller & rule. I think it's tough to pick the winner for 1st parameter position. So should we keep the current order in v1?

packages/runtime/src/binding/property-binding.ts
```diff
@@ -73,15 +73,15 @@ export class PropertyBinding implements IPartialConnectableBinding {
     this.sourceExpression.assign!(flags, this.$scope!, this.locator, value, this.part);
   }
 
-  public handleChange(newValue: unknown, _previousValue: unknown, flags: LifecycleFlags): void {
+  public handleChange(newValue: unknown, previousValue: unknown, flags: LifecycleFlags): void {
```
bigopon: why is this change? wasn't it to help with optimization of JS engine?

packages/runtime/src/renderer.ts
```diff
@@ -348,6 +348,7 @@ export class CustomAttributeRenderer implements IInstructionRenderer {
     const childController = Controller.forCustomAttribute(
       /* viewModel */component,
       /* lifecycle */lifecycle,
+      /* host      */target,
```
bigopon: simple but pretty necessary

packages/validation/src/rule-provider.ts
```diff
@@ -0,0 +1,601 @@
...
+}
+
+export type PropertyAccessor<TObject extends IValidateable = IValidateable, TValue = unknown> = (object: TObject) => TValue;
+export function parsePropertyName(property: string | PropertyAccessor, parser: IExpressionParser): [string, IsBindingBehavior] {
+
+  switch (typeof property) {
+    case "string":
+      break;
+    case "function": {
+      const classic = /^function\s*\([$_\w\d]+\)\s*\{(?:\s*["']{1}use strict["']{1};)?\s*(?:[$_\w\d.['"\]+;]+)?\s*return\s+[$_\w\d]+((\.[$_\w\d]+|\[['"$_\w\d]+\])+)\s*;?\s*\}$/;
```
bigopon: really want to nudge @fkleuver for a function AST + parser so we can evaluate + connect fn without having to rely on proxy 😁

bigopon: and reuse it here

packages/validation/src/rules.ts
```diff
@@ -0,0 +1,250 @@
...
+    if (min !== void 0 && max !== void 0) {
+      this._messageKey = this.isInclusive ? 'range' : 'between';
+    } else {
+      this._messageKey = min !== void 0 ? 'min' : 'max';
+    }
+    this.min = min ?? this.min;
+    this.max = max ?? this.max;
+  }
+  public execute(value: number, _object?: IValidateable): boolean | Promise<boolean> {
+    return value === null || value === undefined || (this.isInclusive
```
bigopon: nit: shouldn't this be
```ts
return value === null
  || value === undefined
  || (this.inclusive
    ? value >= this.min && value <= this.max
    : value > this.min && value < this.max
  );
```

bigopon: just to signals the equivalence of each condition

packages/validation/src/validation-controller.ts
```diff
@@ -0,0 +1,599 @@
...
+  }
+
+  public removeError(result: ValidationResult) {
+    if (this.results.includes(result)) {
+      this.processResultDelta(ValidateEventKind.reset, [result], []);
+    }
+  }
+
+  public addSubscriber(subscriber: ValidationErrorsSubscriber) {
+    this.subscribers.push(subscriber);
```
bigopon: maybe have a guard to prevent adding same subscriber twice. Idempotent is quite common in Aurelia internal API i think

packages/validation/src/validator.ts
```diff
@@ -0,0 +1,81 @@
...
+  public async validate<TObject extends IValidateable = IValidateable>(instruction: ValidateInstruction<TObject>): Promise<ValidationResult[]> {
+    const object = instruction.object;
+    const propertyName = instruction.propertyName;
+    const propertyTag = instruction.propertyTag;
+    const flags = instruction.flags;
+
+    const validateAllProperties = propertyName === void 0;
+    const rules = instruction.rules ?? validationRulesRegistrar.get(object, instruction.objectTag) ?? [];
+
+    const result = await Promise.all(rules.reduce((acc: Promise<ValidationResult[]>[], rule) => {
```
bigopon: I think a potential todo here is like how we have it with lifecycle, only turn the entire thing to promise if one of them returns a promise. that would be so nice. I'm aware that this is first iteration, just saying the obvious since you and @fkleuver probably already thought of this

bigopon:
Not much extra to add, feels pretty solid to me. Amazing work!

packages/runtime/src/binding/property-binding.ts
fkleuver: The underscore prefix might have been an old linting artifact for unused variable something. I don't know of any optimizations for this though, or none that would have any measurable impact at least.

packages/validation/src/validation-controller.ts
fkleuver: Would rather just have a `// TODO` note here that this needs to utilize `SubscriberCollection` once we normalized it in the runtime. Probably not worth the time/energy to fine tune something that will be replaced in a while.

packages/validation/src/validator.ts
fkleuver: Yeah I've spotted a *lot* of potential perf optimizations similar to that as well. That's not necessarily what I want to be focusing on though for this first iteration. Performance is easier to work on once we have a set of proper benchmarks to test against, and doing it in one go as a separate work item rather than as a combined effort in the initial implementation

packages/runtime/src/binding/property-binding.ts
Sayan751: I think it is from the time I tried to implement the binding middleware which @fkleuver made available later as binding interceptors. I cleaned up my changes, or so I thought :) I think it is just the only instance I missed. I will revert it.

packages/validation/src/rule-provider.ts
Sayan751: I am also not happy with this 'parsing'. Any guidance on this will be great :)

fkleuver: @bigopon We can do svelte-like magic with AOT but adding function parsing to the jit expression parser for usage in regular getters would be a lot more involved than you think. We might as well add acorn as a jit dependency at that point. Definitely overkill for this niche situation where the only valid syntax is a simple member or index accessor.

fkleuver: Can we put these regexes in module scope though and try to match arrow first? For perf

packages/validation/src/validation-controller.ts
Sayan751: @fkleuver I am not so sure about that. Usually, the handleChange takes the oldValue, newValue, and the flags. In this case however, there is no concept of old and new values. If you look at the ValidationController#processResultDelta, you'll see that the event data is comprised of the errors that have been removed and the news ones that have been added. It bit different in that sense. Having said that, if it possible to normalize that I will happily do so. If nothing else, it'll make the validation controller less cumbersome.

Sayan751: BTW a easy solution for now is to use a Set for subscribers instead of an array

packages/runtime/src/binding/property-binding.ts
Sayan751: Done :)

packages/validation/src/rules.ts
Sayan751: Done :)

packages/validation/src/validation-controller.ts
Sayan751: Used a Set

AureliaEffect:
JS Framework Benchmark: https://40632-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

Sayan751:
> Among trigger, validation controller & rule. I think it's tough to pick the winner for 1st parameter position. So should we keep the current order in v1?

In a way it is like that. For full options v1 has this syntax: `validateOnTrigger:controller:rules`. In v2 I have this `validate:trigger:controller:rules`. Additionally I enforced the datatypes for the positional attributes. This means that following expression is not allowed: `validate:trigger:rules`.

packages/validation/src/validation-controller.ts
fkleuver: Well yeah part of that normalization is also to split add/remove subscriber (which is the same for all subscription types) and `callSubscribers` (which varies per subscription type)

AureliaEffect:
JS Framework Benchmark: https://40827-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/validation/src/rule-provider.ts
Sayan751: done

packages/validation/src/validation-controller.ts
Sayan751: postponed

AureliaEffect:
JS Framework Benchmark: https://40856-121594173-gh.circle-artifacts.com/0/home/circleci/repo/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://41479-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://41483-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://41839-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/validation/src/configuration.ts
```diff
@@ -0,0 +1,36 @@
...
+    optionsProvider,
+    register(container: IContainer) {
+      const options: ValidationCustomizationOptions = { ValidatorType: StandardValidator, CustomMessages: [], DefaultTrigger: ValidationTrigger.blur };
+      optionsProvider(options);
+
+      return container.register(
+        Registration.callback(ICustomMessages, () => options.CustomMessages),
+        Registration.callback(IDefaultTrigger, () => options.DefaultTrigger),
+        Registration.singleton(IValidator, options.ValidatorType),
+        Registration.singleton(IValidationMessageProvider, ValidationMessageProvider), // TODO enable customization of messages and i18n
```
zewa666: `// TODO` could need a colon as that helps e.g a few vscode extensions to pick it up as todos in a quick overview

packages/validation/src/rules.ts
```diff
@@ -0,0 +1,264 @@
...
+  define<TRule extends BaseValidationRule>(target: Constructable<TRule>, definition: ValidationRuleDefinition): Constructable<TRule> {
+    ValidationRuleAliasMessage.setDefaultMessage(target, definition);
+    return target;
+  },
+  setDefaultMessage<TRule extends BaseValidationRule>(rule: Constructable<TRule>, { aliases }: ValidationRuleDefinition, append: boolean = true) {
+    // conditionally merge
+    const defaultMessages = append ? Metadata.getOwn(this.aliasKey, rule.prototype) as ValidationRuleAlias[] : void 0;
+    if (defaultMessages !== void 0) {
+      const allMessages: Record<string, string> = {
+        ...Object.fromEntries(defaultMessages.map(({ name, defaultMessage }) => [name, defaultMessage])) as Record<string, string>,
```
zewa666: Oh cool, didn't know about this. Is there a polyfill in place for fromEntries?

packages/validation/src/rules.ts
```diff
@@ -0,0 +1,264 @@
...
+
+  protected _messageKey: string = (void 0)!;
+  public get messageKey() { return this._messageKey; }
+  public set messageKey(key: string) {
+    this._messageKey = key;
+    this._message = (void 0)!;
+  }
+
+  public get message(): IInterpolationExpression | PrimitiveLiteralExpression {
+    let message = this._message;
```
zewa666: I'm unsure I follow the reason for the letted message here. Can't you just make use of `this._message`

packages/validation/src/rules.ts
```diff
@@ -0,0 +1,264 @@
...
+ * @see PropertyRule#minLength
+ * @see PropertyRule#maxLength
+ */
+@validationRule({
+  aliases: [
+    { name: 'minLength', defaultMessage: `\${$displayName} must be at least \${$rule.length} character\${$rule.length === 1 ? '' : 's'}.` },
+    { name: 'maxLength', defaultMessage: `\${$displayName} cannot be longer than \${$rule.length} character\${$rule.length === 1 ? '' : 's'}.` },
+  ]
+})
+export class LengthRule extends BaseValidationRule<string> {
```
zewa666: I really haven't worked a lot with these rules, but if there is a possibility that definitions get bound, there might also be again a chance for a `NaN`. Should we check for that as well? Perhaps applies to others like the Range or Size rule as well

AureliaEffect:
JS Framework Benchmark: https://42103-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://42128-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://42134-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/__e2e__/src/startup.ts
```diff
@@ -36,15 +37,16 @@ Intl['RelativeTimeFormat'] = Intl['RelativeTimeFormat'] || RelativeTimeFormat;
...
-        };
-      })
+        } as I18nInitOptions;
+      }),
+      ValidationConfiguration,
+      SutI18N,
+      CustomMessage
     );
-  au.container.register(SutI18N, CustomMessage);
+  // au.container.register(SutI18N, CustomMessage, SutValidation);
```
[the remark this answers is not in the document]
Sayan751: Done

packages/__tests__/2-runtime/ast.spec.ts
```diff
@@ -2027,7 +2029,7 @@ describe('BindingBehaviorExpression', function () {
       const sut = new BindingBehaviorExpression(expr as any, 'mock', args);
 
       const mock = new MockBindingBehavior();
-      const locator = new MockServiceLocator(new Map<any, any>([['au:resource:binding-behavior:mock', mock]]));
+      const locator = new MockServiceLocator(new Map<any, any>([['au:resource:binding-behavior:mock', mock], [IScheduler, TestContext.createHTMLTestContext().scheduler]]));
```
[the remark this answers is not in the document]
Sayan751: Old artifact. Commit reverted.

packages/__tests__/util.ts
```diff
@@ -0,0 +1,35 @@
...
+
+  $it.skip = function (title: string, testFunction: TestFunction<TTestContext>, setupContext?: TSetupContext) {
+    it.skip(title, async function () { await wrap(testFunction, setupContext); });
+  };
+  return $it;
+}
+
+@valueConverter('toNumber')
+export class ToNumberValueConverter {
+  public fromView(value: string): number { return Number(value) || void 0; }
```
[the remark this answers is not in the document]
Sayan751: The tests are kind of based on that. So considering this is just for tests, I would like to keep it this way 😛

packages/__tests__/validation/.eslintrc.js
```diff
@@ -0,0 +1,9 @@
+module.exports = {
+  extends: [
+    '../.eslintrc.js'
+  ],
+  rules: {
+    'no-template-curly-in-string': 'off',
+    'no-useless-escape': 'off'
```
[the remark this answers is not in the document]
Sayan751: The second one I have switched on. But the first one is needed for instances like this: `'${$displayName} is required'`

packages/__tests__/validation/_test-resources.ts
```diff
@@ -0,0 +1,18 @@
+export interface Address {
+  line1: string;
+  line2?: string;
+  city: string;
+  pin: number;
+}
+export interface IPerson {
```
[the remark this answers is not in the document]
Sayan751: Removed.

packages/__tests__/validation/rules.spec.ts
```diff
@@ -0,0 +1,144 @@
...
+    { value: 41,        range: { min: 39,         max: 42        }, isInclusive: false, isValid: true,  key: 'between' },
+    { value: 42,        range: { min: 39,         max: 42        }, isInclusive: false, isValid: false, key: 'between' },
+    { value: 43,        range: { min: 39,         max: 42        }, isInclusive: false, isValid: false, key: 'between' },
+    { value: 38,        range: { min: 39,         max: 42        }, isInclusive: true,  isValid: false, key: 'range'   },
+    { value: 39,        range: { min: 39,         max: 42        }, isInclusive: true,  isValid: true,  key: 'range'   },
+    { value: 40,        range: { min: 39,         max: 42        }, isInclusive: true,  isValid: true,  key: 'range'   },
+    { value: 41,        range: { min: 39,         max: 42        }, isInclusive: true,  isValid: true,  key: 'range'   },
+    { value: 42,        range: { min: 39,         max: 42        }, isInclusive: true,  isValid: true,  key: 'range'   },
+    { value: 43,        range: { min: 39,         max: 42        }, isInclusive: true,  isValid: false, key: 'range'   },
```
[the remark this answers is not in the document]
Sayan751: Done

packages/runtime-html/src/binding/listener.ts
```diff
@@ -9,11 +9,11 @@ import {
   IsBindingBehavior,
   IScope,
   LifecycleFlags,
-  State
+  State,
```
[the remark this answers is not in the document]
Sayan751: Old artifact. Cleaned up

packages/validation/src/configuration.ts
Sayan751: Done

packages/validation/src/rules.ts
Sayan751: I think there are quite few usage instances of new APIs. Polyfill will needs to arranged in general for all. Maybe this too can be postponed. BTW MDN says that only IE does not support this. With Chromium Edge it might die soon.

packages/validation/src/rules.ts
Sayan751: Old artifact. Implementation changed

packages/validation/src/rules.ts
Sayan751: I am not sure what you meant. Can you please elaborate?

packages/__tests__/validation/subscribers/validation-errors-custom-attribute.spec.ts
```diff
@@ -0,0 +1,347 @@
...
+    controllerSpy.methodCalledTimes('validate', 1);
+    assert.equal(handleValidationEventSpy.calls.length, 1, 'incorrect #calls for handleValidationEvent');
+  }
+  function assertSubscriber(controller: ValidationController, ca: ValidationErrorsCustomAttribute) {
+    const subscribers = (controller['subscribers'] as Set<ValidationErrorsSubscriber>);
+    assert.equal((subscribers).has(ca), true);
+    assert.equal(ca.controller, controller);
+  }
+
+  $it('registers only the errors targeted for the containing elements',
```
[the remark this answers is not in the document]
Sayan751: There indeed a lot going on there. Not sure how that can be bifurcated. Added some comments though.

AureliaEffect:
JS Framework Benchmark: https://42196-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/validation/src/rules.ts
zewa666: forget this one all good

packages/validation/src/rules.ts
zewa666: I agree, so perhaps just leaving a TODO here might be enough?

AureliaEffect:
JS Framework Benchmark: https://42233-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/validation/src/rules.ts
Sayan751: Done

AureliaEffect:
JS Framework Benchmark: https://42259-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://42410-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://42471-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/runtime/package.json
```diff
@@ -29,7 +29,7 @@
     "lint": "eslint --cache --ext .js,.ts src/",
     "build": "tsc -b",
     "bundle": "ts-node -P ../../tsconfig.json ../../scripts/bundle.ts umd,esm,system runtime",
-    "dev": "tsc -b -w"
+    "dev": "tsc -b -w --preserveWatchOutput"
```
bigopon: well i support this one. I have cmder open and sometime i just have no clue which packages have been in dev mode

AureliaEffect:
JS Framework Benchmark: https://42732-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://42748-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/runtime/src/binding/ast.ts
```diff
@@ -554,9 +563,16 @@ export class AccessKeyedExpression implements IAccessKeyedExpression {
   public accept<T>(visitor: IVisitor<T>): T {
     return visitor.visitAccessKeyed(this);
   }
+
+  public static fromJSON(jsonObject: Pick<AccessKeyedExpression, 'object' | 'key'>, hydrator: IHydrator): AccessKeyedExpression {
```
bigopon: This one is missed

packages/validation/src/validation-controller.ts
```diff
@@ -0,0 +1,585 @@
...
+   * @internal
+   */
+  registerBinding(binding: BindingWithBehavior, info: BindingInfo): void;
+  /**
+   * Deregisters a binding; i.e. it won't be validated during validate without instruction.
+   * This is usually done via the `validate` binding behavior during unbinding phase.
+   *
+   * @internal
+   */
+  deregisterBinding(binding: BindingWithBehavior): void;
```
bigopon: well, sorry for the nit but we may see a request of having `unregister` as an alias

packages/validation/src/validation-controller.ts
```diff
@@ -0,0 +1,585 @@
...
+    if (propertyInfo !== void 0) {
+      return propertyInfo;
+    }
+
+    const scope = info.scope;
+    let expression = binding.sourceExpression.expression;
+    const locator = binding.locator;
+    let toCachePropertyName = true;
+    let propertyName: string = "";
+    while (expression !== void 0 && !(expression instanceof AccessScopeExpression)) {
```
bigopon: It feels to me if we already have static field `$Type` on the AST, may as well just allow duck typing, or remove the static field

bigopon:
@Sayan751 Looks so good. For the AST, I think if we don't use `static $Type` anywhere but serializer, maybe move them all there?

packages/validation/src/validation-controller.ts
Sayan751: I remember this one :) This is the reason for this naming: https://english.stackexchange.com/a/40095/249107

But can change this if "unregister" as verb is more common.

packages/runtime/src/binding/ast.ts
Sayan751: Thank you :) I will remove it

Sayan751:
The `$TYPE` from AST can probably be moved to the deserializer. Should have done that. Thanks for the review @bigopon :)

AureliaEffect:
JS Framework Benchmark: https://42781-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/validation/src/configuration.ts
```diff
@@ -0,0 +1,64 @@
...
+      const options: ValidationCustomizationOptions = getDefaultValidationConfiguration();
+
+      optionsProvider(options);
+
+      const key = Protocol.annotation.keyFor('di:factory');
+      Protocol.annotation.set(IValidationController as unknown as Constructable, 'di:factory', new options.ValidationControllerFactoryType());
+      Protocol.annotation.appendTo(IValidationController as unknown as Constructable, key);
+
+      container.register(
+        Registration.callback(ICustomMessages, () => options.CustomMessages),
```
fkleuver: Why not `Registration.instance(ICustomMessages, options.CustomMessages),`?

packages/validation/src/configuration.ts
```diff
@@ -0,0 +1,64 @@
...
+  return {
+    optionsProvider,
+    register(container: IContainer) {
+      const options: ValidationCustomizationOptions = getDefaultValidationConfiguration();
+
+      optionsProvider(options);
+
+      const key = Protocol.annotation.keyFor('di:factory');
+      Protocol.annotation.set(IValidationController as unknown as Constructable, 'di:factory', new options.ValidationControllerFactoryType());
+      Protocol.annotation.appendTo(IValidationController as unknown as Constructable, key);
```
fkleuver: Hmm.. I'll make note of this. We probably need a decent static api in the kernel for getting and setting factories flexibly.

packages/validation/src/rule-provider.ts
```diff
@@ -0,0 +1,650 @@
...
+        const { displayName, name } = this.property;
+        let message: string | undefined;
+        if (!isValidOrPromise) {
+          const messageEvaluationScope = Scope.create(flags!, {
+            $object: object,
+            $displayName: this.messageProvider.getDisplayName(name, displayName),
+            $propertyName: name,
+            $value: value,
+            $rule: rule,
+            $getDisplayName: this.messageProvider.getDisplayName.bind(this.messageProvider)
```
fkleuver: Could we make this a class? And I think pass in the messageProvider, then `$getDisplayName` can be a regular method if that works for the `this`

packages/validation/src/rules.ts
```diff
@@ -0,0 +1,272 @@
...
+  aliasKey: Protocol.annotation.keyFor('validation-rule-alias-message'),
+  define<TRule extends IValidationRule>(target: RuleType<TRule>, definition: ValidationRuleDefinition): RuleType<TRule> {
+    ValidationRuleAliasMessage.setDefaultMessage(target, definition);
+    return target;
+  },
+  setDefaultMessage<TRule extends IValidationRule>(rule: Constructable<TRule>, { aliases }: ValidationRuleDefinition, append: boolean = true) {
+    // conditionally merge
+    const defaultMessages = append ? Metadata.getOwn(this.aliasKey, rule.prototype) as ValidationRuleAlias[] : void 0;
+    if (defaultMessages !== void 0) {
+      // TODO: have polyfill for `Object.fromEntries` as IE does not yet support it
```
fkleuver: Don't worry about this. We can bulk-detect all polyfill necessities with linter and later AOT. I'll fix that all in one go before RC.

packages/validation/src/serialization.ts
```diff
@@ -0,0 +1,295 @@
...
+    if (messageKey !== void 0 && messageKey !== null) {
+      rule.messageKey = messageKey;
+    }
+    rule.tag = raw.tag;
+    const when = raw.when;
+    if (when) {
+      if (typeof when === 'string') {
+        const parsed = this.parser.parse(when, BindingType.None);
+        rule.canExecute = (object: IValidateable) => {
+          const flags = LifecycleFlags.none; // TODO? need to get the flags propagated here?
```
fkleuver: Evaluation-specific flags to be aware of:
- `LifecycleFlag.allowParentScopeTraversal` (specific to `replaceable`, so some replaceable integration tests could be useful to double check)
- `LifecycleFlags.mustEvaluate` (in `CallFunctionExpression` - probably never relevant)
- `LifecycleFlags.isStrictBindingStrategy` (in `AccessScopeExpression` - this could matter here because when it's set, null and undefined won't default to empty string, and that affects boolean coercion)

packages/validation/src/validate-binding-behavior.ts
```diff
@@ -0,0 +1,250 @@
...
+    expr: IBindingBehaviorExpression,
+  ) {
+    super(binding, expr);
+    this.scheduler = this.locator.get(IScheduler);
+    this.defaultTrigger = this.locator.get(IDefaultTrigger);
+    this.setPropertyBinding();
+  }
+
+  public updateSource(value: unknown, flags: LifecycleFlags) {
+    // TODO: need better approach. If done incorrectly may cause infinite loop, stack overflow 💣
```
fkleuver: Can you include a snippet that demonstrates infinite loop situation? Then I can help you look into it later asynchronously

packages/validation/src/validate-binding-behavior.ts
```diff
@@ -0,0 +1,250 @@
...
+    const delta = this.processBindingExpressionArgs(flags);
+    this.processDelta(delta);
+  }
+
+  public $unbind(flags: LifecycleFlags) {
+    this.target?.removeEventListener('blur', this);
+    this.controller?.deregisterBinding(this.propertyBinding);
+    this.binding.$unbind(flags);
+    for (const expr of this.connectedExpressions) {
+      if (expr.unbind !== void 0) {
```
fkleuver:
```suggestion
      expr.unbind?.(flags, this.scope, this);
```

packages/validation/src/validation-controller.ts
```diff
@@ -0,0 +1,585 @@
...
+
+  public async revalidateErrors() {
+    const map = this.results
+      .reduce(
+        (acc, { isManual, object, propertyRule, rule, valid }) => {
+          if (!valid && !isManual && propertyRule !== void 0 && object !== void 0 && rule !== void 0) {
+            let value = acc.get(object);
+            if (value === void 0) {
+              value = new Map();
+              acc.set(object, value);
```
fkleuver: 2 lines -> 1 line
```suggestion
              acc.set(object, value = new Map());
```

packages/validation/src/validation-controller.ts
```diff
@@ -0,0 +1,585 @@
...
+            }
+            let rules = value.get(propertyRule);
+            if (rules === void 0) {
+              rules = [];
+              value.set(propertyRule, rules);
```
fkleuver: 2 lines -> 1 line
```suggestion
              value.set(propertyRule, rules = []);
```

packages/validation/src/validation-controller.ts
fkleuver: I'd strongly favor using `$kind`. You could simply do:
```ts
loop: while (expression !== void 0) {
  case ExpressionKind.AccessScope:
    break loop;
```

packages/validation/src/validation-controller.ts
```diff
@@ -0,0 +1,585 @@
...
+  }
+
+  private getPropertyInfo(binding: BindingWithBehavior, info: BindingInfo, flags: LifecycleFlags = LifecycleFlags.none): PropertyInfo | undefined {
+    let propertyInfo = info.propertyInfo;
+    if (propertyInfo !== void 0) {
+      return propertyInfo;
+    }
+
+    const scope = info.scope;
+    let expression = binding.sourceExpression.expression;
```
fkleuver:
```suggestion
    let expression = binding.sourceExpression.expression as IsBindingBehavior;
```

That's a union of all basic types, so you don't need to cast each expression in the `switch` statement down below

packages/validation/src/validator.ts
```diff
@@ -0,0 +1,72 @@
...
+    if (propertyName !== void 0) {
+      return (await rules.find((r) => r.property.name === propertyName)?.validate(object, propertyTag, flags, scope)) ?? [];
+    }
+
+    return (
+      await Promise.all(rules.reduce((acc: Promise<ValidationResult[]>[], rule) => {
+        acc.push(rule.validate(object, propertyTag, flags, scope));
+        return acc;
+      }, []))
+    ).flatMap((r) => r);
```
fkleuver: Can simplify this a bit :)
```suggestion
    return (await Promise.all(rules.map(rule => rule.validate(object, propertyTag, flags, scope)))).flat();
```

packages/runtime/src/binding/ast.ts
Sayan751: Done

packages/validation/src/validation-controller.ts
Sayan751: Done

packages/validation/src/validation-controller.ts
Sayan751: Done. Used the `$kind`.

packages/validation/src/subscribers/validation-container-custom-element.ts
```diff
@@ -0,0 +1,64 @@
...
+      if (index !== -1) {
+        this.errors.splice(index, 1);
+      }
+    }
+
+    for (const { result, targets: elements } of event.addedResults) {
+      if (result.valid) {
+        continue;
+      }
+      const targets = elements.filter(e => this.host.contains(e));
```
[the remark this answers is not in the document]
Sayan751: The `targets` variable is needed anyway in L44 :)

AureliaEffect:
JS Framework Benchmark: https://42855-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/validation/src/validate-binding-behavior.ts
Sayan751: I don't have an example right away. But I found our discussion on Discord on this topic. My point only is that people need to use this pattern (if+else+while) when writing a binding interceptor or chaining those. Otherwise the infinite loop can easily be possible from the terminal BB, where the interceptor edge points to itself.

packages/validation/src/subscribers/common.ts
```diff
@@ -0,0 +1,2 @@
+// polyfill for Node.js
+export const DOCUMENT_POSITION_PRECEDING = typeof Node !== 'undefined' ? Node.DOCUMENT_POSITION_PRECEDING : 2;
```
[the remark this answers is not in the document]
Sayan751: Done

packages/validation/src/configuration.ts
Sayan751: Done

packages/validation/src/rule-provider.ts
Sayan751: Done

packages/validation/src/validate-binding-behavior.ts
Sayan751: Done

packages/validation/src/validation-controller.ts
Sayan751: Done

packages/validation/src/validation-controller.ts
Sayan751: This is cool :) Changed it.

packages/validation/src/validator.ts
Sayan751: Done

packages/validation/src/validation-controller.ts
Sayan751: Done

AureliaEffect:
JS Framework Benchmark: https://42862-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/validation/src/serialization.ts
Sayan751: Added some tests here: https://github.com/aurelia/aurelia/pull/811/commits/ad7fcd942e05cf94aebac90ae5c49b0ff4a9fb9f

AureliaEffect:
JS Framework Benchmark: https://42928-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://42938-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://42946-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://42959-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

PR #816, opened 2020-02-19
I18n zero undefined null

zewa666:
related PR from v1 https://github.com/aurelia/i18n/pull/314

### DIFF
--- a/packages/i18n/src/t/translation-binding.ts
+++ b/packages/i18n/src/t/translation-binding.ts
@@ -55,7 +55,7 @@ export class TranslationBinding implements IPartialConnectableBinding {
   public parametersExpr?: IsExpression;
   private readonly i18n: I18N;
   private readonly contentAttributes: readonly string[] = contentAttributes;
-  private keyExpression!: string;
+  private keyExpression!: string | undefined | null;
   private translationParameters!: i18next.TOptions;
   private scope!: IScope;
   private isInterpolatedSourceExpr!: boolean;
@@ -166,7 +166,11 @@ export class TranslationBinding implements IPartialConnectableBinding {
   }
 
   private updateTranslations(flags: LifecycleFlags) {
-    const results = this.i18n.evaluate(this.keyExpression, this.translationParameters);
+    // zero needs to be a valid value
+    const keyExpressionSanitized = this.keyExpression === null || this.keyExpression === undefined
+      ? ""
+      : this.keyExpression;
+    const results = this.i18n.evaluate(keyExpressionSanitized, this.translationParameters);
     const content: ContentValue = Object.create(null);
     this.unobserveTargets(flags);
 
--- a/packages/i18n/src/t/translation-binding.ts
+++ b/packages/i18n/src/t/translation-binding.ts
@@ -55,7 +55,7 @@ export class TranslationBinding implements IPartialConnectableBinding {
   public parametersExpr?: IsExpression;
   private readonly i18n: I18N;
   private readonly contentAttributes: readonly string[] = contentAttributes;
-  private keyExpression!: string | undefined | null;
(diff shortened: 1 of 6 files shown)


packages/i18n/src/t/translation-binding.ts
```diff
@@ -55,7 +55,7 @@ export class TranslationBinding implements IPartialConnectableBinding {
   public parametersExpr?: IsExpression;
   private readonly i18n: I18N;
   private readonly contentAttributes: readonly string[] = contentAttributes;
-  private keyExpression!: string;
+  private keyExpression!: string | undefined | null;
```
zewa666: I have no idea why the keyExpression was created with a !, guess to suppress some TS issues. Anyways when `t.bind="whatever"` is bound to a non-string, pretty much anything can land in there

zewa666:
I've added a test which for now is failing due to trying to split() on a non-string. So it might be worth to look into this further and get the types of keyExpression right.

packages/i18n/src/t/translation-binding.ts
Sayan751: The `!` assertion is to avoid the truthy assertion for future (anything after `$bind`). I think we should not allows binding `0` at all. If the key expression does not evaluate to string (refer `this.keyExpression = this.expr.evaluate(flags, scope, this.locator, part) as string;` in `$bind`), we should simply throw an error. With this place, you can have this 
```typescript
const results = this.i18n.evaluate(this.keyExpression ?? '', this.translationParameters);
```
instead of 
```typescript
const keyExpressionSanitized = this.keyExpression === null || this.keyExpression === undefined
  ? ""
  : this.keyExpression;
const results = this.i18n.evaluate(keyExpressionSanitized, this.translationParameters);
```

IMO a non-string i18n key does not make any sense. The test is currently failing because `0.split` does not exist.

AureliaEffect:
JS Framework Benchmark: https://41746-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

zewa666:
Very nice, the Reporter instead Error usage is left. Do you intend to do this at another time? Otherwise this is lgtm

Sayan751:
> Reporter instead Error usage is left. Do you intend to do this at another time?

I think the related infra is not there yet. So by convention I am marking the usages with TODOs, so that later these can be replace.

EisenbergEffect:
@Sayan751 Yes, don't hold things up for Reporter. You can throw errors or use what you like, but mark the areas with TODOs and we'll clean that up in the future when the new infrastructure is in place.

Sayan751:
@EisenbergEffect Cool! Doing that.

Sayan751:
Need to update the docs as well

Sayan751:
@zewa666 added 2 new tests and updated the doc. From my perspective, there is nothing more to be done. If I am missing something, let me know.

AureliaEffect:
JS Framework Benchmark: https://42479-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

zewa666:
I'm unsure about he Docs saying its an empty string that results. What would be more of interest for the user is whether an empty textnode is created or not. But aside that all fine

Sayan751:
@zewa666 correction done :)

AureliaEffect:
JS Framework Benchmark: https://42485-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

zewa666:
Lgtm

AureliaEffect:
JS Framework Benchmark: https://42672-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

PR #821, opened 2020-03-16
feat: support binding for Set/Map for CheckedObserver

bigopon:
# Pull Request

## 📖 Description

Provide support for binding multiple checkbox to Set/Map, in addition to Array

### 🎫 Issues

N/A

## 👩‍💻 Reviewer Notes

N/A

## 📑 Test Plan

More sophisticate usage with mater & Set/Map


fkleuver:
@bigopon Seems like an excellent idea. Makes it consistent with the repeater as well. Gets me wondering if there would be a way to share some logic between the two.. maybe something to look into later.
Does this work as it is right now? Anything specific you want me to look at?

AureliaEffect:
JS Framework Benchmark: https://42794-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

bigopon:
@fkleuver I was asking how we would feel about this addition mainly. If it seems good, then i'll finish the rest with some more complicated scenarios with regards to matcher and seal this PR. Will also need to work on some doc

bigopon:
@fkleuver ready to go. Please help review

AureliaEffect:
JS Framework Benchmark: https://42837-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://42906-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/runtime-html/src/observation/checked-observer.ts
```diff
@@ -137,44 +141,144 @@ export class CheckedObserver implements IAccessor {
...
+    const elementValue = Object.prototype.hasOwnProperty.call(obj, 'model') as boolean ? obj.model : obj.value;
     const isRadio = obj.type === 'radio';
     const matcher = obj.matcher !== void 0 ? obj.matcher : defaultMatcher;
 
     if (isRadio) {
       obj.checked = !!matcher(currentValue, elementValue);
     } else if (currentValue === true) {
       obj.checked = true;
-    } else if (Array.isArray(currentValue)) {
+    } else if (currentValue instanceof Array /* Array.isArray does not account for proxy */) {
```
bigopon: I was always wondering why we are using `instanceof Array` in our code, now I understand why. This does not need to go with this PR, but it wouldn't hurt. @fkleuver your call.

AureliaEffect:
JS Framework Benchmark: https://43005-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://43042-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://43168-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/runtime-html/src/observation/checked-observer.ts
```diff
@@ -12,9 +11,12 @@ import {
   subscriberCollection,
   IScheduler,
   ITask,
+  getCollectionObserver,
+  ILifecycle,
 } from '@aurelia/runtime';
 import { IEventSubscriber } from './event-manager';
 import { ValueAttributeObserver } from './value-attribute-observer';
+import { RepeatableCollection } from '@aurelia/runtime/dist/observation/observer-locator';
```
fkleuver, line 19: ⛔️

bigopon: hmm, what should it be here. Should I create a local type for Map/Set/Array union?

bigopon: ✔

fkleuver: That's fine or you can export it. The main thing is to make sure you don't import from dist as this can break the runtime output sometimes

AureliaEffect:
JS Framework Benchmark: https://43735-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

PR #825, opened 2020-03-21
route recognizer rewrite / move to separate package

fkleuver:
# Pull Request


packages/route-recognizer/src/index.ts
```diff
@@ -0,0 +1,515 @@
+import { IQueryParams, parseQueryString } from '@aurelia/kernel';
```
EisenbergEffect: These live in kernel today, but is anything else outside of the router or route recognizer using them? If not, could they be moved to the route recognizer package and the dependency on kernel removed? This would make the recognizer more reusable as its weight would be significantly reduced if it doesn't have to rely on the kernel. (I guess things could be tree-shaken by a bundler but I'm also thinking of no-bundler situations and organizations that might have to get every dependency approved by security team, etc.)

AureliaEffect:
JS Framework Benchmark: https://43372-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/route-recognizer/src/index.ts
fkleuver: @EisenbergEffect Sorry, I missed this notification. Yeah currently the router is using it I think, but this is also in functionality that should become the route-recognizer's responsibility. The plan is definitely to consolidate it and I think we can remove the dependency from the kernel, good point.

AureliaEffect:
JS Framework Benchmark: https://43441-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://43534-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://43546-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://43648-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

jwx:
Out of curiosity, why have our own query string routines rather than using/following `URLSearchParams`?

fkleuver:
Hmm, I wasn't aware of that. This is just inheritance from aurelia v1. I suppose back then it wasn't standardized yet

AureliaEffect:
JS Framework Benchmark: https://43676-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://43687-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

PR #830, opened 2020-04-01
move metadata to separate package

fkleuver:
# Pull Request

## 📖 Description

Metadata itself has no dependencies and could easily be useful as a standalone package, so for this reason as well as to maintain some degree of consistency with v1, it'll go back to its own package now.


AureliaEffect:
JS Framework Benchmark: https://43802-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

PR #831, opened 2020-04-01
cleanup & move scheduler to separate package

fkleuver:
# Pull Request

## 📖 Description

This PR is intended to move the scheduler from the runtime package to a separate scheduler package that the runtime takes a dependency on. This way, it can be used without taking a dependency on the large runtime.
It also cleans up a few implementation details of the scheduler to make it smaller and more performant.


AureliaEffect:
JS Framework Benchmark: https://43889-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://43966-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

PR #835, opened 2020-04-06
docs: add a contributing file for cli developers

xenoterracide:
this will be easier to find than the reference in the README,
if you have only cloned the repo.

I had this problem, I basically never noticed the contributing docs, because I was just code hunting at first.


fkleuver:
Good call, thanks!

PR #838, opened 2020-04-08
feat(validation): package bifurcation + changeOrEvent trigger behavior change

Sayan751:
# Pull Request

## 📖 Description

- The validation plugin is bifurcated into `@aurelia/validation` and `@aurelia/validation-html`. The later wraps the former. The `@aurelia/validation-i18n` wraps the `@aurelia/validation-html` package.
- This PR incorporates the change in default behavior in the `changeOrEvent` trigger, done for au1 (refer: https://github.com/aurelia/validation/pull/544). This also introduces the `focusout` trigger.
- Usage of `BaseValidationRule` is minimized in favor of `IValidationRule`.
- The change/customization of the `validation-container` CE is facilitated. As the CE uses shadowDOM, it is difficult to create flexible set of CSS custom properties (read variables) that supports every use-cases in devland. Instead, devs can now use a template as per their UI need.
- Incorporates the fix from https://github.com/aurelia/validation/pull/547
- Fixed validation controller injection issue. In the subscribers, made the scopedController injection optional, as it is possible to avoid the scoped controller usage completely and live of explicitly bound validation controllers.


AureliaEffect:
JS Framework Benchmark: https://44410-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://44527-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://44546-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://44558-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://44564-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/validation-html/src/configuration.ts
```diff
@@ -0,0 +1,62 @@
...
+        }),
+        Registration.instance(IDefaultTrigger, options.DefaultTrigger),
+        ValidateBindingBehavior,
+      );
+      if (options.UseSubscriberCustomAttribute) {
+        container.register(ValidationErrorsCustomAttribute);
+      }
+      const template = options.SubscriberCustomElementTemplate;
+      if (template) { // we need the boolean coercion here to ignore null, undefined, and ''
+        container.register(CustomElement.define({ ...defaultContainerDefinition, template }, ValidationContainerCustomElement));
```
bigopon, line 51: Different sections of an app may not always use the same template. While it can still be solved with an `if`, or a local registration of `<validation-container/>`, maybe it can also be a test subject for a smoother API around retrieving the template contextually

Sayan751: @bigopon Thanks for the review :) 

This is a nice point, and frankly it didn't occur to me. Now come to think of it, there are several ways to achieve this.

1. Using a different registration like you said. That is quite easy so to speak. Devs need to do a registration, and use the alias in view like `<custom-validation-container>...</custom-validation-container>`
    ```typescript
    import { 
      ValidationContainerCustomElement,
      defaultContainerDefinition 
    } from '@aurelia/validation';

    ///

    container.register(
      CustomElement.define({ 
        ...defaultContainerDefinition, 
        name: 'custom-validation-container', 
        template: customTemplate }, 
        ValidationContainerCustomElement));
    ```

1. There are native ways to customize the template I believe, if we go via single template, for single CE approach. [This shows an example](https://developers.google.com/web/fundamentals/web-components/shadowdom#contextstyling) of styling shadowDOMs based on context. Devs can easily provide the required styles via the registered template, in order to contextualize it.

Moreover, using one alternative does not prevent another one! Does this address your concern?

bigopon: For 1, maybe that's better. I was thinking of something fancy for `<validation-container/>`  hooking into created and return different definition, based on different contexts

AureliaEffect:
JS Framework Benchmark: https://44988-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://45002-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://45010-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://45020-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

PR #839, opened 2020-04-08
Topic/ide testing improvements

xenoterracide:
due to problems getting tests debugging locally, I've created this

features include

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -25,6 +25,61 @@
     {
       "type": "node",
       "request": "launch",
+      "name": "Launch current mocha test",
+      "runtimeArgs": [
+        "mocha"
+      ],
+      "outputCapture": "std",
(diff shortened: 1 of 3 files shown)


.vscode/launch.json
```diff
@@ -22,6 +22,61 @@
...
+        "esm",
+        "--recursive",
+        "--globals",
+        "expect",
+        "--timeout",
+        "999999"
+      ],
+      "cwd": "${workspaceRoot}",
+      "internalConsoleOptions": "openOnSessionStart"
+    },
```
fkleuver, line 79: Why specifically are the kernel tests still in this config?

xenoterracide: they weren't before, I added them so I could run them all at once, though I don't really require them if you want me to remove them, more important is the ability to run a single file

PR #856, opened 2020-04-20
Add Github Actions for node tests

xenoterracide:
This is not meant to be a replacement for CircleCI the purpose here is that a person (like me) who forked the repo (on github ) can have tests run automatically. Currently the Node tests are all I had the patience to set up. I also made some improvements to `package.json` mostly because I needed a way to run the tests that had a pattern. I then used that pattern to DRY up some of the existing code.


fkleuver:
We've been meaning to look into this for a while, so thanks for doing so :)

How can I see the results of the actions?

Just one note from a quick first look, I'd prefer to avoid caching if possible, and for each "thing" to work completely in isolation from start to finish. This keeps the workflow simpler and also will speed it up when we reduce the footprint of our dependencies in the future

xenoterracide:
you can't see them here yet because it's not merged, https://github.com/xenoterracide/aurelia/actions/runs/82609579

if we aren't using the "cache" feature, then I think it's wise to do the upload/download thing for node_modules too, as this will still allow the `install` to produce that code for the other "jobs"

xenoterracide:
closing this as it's part of #837 but a separate commit

PR #860, opened 2020-04-21
Listener binding fix

fkleuver:
# Pull Request

## 📖 Description

Fixes a slip-up in the Listener binding (responsible for `trigger`, `delegate` and `capture` commands) so that it works properly in combination with binding behaviors.

### 🎫 Issues

Reported on discord

## 👩‍💻 Reviewer Notes

Fix is only 1 line

## 📑 Test Plan

Added a few tests specifically for debounce (for regular bindings and event bindings)

## ⏭ Next Steps

More tests are still necessary, also for Throttle, but at least the common cases are now verified


AureliaEffect:
JS Framework Benchmark: https://45793-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

PR #866, opened 2020-04-24
don't cache lazy,  allow configuration of the container

xenoterracide:
docs(di): add docs for lazy and optional
    feat: allow configuration of Container
    test: split out DI.get tests into their own file
    feat: stop lazy caching, let container control lifecycle


bigopon:
This looks pretty nice changes for the DI itself. Though for Aurelia apps, it probably won't be used much. Thanks @xenoterracide

PR #880, opened 2020-05-16
Logging improvement

Sayan751:
# Pull Request

## 📖 Description

This PR introduces the following improvements for logging. 
1. The `console` logging can be deactivated by simply providing `null` or `undefined` for the console implementation.
```typescript
container.register(LoggerConfiguration.create())
// OR
container.register(LoggerConfiguration.create({$console: null}));
```
2. Developers can now register Sinks explicitly while registering `LoggerConfiguration`.
```typescript
container.register(LoggerConfiguration.create({sinks: [SinkClass1, SinkClass2]}));
```
3. Sinks can declare their handling capabilities using the `sink` decorator. For the following example, the sink instance won't be notified of events other than `error`.
```typescript
@sink({ handles: [LogLevel.error] })
class EventLog implements ISink {
  public readonly log: ILogEvent[] = [];
  public handleEvent(event: ILogEvent): void {
    this.log.push(event);
  }
}
```
The use of `@sink` decorator is completely optional. By default, the sinks will be notified on every kind of events.
4. Lastly the `ISink#emit` is renamed to `ISink#handleEvent`.

### DIFF
--- a/packages/__tests__/1-kernel/logger.spec.ts
+++ b/packages/__tests__/1-kernel/logger.spec.ts
@@ -5,6 +5,10 @@ import {
   ILogger,
   LogLevel,
   ColorOptions,
+  ISink,
+  ILogEvent,
+  sink,
+  DefaultLogger,
 } from '@aurelia/kernel';
 import { assert, eachCartesianJoin } from '@aurelia/testing';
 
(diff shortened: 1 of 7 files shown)


AureliaEffect:
JS Framework Benchmark: https://47693-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

Issue #822, opened 2020-03-18
TODO: add static kernel api for getting/setting type-associated factories


fkleuver:
Related to comment: https://github.com/aurelia/aurelia/pull/811#discussion_r394164268

Sayan751:
Resolved with #881

PR #881, opened 2020-05-16
fix(di): registerFactory #822

Sayan751:
# Pull Request


AureliaEffect:
JS Framework Benchmark: https://47726-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

PR #884, opened 2020-05-19
docs(app-basics): html only custom elements

Vheissu:
Added in the documentation for HTML custom elements, including how to specify bindable properties and change the binding mode.

### DIFF
--- a/docs/user-docs/app-basics/components-revisited.md
+++ b/docs/user-docs/app-basics/components-revisited.md
@@ -21,7 +21,7 @@ Here is a basic example that will render a heading one element with a value of `
 Saving this as `my-element.html` will result in a component that will be referenced using its tag `<my-element></my-element>`. To import the component in your application, use the `<import>` element.
 
 ```HTML
-<import from="./my-element.html" />
+<import from="./my-element.html"></import>
 
 <my-element></my-element>
 ```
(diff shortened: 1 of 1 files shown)


AureliaEffect:
JS Framework Benchmark: https://47910-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

docs/user-docs/app-basics/components-revisited.md
```diff
@@ -6,7 +6,49 @@
...
+
+<my-element></my-element>
+```
+
+### With Bindable Properties
+
+In many instances, you'll want a custom element which supports one or more bindable properties. These properties allow you to pass in data to the component itself. Taking the above example, let's allow the text to be changed and we will save it as `heading-one.html` instead.
+
+```HTML
+<bindable name="text" />
```
fkleuver, line 34: @3cp Is this something the convention plugin is supposed to do?

3cp: Yes, conventions got that.

BTW, @Vheissu I voiced up many times in the part: **we should not write self-closing in HTML5 because it's not HTML5**, it's a XML spec which HTML5 didn't support.

Considering following example:
```html
<div />
<div />
```
Do you think HTML5 take it as
```html
<div></div>
<div></div>
```

Nope, HTML5 simply took it as
```html
<div>
  <div>
  </div>
</div>
```

Using self-closing in HTML5 is very misleading, although there is no harm for our meta tags as conventions removes all of them during preprocessing.

My argument is that we do not demo "broken HTML5" in our doc or example code. On the other side, the conventions implementation tolerates both `<bindable name="f" />` and `<bindable name="f">` without proper closing, but the correct form is always `<bindable name="f"></bindable>`.

3cp: `<bindable name="f">` is better than `<<bindable name="f" />`, if we want to promote code with less verbosity.

We could explicitly document that users do not need to close those meta tags, they can be used as if they are native void elements like `<input>` and `<meta>`.

The list are
```html
<import from="./a">
<bindable name="a">
<use-shadow-dom mode="open">
<containerless>
<alias name="a">
```

AureliaEffect:
JS Framework Benchmark: https://47940-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

Issue #666, opened 2019-10-06
Enhance api


fkleuver:
Placeholder issue so we don't forget. To be prioritized

Sayan751:
### Use cases

1. Enhance a not-enhance part in an enhance Aurelia app. Examples:

```html
<body>
  <app></app> <!-- already enhanced -->
  <not-enhanced></not-enhanced>
</body>
```

```html
<head>
  <!--not enhanced-->
  ${awesomeStuff}
</head>
<body>
  <app></app> <!-- already enhanced -->
</body>
```

2. Enhance a not-enhance part inside an enhanced part. For example hydrating bound raw HTML strings.

```html
<app>
<!-- already enhanced -->
  ...
    ...
      <div inner-html.bind="hydratableHtmlString"></div>
</app>
```

### API

```typescript
// Injected Aurelia instance; otherwise the plugin might needs to re-registered for enhance
let au: Aurelia;

// alternative#1
await au
  .enhance(element, viewModelInstanceOrClass)
  .start();

// alternative#2
await au
  .enhance({host: element, component: viewModelInstanceOrClass}) // <-- like au.app() it accepts ISinglePageApp
  .start();
```
If an enhance root is tried for re-enhancement, that can be ignored with a mere warning.

Question: Does it make sense to create a child `Aurelia` instance on `enhance` and return that instead? I think do so will not interfere with the workflow of root `Aurelia` instance. I have no evidence to prove any advantages of that, just throwing out ideas.

bigopon:
In v1, calling `enhance` returns a View, and I think it makes sense. For v2, that'd be our Controller. As you may want to do things that are independent with an Aurelia instance

Sayan751:
Closed with #888

PR #888, opened 2020-05-23
feat: enhance API

Sayan751:
# Pull Request

## 📖 Description

This PR introduces the enhance API (#666). Like Au1, enhance can be used to hydrate any existing DOM node. The `enhance` is possible for following cases:

- Element root outside hydrated app root.
  ```html
  <app></app> <!-- already hydrated-->
  <enhance-target></enhance-target> <!--not hydrated-->
  ```
- Element root outside hydrated app root.
  ```html
  <app>
  <!-- already hydrated-->
    <div innerhtml.bind="'<div>\${message}</div>'"></div><!-- hydration target -->
  </app>
  ```

The API can be used like this:

```typescript
const au = new Aurelia()
    .register(configuration);
au.enhance({
  host: document.querySelector('enhance-root'), 
  component: classOrInstance
});
await au.start().wait();
```
## 👩‍💻 Reviewer Notes

- ~~The API directly grabs the child node instances from the enhance root.~~
- ~~`NodeList` instance checking is implemented via string coercion:  https://github.com/aurelia/aurelia/pull/888/files?file-filters%5B%5D=.ts#diff-5774ba38006e0f79ec112022c1cd197cR284-R287~~

## 📑 Test Plan

Couple of integrations tests are already put in place.


AureliaEffect:
JS Framework Benchmark: https://48157-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://48199-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/runtime/src/templating/controller.ts
```diff
@@ -303,6 +339,7 @@ export class Controller<
     definition: CustomElementDefinition,
     parentContainer: IContainer,
     parts: PartialCustomElementDefinitionParts | undefined,
+    toEnhance: boolean = false,
```
bigopon, line 342: I think for private method, we can enforce strictness so that there's unnecessary mistakes/computation

packages/runtime/src/aurelia.ts
```diff
@@ -158,6 +163,29 @@ export class CompositionRoot<T extends INode = INode> {
     const lifecycle = container.get(ILifecycle);
     this.controller = Controller.forCustomElement(instance, lifecycle, this.host, container, void 0, this.strategy as number);
   }
+
+  private enhance(): void {
+    const container = this.container;
+    const config = this.config;
+    const component = config.component as Constructable | ICustomElementViewModel<T>;
+    const template = (this.host as INode as Element).childNodes;
```
bigopon, line 171: Does this mean that if we want to enhance an element, we have to give the parent of that node to enhance?

bigopon: if we are to enhance from whatever node (host) that is given, we don't need to deal with casting to `Element`, which doesn't belong to runtime as well

bigopon:
Very nice 👍  I left a few comments

packages/runtime/src/aurelia.ts
Sayan751: > Does this mean that if we want to enhance an element, we have to give the parent of that node to enhance

Yes. I think it is also analogous to the `.app()` as the host itself is not hydrated. One problem to do that otherwise is this part here:
https://github.com/aurelia/aurelia/blob/2e3f814610e1ba41077c69359a2fe7d2ac8cb1e8/packages/jit-html/src/template-element-factory.ts#L91-L96
If the `host` is used directly as template, it gets detached from the parent, and later when re-attaching is attempted, the following error is thrown:

```shell
dom.js:326 Uncaught (in promise) DOMException: Failed to execute 'appendChild' on 'Node': The new child element contains the parent.
    at FragmentNodeSequence.appendTo (http://localhost:9500/main.js:16060:20)
    at HostProjector.project (http://localhost:9500/main.js:18650:15)
    at Controller.mountCustomElement (http://localhost:9500/main.js:31296:24)
    at Controller.mount (http://localhost:9500/main.js:30951:22)
    at MountQueue.process (http://localhost:9500/main.js:24449:21)
    at AttachedQueue.end (http://localhost:9500/main.js:24277:34)
    at Controller.attachCustomElement (http://localhost:9500/main.js:31205:36)
    at Controller.attach (http://localhost:9500/main.js:30915:22)
    at Activator.attach (http://localhost:9500/main.js:20094:103)
    at Activator.activate (http://localhost:9500/main.js:20074:18)
```

Theoretically, the 'host' and the 'template' should be 2 separate entities. Of course I can add `enhance`-specific logic in `template-element-factory`, but I am not sure if that's a good idea.

packages/runtime/src/templating/controller.ts
Sayan751: Good catch! Fixed :)

AureliaEffect:
JS Framework Benchmark: https://48221-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

bigopon:
btw, you got the number 888 of our PRs/issues 🥳

Sayan751:
> btw, you got the number 888 of our PRs/issues 🥳
> 
> EDIT: and the issue number of this PR is 666 😮

I know, what are the odds!! 😈 😈

packages/runtime/src/aurelia.ts
bigopon: For https://github.com/aurelia/aurelia/blob/2e3f814610e1ba41077c69359a2fe7d2ac8cb1e8/packages/jit-html/src/template-element-factory.ts#L91-L96

The nodes are moved out of the document, or wherever they are, put into a template, and thus, may cause the loss of state: input caret position, selection, scroll state etc...
I think we can tweak the compiler, so that it'll only compile what it's given, instead of trying to convert the input into the way it can work with.
At least it's only safe to do so if the template given for a compilation is a string.

Sayan751: Made the changes. Hydration of the root seems to be working.

AureliaEffect:
JS Framework Benchmark: https://48245-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://48276-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/runtime/src/renderer.ts
```diff
@@ -58,7 +58,7 @@ import { BindingBehaviorExpression } from './binding/ast';
 import { BindingBehaviorFactory, BindingBehaviorInstance, IInterceptableBinding } from './resources/binding-behavior';
 
 export interface ITemplateCompiler {
-  compile(partialDefinition: PartialCustomElementDefinition, context: IContainer): CustomElementDefinition;
+  compile(partialDefinition: PartialCustomElementDefinition, context: IContainer, wrap?: boolean): CustomElementDefinition;
```
bigopon, line 61: I think this is a bit cryptic, in terms on self explanation. In v1, we have view compile instruction, and i remember early version of the compiler interface had it as well. Maybe we can bring it back? Or at least a number to store all instructions maybe

Sayan751: That's a very good point. In validation I also had this notion of validation "instruction". That is bit different from the perspective that the instruction class encapsulates the data. Do you think we can do something similar here as well?

https://github.com/aurelia/aurelia/blob/685df881206668f4b26185c8e10aef23d0f97ba3/packages/validation/src/validator.ts#L6-L26

https://github.com/aurelia/aurelia/blob/685df881206668f4b26185c8e10aef23d0f97ba3/packages/validation/src/validator.ts#L52-L66

Sayan751: removed

AureliaEffect:
JS Framework Benchmark: https://48310-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://48325-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/jit-html/src/template-compiler.ts
```diff
@@ -117,7 +117,9 @@ export class TemplateCompiler implements ITemplateCompiler {
 
     const binder = new TemplateBinder(context.get(IDOM), resources, attrParser, exprParser, attrSyntaxModifier);
 
-    const template = factory.createTemplate(definition.template) as HTMLTemplateElement;
+    const template = definition.enhance === true
```
bigopon, line 120: This feels much better than always `factory.createTemplate()`

packages/runtime-html/src/dom.ts
```diff
@@ -430,7 +430,9 @@ export class FragmentNodeSequence implements INodeSequence {
       }
     } else {
       this.isMounted = true;
-      parent.appendChild(this.fragment);
+      if (!parent.contains(this.fragment)) {
```
bigopon, line 433: can we have some comment on which cases we need to perform this check? I'm unable to guess this.

packages/__tests__/2-runtime/enhance.spec.ts
```diff
@@ -0,0 +1,144 @@
...
+      class App2 {
+        private readonly r1!: HTMLDivElement;
+        private readonly r2!: HTMLDivElement;
+
+        public async afterAttach() {
+          await this.enhance(this.r1);
+        }
+
+        private async enhance(host = this.r2) {
+          await new Aurelia(TestContext.createHTMLTestContext().container)
```
bigopon, line 109: maybe there should be another test for reusing an existing container?

Sayan751: Added more tests, same and child container seems to be working as well.

packages/runtime-html/src/dom.ts
Sayan751: For enhance the `host` and `fragment` will be same. However, instead of equality check I think that the more general `contains` check makes more sense in this case, as we might generally want to skip the append action if the host already has the child.

AureliaEffect:
JS Framework Benchmark: https://48384-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/__tests__/2-runtime/enhance.spec.ts
Sayan751: @bigopon However, I think it might be safest to throw an error on `new Aurelia(container)` if `container.has(Aurelia, true)` satisfies. Otherwise it might create registration conflict. Throwing an appropriate error upfront might be better than facing unexpected issues down in the line.

@fkleuver what do you think?

packages/runtime/src/aurelia.ts
```diff
@@ -90,6 +96,19 @@ export class CompositionRoot<T extends INode = INode> {
     }
   }
 
+  private setEnhanceDefinition(config: ISinglePageApp<T>) {
```
bigopon, line 99: my last comment on this would be the name of this method. it feels a bit out of place since all we have is build/create/compile. Set doesn't feel fit. Other than this, this looks great

packages/runtime/src/templating/controller.ts
```diff
@@ -215,12 +215,13 @@ export class Controller<
     parentContainer: IContainer,
     parts: PartialCustomElementDefinitionParts | undefined,
     flags: LifecycleFlags = LifecycleFlags.none,
+    definition: CustomElementDefinition | undefined = void 0,
```
bigopon, line 218: @fkleuver what do you think about this

Sayan751: This materialized from the intention of streamlining things a bit. Previously the create and enhance methods were separate and so were the Controller counterparts. But both end up using the same 99% processing pipeline. Therefore, I sort of cached the definition for Controller later.

AureliaEffect:
JS Framework Benchmark: https://48493-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/runtime/src/templating/controller.ts
fkleuver: The definition is already cached directly in metadata based on the `constructor` as the key.

So for an additional caching layer I don't see any benefits here.
Otherwise, this tells me there is a situation where you need to pass in a definition that is different than the one that is immediately associated with the constructor (or the constructor has no definition at all and you don't want to associate it either). Is that correct?

packages/runtime/src/aurelia.ts
```diff
@@ -266,6 +283,16 @@ export class Aurelia<TNode extends INode = INode> {
     return this.task.wait() as Promise<void>;
   }
 
+  private configureRoot(config: ISinglePageApp<TNode>, toEnhance?: boolean): Omit<this, 'register' | 'app' | 'enhance'> {
```
fkleuver, line 286: Why `toEnhance` instead of just `enhance`?

packages/runtime/src/aurelia.ts
fkleuver: I can certainly understand the intuition to encapsulate this, but it's only used inside the constructor and not *that* big. Just one fairly atomic operation, unlike e.g. `hydrateCustomElement`. Maybe just keep this inline in the ctor?

AureliaEffect:
JS Framework Benchmark: https://50307-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/runtime/src/aurelia.ts
```diff
@@ -90,6 +96,19 @@ export class CompositionRoot<T extends INode = INode> {
...
+    const component = config.component as Constructable | ICustomElementViewModel<T>;
+    if (CustomElement.isType(component)) {
+      CustomElement.define(
+        this.enhanceDefinition = CustomElementDefinition.create({ ...CustomElement.getDefinition(component), template: this.host, enhance: true }),
+        component);
+    } else {
+      CustomElement.define(
+        this.enhanceDefinition = CustomElementDefinition.create({ name: CustomElement.generateName(), template: this.host, enhance: true }),
+        class { });
+    }
```
fkleuver, line 109: This looks good functionally, but it could be simplified a little bit because some stuff the `define` api already does for you:
```ts
    if (CustomElement.isType(component)) {
      this.enhanceDefinition = CustomElement.getDefinition(
        CustomElement.define({ ...CustomElement.getDefinition(component), template: this.host, enhance: true },
        component)
      );
    } else {
      this.enhanceDefinition = CustomElement.getDefinition(CustomElement.define({ template: this.host, enhance: true }));
    }
```

Btw, what if the component instance has hooks on it? Shouldn't we make that work?

packages/runtime-html/src/dom.ts
fkleuver: The idea is good but I'd prefer then if we throw instead, otherwise people try to do something, they don't know they're doing anything wrong, and nothing happens and they won't know why. Silent failures are the worst.
But if I'm not mistaken, `parent.appendChild` already throws if it's contained in the tree, so why not just let the DOM api throw? Is there a specific use case where we need to allow this?

packages/__tests__/integration/integration.spec.ts
```diff
@@ -1,942 +1,948 @@
...
 describe('app', function () {
 
-  $it('has some readonly texts with different binding modes', function ({ host }) {
-    for (let i = 0; i < 4; i++) {
-      const selector = `read-only-text#text${i}`;
-      assert.html.textContent(selector, `text${i}`, `incorrect text for ${selector}`, host);
-    }
-  });
+  eachCartesianJoin([
+    ['app', 'enhance'] as ('app' | 'enhance')[],
```
fkleuver, line 15:
```suggestion
    ['app', 'enhance'] as ['app', 'enhance'],
```
Just a bit simpler and a bit more accurate. Not really important though :)

packages/__tests__/2-runtime/enhance.spec.ts
```diff
@@ -0,0 +1,171 @@
...
+
+        assert.html.textContent('div', message, 'div', host);
+
+        host.querySelector('button').click();
+        const scheduler = container.get(IScheduler);
+        scheduler.getPostRenderTaskQueue().flush();
+
+        assert.html.textContent('div:nth-of-type(2)', message, 'div:nth-of-type(2)', host);
+
+        await au.stop().wait();
```
fkleuver, line 166: I think we should have some idempotency tests as well, unless I missed them. So `start` + `stop` a second time, verifying that everything is still the same etc, if that makes sense?

packages/runtime/src/aurelia.ts
Sayan751: It's like the yes/no question, "to enhance or not enhance" :) But can change, if it does not compatible with current code style.

packages/__tests__/2-runtime/enhance.spec.ts
fkleuver: Yeah probably best to throw, either that or we need to use an `InstanceProvider` manually and replace the instance. But that wouldn't be compatible with multi instance situations though. People should just create forked child containers in those cases

packages/runtime-html/src/dom.ts
Sayan751: Actually it does throw. We need this check only for enahnce. I felt propagating the enhance flag all the way to dom bit of an overkill. That's why I felt this check to be a reliable proxy. But the changed version looks much cleaner.

packages/runtime/src/aurelia.ts
Sayan751: Done

packages/runtime/src/aurelia.ts
Sayan751: Done

packages/__tests__/integration/integration.spec.ts
Sayan751: Done

AureliaEffect:
JS Framework Benchmark: https://50338-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/__tests__/2-runtime/enhance.spec.ts
Sayan751: Thrown error for now.

packages/runtime/src/templating/controller.ts
Sayan751: If the `component` is not a type at all (raw object; `CustomElement.isType(component)` returns `false`), then the definition is not associated with the viewModel instance. From that perspective, it makes sense to pass on the definition directly.

AureliaEffect:
JS Framework Benchmark: https://50349-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/runtime/src/aurelia.ts
Sayan751: Great point. Missed that. Made the change.

packages/runtime/src/templating/controller.ts
fkleuver: Ah I see, that makes sense yeah. Then I'd say add a comment like this:
```ts
// This is passed in when `instance.constructor` is not a custom element type
definition: CustomElementDefinition | undefined = void 0,
```

packages/runtime/src/aurelia.ts
fkleuver: Lol :) yeah ok, I understand how you get to that.
As a rule of thumb we prefix boolean properties with either `is` (to describe a state) or modal verbs like `can`, `should` (to describe ability, obligation, etc).
When that doesn't necessarily make sense, I tend to just use the word itself without any prefix/suffix. `shouldEnhance` could work here, but in this case feels a bit weird/wrong to me

packages/__tests__/2-runtime/enhance.spec.ts
Sayan751: Done

AureliaEffect:
JS Framework Benchmark: https://50366-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://50492-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://50502-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

PR #889, opened 2020-05-23
Topic/register in requester

xenoterracide:
* add `@scoped` decorator, was unable to get `@singleton(true)` this is facilitated by adding a `registerInRequester` property to the decorated injectable
* more fixes to github actions, apparently something was bad caching
* remove registerInRequester container configuration


packages/kernel/src/di.ts
```diff
@@ -376,11 +375,12 @@ export const DI = {
    * Foo.register(container);
    * ```
    */
-  transient<T extends Constructable>(target: T & Partial<RegisterSelf<T>>): T & RegisterSelf<T> {
+  transient<T extends Constructable>(target: T & Partial<RegisterSelf<T>>, registerInRequester: boolean = false): T & RegisterSelf<T> {
```
fkleuver, line 378: Is there a real use case for this or is this just a consistency thing?

xenoterracide: registerInRequester isn't optional, so I made it consistent instead of just setting it to false.

xenoterracide: I forgot this was there, I suppose I should update this to match singleton? or I could always just set it to false

xenoterracide: removed the parameters

packages/__tests__/1-kernel/di.get.spec.ts
```diff
@@ -16,21 +16,60 @@ describe('DI.get', function () {
...
       container.register(Registration.transient(Bar, Bar));
       const bar0 = container.get(Foo).provider();
       const bar1 = container.get(Foo).provider();
 
       assert.notStrictEqual(bar0, bar1);
     });
+  });
+
+  describe('@scoped', function () {
+    @singleton({ scoped: true })
```
fkleuver, line 36: Could you also add an explicit test with `scoped: false` just in case? Because I don't see any `.registerInRequester === true` anywhere, so I'm a bit confused as to how `true` is distinguished from `false` lol

xenoterracide: uh.... so much egg on face... apparently it wasn't, I was (foolishly?) thinking there was an existing test that ensured that registering in a child actually checked that it went to the root.

Test and fix added.

PR #895, opened 2020-06-05
feat: local template

### LINKED ISSUE #891: [RFC]: Local template
# 💬 RFC

Proposal for supporting local templates, and related discussion.

## 🔦 Context

In real-life Aurelia *apps* it is quite common that a custom element ends up getting used in one single custom element. This scenario excludes the reusable component libraries, offering a bunch custom elements.

For example, let's consider this simplistic example of a `person-template`.

```html
<!--person-template-->
${person.lastName}, ${person.firstName}
```
```html
<!--user-info-->

<!--initial display-->
<person-template person.bind="person"></person-template>

<!--in popup-->
<person-template person.bind="person"></person-template>
```
In this hypothetical scenario, this can be used in a CE that shows logged-in user's info in a popup like menu. Let's also assume that the `person-template` CE gets used multiple times in the user info CE. Therefore, having the person template as a CE makes it easier to reuse. However, there is a fair chance that such specialized templates get used only in one CE. Therefore, creating a full-fledged CE for that purpose seems a bit overkill.

An alternative idea is to support local named templates.

## 💻 Examples

With the named template, the example above will look something like this:

```html
<!--user-info-->
<!--define local template-->
<template id="person-template">
  <bindable property="person" mode="toView"></bindable>
  ${person.lastName}, ${person.firstName}
</template>

<!--usages-->
<!--initial display-->
<person-template person.bind="person"></person-template>

<!--in popup-->
<person-template person.bind="person"></person-template>
```

This removes the need of having the template in form of a separate custom element. This approach can be considered analogous to [local functions](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/local-functions). Moreover, local templates offer more freedom in terms of refactoring, as it is already known that this is not used in elsewhere, and the ripple of change is limited to this particular CE.

### More examples

* Local template w/o `bindable`s

```html
<template id="local-template">
  template body
</template>

<local-template></local-template>
```

* Local template with `bindable`s

```html
<template id="local-template">
  <bindable property="prop" mode="bindingMode" attribute="attr"></bindable>
  template body using 'prop'
</template>

<local-template attr.binding-mode="value"></local-template>
```

Note that the attributes used in `bindable`s refer to this:

https://github.com/aurelia/aurelia/blob/685df881206668f4b26185c8e10aef23d0f97ba3/packages/runtime/src/templating/bindable.ts#L17-L24

For a local template, `callback`, `primary`, and `set` probably do not make much sense. The rest should be supported in the `bindable`s in a local template.

## Goals and Non-Goals

From the discussion so far, we can draw the line between the goals and non-goals. 

The **goal is** to reduce the boilerplate code and to promote highly cohesive, better encapsulated components/custom elements.

> Keep the things together that varies together.
> \-Old jungle sayings

The **goal is not** to reuse these local templates in another custom elements. In such cases, the template can be pulled out to a separate full-fledged custom element.

## Open questions

1. What attribute to use as the local template identifier? The examples above use the `id` property as the template identifier. This is also aligned with the [standard usage of `template`s](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/template). Alternatives can be `<template name="...">`, `<template custom-element="...">`, etc. 
1. Should a local template, defined in a parent CE, be allowed to be used in nested child CE? Consider the following example.
  ```html
  <!--custom-element-->
  <template id="local-template">
  ...
  </template>

  <another-custom-element>
    <!-- this can be a custom element defined externally or another local template -->
  </another-custom-element>
  ```

  ```html
  <!--another-custom-element-->
  <local-template></local-template>
  ```

  If `another-custom-element` is used elsewhere (other than `custom-element`), then `local-template` might end up getting ignored. The question however is that, when it is used in `custom-element` whether or not to allow such usage of local templates? As outlined above this is also a non-goal.

## Solution Idea

The preliminary idea is to create a custom element with the template body, using the template identifier as custom-element name. The `bindable`s can be stripped from the template definition/body and appropriate metadata can be defined on an anonymous class. The same class is also used while defining the custom element definition. This new custom element can then be registered with the current render context, thereby making it available for future compilation steps.

**Note#1:** Registering with the render context will possibly make the new CE available for the child CEs as well. This brings us back the Q#2 as stated above. If the usage shown in the question is allowed then the advantage of "limited ripple of change" is compromised.
**Note#2:** As the template compilation goes top-to-bottom, any usage of the local template before the definition of the template can end up being non-hydrated. The question is whether that's a reasonable restriction, or the processing of the local templates should be prioritized and thereby effectively hoisting those templates?

[Inspiration from olden days](https://knockoutjs.com/documentation/template-binding.html) :)

Sayan751:
# Pull Request

## 📖 Description

Closes #891


packages/__tests__/5-jit-html/template-compiler.spec.ts
```diff
@@ -1227,3 +1231,253 @@ describe(`TemplateCompiler - combinations`, function () {
     });
   });
 });
+
+describe('TemplateCompiler - local templates', function () {
```
bigopon, line 1235: probably we should have at least 1 main test: the template of a custom element is a single template element, with `as-custom-element` attribute on it. Need to make some decision about it though.

packages/jit-html/src/template-compiler.ts
```diff
@@ -146,6 +153,81 @@ export class TemplateCompiler implements ITemplateCompiler {
...
+        case 'twoWay':
+          return BindingMode.twoWay;
+        case 'default':
+        default:
+          return BindingMode.default;
+      }
+    }
+
+    for (const localTemplate of localTemplates) {
+      if (localTemplate.parentNode !== root) {
```
bigopon, line 196: we may want to issue some other warning as well, maybe if we detect any other attributes on this? Since they all are gonna be ignored

packages/jit-html/src/template-compiler.ts
```diff
@@ -146,6 +153,81 @@ export class TemplateCompiler implements ITemplateCompiler {
...
+        throw new Error('Local templates needs to be defined directly under root.'); /* TODO: use reporter/logger */
+      }
+      const name = getTemplateName(localTemplate);
+
+      // eslint-disable-next-line @typescript-eslint/class-name-casing
+      const localTemplateType = class { };
+      const content = localTemplate.content;
+      const bindables = toArray(content.querySelectorAll('bindable'));
```
bigopon, line 204: if it's an element, maybe name it so? `bindableEls`?

AureliaEffect:
JS Framework Benchmark: https://48860-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://48847-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/__tests__/5-jit-html/template-compiler.spec.ts
Sayan751: That's practically an empty template; i.e. nothing will be rendered for that custom element. As Au2 is not generally much restrictive, a mere warning about such situation would suffice IMO.

Sayan751: I understood it bit differently. I think it is best to throw in case of following case:

```html
<template as-custom-element="...">
...
</template>
```

There is another case.
```html
<template as-custom-element="...">
...
</template>
<template as-custom-element="...">
...
</template>
```
I think in both cases, there is no point in compiling the parent custom element. I think throwing error in both cases will be acceptable.

bigopon: Throwing is a bit restrictive, but I'm not opposing that completely. Can always ease it later. I just meant we don't leave edge cases open.

bigopon:
This looks great to me, cant wait for your awesome doc skill on this 😁

packages/__tests__/5-jit-html/template-compiler.spec.ts
Sayan751: I have thrown for now both the cases, as both the cases are meaningless. The error can be prevented by adding just any other trivial element. I propose to revisit this if devs find it annoying.

packages/jit-html/src/template-compiler.ts
Sayan751: Done

packages/jit-html/src/template-compiler.ts
Sayan751: Done

Sayan751:
@bigopon Thanks for the review, and for reminding me that I need to write the docs :)

packages/__tests__/5-jit-html/template-compiler.spec.ts
```diff
@@ -1227,3 +1236,295 @@ describe(`TemplateCompiler - combinations`, function () {
...
+
+    await au.start().wait();
+
+    assert.html.textContent(host, expectedContent);
+
+    await au.stop().wait();
+    ctx.doc.body.removeChild(host);
+  });
+
+  it('works with nested templates', async function () {
```
bigopon, line 1422: the spec name of this test gives an interesting scenario:
```html
<template as-custom-element="one">
  <template as-custom-element="one-two">
  </template>
  <one-two>
</template>
<template as-custom-element="two">
  <template as-custom-element="two-two">
  </template>
  <two-two>
</template>
```
Do we want to allow infinite level of nesting for local template? I think the way it works currently does allow it

AureliaEffect:
JS Framework Benchmark: https://48871-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/__tests__/5-jit-html/template-compiler.spec.ts
Sayan751: It does work. Added a test to showcase that.

AureliaEffect:
JS Framework Benchmark: https://48883-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://48894-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://48904-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/jit-html/src/template-compiler.ts
```diff
@@ -146,6 +169,110 @@ export class TemplateCompiler implements ITemplateCompiler {
...
+      const properties: Set<string> = new Set();
+      const attributes: Set<string> = new Set();
+      for (const bindableEl of bindableEls) {
+        if (bindableEl.parentNode !== content) {
+          throw new Error('Bindable properties of local templates needs to be defined directly under root.'); /* TODO: use reporter */
+        }
+        const property = bindableEl.getAttribute(LocalTemplateBindableAttributes.property);
+        if (property === null) { throw new Error(`The attribute 'property' is missing in ${bindableEl.outerHTML}`); /* TODO: use reporter */ }
+        const attribute = bindableEl.getAttribute(LocalTemplateBindableAttributes.attribute) ?? void 0;
+        if (attributes.has(attribute!) || properties.has(property)) {
```
fkleuver, line 244:
```suggestion
        if ((attribute !== null && attributes.has(attribute)) || properties.has(property)) {
```

And just keep `attribute` `null` instead of `?? void 0`?

packages/jit-html/src/template-compiler.ts
```diff
@@ -146,6 +169,110 @@ export class TemplateCompiler implements ITemplateCompiler {
...
+    }
+
+    for (const localTemplate of localTemplates) {
+      if (localTemplate.parentNode !== root) {
+        throw new Error('Local templates needs to be defined directly under root.'); /* TODO: use reporter */
+      }
+      const name = getTemplateName(localTemplate);
+
+      // eslint-disable-next-line @typescript-eslint/class-name-casing
+      const localTemplateType = class { };
```
fkleuver, line 231:
```suggestion
      const localTemplateType = class LocalTemplate { };
      Reflect.defineProperty(localTemplateType, 'name', { value: pascalCase(name), configurable: true });
```

Just a bit of a nitpick, but would make it a bit more debugger friendly

packages/jit-html/src/template-compiler.ts
```diff
@@ -146,6 +169,110 @@ export class TemplateCompiler implements ITemplateCompiler {
...
+      const content = localTemplate.content;
+      const bindableEls = toArray(content.querySelectorAll('bindable'));
+      const bindableInstructions = Bindable.for(localTemplateType);
+      const properties: Set<string> = new Set();
+      const attributes: Set<string> = new Set();
```
fkleuver, line 236:
```suggestion
      const attributes = new Set<string>();
```
More nitpicks

AureliaEffect:
JS Framework Benchmark: https://50316-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/jit-html/src/template-compiler.ts
```diff
@@ -146,6 +169,110 @@ export class TemplateCompiler implements ITemplateCompiler {
...
+    }
+    const localTemplateNames: Set<string> = new Set();
+
+    function getTemplateName(localTemplate: HTMLTemplateElement): string {
+      const name = localTemplate.getAttribute(localTemplateIdentifier);
+      if (name === null || name === '') {
+        throw new Error('The value of "as-custom-element" attribute cannot be empty for local template'); /* TODO: use reporter/logger */
+      }
+      if (localTemplateNames.has(name)) {
+        throw new Error(`Duplicate definition of the local template named ${name}`); /* TODO: use reporter/logger */
```
fkleuver, line 202:
```suggestion
        throw new Error(`Duplicate definition of the local template named ${name}`);
```

Throwing is fine. We're not going to use the reporter, it will be removed, so no need to use these TODOs anymore

AureliaEffect:
JS Framework Benchmark: https://50398-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/jit-html/src/template-compiler.ts
Sayan751: Done

packages/jit-html/src/template-compiler.ts
Sayan751: Done

packages/jit-html/src/template-compiler.ts
Sayan751: Done

AureliaEffect:
JS Framework Benchmark: https://50406-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

Issue #906, opened 2020-07-04
e2e test scenarios (todos)


fkleuver:
Placeholder issue for things that cannot directly be unit or integration tested because they need specific project setups:
- Verify that multiple versions of the same `@aurelia/metadata` module is detected and thrown properly
- Verify that `reflect-metadata` can be prevented from applying itself by specific `applyMetadataPolyfill` options

Vheissu:
Given we removed the `@aurelia/metadata` error message and other changes since this was created, is this issue still needed @fkleuver @bigopon ?

bigopon:
It's fine now. I've resolved it. Metadata will try reusing the existing weakmap if it detects any.

Issue #902, opened 2020-06-22
aurelia/metadata conflicts with reflect-metadata which is required by typeorm


xenoterracide:
```
    Conflicting reflect.metadata polyfill found. If you have 'reflect-metadata' or any other reflect polyfill imported, please remove it, if not (or if you must use a specific polyfill) please file an issue
at https://github.com/aurelia/aurelia/issues so that we can look into compatibility options for this scenario.

      at Object.applyMetadataPolyfill (node_modules/@aurelia/metadata/src/index.ts:1032:11)
      at node_modules/@aurelia/kernel/src/di.ts:3:1
      at node_modules/@aurelia/kernel/dist/umd/di.js:3:17
      at Object.<anonymous> (node_modules/@aurelia/kernel/dist/umd/di.js:9:3)
      at node_modules/@aurelia/kernel/src/index.ts:1:1
      at node_modules/@aurelia/kernel/dist/umd/index.js:3:17
````

fkleuver:
Is there a specific reason why `reflect-metadata` must be used? Have you tried removing it?

xenoterracide:
we removed it from our direct dependencies, but it's pulled in as a transient dependency

fkleuver:
>  but i'd suspect they use the reflect-metadata APIs

That's fine, it's just a polyfill. `@aurelia/metadata` provides the same polyfill and is therefore able to replace `reflect-metadata`. The reason for this error is because our polyfill addresses two minor issues that surfaced in `reflect-metadata` and we need those fixes in order for things to work.
I have a workaround for this in the making where our polyfill can be manually invoked to force-overwrites any existing one, I'll try to get it in this week.

xenoterracide:
any chance you could enumerate those minor issues, just so I understand them in case it causes any problems with typeorm? also not entirely certain how they conflict since they should be being used on different classes? but maybe not...?

fkleuver:
> any chance you could enumerate those minor issues, just so I understand them in case it causes any problems with typeorm?

`tslib`'s static property polyfill is slightly more permissive, so some things that work with just `tslib` will throw when `reflect-metadata` is installed. I only realized this several iterations later when I forgot the specifics, but I'll look into it again and make a note about it in our docs before we go RC.

TypeORM has been around for a long time, I don't imagine you'll run into any `reflect-metadata` related issues that they haven't either already worked around or have existing issues for that can easily be found with a search.

[...]

Issue #833, opened 2020-04-02
Metadata.getOwn does not behave in the same way as Reflect.getOwnMetadata, when passing 'design:paramtypes'


xenoterracide:
I've written this little "test"
```
const reflectKeys = Reflect.getMetadataKeys(OrderService).map( (k) => Reflect.getOwnMetadata(k, OrderService));
  let getOwnKeys = Metadata.getOwnKeys(OrderService).map((k) => Metadata.getOwn(k, OrderService));
  const dependencies = DI.getDependencies(OrderService);

  root.get<Logger>(LoggerDefault).info( 'Reflect %O Metadata %O Dependencies %O', reflectKeys, getOwnKeys, dependencies);
```
```
info:     Reflect [
  [
    [Function: AmqpSender] { register: [Function: register] },
    [Function: Object] --- must be 'LoggerDefault`
  ]
] Metadata [
  [ <1 empty item>, 'LoggerDefault' ],
  [ 'au:annotation:di:paramtypes', 'au:annotation:di:dependencies' ]
] Dependencies [ <1 empty item>, 'LoggerDefault' ]
```

So here's `AmqpSender`
```ts
@singleton()
export class AmqpSender implements Sender<any, void> {
  constructor(
    @inject(LoggerDefault) private readonly log: Logger,
    @inject(AmqpChannel) private readonly channel: Channel,
  ) {}

  async send(exchange: Exchange, message: AggregateEvent<any, any>): Promise<void> {
    this.log.silly('publishing to %s %j', exchange, message);
    if (!this.channel.publish(exchange, message.type, Buffer.from(JSON.stringify(message)))) {
      this.log.error('failed to publish %s %j', exchange, message);
    }
  }
}
```

and more importantly, what's needed from `OrderService`

```ts
@singleton()
export class OrderService {
  constructor(
    private readonly sender: AmqpSender,
    @inject(LoggerDefault) private readonly log: Logger,
  ) {}
```

this is what's in `tsconfig`

```
{
  "compilerOptions": {
    "outDir": "build",
    "target": "es2018",
    "module": "commonjs",
    "lib": [
      "es2018"
    ],
    "moduleResolution": "node",
    "declaration": true,
    "strict": true,
    "esModuleInterop": true,
    "emitDecoratorMetadata": true,
    "experimentalDecorators": true,
    "sourceMap": true,
    "resolveJsonModule": true
  },
  "include": [
    "src/**/*"
  ],
  "exclude": [
    "node_modules"
[...]

PR #907, opened 2020-07-04
Metadata fixes

fkleuver:
# Pull Request

## 📖 Description

This PR exports the `applyMetadataPolyfill` as a function with options to override the defaults on how the polyfill is applied, namely:
- `reflect` - the `Reflect` object to apply the polyfill to
- `throwIfConflict` - whether to throw if any of the polyfilled properties already exist (default: `true`)
- `forceOverwrite` - whether to apply the polyfill if any of the polyfilled properties already exist (default: `false`)
- `writable` - the `writable` value to pass into `Reflect.defineProperty` when applying the polyfill (default: `true`)
- `configurable` - the `configurable` value to pass into `Reflect.defineProperty` when applying the polyfill (default: `true`)

There is a specific check for whether the polyfill has already been applied by `@aurelia/metadata` before, and if so, `applyMetadataPolyfill` becomes a no-op *unless* it was applied by a different *version* of `@aurelia/metadata`, in which case it will always throw.
The error is more specific in that case: `"Conflicting @aurelia/metadata module import detected. Please make sure you have the same version of all Aurelia packages in your dependency tree."`

The function is still called in module scope by `di` with the above defaults, but you can get "ahead" of it by importing and calling it first thing in your entry file, like so:
```ts
import { applyMetadataPolyfill } from '@aurelia/metadata';
applyMetadataPolyfill(Reflect, false, true, false, false);
```

This will apply the polyfill and make the properties non-configurable, effectively "blocking" `reflect-metadata` and ensuring everything will work in harmony.

### 🎫 Issues

Something I specifically ran into: when debugging Aurelia in node with vs code, `reflect-metadata` is loaded (not sure why), causing code to break exclusively in debugger mode while working perfectly fine in normal cli mode (you can imagine that was a lengthy troubleshooting session)
The fix: apply the above function at the top of `packages/__tests__/setup-node.ts`.

Another example: https://github.com/aurelia/aurelia/issues/902
Also related: https://github.com/aurelia/aurelia/issues/833

## 👩‍💻 Reviewer Notes

Changes are simple enough to speak for themselves I think

## 📑 Test Plan

Other than existing tests passing, we'll need some specific project setups to verify this stuff. I've already verified it locally, we just need e2e tests running on CI to prevent future regressions. Logged it in this issue: https://github.com/aurelia/aurelia/issues/906

## ⏭ Next Steps

Add e2e tests to prevent regression as per https://github.com/aurelia/aurelia/issues/906

### DIFF
--- a/packages/metadata/src/index.ts
+++ b/packages/metadata/src/index.ts
@@ -1022,7 +1022,7 @@ function $applyMetadataPolyfill(
   writable: boolean,
   configurable: boolean,
 ): void {
-  def(Metadata, '$Internal', metadataInternalSlot, writable, configurable);
+  def(reflect, internalSlotName, metadataInternalSlot, writable, configurable);
 
   def(reflect, 'metadata', metadata, writable, configurable);
   def(reflect, 'decorate', decorate, writable, configurable);
--- a/packages/metadata/src/index.ts
+++ b/packages/metadata/src/index.ts
@@ -995,13 +995,21 @@ export const Metadata = {
   delete: $delete,
 };
 
-function def(obj: object, key: string, value: unknown): void {
-  Reflect.defineProperty(obj, key, {
-    writable: true,
+function def(
+  obj: object,
+  key: string,
+  value: unknown,
+  writable: boolean,
+  configurable: boolean,
+): void {
+  if (!Reflect.defineProperty(obj, key, {
+    writable,
     enumerable: false,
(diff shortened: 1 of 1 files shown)


johnmanko:
Should calling `applyMetadataPolyfill` only be called with running with a test build using webpack dev server?  Should this be skipped on a "production" build?

Issue #898, opened 2020-06-12
Global option to On/Off Shadow DOM for whole application


HamedFathi:
#  🙋   Feature Request

I want to be able to globally enable/disable Shadow DOM for all elements. As I talked with @fkleuver, I made this issue as a reminder.

Sayan751:
The PR for `au-slot` is merged. That mimics the slot behavior without shadow DOM. The usage of native `<slot>` brings shadow DOM as expected. Primary examples of `au-slot`s can be found in the PR, but the docs PR will follow. 

Closing this in favor of `au-slot`. Feel free to open and/or continue discussion.

Issue #654, opened 2019-10-03
Make Shadow DOM optional for components with slots


rluba:
# 🙋 Feature Request

I’ve tried to convert one of our internal apps to vNext in the past few days. One thing that causes problems for us is that components with `<slot>`s now use the Shadow DOM, which…

1. prevents our style sheet from affecting the component.
2. forces us to include _another_ polyfill (for IE11 and MS Edge), even though we gain no benefit from the Shadow DOM.

## 🤔 Expected Behavior

I’d be really great if the Shadow DOM was Opt-In. Especially because it has other side-effects like #653. Slots worked great in Aurelia v1 without the Shadow DOM (although I expect this was a lot of work…)

## 🔦 Context

The `<slot>` of Aurelia v1 was a perfect equivalent of AngularJS’s `<transclude>`. They filled an important role without causing all the trouble that the Shadow DOM does  – in a homogenous environment.

I do understand that the Shadow DOM might be useful for heterogenous environments (i.e. an Aurelia component inside a CMS). But for "Aurelia owns the whole page" scenarios, its benefits are dwarfed by its side-effects.

3cp:
This is a very good argument to bring back non-shadowdom slot implementation from vcurrent.

fkleuver:
> prevents our style sheet from affecting the component.

With shadydom polyfill you can force the polyfill to override native behavior, so you'll keep your styles.

> forces us to include another polyfill (for IE11 and MS Edge), even though we gain no benefit from the Shadow DOM.

v1 slot emulation is quite a lot of code, so v1 is as much a polyfill in that sense as any other polyfill. Shadydom is just a separate install so it doesn't "feel" like "installing aurelia"

> This is a very good argument to bring back non-shadowdom slot implementation from vcurrent.

Let me give you some arguments against it:
- The costs of v1 slot emulation are immense for maintainability and it wasn't even spec compliant.
- Other frameworks who provide slot emulation have numerous longstanding open issues because of this as well. Many of which never get addressed.
[...]

Issue #399, opened 2019-01-20
Content Projection with Multiple Slots


shahabganji:
# 🙋 Feature Request

There are times you need to let the user of your component to inject its customized contents to the templates of your component, this is when content projection comes into play; that being said, I think some times we want that content to be rendered in multiple places but one of them will be shown, or even all, the surrounding HTML tags are different though, with vCurrent it is kinda impossible to have multiple slots without **name attribute**, multiple default slots :see_no_evil: 

## 🤔 Expected Behavior

consider I have `some-component`:

```html
<template>
  <button if.bind="someConditions">
     <slot></slot>
</button>
<a if.bind="someOtherConditions">
     <slot></slot>
</a>
</template>
```

and when I use this component, I expected: 

```html
<some-component some-conditions.bind="true"></some-component>
<some-component some-other-conditions.bind="true"></some-component>
<some-component some-other-conditions.bind="true" some-conditions.bind="true"></some-component>
```

I expect the first component to render a **`button`**, the second one to render an **anchor** tag, and the last one to render **both a button and an anchor** .

## 😯 Current Behavior

* The aurelia templating engine throws an exception when observing multiple `slot` tags without name.

## 💁 Possible Solution

No Idea! :see_no_evil:  replace all the slots if the conditions met? :eyes: 

## 🔦 Context

A real world scenario would be bootstrap's **`list-group`** to implement all of its variants via a single component, consider [this](https://getbootstrap.com/docs/4.2/components/list-group/#basic-example) variant and [this](https://getbootstrap.com/docs/4.2/components/list-group/#links-and-buttons) one for instance.

## 💻 Examples

a `list-group` component

```html
<template>
    <ul class="list-group" show.bind="someCondition">
        <slot></slot>
    </ul>
    <div class="list-group" show.bind="!someCondition">
        <slot></slot>
    </div>
</template>
```

the `list-group-item` component

```html
<template>
    <li if.bind="someCondition">
        <slot></slot>
    </li>
    <a if.bind="!someCondition">
        <slot></slot>
    </a>
</template>
```

and the usage: 

```html
<list-group someCondition.bind="true">
    <list-group-item someCondition.bind="true">First</list-group-item>
    <list-group-item someCondition.bind="true">Second</list-group-item>
</list-group>
```

fkleuver:
What you are looking for should be possible with `replaceable part` / `replace-part` although they need to be named:

```html
<template>
    <li if.bind="someCondition" replaceable part="list-item1">
[...]

PR #916, opened 2020-08-03
feat: au-slot #900

### LINKED ISSUE #900: [RFC]: au-slot, a replacement for the replaceable-parts
# 💬 RFC

This RFC proposes an alternative and enhanced syntax for replaceable parts. It necessitated from the need of emulating the native slots to allow CSS penetration. See #654, and #898. However, emulating slots based on global, and custom-element-level configuration brings its own set of complications in terms of templating, binding, and managing scopes. Apart from that, the idea of emulating slots, feels going towards opposite direction of Aurelia's spirit of keeping things closer to the native behavior.

With that spirit, this RFC proposes not to emulate slot at all (hear me out 😃) and rather provide an alternative syntax to support the actual use cases. From a practical standpoint, deciding whether a custom element is going to use slot/shadow DOM or not is a non-trivial decision, as the custom element needs to be styled accordingly. That is almost always a non-trivial task in a real-life application. Therefore, the assumption that a flag can be set to emulate or provide native behavior is bit impractical.

## 🔦 Context

In Aurelia2 new syntax of `au-slot` is being proposed instead. This provides slot like behavior but with CSS penetration. From this perspective it is similar to replaceable-part syntax. In fact `au-slot` is proposed to be an enhancement over replaceable-part with the addition of controlling the binding scope. Moreover, it explicitly points out that devs are not using the native slot, and therefore there is no question of emulating the same.

## 💻 Examples

**Example#1:** Simplest case with any number of content from the custom element as well as an optional default and any number of named 'slot's.

`my-element.html`
```html
<content-from-my-element></content-from-my-element> <!-- <- there are 0 or more of this -->
<au-slot>
    <!-- [default slot with optional fallback content] -->
</au-slot>
<au-slot name="s1">
    <!-- [named slot with optional fallback content] -->
</au-slot>
```

`my-app.html`
```html
<my-element>
    this goes to default slot
    <div au-slot="s1">
        this goes to slot s1
    </div>
    <span> this goes to default slot as well </span>
    <div au-slot="s1">
        this goes to slot s1 as well
    </div>
</my-element>
```

This produces loosely the following output.

```html
<my-element>
    <content-from-my-element></content-from-my-element>
    <au-slot>
        this goes to default slot
        <span> this goes to default slot as well </span>
    </au-slot>
    <au-slot name="s1">
        <div>
            this goes to slot s1
        </div>
        <div>
            this goes to slot s1 as well
        </div>
    </au-slot>
</my-element>
```

**Example#2:** Controlling binding scopes with `$host`.

`my-element.ts`
```typescript
export class MyElement {
    public message: string = "John Doe";
}
```
`my-element.html`
```html
<au-slot name="s1">
    Hello ${message}
</au-slot>
<au-slot name="s2">
    Hello ${message}
</au-slot>
```

`my-app.ts`
```typescript
export class MyApp {
    public message: string = "Max Mustermann";
}
```

`my-app.html`
```html
<my-element>
    <div au-slot="s1">
        Bye ${$host.message} <!-- grabs the inner scope with `$host` -->
    </div>
    <div au-slot="s2">
        Hi ${message}
    </div>
</my-element>
<my-element>
    <div au-slot="s1">
        Bye ${message}
    </div>
</my-element>
```

This produces loosely the following output.

```html
<my-element>
    <au-slot name="s1">
        <div>
            Bye John Doe        <!-- Uses the scope of my-element instance, and does not traverse to parent scope as message is available there -->
        </div>
    </au-slot>
    <au-slot name="s2">
        <div>
            Hi Max Mustermann   <!-- Uses the scope of my-app instance -->
        </div>
    </au-slot>
</my-element>
<my-element>
    <au-slot name="s1">
        <div>
            Bye John Doe    
        </div>
    </au-slot>
    <au-slot name="s2">
        <div>
            Hello John Doe     <!-- Parent does not project to s2; fallback uses own scope-->
        </div>
    </au-slot>
</my-element>
```

**Scoping rules in a nutshell:**
- IF parent projects to a slot
  - AND `$host` is used in the expression. the traversal starts with the inner scope.
  - ELSE (`$host` is not used) the parent scope is used for binding.
- ELSE (the parent does not project to a particular slot) the fallback content uses the inner scope for binding.

[Check this example out](https://github.com/aurelia/aurelia/issues/900#issuecomment-646727706).

**Example#3:** Repeater

```html
<!-- au-grid with @bindable items -->
<table>
    <thead>
        <tr>
            <au-slot name="header">
            </au-slot>
            <th>
                Actions
            </th>
        </tr>
    </thead>
    <tbody>
        <tr repeat.for="item of items">
            <au-slot name="content">
            </au-slot>
            <td>
                <button>edit</button>
                <button>delete</button>
            </td>
        </tr>
    </tbody>
</table>

<!-- Usage -->
<au-grid items.bind="items">
    <th au-slot="header">First Name</th>
    <td au-slot="content">${$host.item.FirstName}</td> <!-- $host lets us grab the item from repeater scope -->

    <th au-slot="header">Last Name</th>
    <td au-slot="content">${$host.item.LastName}</td>

    <th au-slot="header">Foo Bar</th>
    <td au-slot="content">
      <template if.bind="someConditionFromUsageVm">
          ${item.foo}
      </template>
      <template else>
        ${item.bar}
      </template>
    </td>
</au-grid>
```

## 📃 Summary

1. The syntax is closer to the slot, as the `replaceable` does not support projecting multiple roots to a single slot. In case of the example#1, `replaceable` would have projected only `<div>this goes to slot s1 as well</div>`.
2. With the new `au-slot` in place, the `replaceable` template controller and the replace parts can be removed altogether.
3. Provides an alternative to slot, where devs can have CSS penetration as well as use the parent (outer) scope for binding. This is unlike the `replaceable` where the inner scope is used for binding, and the parent scope is only traversed if the property is not found in inner scope.

Let me know your feedback on this.

#### Edits
- [2020-06-20]: Updated the scoping examples with `$host` and added the repeater example.

Sayan751:
# Pull Request

## 📖 Description

This PR replaces the `replaceable` `part`  infra with the new `au-slot` custom element. The `au-slot` syntax of content projection is very similar to the native `slot` syntax. The primary difference is that it is a normal Aurelia 2 custom element without any shadowDOM. This means that CSS styles from outer element can penetrate the inner component boundary. Moreover, the use of template controllers like `if`,`else`, `repeat.for` works as well in combination with `au-slot`. For more context, refer #900.

### Examples
For the complete list of examples please refer: https://github.com/aurelia/aurelia/pull/916/files#diff-5c765f5d568c9f5d673e41674ae059e4.

Note that in the examples below, the local templates are used to provide the definition of the custom elements. However, `au-slot` certainly is not limited/tied to local templates in anyway, and works with normal custom elements as well. Local templates are used here to provide concise examples.

#### Example#1: vanilla content projection

```html
<template as-custom-element="foo-bar">
  <au-slot>fallback for default slot</au-slot> <!-- an au-slot w/o a name is always named 'default' -->
  <au-slot name="s1">fallback for s1 slot</au-slot>
</template>

<foo-bar>
  <div au-slot>projection to default slot</div> <!-- for default slot, the name can be omitted -->
  <div au-slot="s1">projection to s1 slot</div>
  <div au-slot>This is projected to the default slot as well</div>
</foo-bar>
<foo-bar>
  <div au-slot="default">projection to default slot</div> <!-- using the 'default' slot name explicitly works as well -->
  <!-- no projection to s1 slot is provided, thus the fallback is used in this case -->
</foo-bar>
```

#### Example#2: Host scope

```html
<template as-custom-element="foo-bar">
  <bindable property="name"><bindable>
  <au-slot></au-slot>
</template>

<!-- this VM also has a `name` property with value 'foo' -->
<foo-bar name="bar"><span au-slot>Hello ${name}</span></foo-bar> <!-- 'Hello foo' -->
<foo-bar name="bar"><span au-slot>Hello ${$host.name}</span></foo-bar> <!-- 'Hello bar'; access the inner scope -->
```

#### Example#3: Dynamic slots using template controllers

```html
<template as-custom-element="foo-bar">
  <bindable property="showS1"><bindable>
  <bindable property="people"><bindable>

  <au-slot name="s1" if.bind="showS1">s1fb</au-slot>
  <au-slot name="s2" else>s2fb</au-slot>
  <au-slot name="list">
    <au-slot name="list-row" repeat.for="person of people">
      ${person.firstName} ${person.lastName}
    </au-slot>
  </au-slot>
</template>

<foo-bar show-s1.bind="false" people.bind="people">
  <span au-slot="s1"> This projection will not be shown <span>
  <template au-slot="list-row">${person.lastName}, ${person.firstName}</template>
  <!-- The list slot can also be replaced if needed. -->
</foo-bar>
```

More examples will follow in the doc PR (coming soon). 

### 🎫 Issues

This PR closes #900, #654, #898, #399.

## 👩‍💻 Reviewer Notes

**Implementation outline:**
- The template binder provides the information regarding the projections provided by a CE in the element symbol instance.
- The template compiler then compiles the projections and creates a mapping between a CE instruction and the projections targeted for it.
- The controller registers this map to the projection provider with the its `scope`.
- Later when the inner CE instruction (where projections are targeted) is hydrated, the targeted projections can be retrieved. The template compiler takes care of deciding the type of content: projection or fallback.
- A new keyword named `$host` is introduced to access the inner scope. Accordingly, the expression parser, the expression implementations are changed to respect the host scope. The host scope is always provided by the `au-slot`. So in a way, the `$host` keyword is tied to `au-slot`. This discourages usage of the `$host` keyword outside the context of `au-slot`.
- The `parentOverrideContext` is removed in favor of `parentScope`. That is used for scope, and context traversal.

**Current limitations:**
- It is not yet possible to mix the scopes at will. Following is an example. The reason the following example does not work is the following. The parent scope path is as follows: `au-slot#3` -> `coll-vwr` -> repeater ->`my-element` -> `app`. The `h` is registered in the overriding context of `au-slot#2` and that does not lie on the path. If the parent scope path can be modified like this `... repeater -> au-slot#2 -> ...`, such scope mixing could have worked (naively/hypothetically speaking). But that is more involved. As this might be an edge case,  it is ignored for now. 
```html
<!--my-element-->
<au-slot name="grid">
  <au-slot name="header">
    <h4>First Name </h4>
    <h4>Last Name</h4>
  </au-slot>
  <span></span>
  <template repeat.for="person of people">
    <au-slot name="content">
      <div>${person.firstName}</div>
      <div>${person.lastName}</div>
    </au-slot>
    <span></span>
  </template>
</au-slot>

<!---------------------------------->

<!-- usage in my-app -->
<!-- outer scope -->
<my-element people.bind="people">
  <template au-slot="header"> <!-- au-slot#1 -->
    <h4>First Name </h4>
    <h4>Last Name</h4>
    <h4>Pets</h4>
  </template>
  <template au-slot="content"> <!-- au-slot#2 -->
    <let h.bind="$host"></let> <!-- we are let binding here -->
    <div>Person#${$host.$index} ${h.person.firstName}</div>
    <div>${h.person.lastName}</div>
    <coll-vwr collection.bind="h.person.pets">
      <ul au-slot="colleslawt"> <!-- au-slot#3 -->
        <li repeat.for="item of $host.collection">${item} - ${h.person.firstName}</li> <!-- ${h.person.firstName} does not work as expected -->
      </ul>
    </coll-vwr>
  </template>
</my-element>
```
- It not yet possible to combine template controller with projections (that is with `[au-slot]`; template controllers works with `<au-slot>`). For example the following attempt will result a binder error.
```html
<foo-bar><span au-slot if.bind="someCondition">does not work!</span></foo-bar>
```
However, there is an easy workaround for this.
```html
<foo-bar><template au-slot><span if.bind="someCondition">does not work!</span></template></foo-bar>
```
## 📑 Test Plan

- A bunch of data driven integration tests has been added for au-slot.
- Apart from that, tests also has been added for the following internal components
  - Expression parser
  - Expression evaluation (AST)
  - BindingContext resolution
  - TemplateBinder
  - TemplateCompiler
- Tests involving integration with other packages, such as i18n, and validation has also been added. 
- Tests also involve `let` binding. But as already discussed before, there are some limitation in this area.

## ⏭ Next Steps

Docs, docs docs.


AureliaEffect:
JS Framework Benchmark: https://51529-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://51559-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://51623-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://51671-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://51694-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://51721-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://51761-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://51773-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://51850-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://51865-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://51880-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://51902-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://51909-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://51926-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://51968-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://52002-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://52013-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://52032-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://52036-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://52044-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://52073-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

bigopon:
Great work!!! My initial feedback will be around the PR description. Will continue with the code changes soon.

For the limitations, the part
```
<li repeat.for="item of $host.collection">${item} - ${h.person.firstName}</li> <!-- ${h.person.firstName} does not work as expected -->
```
is interesting. Maybe the word "mix" can be expanded a bit so that it's clearer what to fix next? Can you help adding a few line summarizing why it's not working yet?

For "Test plan" part, there's no mention of `<let/>`, maybe add it? There seems to be tests for `<let/>` already though.

Sayan751:
Thanks @bigopon. Added few more details.

packages/__tests__/2-runtime/au-slot.spec.ts
```diff
@@ -0,0 +1,1015 @@
...
+      `<my-element><template au-slot="default">d</template><template au-slot="s1">p1</template></my-element>`,
+      [
+        createMyElement(`static <au-slot>default</au-slot> <au-slot name="s1">s1</au-slot> <au-slot name="s2">s2</au-slot>`),
+      ],
+      { 'my-element': 'static d p1 s2' },
+    );
+
+    yield new TestData(
+      'supports n-1 projections',
+      `<my-element> <div au-slot="s2">p20</div> <div au-slot="s1">p11</div> <div au-slot="s2">p21</div> <div au-slot="s1">p12</div> </my-element>`,
```
bigopon, line 118: nice test 👍

packages/__tests__/2-runtime/au-slot.spec.ts
```diff
@@ -0,0 +1,1015 @@
...
+      [
+        createMyElement(`static <au-slot>default</au-slot> <au-slot name="s1">s1</au-slot> <au-slot name="s2">s2</au-slot>`),
+      ],
+      { 'my-element': `static default <div>p11</div><div>p12</div> <div>p20</div><div>p21</div>` },
+    );
+
+    yield new TestData(
+      'au-slot name with space works',
+      `<my-element><div au-slot="slot one">p</div></my-element>`,
```
bigopon, line 127: this reads almost like namespace to me instead of name with space. maybe add a test for name with `.` in name too lol

packages/__tests__/2-runtime/au-slot.spec.ts
```diff
@@ -0,0 +1,1015 @@
...
+      [
+        createMyElement(`<au-slot></au-slot><au-slot name="s1">s1fb</au-slot>`),
+      ],
+      { 'my-element': '<div>p</div>s1fb' },
+    );
+
+    // tag: mis-projection
+    yield new TestData(
+      'projection w/o [au-slot] causes mis-projection',
+      `<my-element><div>p</div></my-element>`,
```
bigopon, line 146: This is a difference with v1 in behavior. In v1, all content of the host is cleared out, compiled into a factory instead of left where they are. We probably don't have to handle compatibility around here now, but eventually need to either doc it or handle it like v1.
cc @fkleuver @EisenbergEffect

packages/__tests__/2-runtime/au-slot.spec.ts
```diff
@@ -0,0 +1,1015 @@
...
+    {
+      @customElement({ name: 'my-element', isStrictBinding: true, template: `static <au-slot>default</au-slot> <au-slot name="s1" if.bind="showS1">s1</au-slot> <au-slot else name="s2">s2</au-slot>` })
+      class MyElement {
+        @bindable public showS1: boolean = true;
+      }
+      yield new TestData(
+        'works with template controller - if-else',
+        `
+        <my-element show-s1.bind="false"> <div au-slot="s2">p21</div> <div au-slot="s1">p11</div> </my-element>
+        <my-element show-s1.bind="true" > <div au-slot="s2">p22</div> <div au-slot="s1">p12</div> </my-element>
```
bigopon, line 223: Very nice, the behavior is really close to native slot

packages/__tests__/2-runtime/au-slot.spec.ts
```diff
@@ -0,0 +1,1015 @@
...
+        ],
+        { 'my-element': `static default <div>p21</div>`, 'my-element+my-element': `static default <div>p12</div>` },
+        async function ({ host, scheduler }) {
+          const el1 = host.querySelector('my-element');
+          const el2 = host.querySelector('my-element+my-element');
+          const vm1: MyElement = CustomElement.for<Element, MyElement>(el1).viewModel;
+          const vm2: MyElement = CustomElement.for<Element, MyElement>(el2).viewModel;
+
+          vm1.showS1 = !vm1.showS1;
+          vm2.showS1 = !vm2.showS1;
```
bigopon, line 236: I think there should be an assertion like this, but before the values get reverted as well.
```suggestion
          assert.html.innerEqual(el1, `static default <div>p21</div>`, 'my-element.innerHTML');
          assert.html.innerEqual(el2, `static default <div>p12</div>`, 'my-element+my-element.innerHTML');

          vm1.showS1 = !vm1.showS1;
          vm2.showS1 = !vm2.showS1;
```

packages/__tests__/2-runtime/au-slot.spec.ts
```diff
@@ -0,0 +1,1015 @@
...
+    {
+      @customElement({
+        name: 'my-element', isStrictBinding: true, template: `
+      <au-slot name="grid">
+        <au-slot name="header">
+          <h4>First Name</h4>
+          <h4>Last Name</h4>
+        </au-slot>
+        <template repeat.for="person of people">
+          <au-slot name="content">
```
bigopon, line 273: @EisenbergEffect this test should make it easy to assert if the new `<au-slot/>`/`[au-slot]` behaviors/capabilities are on par with the `[replaceable]`/`[part]`. Can you chime in?

packages/runtime/src/renderer.ts
```diff
@@ -280,15 +273,21 @@ export class CustomElementRenderer implements IInstructionRenderer {
     controller: IRenderableController,
     target: INode,
     instruction: IHydrateElementInstruction,
-    parts: PartialCustomElementDefinitionParts | undefined,
   ): void {
-    parts = mergeParts(parts, instruction.parts);
+
+    let viewFactory: IViewFactory | undefined;
```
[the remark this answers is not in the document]
Sayan751: Not sure if a viewFactory can be conceptualized like that. I think at most though the name can be changed to `auSlotViewFactory`, as otherwise this is `undefined`. @fkleuver What do you think?

packages/__tests__/2-runtime/au-slot.spec.ts
Sayan751: Once this PR is merged (i.e. we mostly agree on the `au-slot` behavior), I will do a doc PR with plenty examples, including the edge cases. Having said that I too thought about it, but wasn't sure  how to tackle it.

I believe that the reasonable thing to do in this case would be to put such elements into the default slot. If the CE has a default slot, it will be rendered there, otherwise will eventually be ignored. One simple rule is to check if the parent node is a CE, and if so, put the current node in the projection map. What do you think?

packages/__tests__/2-runtime/au-slot.spec.ts
Sayan751: It's already there in the line before the `async function` (currently L228) :)

packages/__tests__/2-runtime/au-slot.spec.ts
Sayan751: Just wanted emphasize on the point that there are tests showing the replacement of the repeater template as well as the inner slots in the repeater template.

packages/__tests__/2-runtime/au-slot.spec.ts
Sayan751: Done :)

AureliaEffect:
JS Framework Benchmark: https://52261-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/__tests__/2-runtime/au-slot.spec.ts
Sayan751: Nope. Tried that and it brings more problem than anticipated. I would like to leave it like this for now, and doc it.

Sayan751: Also on a more far stretched example, what if the CE has a default native slot in it along with au-slot? Removing the nodes will create a conflict there.

AureliaEffect:
JS Framework Benchmark: https://52892-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/runtime/src/templating/render-context.ts
```diff
@@ -344,7 +342,7 @@ export class RenderContext<T extends INode = INode> implements IComponentFactory
   // #endregion
 
   // #region IRenderContext api
-  public compile(): ICompiledRenderContext<T> {
+  public compile(targetedProjections: RegisteredProjections | null): ICompiledRenderContext<T> {
```
bigopon, line 345: Is `availableProjections` correct understanding?

packages/runtime/src/renderer.ts
```diff
@@ -280,15 +273,21 @@ export class CustomElementRenderer implements IInstructionRenderer {
...
+
+    const slotInfo = instruction.slotInfo;
+    if (slotInfo!==null) {
+      const projectionCtx = slotInfo.projectionContext;
+      viewFactory = getRenderContext(projectionCtx.content, context).getViewFactory(void 0, slotInfo.type, projectionCtx.scope);
```
bigopon, line 283: I just realized with this, we should be able to enable multiple slot with the same name for `<au-slot/>`, as per highlighted in this comment https://github.com/aurelia/aurelia/issues/399#issuecomment-456294198

Instead of:
```html
<template>
  <div replaceable part="copy">
  </div>

  <div replaceable part="copy">
  </div>

  <div replaceable part="copy">
  </div>
</template>
```

We can have
```html
<template>
  <au-slot name="copy"></au-slot>

  <au-slot name="copy"></au-slot>

  <au-slot name="copy"></au-slot>
</template>
```
And all projection for `copy` will be multiplied by 3

packages/runtime/src/templating/render-context.ts
Sayan751: That's right. But the word 'available' is sort of ambiguous in this context.

packages/runtime/src/renderer.ts
Sayan751: That's correct.

AureliaEffect:
JS Framework Benchmark: https://54036-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://54091-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://54123-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://54130-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://54217-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://54256-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://54274-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

Issue #917, opened 2020-08-07
Bug: realworld example not working


mroeling:
# 🐛 Bug Report

Realworld example is not working atm.

## 🤔 Expected Behavior

It should at least build properly and show the application in its current state.

## 😯 Current Behavior

The build fails. After build fix error in main.ts, 

```
ERROR in /.../au2-realworldapp/test/realworld/src/app.ts
./src/app.ts
[tsl] ERROR in /.../au2-realworldapp/test/realworld/src/app.ts(48,7)
      TS2769: No overload matches this call.
  The last overload gave the following error.
    Argument of type '() => true | never[]' is not assignable to parameter of type 'HookFunction'.
      Type '() => true | never[]' is not assignable to type 'BeforeNavigationHookFunction'.
        Type 'true | never[]' is not assignable to type 'Promise<boolean | ViewportInstruction[]>'.
          Type 'true' is not assignable to type 'Promise<boolean | ViewportInstruction[]>'.
```

## 💁 Possible Solution

Working on fix, creating this issue to refer in PR. 

## 🔦 Context

Not applicable

## 💻 Code Sample

Not applicable

## 🌍 Your Environment

Starting application from within the folder, so all according to settings within the folder. Tried both Yarn and Npm, but no differences.

| Software         | Version(s) |
| ---------------- | ---------- |
| Aurelia          | master (d425f66) |
| Operating System | WIN10/WSL2/Ubuntu 20.04 |

fkleuver:
Thanks for reporting this. The reason for this failure is due to a breaking change in TS 3.8 or 3.9 (don't recall which one) and this manifestation of that breaking changed slipped our attention due to the realworld app not being included in our CI/CD process. So, the TODO here is to fix the typings and then to include it in CI/CD. Will get to that before alpha.

fkleuver:
This is fixed

PR #918, opened 2020-08-08
fix(test/realworld): improve routerhook and route callbacks

mroeling:
# Pull Request

## 📖 Description

The intention was to be able to start the realworld app. A couple of errors prevented this,
at first a type error in the authorization router hook, and after that the runtime error
to check if the user is logged in. The callbacks are defined in the app class,
but the scope of the callback is in the navigation classes. The change fixes the scope for the callback.

### 🎫 Issues

Fix for #917

## 👩‍💻 Reviewer Notes

-

## 📑 Test Plan

No changes in tests needed

## ⏭ Next Steps

-


test/realworld/src/app.ts
```diff
@@ -86,30 +86,30 @@ export class App implements IViewModel {
         title: 'Home',
       },
       {
-        condition: this.authenticated,
```
fkleuver, line 89: @jwx Seems odd that this ever worked in the first place. Do you recall if this worked? *should* it work?

test/realworld/src/app.ts
```diff
@@ -47,10 +47,10 @@ export class App implements IViewModel {
     this.router.addHook(
       () => {
         if (this.state.isAuthenticated) {
-          return true;
+          return Promise.resolve(true);
```
fkleuver, line 50: @jwx This shouldn't be necessary right?

mroeling: FYI, this was actually because of the build error I had. The other changes were because of runtime errors.

jwx: Probably because the method isn't `async`. (@fkleuver Not once our latest changes are in, no.)

test/realworld/src/app.ts
jwx: Yeah, it does seem a bit odd. Can't recall if it did, but am inclined to think that it shouldn't since
```typescript
  public authenticated(): boolean { ... }
```
is a method. Now, if it'd been
```typescript
  public authenticated = (): boolean => { ... }
```
it should work.

jwx:
Both the router hooks and the `au-nav` have upcoming changes that will affect `realworld`, but if `realworld` is currently broken and this PR fixes that then I'm all for it.

fkleuver:
Code has changed a fair bit since this PR was opened, realworld is now also up to date

PR #920, opened 2020-08-10
docs(components): beforeBind/afterUnbind pair

zewa666:
explains where an EventAggregator or RxJS subscription should be created and disposed

# Pull Request


AureliaEffect:
JS Framework Benchmark: https://56007-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

fkleuver:
Already updated I believe

PR #921, opened 2020-08-13
Karma tests filter

zewa666:
# Pull Request

## 📖 Description

This allows to startup inidividual package tests only compared to running all of them with `npm run test-chrome`.
The reason is that the currently preferred way is to have to do manual modifications with the `setup-browser.ts` file which has a few drawbacks:

* Pollutes the change tracking of Git
* Might be unintentionally commited
* Lots of keystrokes

This proposed change does not remove the previous manual approach but merely adds an alternative approach for people preferring not touching files ;)

## 📑 Test Plan

Running `npm run test-browser` should still execute all tests, whereas e.g
running `npm run test-browser -- --pkg=store` will only execute the tests for the store package.

### DIFF
--- a/packages/__tests__/karma.conf.js
+++ b/packages/__tests__/karma.conf.js
@@ -1,4 +1,5 @@
 const path = require('path');
+const { DefinePlugin } = require('webpack');
 
 const basePath = path.resolve(__dirname);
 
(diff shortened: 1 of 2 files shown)


fkleuver:
Karma setup significantly different now, can be changed more easily to include specific folders etc

PR #923, opened 2020-08-17
docs(readme) updated form section

avrahamcool:
removing the `id`s of the input/select. for a cleaner approach. [without losing the ability to click on the label for focusing]

## 📖 Description
updating the demo of form section, so it won't have unnecessary `id` for elements.
using `id`s there can confuse newcomers into thinking that the binding require some kind of `id` in order to work.
the `id` is there purely for the focus in the label to work.
this can be done in a cleaner way - by nesting the input/select inside the label.

## 👩‍💻 Reviewer Notes
in order to keep thing cleaner, and "production like", i've also surrounded the text inside a span.


avrahamcool:
BTW: unit test are failing regardless of my change.
(iv'e made changes in the readme.md file only).

PR #931, opened 2020-08-26
feat: hide the default function

xenoterracide:
Signed-off-by: Caleb Cushing <xenoterracide@gmail.com>

`merge-anything` works, I tried updating `mergeObjects` to have a compilable signature, no success, I could try `Objects.assign` or even the spread operator. the Goal here is to hide the `DefaultContainerConfiguration` and only require users to input the changes they want from that.


packages/kernel/src/di.ts
```diff
@@ -145,22 +145,21 @@ function cloneArrayWithPossibleProps<T>(source: readonly T[]): T[] {
...
-export const DefaultContainerConfiguration: IContainerConfiguration = {
+const DefaultContainerConfiguration: Required<IContainerConfiguration> = {
   defaultResolver: DefaultResolver.singleton,
 };
 
 export const DI = {
-  createContainer(config: IContainerConfiguration = DefaultContainerConfiguration): IContainer {
-      return new Container(null, config);
+  createContainer(config: IContainerConfiguration = {}): IContainer {
+      return new Container(null, merge(DefaultContainerConfiguration, config));
```
bigopon, line 162: 2 issues:
- a policy that we have is no 3rd party dep, unless we have to.
- this is mutating the object `DefaultContainerConfiguration`, making subsequent call to this `createContainer` unpredictable.

bigopon:
@xenoterracide I think specifying only what users need can also be done in userland, what issues are you having?

packages/kernel/src/di.ts
xenoterracide: > a policy that we have is no 3rd party dep, unless we have to.
I know, but I'm having trouble getting the signature to pass with `mergeObjects`

> this is mutating the object DefaultContainerConfiguration, making subsequent call to this createContainer unpredictable.

actually, merge-anything doesn't mutate, it returns a new thing.

xenoterracide:
@bigopon just thinking the API is ugly and over exposed if there is more than 1 item in the Default. I ran into annoyances with it as was trying to get the debugging thing down.

PR #936, opened 2020-09-17
remove pull_request

xenoterracide:
@bigopon would prefer to not have github actions run twice on a pull request.

# Pull Request

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,5 +1,5 @@
 name: Github Actions
-on: [push, pull_request]
+on: [push]
 jobs:
 
   lint:


bigopon:
@xenoterracide it seems when we remove pull request, push also not run. Is this behaviour wanted, compared to the purpose of the previous PR?

xenoterracide:
it ran on my fork https://github.com/xenoterracide/aurelia/actions/runs/259728449, which was more or less the purpose, having tests run without having submitted even a draft PR yet. however this means they don't show up here on the PR. A reason you're probably getting a lot of double runs internally is because you're writing to the main repo so both happen. this is basically a github stupid in my mind. I don't know of a way to get both behaviors but only have it run once per sha.

bigopon:
@xenoterracide thanks. I guess this is good to go then, unless you want to tweak the config @fkleuver

bigopon:
A lot better in PR now, thanks @xenoterracide

Issue #303, opened 2018-11-19
[RFC] [Feature] switch (/ case) template controller


fkleuver:
TODO

Sayan751:
# RFC

This RFC proposes the syntaxes for the `switch`/`case` template controller.
At this point, there are `if`/`else` template controllers to facilitate the conditional attach/detach of elements to/from the DOM.
However, the semantics of switch-case is missing.
Let's look at some examples.

## Examples

### Example#1: enum
A typical use-case for switch is to deal with `enum`s.
The following `OrderStatus` `enum` is given.

```typescript
const enum Status {
  received = 'received',
  processing = 'processing',
  dispatched = 'dispatched',
  delivered = 'delivered',
}
```

If we are tasked with showing a status text according to the `status: Status`, then with `if` template controller we might end up with the following markup.

```html
<span if.bind="status === 'received'">Order received.</span>
<span if.bind="status === 'processing'">Processing your order.</span>
<span if.bind="status === 'dispatched'">On the way.</span>
<span if.bind="status === 'delivered'">Delivered.</span>
```

> Note that the markup is kept simplistic for the ease of discussion.
> And yes, you can use i18n to have contextualize the text, but that is out of scope of this discussion.

The markup above will produce the desired result.
However, apart from having a bit cumbersome syntax, the semantic is somewhat lost.
With the proposed `switch`/`case` template controller, the above can be rewritten as the following.

```html
<au-switch expression.bind="status">
  <span case="received">Order received.</span>
  <span case="processing">Processing your order.</span>
  <span case="dispatched">On the way.</span>
  <span case="delivered">Delivered.</span>
</au-switch>
```

The goal is to render the first element that matches the `[expression]` of `au-switch`.
The rendered output should be same as the one with `if` template controllers.

There is one important difference between the two examples.
With the `switch`/`case` example, we have one dynamic expression (`status`), and four literal expressions for for `case`s.
As the literal expression does not change, we don't need to observe/connect to those, rather connecting to the `[expression]` of the `switch` should suffice.
Moreover, as with the switch syntax the first matched element is rendered, reevaluation of all four `cases` might not always be necessary when `status` changes.
That is not the case for the markup with `if` template controllers, as in that case when `status` changes, all four bindings are reevaluated.

### Example#2: another comparison with `if`/`else`

Sometimes we end up with a if-else ladder like the following.

```html
<some-el1 if.bind="expr1"></some-el1>
<template else>
[...]

PR #942, opened 2020-09-26
feat: switch #303

Sayan751:
# Pull Request


AureliaEffect:
JS Framework Benchmark: https://53622-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://53643-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://53705-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://53769-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://53891-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/runtime/src/renderer.ts
```diff
@@ -399,9 +400,17 @@ export class TemplateControllerRenderer implements IInstructionRenderer {
 
     Metadata.define(key, childController, renderLocation);
 
-    if (instruction.link) {
-      const controllers = controller.controllers!;
-      (component as { link(componentTail: IController): void}).link(controllers[controllers.length - 1]);
+    switch (instruction.linkType) {
```
bigopon, line 403: 👍 so probably there will be a link resolver and our infra here will be perfectly valid for app to reuse. Imagine some scenario like this, which is copied from Svelte syntax:
```html
<div async.bind="promise>
  Loading....
</div>
// the following line should work like with, with some extra capability to make it behave like then
<div then.bind="resolvedData">
  ${resolvedData.id}
</div>
<div catch.bind="err">
  ${err.message}
</div>
```

Sayan751: @bigopon This is a great idea. I am thinking of changing the `LinkType` enum to have `parent` and `sibling` instead of the current ones. And put that also into CA definition, instead of computing that from the CA name. In that way, devs can actually advertise which kind of linking is needed. Ofc we support only a defined set of linking, which are for now limited to `parent` and `sibling`. We need to document this, as how devs can leverage the linking. I think that would work for now.

bigopon: `parent`/`sibling` sounds like a nice starting point. For `parent`, it's probably easy to determine, since a node only ever has 1 parent maximum, in a tree. Sibling, however, probably trickier, as sometimes we may want to "jump"

Sayan751: Nope these are not in terms of DOM, rather in terms of controller. As discussed also previously with @fkleuver that would require more work, and it is better to be postponed. IMO the link type will be a trickier feature to implement correctly that is also performant. I will for now only introduce the linktype in the CA definition, which should save us the link type computation based on CA's name.

bigopon: > Nope these are not in terms of DOM, rather in terms of controller

Yes, but for `async/then/catch`, shouldn't the catch "jump" to link with the `async` instead of the `then`? Ability to jump, though, may be that popular, and this could be the only place where it's needed.

Sayan751: I would suggest the following syntax instead 😛 

```html
<template async.bind="promise">
  <span unsettled> Running... 🏃 </span>
  <span then.bind="resolvedData"> All resolved 🎉 </span>
  <span catch.bind="error"> Oh no! 😡 </span>
</template>
```

You see that I am in switch mode :)

bigopon: image

bigopon: image

AureliaEffect:
JS Framework Benchmark: https://53904-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://53917-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://54652-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://54984-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/__tests__/2-runtime/switch.spec.ts
```diff
@@ -0,0 +1,1721 @@
...
+    );
+
+    yield new TestData(
+      'supports multi-case',
+      {
+        initialStatus: Status.processing,
+        template: `
+    <template>
+      <template switch.bind="status">
+        <span case.bind="['received', 'processing']">Processing.</span>
```
bigopon, line 673: This is interesting ... case. Treating array as cases by default is probably the expectation half of the times at least. While there's probably edge case, I think this is more preferred

packages/__tests__/2-runtime/switch.spec.ts
```diff
@@ -0,0 +1,1721 @@
...
+        <template switch.bind="status">
+          <span case="received">Order received.</span>
+          <span case="dispatched">On the way.</span>
+          <span case="processing">Processing your order.</span>
+          <span case="delivered">Delivered.</span>
+          <span>foobar</span>
+          <span if.bind="true">foo</span><span else>bar</span>
+          <span if.bind="false">foo</span><span else>bar</span>
+          <span repeat.for="i of 3">\${i}</span>
+          <my-echo message="awesome possum"></my-echo>
```
bigopon, line 1078: well now i'm interested in possum

packages/__tests__/2-runtime/switch.spec.ts
```diff
@@ -0,0 +1,1721 @@
...
+      }
+    );
+
+    yield new TestData(
+      '*[switch][repeat.for] works',
+      {
+        initialStatus: Status.delivered,
+        template: `
+      <template>
+        <div repeat.for="s of ['received', 'dispatched']" switch.bind="s">
```
bigopon, line 1305: This is interesting test, it begs:
```html
<div switch.bind="s">
  <span case="received"  repeat.for="s of ['received', 'dispatched']">${s}</span>
```

packages/__tests__/2-runtime/switch.spec.ts
```diff
@@ -0,0 +1,1721 @@
...
+    yield new TestData(
+      '*[switch]>CE>*[case] works',
+      {
+        initialStatus: Status.dispatched,
+        template: `
+      <template as-custom-element="foo-bar">
+        foo bar
+      </template>
+
+      <template switch.bind="status">
```
bigopon, line 1596: wow this is interesting 🤔 is this legal lol

packages/runtime/src/resources/custom-attributes/switch.ts
```diff
@@ -0,0 +1,329 @@
...
+  }
+}
+
+@templateController({ name: 'case', linkType: TemplateControllerLinkType.parent })
+export class Case<T extends INode = Node> implements ICustomAttributeViewModel<T> {
+  public readonly id: number = nextId('au$component');
+  public readonly $controller!: ICustomAttributeController<T, this>; // This is set by the controller after this instance is constructed
+
+  @bindable public value: any;
+  @bindable({ mode: BindingMode.oneTime }) public fallThrough: boolean = false;
```
bigopon, line 259: we have `set` converter for bindable, so this can be converted to:
```ts
  @bindable({
    set: v => {
      // well
      switch (v) {
        // well
        case 'true':  return true;
        case 'false': return false;
        default: return !!v;
      }
    },
    mode: BindingMode.oneTime
  })
  public fallThrough: boolean = false;
```
so we can do
```html
<span case="value.bind: v; fall-through: true">
<span case="value.bind: v; fall-through: false">
```
instead of
```html
<span case="value.bind: v; fall-through.bind: true">
```

This probably needs to be expanded to if/else too (for the `cache` property)

packages/runtime/src/resources/custom-attributes/switch.ts
```diff
@@ -0,0 +1,329 @@
...
+  protected linkToSwitch(auSwitch: Switch<T>) {
+    auSwitch.cases.push(this);
+  }
+}
+
+@templateController({ name: 'default-case', linkType: TemplateControllerLinkType.parent })
+export class DefaultCase<T extends INode = Node> extends Case<T>{
+
+  protected linkToSwitch($switch: Switch<T>) {
+    $switch.defaultCase = this;
```
bigopon, line 327: we probably will need to throw if there's already a default case too

bigopon (commented):
Finished the first round, amazing work 🚀

packages/__tests__/2-runtime/switch.spec.ts
bigopon: It's quite a useless test, as no one does this. pls ignore this

packages/__tests__/2-runtime/switch.spec.ts
bigopon: I meant this test, the cross level switch/case combo test show case is good 👍

packages/runtime/src/resources/custom-attributes/switch.ts
```diff
@@ -0,0 +1,329 @@
...
+
+  public valueChanged(_newValue: boolean, _oldValue: boolean, flags: LifecycleFlags): void {
+    if ((this.$controller.state & State.activated) === 0) {
+      return;
+    }
+    void (this.queue(() => this.swap(flags, this.value)));
+  }
+
+  public caseChanged($case: Case<T>, flags: LifecycleFlags): void {
+    void (this.queue(async () => this.handleCaseChange($case, flags)));
```
fkleuver, line 109: Don't need this `async` right? If the linter complains just suppress it in this case. I think we just need to disable that rule actually..

packages/runtime/src/resources/custom-attributes/switch.ts
```diff
@@ -0,0 +1,329 @@
...
+    }
+
+    // compute the new active cases
+    const newActiveCases: Case<T>[] = [];
+    let fallThrough = $case.fallThrough;
+    if (!fallThrough) {
+      newActiveCases.push($case);
+    } else {
+      const cases = this.cases;
+      const idx = cases.findIndex((c) => c === $case);
```
fkleuver, line 144:
```suggestion
      const idx = cases.indexOf($case);
```

packages/runtime/src/resources/custom-attributes/switch.ts
```diff
@@ -0,0 +1,329 @@
...
+        this.activeCases = newActiveCases;
+        if (newActiveCases.length === 0) { return; }
+        return this.activateCases(flags);
+      }
+    );
+  }
+
+  private activateCases(flags: LifecycleFlags): void | Promise<void> {
+    const controller = this.$controller;
+    if ((controller.state & State.activated) === 0) { return; }
```
fkleuver, line 191: Why not `if (controller.isActive)`? Did you find an issue with that?

packages/runtime/src/resources/custom-attributes/switch.ts
```diff
@@ -0,0 +1,329 @@
...
+  private clearActiveCases(flags: LifecycleFlags, newActiveCases: Case<T>[] = []): void | Promise<void> {
+    const cases = this.activeCases;
+    const numCases = cases.length;
+
+    if (numCases === 0) { return; }
+
+    if (numCases === 1) {
+      const firstCase = cases[0];
+      if (!newActiveCases.includes(firstCase)) {
+        cases.splice(0);
```
fkleuver, line 217:
```suggestion
        cases.length = 0;
```

packages/runtime/src/resources/custom-attributes/switch.ts
```diff
@@ -0,0 +1,329 @@
...
+
+    return onResolve(
+      resolveAll(...cases.reduce((acc: (void | Promise<void>)[], $case) => {
+        if (!newActiveCases.includes($case)) {
+          acc.push($case.deactivate(flags));
+        }
+        return acc;
+      }, [])),
+      () => {
+        cases.splice(0);
```
fkleuver, line 231:
```suggestion
        cases.length = 0;
```

packages/runtime/src/resources/custom-attributes/switch.ts
```diff
@@ -0,0 +1,329 @@
...
+      return;
+    }
+
+    return onResolve(
+      resolveAll(...cases.reduce((acc: (void | Promise<void>)[], $case) => {
+        if (!newActiveCases.includes($case)) {
+          acc.push($case.deactivate(flags));
+        }
+        return acc;
+      }, [])),
```
fkleuver, line 229: I think this would be a bit simpler without reduce

```ts
resolveAll(
  ...cases.filter(c => !newActiveCases.includes(c)).map(c => c.deactivate(flags)),
  () => {
// ...
```

packages/runtime/src/resources/custom-attributes/switch.ts
```diff
@@ -0,0 +1,329 @@
...
+      () => {
+        cases.splice(0);
+      }
+    );
+  }
+
+  private queue(action: () => void | Promise<void>): void {
+    let promise: void | Promise<void> = void 0;
+    const unset = () => {
```
fkleuver, line 238: If it's only used once, can we just inline it maybe?

packages/runtime/src/resources/custom-attributes/switch.ts
```diff
@@ -0,0 +1,329 @@
...
+    );
+  }
+}
+
+@templateController({ name: 'case', linkType: TemplateControllerLinkType.parent })
+export class Case<T extends INode = Node> implements ICustomAttributeViewModel<T> {
+  public readonly id: number = nextId('au$component');
+  public readonly $controller!: ICustomAttributeController<T, this>; // This is set by the controller after this instance is constructed
+
+  @bindable public value: any;
```
fkleuver, line 258:
```suggestion
  @bindable public value: unknown;
```

packages/runtime/src/resources/custom-attributes/switch.ts
```diff
@@ -0,0 +1,329 @@
...
+  private $switch!: Switch<T>;
+  private isObserving: boolean = false;
+
+  public constructor(
+    @IViewFactory private readonly factory: IViewFactory<T>,
+    @IObserverLocator private readonly locator: IObserverLocator,
+    @IRenderLocation location: IRenderLocation<T>,
+  ) {
+    const view = this.view = this.factory.create();
+    view.setLocation(location, MountStrategy.insertBefore);
```
fkleuver, line 271: Just a nit, but these 2 lines can be shortened a bit:
```ts
(this.view = factory.create()).setLocation(location, MountStrategy.insertBefore);
```

packages/runtime/src/definitions.ts
```diff
@@ -136,7 +142,7 @@ export interface IHydrateTemplateController extends IHydrateInstruction {
   res: string;
   instructions: ITargetedInstruction[];
   def: PartialCustomElementDefinition;
-  link?: boolean;
+  linkType: TemplateControllerLinkType;
```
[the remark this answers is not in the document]
bigopon: Nice idea 👍 , will this pose any challenge to AOT? if not, probably it's a really good one, because it feels clearer linking step

packages/runtime/src/renderer.ts
```diff
@@ -398,9 +399,17 @@ export class TemplateControllerRenderer implements IInstructionRenderer {
...
-      const children = controller.children!;
-      (component as { link(componentTail: IController): void}).link(children[children.length - 1]);
+    switch (instruction.linkType) {
+      case TemplateControllerLinkType.sibling: {
+        const children = controller.children!;
+        (component as { link(componentTail: IController): void }).link(children[children.length - 1]);
+        break;
+      }
+      case TemplateControllerLinkType.parent: {
+        const switchController = (controller as Controller).parent!;
```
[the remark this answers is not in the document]
Sayan751: Good catch. I should change the name. That's an old artifact. The actual check is done in `Case#link`.

packages/__tests__/2-runtime/switch.spec.ts
Sayan751: As a matter of fact, this behavior can be removed in favor of bit more verbose alternative.

```html
<template case="value.bind: 'received'; fall-through.bind='true'"></template>
<span case="processing"> Processing. </span>
```

I implemented this, as the syntax is more concise.

packages/__tests__/2-runtime/switch.spec.ts
Sayan751: I was unsure about "to restrict, or not to restrict". And as a wise man once told me, the responsibility lies on user, I left it as it is 😛 Let me know whether we need to enforce the restriction.

packages/runtime/src/resources/custom-attributes/switch.ts
Sayan751: Ignoring browser's internal optimization, with `.filter().map()`  2 iterations over the array is evident, which is not the case with `.reduce()`. That's why I tend to prefer `.reduce()` order `.filter().map()`.

If there not any known problem with this, I would like to keep the `.reduce()`.

packages/runtime/src/resources/custom-attributes/switch.ts
```diff
@@ -0,0 +1,329 @@
...
+      return view.deactivate(view, this.$controller, flags);
+    });
+    return this.promise;
+  }
+
+  public valueChanged(_newValue: boolean, _oldValue: boolean, flags: LifecycleFlags): void {
+    if ((this.$controller.state & State.activated) === 0) {
+      return;
+    }
+    void (this.queue(() => this.swap(flags, this.value)));
```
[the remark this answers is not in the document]
Sayan751: Good catch! Old artifact; removed.

packages/runtime/src/resources/custom-attributes/switch.ts
Sayan751: Good catch! Old artifact; removed.

packages/runtime/src/resources/custom-attributes/switch.ts
Sayan751: Done.

packages/runtime/src/resources/custom-attributes/switch.ts
Sayan751: Old artifact. Changes done.

packages/runtime/src/resources/custom-attributes/switch.ts
Sayan751: Done.

packages/runtime/src/resources/custom-attributes/switch.ts
Sayan751: Done.

packages/runtime/src/resources/custom-attributes/switch.ts
Sayan751: Done. The `let promise = ...;` needs to be there. Otherwise it is throwing error saying `promise` is accessed before it is assigned.

packages/runtime/src/resources/custom-attributes/switch.ts
Sayan751: Done.

packages/runtime/src/resources/custom-attributes/switch.ts
Sayan751: Done.

AureliaEffect:
JS Framework Benchmark: https://55189-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/runtime/src/definitions.ts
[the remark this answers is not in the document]
bigopon: on a 2nd thought, maybe this could be a static method on the class itself? that way we don't have to trade serializability of the instruction for the flexibility, if we pass everything to the link in the end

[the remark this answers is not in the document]
Sayan751: Also my thought to not include this method in definition. Talked with Fred yesterday. I think though it don't need to bbe static. With an instance method we can simply do this from renderer: `component.link?.(flags, parentContext, controller, childController, target, instruction);`

packages/runtime/src/renderer.ts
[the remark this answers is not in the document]
Sayan751: The linking is moved out of renderer.

packages/runtime/src/definitions.ts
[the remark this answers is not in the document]
Sayan751: Changes done.

AureliaEffect:
JS Framework Benchmark: https://55507-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/runtime/src/resources/custom-attributes/switch.ts
Sayan751: Done.

packages/runtime/src/resources/custom-attributes/switch.ts
Sayan751: Good catch! Done.

AureliaEffect:
JS Framework Benchmark: https://55519-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/runtime/src/resources/custom-attributes/switch.ts
```diff
@@ -0,0 +1,350 @@
...
+        }
+        return acc;
+      }, [])),
+      () => {
+        cases.length = 0;
+      }
+    );
+  }
+
+  private queue(action: () => void | Promise<void>): void {
```
bigopon, line 237: This is nice 👍

packages/runtime/src/resources/custom-attributes/switch.ts
```diff
@@ -0,0 +1,350 @@
...
+    } else {
+      throw new Error('The parent switch not found; only `*[switch] > *[case|default-case]` relation is supported.');
+    }
+  }
+
+  public isMatch(value: unknown, flags: LifecycleFlags): boolean {
+    const $value = this.value;
+    if (Array.isArray($value)) {
+      if (!this.isObserving) {
+        const observer = this.locator.getArrayObserver(flags, $value);
```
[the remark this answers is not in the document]
bigopon: Reading the `handleCaseChange` again, i see why it's called here. Seems a bit tricky to find the correct place to unsubscribe

bigopon (commented):
Couldn't sleep, managed to read the implementation again 😃

packages/runtime/src/resources/custom-attributes/switch.ts
```diff
@@ -0,0 +1,350 @@
...
+    return this.promise;
+  }
+
+  public valueChanged(_newValue: boolean, _oldValue: boolean, flags: LifecycleFlags): void {
+    if (!this.$controller.isActive) { return; }
+    this.queue(() => this.swap(flags, this.value));
+  }
+
+  public caseChanged($case: Case<T>, flags: LifecycleFlags): void {
+    this.queue(async () => this.handleCaseChange($case, flags));
```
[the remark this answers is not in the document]
Sayan751: This is mostly because of this eslint  rule: https://github.com/typescript-eslint/typescript-eslint/blob/v2.31.0/packages/eslint-plugin/docs/rules/promise-function-async.md.

packages/runtime/src/resources/custom-attributes/switch.ts
```diff
@@ -0,0 +1,350 @@
...
+      const cases = this.cases;
+      const idx = cases.indexOf($case);
+      for (let i = idx, ii = cases.length; i < ii && fallThrough; i++) {
+        const c = cases[i];
+        newActiveCases.push(c);
+        fallThrough = c.fallThrough;
+      }
+    }
+
+    await this.clearActiveCases(flags, newActiveCases);
```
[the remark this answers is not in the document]
Sayan751: Yeah true that it is bit inconsistent. But I think avoiding async/await completely makes the code needlessly verbose and complex to understand, especially when the function is wrapped here: https://github.com/aurelia/aurelia/blob/topic/303-au-switch/packages/runtime/src/resources/custom-attributes/switch.ts#L113

packages/runtime/src/resources/custom-attributes/switch.ts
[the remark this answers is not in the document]
Sayan751: Good catch! I missed the "unsubscribe" part. I think that can be done in the "afterUnbindChildren" hook.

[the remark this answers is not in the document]
Sayan751: With 2nd thought, it is pushed to the `dispose` hook.

AureliaEffect:
JS Framework Benchmark: https://55866-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/__tests__/2-runtime/switch.spec.ts
Sayan751: Just to clarify, it will end up throwing an error. Added couple of nonsense examples in the tests.

AureliaEffect:
JS Framework Benchmark: https://55896-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/runtime/src/resources/custom-attributes/switch.ts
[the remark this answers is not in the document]
bigopon: Yes, that is my view towards this as well. Though as I understood, we want to enable "single call app", which means if there's nothing async in the app, the whole app can be rendered in a function call, and immediately available in the next line, instead of having to wait for a micro task, or a promise to resolve.

packages/runtime/src/resources/custom-attributes/switch.ts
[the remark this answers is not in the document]
bigopon: There is this case:
```html
<span cases.bind="someArray">
```
and if `someArray` is changed from one array to another, we gonna need to dispose the old array when observing the new array. I think this is where it gets tricky, because there's no need to start the observation too eagerly, and there's a need to keep the code simple.

packages/runtime/src/resources/custom-attributes/switch.ts
bigopon: agree with Sayan, most of the time it's simple to read. But our mind is inconsistent, for example reading `.reduce` at 4am is probably not good 😵

packages/__tests__/2-runtime/switch.spec.ts
bigopon: I'm not even sure where to start the restriction lol. Immediate levels in DOM hierarchy doesn't sound like feasible, or does it? In anyway, I think we can leave this to your awesome doc 😬

packages/runtime/src/resources/custom-attributes/switch.ts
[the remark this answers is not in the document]
Sayan751: I agree with you. My thought on this is that should already be possible in the current state, as the private method called is wrapped from the change handler. I think the old link was rendered confusing after I pushed new changes.

https://github.com/aurelia/aurelia/blob/4deb2b25b0ab74a0bd3e132fb131cf8f9d84d7c9/packages/runtime/src/resources/custom-attributes/switch.ts#L115-L163

packages/runtime/src/resources/custom-attributes/switch.ts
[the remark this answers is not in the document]
Sayan751: Thanks @bigopon for pointing this out. Missed it completely. I will make the change.

AureliaEffect:
JS Framework Benchmark: https://56021-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/runtime/src/resources/custom-attributes/switch.ts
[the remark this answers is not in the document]
Sayan751: Done :)

AureliaEffect:
JS Framework Benchmark: https://56142-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://56167-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

docs/user-docs/getting-started/conditional-rendering.md
```diff
@@ -114,3 +114,303 @@ When unchecked, our "Hello World" div will have the `aurelia-hide` class, which
...
+
+### Miscellaneous examples
+
+This section includes some more interesting examples that you might encounter in real life, and the statutory warnings.
+
+- Another usage of switch that we often see in the wild, is to use a static expression from `switch` and more dynamic expression for `case`. Therefore, following is a valid usage of `switch`.
+
+  {% code title="my-app.html" %}
+  ```markup
+  <template>
```
bigopon, line 278: fizz buzz purely in html for fun?
```html
  <template repeat.for="num of 100">
    <div switch.bind="true">
      <span case.bind="num % 3 === 0 && num % 5 === 0">FizzBuzz</span>
      <span case.bind="num % 3 === 0">Fizz</span>
      <span case.bind="num % 5 === 0">Buzz</span>
    </template>
  </div>
```

Sayan751: Why not! :)

AureliaEffect:
JS Framework Benchmark: https://56195-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

bigopon (approved):
Awesome work

AureliaEffect:
JS Framework Benchmark: https://56206-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/runtime/src/resources/custom-attributes/switch.ts
```diff
@@ -0,0 +1,373 @@
...
+    @IRenderLocation private readonly location: IRenderLocation<T>,
+  ) { }
+
+  public beforeBind(initiator: IHydratedController<T>, parent: IHydratedParentController<T>, flags: LifecycleFlags): void | Promise<void> {
+    const view = this.view = this.factory.create(flags, this.$controller);
+    view.setLocation(this.location, MountStrategy.insertBefore);
+
+    const $controller = this.$controller;
+
+    this.queue(() => onResolve(
```
fkleuver, line 80: This bit actually needs to be sitting in `afterAttach` (same goes for `activateCases` in `afterBind`).

The idea is like this:
```
parent.beforeBind
parent.afterBind
parent.afterAttach <- parent now in DOM
  child.beforeBind
  child.afterBind
  child.afterAttach <- child now in DOM
  child.afterAttachChildren
parent.afterAttachChildren
```

Now, the `afterAttachChildren` bit won't be possible right now because the cases would actually need to be children of the switch controller for that. That's not currently the case and to be able to achieve that we would need to change a few things in the controller. Maybe for later.

So the best outcome for now will be this:
```
parent.beforeBind
parent.afterBind
parent.afterAttach <- parent now in DOM
parent.afterAttachChildren <-- a bit of a lie, but since this controller does not have a backing view model, the user will never be affected
  child.beforeBind
  child.afterBind
  child.afterAttach <- child now in DOM
  child.afterAttachChildren
```

However, with this current impl, it will be like this:
```
parent.beforeBind
  child.beforeBind
  child.afterBind
  child.afterAttach
  child.afterAttachChildren
parent.afterBind
  // didn't swap() already activate the cases..?
parent.afterAttach
parent.afterAttachChildren
```

Which is definitely not right

packages/runtime/src/resources/custom-attributes/switch.ts
```diff
@@ -0,0 +1,373 @@
...
+      view.activate(view, $controller, flags, $controller.scope, $controller.hostScope),
+      () => this.swap(flags, this.value)
+    ));
+    return this.promise;
+  }
+
+  public afterBind(initiator: IHydratedController<T>, parent: IHydratedParentController<T>, flags: LifecycleFlags): void | Promise<void> {
+    this.queue(() => this.activateCases(flags));
```
fkleuver, line 88: If you move the swap to `afterAttach` then this shouldn't be needed, as any kind of fromView binding (assuming that's what you're trying to account for) would be included by then

packages/runtime/src/resources/custom-attributes/switch.ts
```diff
@@ -0,0 +1,373 @@
...
+    return this.promise;
+  }
+
+  public beforeDetach(initiator: IHydratedController<T>, parent: IHydratedParentController<T>, flags: LifecycleFlags): void | Promise<void> {
+    const cases = this.activeCases;
+
+    const length = cases.length;
+    if (length === 0) { return this.promise; }
+
+    this.queue(length === 1
```
fkleuver, line 98: These bits need to move to `afterUnbind` for consistency with the way the controllers currently work.
I'll be inverting the deactivation in some fashion, after figuring out the last few edge cases with partial cancellation and animations, but it's never going to be the plan to have deactivation happen completely bottom-up, because the DOM generally needs to be detached top-down for performance reasons.

docs/user-docs/getting-started/conditional-rendering.md
```diff
@@ -114,3 +114,304 @@ When unchecked, our "Hello World" div will have the `aurelia-hide` class, which
...
+  </template>
+
+  <foo-bar switch.bind="status">
+    <template au-slot="s1">
+      <span case="dispatched">On the way.</span>
+      <span case="delivered">Delivered.</span>
+    </template>
+  </foo-bar>
+
+  <!-- The following example does produce some sort of output; but such usage is not recommended. -->
```
fkleuver, line 401:
```suggestion
  <!-- The following example does produce some sort of output; but such usage is not supported. -->
```
We need to be clear about what we do or don't support, and I think we should not be supporting any kind of child of `switch` other than a `case`. Even if it would technically work, it would also produce inconsistent lifecycle invocation orders, and to fix that the implementation would become significantly more complex (it's possible, but very complex). Imagine this:

```html
  <template switch.bind="status">
    <foo-bar>
      <span case="dispatched">On the way.</span>
      <span case="delivered">Delivered.</span>
    </foo-bar>
  </template>
```

Based on the current "best possible" implementation (as per https://github.com/aurelia/aurelia/pull/942/files#r502774546), here's the kind of thing that might happen if we were to link the cases up to the switch somehow:
```
switch.beforeBind
switch.afterBind
switch.afterAttach
  foo-bar.beforeBind
  foo-bar.afterBind
  foo-bar.afterAttach
  foo-bar.afterAttachChildren <- lies! the cases haven't even begun activating yet, let alone attach
  case1.beforeBind (etc..)
```

Sayan751: Done

packages/runtime/src/resources/custom-attributes/switch.ts
Sayan751: Moved the case deactivation to `Case#afterUnbind`. IMO that makes the `Switch#afterUnbind` simpler, especially when the `Case#afterUnbind` will be triggered anyway.

packages/runtime/src/resources/custom-attributes/switch.ts
Sayan751: Done.

packages/runtime/src/resources/custom-attributes/switch.ts
Sayan751: That makes sense. Also great explanation! Made the changes.

AureliaEffect:
JS Framework Benchmark: https://56215-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://56226-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/runtime/src/configuration.ts
```diff
@@ -165,7 +166,12 @@ export const DefaultResources = [
   ElseRegistration,
   RepeatRegistration,
   WithRegistration,
+  Switch,
+  Case,
+  DefaultCase,
```
fkleuver, line 171: We should be consistent here: either change these to `xxRegistration` as well (like the rest), or remove the `Registration` suffix from everything else and export them as their normal types.

The reasoning for exporting them as registrations is twofold:
1) The type `IRegistry` "hides" from the public the fact that they are classes. We don't necessarily want people inheriting from our built-in resources and we definitely don't want people mutating the prototypes (which happens in v1 a fair bit). They either set themselves up for unexpected breaking changes and/or force us to keep internal implementation details around that we never intended to expose. It's really important that we retain the freedom to refactor stuff as needed.
2) The common convention `Registration` "groups" things together in a more discoverable manner: intellisense will make it clear which kinds of things you can pass into `register` methods (and not do anything else with), and which things are potentially injectable (by prefixed with `I`) etc.

I'm open to other ideas for solving these issues (where 1 is primary and 2 is secondary), but in principle I'd prefer if we can stick to the `Registration` convention. You can still forcefully cast it to `Function` or something again in the tests if really needed (though I'd also kind of prefer avoiding inheritance in our tests for much the same reason..)

Sayan751: Done

AureliaEffect:
JS Framework Benchmark: https://56693-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/runtime/src/resources/custom-attributes/switch.ts
```diff
@@ -0,0 +1,373 @@
...
+  public readonly promise: Promise<void> | void = void 0;
+
+  public constructor(
+    @IViewFactory private readonly factory: IViewFactory<T>,
+    @IRenderLocation private readonly location: IRenderLocation<T>,
+  ) { }
+
+  public beforeBind(initiator: IHydratedController<T>, parent: IHydratedParentController<T>, flags: LifecycleFlags): void | Promise<void> {
+    const view = this.view = this.factory.create(flags, this.$controller);
+    view.setLocation(this.location, MountStrategy.insertBefore);
```
fkleuver, line 80: Why is this not in the constructor btw?

AureliaEffect:
JS Framework Benchmark: https://56723-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

PR #845, opened 2020-04-15
WIP: Lifecycle hooks consistency & async scheduler tasks

fkleuver:
# Pull Request

## 📖 Description

In this PR:
- Additional lifecycle hooks added on specific timings, and minor changes to the timings and order of some hooks. Purpose is to improve consistency and make them more compatible with scheduling, time-slicing, task cancellation and animation.
  - Combined `bind` and `attach` APIs on the controller interface into `activate`
  - Combined `detach` and `unbind` APIs into `deactivate`
  - In the built-in resources, `activate` is called during `afterAttach` and `deactivate` is called during `afterUnbind`. This means that, rather than lifecycle invocations going phase-by-phase (first all `bind`, then all `attach`, etc), they now go controller-by-controller. This allows controllers to be scheduled and time-sliced, makes first-class support for render-blocking animations and suspense (in cleaner forms) possible, simplifies integration with the router and generally reduces the "attack surface", or more specifically, the ways in which the API can be used wrongly.
  - All lifecycle hooks are now awaited if the component returns a promise. This means:
    - When returning a promise from `afterAttach` (that resolves after e.g. an animation is done), this will block activation of child components until the promise resolves. The same already applied for binding hooks
    - When returning a promise from `beforeDetach`, this will block unmounting of the component itself as well as deactivation of child components until the promise resolves (likewise, the same already applied for unbinding hooks)
  - `afterBind` changed to the timing of "after binding self, but before binding children"
  - `afterUnbindChildren` added at the original timing of `afterUnbind`
  - `afterUnbind` changed to the timing of "after unbinding self, but before unbinding children"
  - unbind lifecycle changed from depth-first to breadth-first to make the above possible and to be consistent with the bind lifecycle
  - `afterAttachChildren` added at the original timing of `afterAttach`
  - `afterAttach` added changed to the timing of "after attaching self, but before attaching children"
  - removed `beforeAttach` because it has no use anymore (timing would be identical to `afterBind`)
  - removed `afterDetach` because it has no use anymore (timing would be identical to `beforeUnbind`)

  - `dispose` added which is called after `afterUnbindChildren` (the last "out" hook), if and only if `.release()` is called on the controller (or one of its parents) prior to deactivation, and no parent view has caching enabled (in that case that view along with its whole subtree is returned to cache instead).

- Added an async task api to the scheduler, for tasks that need to be awaited internally. This will be useful for, among others:
[...]

PR #943, opened 2020-09-26
feat(scheduler): awaitable async task support

fkleuver:
# Pull Request

## 📖 Description

In this PR I cherry-picked the scheduler-related commits from https://github.com/aurelia/aurelia/pull/845 that pertain only to adding async support.

Adds a new `TaskQueueOption.async: boolean | 'auto'

- If `true`, the return value of the callback will be treated as a promise. Consecutive tasks will be run only after the promise resolved.

- If `'auto'`, the return value of the callback will be treated as a promise only if is an instance of a promise. Appropriate when it's not certain upfront whether the task callback will return a promise or not.

- If `false` (default), the scheduler will treat the task as a synchronous one and will not await it before starting the next task.

## 👩‍💻 Reviewer Notes

I left in some trace logging helper code that I needed from time to time in pinpointing what happens to scheduled tasks in certain edge cases. These calls are no-ops unless `enabled = false` is changed to `true` in the code. I've deliberately left it as a non-configurable private api as the overhead may be too big for keeping tracing around in this much detail. So, it will probably be removed or significantly reduced before RC.

## 📑 Test Plan

Added additional combinatory tests on top of the existing ones, testing various input combinations and more complex scenarios

## ⏭ Next Steps

Immediate follow-up to this would be other standalone pieces of aforementioned PR.

As for the scheduler specifically, it seems like the current priorities (`microTask`, `macroTask`, `renderTask` etc) don't make sense as being interchangeable for any given task. There should be some notion of prioritization, but not in the form of fundamentally different types of timings. So there will be a rework upcoming for this. Impact on API should be minimal (not a huge breaking change), but it should add a lot more capabilities in particular for things like timeslicing.

Also need the ability to group async tasks for parallelism while still keeping them awaitable. Maybe hierarchical task queues? Needs more thought and experimentation

### DIFF
--- a/packages/__tests__/2-runtime/scheduler.spec.ts
+++ b/packages/__tests__/2-runtime/scheduler.spec.ts
@@ -1,8 +1,33 @@
+/* eslint-disable @typescript-eslint/promise-function-async */
+/* eslint-disable @typescript-eslint/restrict-template-expressions */
+/* eslint-disable @typescript-eslint/require-await */
 import { TestContext, assert } from '@aurelia/testing';
 import { TaskQueuePriority, QueueTaskTargetOptions, ITask } from '@aurelia/runtime';
 
+function createExposedPromise() {
+  let resolve: () => void;
+  let reject: (err: any) => void;
+
+  const promise = new Promise(function ($resolve, $reject) {
+    resolve = $resolve;
+    reject = $reject;
+  });
+
+  return {
+    resolve,
+    reject,
+    promise,
+  };
+}
+
 function round(num: number) {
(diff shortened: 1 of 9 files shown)


AureliaEffect:
JS Framework Benchmark: https://53630-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://53653-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

PR #948, opened 2020-10-01
refactor(repeat): computed contextual properties

bigopon:
# Pull Request

## 📖 Description

Not all of the repeat contextual properties are always used. Change the computed one to getters so we don't have to calculate much in advanced, this could help with startup speed. Also creating dedicated class for repeat scope/override context. This idea is from FAST.

## ⏭ Next Steps

Observation should work like normal (will be computed instead of setter), though we know that each of the computed properties depends on the `$index` and `$length`, can just signal that to via `getter.getObserver`, maybe.


bigopon:
This should come in after we bring back the `.dependencies` option for observer locator to match v1 behavior, then special computed contextual properties can be defined per override context of the repeat.

PR #954, opened 2020-10-02
au-compose api completion

fkleuver:
# Pull Request


fkleuver:
Turns out there were a couple more issues with `au-compose` that need to be addressed, so pushing this back a bit

PR #956, opened 2020-10-03
report InterfaceSymbol friendly name when converted to string

fkleuver:
# Pull Request

## 📖 Description

In some situations you still see the whole function implementation of `InterfaceSymbol` when it's added to a log or error stack trace, which isn't helpful. With this change it'll be easier to inspect declared dependencies.


AureliaEffect:
JS Framework Benchmark: https://54304-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

PR #961, opened 2020-10-05
Ast cleanup

fkleuver:
# Pull Request

## 📖 Description

- Removed AST interfaces that weren't adding much (references are replaced by the actual classes now)
- Removed unused helper functions and type unions
- Made a few small perf tweaks to reduce polymorphism and reduce memory usage


AureliaEffect:
JS Framework Benchmark: https://54723-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

PR #970, opened 2020-10-08
Fix/benchmark

bigopon:
Trying to see if just copy pasting latest benchmark code will work


bigopon:
Will fix this later, probably post alpha

PR #971, opened 2020-10-09
Cleanup

fkleuver:
# Pull Request

## 📖 Description

This is a big cleanup PR, the intent is to cleanup more or less in the broadest sense of the word, anything that's unnecessary, impractical, somehow in the way, etc.

**BREAKING CHANGES**
- `Tracer` removed from `@aurelia/kernel`. This may be added back in again later in some other shape, but at the moment we are not using it anymore and we want to prevent anyone from starting to rely on it before we've worked out a nice way of providing tracing (if we're going to do it at all)
- `Reporter` removed from `@aurelia/kernel` - for the most part, ditto. It's not yet practical to have formalized error codes before the API's have been stabilized, so we may revisit this in the future when we do have a better picture of common/typical errors.
- `Activator` class removed from `@aurelia/runtime` - this turned out to be an unnecessary layer between the Controller and CompositionRoot. It's unlikely anyone would have depended on this as it required a very specific setup that depended on some internal state, but mentioning it for completeness sake.
- `QueueTaskOption#async` renamed to `suspend`. It essentially retains the same functionality. However, promises returned from task callbacks are now internally awaited regardless, albeit by default only for yielding. `suspend` now merely controls whether the promise blocks subsequent tasks or not. This should be much more useful in practice this way.
- Removed the `idle` priority from the scheduler. This is a first step towards making the scheduler a more customizable and well-defined api. The `idle` priority may be added again in a different shape later down the road.
- Removed the `ILifecycleTask` interface and all of its implementations. Things that previously returned an `ILifecycleTask` (most notably `Aurelia#start` and `Aurelia#stop`) now return `void | Promise<void>` and therefore those calls can no longer be chained with `.wait()`. Simply removing the `.wait()` calls will do the trick.
- Removed the `IStartTaskManager` and merged its functionality into `CompositionRoot`.

**Misc**
- Removed redundant/unused types, reusing more stuff where possible and practical
- Merged most runtime component class registrations into their exported interfaces, reducing the number of exports and required explicit registrations.
- Added an injectable for `IAurelia` and `ICompositionRoot`


AureliaEffect:
JS Framework Benchmark: https://56235-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://56261-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://56304-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://56314-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://56332-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

PR #980, opened 2020-10-18
docs: au-slot and some more tests

Sayan751:
# Pull Request


AureliaEffect:
JS Framework Benchmark: https://56927-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

docs/user-docs/app-basics/components-revisited.md
```diff
@@ -4,6 +4,1036 @@
...
+{% tab title="foo-bar.ts" %}
+```typescript
+export class FooBar {
+  @bindable public foo: string;
+}
+```
+{% endtab %}
+{% endtabs %}
+
+If you are not really convinced about the necessity of the `$host` scope, and wondering about a real-world use-case, let's consider the following example.
```
bigopon, line 400: I think we can word it in a more neutral way. Maybe If it's still unclear about the necessity of ...

docs/user-docs/app-basics/components-revisited.md
```diff
@@ -4,6 +4,1036 @@
...
+
+  <h3 au-slot="header">Tab2</h3>
+  <div au-slot="content">Tab2 content</div>
+
+  <!--...-->
+</my-tabs>
+```
+{% endcode %}
+
+This helps keeping things closer that belong together.
```
bigopon, line 525: 👍

docs/user-docs/app-basics/components-revisited.md
```diff
@@ -4,6 +4,1036 @@
...
+
+### Duplicate slots
+
+Having slots with same name is also perfectly alright.
```
bigopon, line 529: Having more than one `<au-slot/>` with...?

docs/user-docs/app-basics/components-revisited.md
```diff
@@ -4,6 +4,1036 @@
...
+-->
+```
+{% endcode %}
+
+Another interesting example of `repeat.for` is already shown in the [Binding scope](#binding-scope) section in context of `$host` keyword that you must check, if not already done.
+
+**Examples using `with`**
+
+The `with` template controller can also be used naturally with `au-slot`.
+A basic example is shown below.
```
bigopon, line 775: 👍

bigopon (approved):
This PR, and your constant effort on docs make me a bit embarrassed with how I've been doing doc 😓 Very nice work! 🚀

docs/user-docs/app-basics/components-revisited.md
Sayan751: Yup. Changed it.

docs/user-docs/app-basics/components-revisited.md
Sayan751: Changed

Sayan751:
Thanks @bigopon for the review. My tendency to write the docs only shows that I used Aurelia 1 since 2015 😛

AureliaEffect:
JS Framework Benchmark: https://56980-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://57343-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://58002-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

docs/user-docs/app-basics/components-revisited.md
```diff
@@ -4,6 +4,1052 @@
 
 // TODO
 
+## au-slot
+
+Aurelia provides another way of view projection with `au-slot`.
```
bigopon, line 9: I think we shouldn't be using a different term here. In v1, and probably in many other frameworks/docs, it's content projection. Maybe change to that? Also the term "content" won't click right away, but when it does, it will help users have better understanding of what we are trying to do with `[au-slot]`/`<au-slot/>` pair

docs/user-docs/app-basics/components-revisited.md
```diff
@@ -4,6 +4,1052 @@
...
+However, if you still want to have slot-like behavior, then you can use `au-slot`, as that makes the styling those custom elements/components easier.
+Instead of using shadow DOM, the resulting view is composed purely by Aurelia compilation pipeline.
+There are other aspects of `au-slot` as well which will be explored in this section with examples.
+
+{% hint style="info" %}
+- An obvious question might be "Why not simply 'turn off' shadow DOM, and use the `slot` itself"? We feel that goes to the opposite direction of Aurelia's promise of keeping things as close to native behavior as possible. Moreover, using a different name like `au-slot` makes it clear that the native slot is not used in this case, however still bringing slotting behavior to use.
+- If you have used the `replaceable` and `replace part` before or with Aurelia1, it is replaced with `au-slot`.
+- The following examples use [local templates](#local-templates) for succinctness only; the examples should also work for full-fledged custom elements.
+{% endhint %}
```
bigopon, line 21: In a near future, when @fkleuver has finished the `@processContent` feature, we can add some more example of how to use it to achieve `au-slot` effect with normal slotting code:
```ts
@processContent((node, compiler) => {
  const contentTemplate = compiler.createElement('template');
  contentTemplate.setAttribute('au-slot', '');
  while (node.firstChild != null) {
    contentTemplate.content.appendChild(node.firstChild);
  }
  node.appendChild(contentTemplate);
  // if not returning false, then the compiler should start compiling the template
})
```

packages/__tests__/3-runtime-html/au-slot.spec.ts
```diff
@@ -186,6 +186,20 @@ describe('au-slot', function () {
       ],
       { 'my-element': `<div>inner</div> <div>root</div>` },
     );
+
+    yield new TestData(
```
bigopon, line 190: nice extra tests

packages/__tests__/3-runtime-html/integration.spec.ts
```diff
@@ -0,0 +1,536 @@
...
+            <template repeat.for="i of 3">
+              <template repeat.for="i of 3">
+                <let gp.bind="$parent.$parent" p.bind="$parent" k.bind="gp.i" j.bind="p.i"></let>
+                \${k + j + i}
+              </template>
+            </template>
+          </template>
+        </template>`,
+
+      // The following template does not supported; kept here for docs purpose.
```
bigopon, line 500: `does not support` -> `is not supported`

packages/runtime/src/observation/binding-context.ts
```diff
@@ -96,33 +98,43 @@ function chooseContext(scope: Scope, name: string, ancestor: number) {
...
-  return null;
+  return [null, false];
 }
 
 export class Scope {
   private constructor(
     public parentScope: Scope | null,
     public bindingContext: IBindingContext,
     public overrideContext: IOverrideContext,
+    public readonly isComponentBoundary: boolean,
```
bigopon, line 137: We may be able to turn this into a real flag. For the `$host` traversal as well, to overcome the limitation of "cannot mix the scope of slotted content at will".
Example:
```ts
export const enum ScopeFlag {
  boundary     = 0b0_001,
  hostBoundary = 0b0_010,
}
```
And by default, a component is both a host boundary, and a boundary, so that we can switch back to different scope branch:

```html
<my-element>
  ${$host.message etc}
</my-element>
```

I'm not sure if I understood the limitation correctly, can you verify this potential way of doing it?

Sayan751: That's an interesting point of view. Although I am not sure I understood it correctly. Previously while traversing up the scope, there was not restriction, how far up the traversal can go. I am ignoring the `$parent` qualified expressions. For that, scope jump is bit different. This new flag is put there to prevent traversing all the way up to the root scope.

Secondly, as you said, "host boundary" is also a component boundary. From that respect, not sure what more restriction/advantage it will bring in preventing the scope traversal to reach to the root. 

As I said in the beginning, I might have understood it wrong. Would be nice, if you can elaborate further.

AureliaEffect:
JS Framework Benchmark: https://58096-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/runtime/src/observation/binding-context.ts
bigopon: Oops, sorry I missed this comment.

I just realized that we are employing a different way to build scope tree now. In v1, we have a special property name `parentOverrideContext` that points to the parent "scope".

With our current approach, we always have the scope tree normalize, and have a flag `allowParentScopeTraversal` to define the boundary.

I think with this approach, we may still have an issue around distinguishing between intentional cross boundary traversal, but it can be solved later. For now, I think this is a more prominent bug that should be fixed.

packages/runtime/src/observation/binding-context.ts
```diff
@@ -81,7 +83,7 @@ export class BindingContext implements IBindingContext {
   }
 }
 
-function chooseContext(scope: Scope, name: string, ancestor: number) {
+function chooseContext(scope: Scope, name: string, ancestor: number): [IBindingContext, boolean] | [undefined | null, false] {
```
fkleuver, line 86: In this part of the framework, every extra object created will have significant perf impact. We really need to avoid that here. Can we perhaps use a module-scoped variable for the boolean since all of this stuff is synchronous anyway?


In this part of the framework we really need to avoid creating objects for passing extra info around.

AureliaEffect:
JS Framework Benchmark: https://58366-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/runtime/src/observation/binding-context.ts
Sayan751: I too am not very comfortable with  this particular change. I think I can move the `in` check in the `BindingContext#get` itself, avoiding duplicate codes.

packages/runtime/src/observation/binding-context.ts
Sayan751: > intentional cross boundary traversal

That's interesting! I think that's still possible using `$parent` keyword, or a chain of those. Note that with ancestor jump (or the explicit scope traversal), this restriction is not applied.

packages/runtime/src/observation/binding-context.ts
Sayan751: Refactoring done!

AureliaEffect:
JS Framework Benchmark: https://58490-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

PR #983, opened 2020-10-24
chore: ensure test app build with Gh action

bigopon:
Current test apps are not ensured to be up to date, so sometimes it's a bit hard to test some new features/verify new changes. Add a script to ensure test apps are built as part of CI to prevent breakage.


AureliaEffect:
JS Framework Benchmark: https://57705-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://57832-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

PR #987, opened 2020-10-27
Platform fix

fkleuver:
# Pull Request

## 📖 Description

Instead of eagerly crashing when some globals are missing (such as `fetch` and `queueMicroTask`), fall back to using a function that only throws when it's actually used.

## 📑 Test Plan

@3cp will verify this in the jest setup.

## ⏭ Next Steps

We'll be working to bring more e2e test setups into this repo as part of finalizing the testing package.

### DIFF
--- a/packages/platform-browser/src/index.ts
+++ b/packages/platform-browser/src/index.ts
@@ -2,6 +2,12 @@ import { Platform, TaskQueue } from '@aurelia/platform';
 
 const lookup = new Map<object, BrowserPlatform>();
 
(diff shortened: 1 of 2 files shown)


AureliaEffect:
JS Framework Benchmark: https://57806-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

3cp:
makes skeletons e2e are fully passed now.

PR #992, opened 2020-10-31
Lifecycle hook finalization

fkleuver:
# Pull Request

## 📖 Description

This is the 3rd time in a relative short period that hooks are being renamed and/or timings changed - apologies to those inconvenienced by it. It took some trial-and-error, discussions and a fresh pair of eyes to see what made sense and what didn't.
They won't be renamed anymore now, and the timings won't change anymore either. These names and timings are final.

As for the reasoning behind the renames: the previous rename was a not-too-great attempt at trying to make the hooks self-explanatory, but we concluded that no truly self-explanatory (but still somewhat succinct) names are possible for all hooks.
Consequently, we're now back to some of the initial names we had when we started working on v2. They're short and clean, and their meanings will be clarified in the documentation (with the help of a visual diagram, soon to be added)

### Creation hooks
- Renamed `beforeCompile` to `hydrating` (about to hydrate/compile self)
- Renamed `beforeCompileChildren` to `hydrated` (just hydrated/compiled self, about to hydrate children)
- Renamed `afterCompileChildren` to `created` (all done - the timing of this hook is the same as the `created` hook in v1)
### Activation hooks
- Renamed `beforeBind` to `binding` (incoming bindable properties are set)
- Renamed `afterBind` to `bound` (from-view, let and ref bindings (basically anything coming from the view) are set here)
- Renamed `afterAttach` to `attaching` (the component itself is attached but its descendant components not yet)
  - Promises returned from `attaching` will no longer "suspend" activation of children.
    Motivation: things you do in `binding` / `bound` (e.g. data loading) typically require suspense, therefore those promises do suspend child activation. However, in `attaching` (which is called after the element itself is attached but before its children are) you might initiate a css animation and return a promise and that sort of thing. Waiting for that to be done before starting child activation does not benefit the end-user experience. The promise *is* awaited before `attached` is called, and is also included in the promise returned from `Controller#activate`, so there are no floating promises.
- Renamed `afterAttachChildren` to `attached` (all descendant components are attached and their `attaching` hooks awaited)
### Deactivation hooks
Note: `deactivate` used to be top-down but is now effectively bottom-up. Some tweaks for performance and UX considerations are also implemented (clarified down below)
- Renamed `beforeDetach` to `detaching` (about to detach self)
  - Promises returned from `detaching` will not "suspend" parent deactivation. This is the "opposing" hook of `attaching` and as such, the reasoning is the same: should you want to initiate a transition on a parent and a child, you probably want them to happen smoothly at the same time (instead of the parent waiting for the child). If you really must sequence them in some other way, using timeouts will allow you to do that (the other way around would have been a lot trickier / error-prone due to the necessity of floating promises)
  - After all `attaching` hooks are awaited, the top-level controller (initiator) is detached from the DOM *first*, and then the child tree is detached in a bottom-up fashion again. So the top-level controller explicitly gets special treatment and this is purely for performance reasons: a detached DOM is quicker to peel apart than an attached DOM.
- Renamed `beforeUnbind` to `unbinding` (about to unbind self)
  - Promises returned from `unbinding` will not "suspend" parent deactivation. Unlike `binding` and `bound`, there is typically no hierarchical dependency for "unbind" logic to happen in any particular order. The most common case of using promises here would be to save pending changes. In the probably rare case that a parent and child would independently save pending changes, you would probably want that to happen in parallel, rather than the parent waiting for the child. Like all the other hooks though, the promise is added to the return values of the initial `deactivate` call to ensure there are no floating promises and everything can be awaited.
- Removed the `afterUnbind` hook because there was no clear purpose for it now that we have `dispose`. This removal simplifies the controller and reduces the overhead of hook invocations
- Removed the `afterUnbindChildren` hook for the same reasons.

### Other
- Renamed `CompositionContext` back to `RenderContext` again
- Renamed `getCompositionContext` back to `getRenderContext` again
- Removed `Composer` (was previously `Renderer`), embedded its functionality into `RenderContext`
- Renamed `@instructionComposer` (previously `@instructionRenderer`) to simply `@renderer`
- Removed the `cancel` hook and removed partial cancellation. This is a temporary measure to reduce the complexity and number of required test cases so that we can go alpha more quickly. Adding it back in again later will not be a breaking change.
- Removed `ProjectorLocator` and the `ElementProjector`s. These abstractions turned out to be overkill and the functionality is now in the controllers.
- Removed `MountStrategy` and replaced it with `setShadowRoot` and `setHost` (making for a total of 3 distinct methods to set the mount target). Introduced a `MountTarget` enum that indicates whether to attach to a location, an ordinary host, or a shadow root.
- Removed the `ResourceModel` class and moved the `find` and `create` methods to the container.

## 👩‍💻 Reviewer Notes

This test file may be helpful in understanding how things are tied together now:
[`packages/__tests__/3-runtime-html/controller.hook-timings.integration.spec.ts`](https://github.com/aurelia/aurelia/pull/992/files#diff-ac7303ff9a1cd358311d54ef1b3716510a3ffbda18c94a03a0ca7c2cdd591377)

The most significant changes are in the controller:
[`packages/runtime-html/src/templating/controller.ts`](https://github.com/aurelia/aurelia/pull/992/files#diff-fc14a828edc9f7a5e8ed01d74efaa387440df42c97a1ff446259087184a7a145)

## 📑 Test Plan

Added a number of controller tests that perform fine-grained assertions on the lifecycle hook invocations and timings. They're by no means exhaustive from the perspective of the runtime as a whole (e.g. it's possible there are still some erroneous edge cases in template controllers that perform swaps), but they do verify that the controllers themselves work exactly as intended, and that all the template controllers are at least wired up correctly.

## ⏭ Next Steps

- Testing: mutation tests need to be added for `if`, `repeat`, `switch`, and more tests need to be added for `au-slot` and `au-compose`.
- Still a few compiler refactors/simplifications upcoming


AureliaEffect:
JS Framework Benchmark: https://58143-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://58153-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

bigopon:
Just a note that this PR should come in with doc change, even just with the old description, and new method names only. I think we've finally nailed it with the name, related to describing what the framework does, and what user can accept/learn/memorize.

I think the PR description is ok for where it is, but for future doc, the explanation should focus around 2 questions, instead of its current format:
- will this be awaited
- will this be awaited in sequence or parallel, and how  at each level (parent -> (all children at once/parallel) -> grandchild etc...)
At least I believe it's easier to digest after one or two rounds of reading.

AureliaEffect:
JS Framework Benchmark: https://58218-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://58240-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

PR #994, opened 2020-10-31
Refactor/cleanup old proxy impl

bigopon:
# Pull Request

## 📖 Description

Remove existing proxy implementation, and binding strategy around it.

**Motivation**: 
Working with proxy requires user awareness in quite a number of scenarios, and many are common. Our original direction was about automatically wrapping view model instances in a proxies and use this to enable simple tracking. While this can work well if user are careful around some certain scenarios, this will not work well with Aurelia unobtrusive programming model, also increase the supporting cost around questions, app style etc... This should also help avoid another potential issue of Proxy, that is bad performance.

A better approach to enable more powerful observation system, while not affecting performance, would be explicit observation, similar to how Knockout works.


AureliaEffect:
JS Framework Benchmark: https://58210-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://58291-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://58301-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://58310-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://58320-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://58329-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://58350-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

PR #997, opened 2020-11-03
Change target to ES2020

fkleuver:
# Pull Request

## 📖 Description

Since we're using optional chaining and nullish coalescing quite a lot now, and them being implemented in all major browsers, switching to ES2020 should improve the perf of the build output a bit.

Setting CircleCI as well as the required engine to node 14.x as this is now the LTS version.


fkleuver:
Nope. Nope. Nope. Still too much ancient rubbish in the dev dep tree that won't co-operate 😩

PR #999, opened 2020-11-05
Refactor/select observer

bigopon:
# Pull Request

## 📖 Description

A small cleanup to remove task from observers, and align the behavior around `.noFlush` flag. Also make `SelectValueObserver` working mechanism based on observation start/stop instead of bind/unbind


AureliaEffect:
JS Framework Benchmark: https://58765-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://58784-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

PR #1002, opened 2020-11-07
chore(switch): got rid of the async case change handler

Sayan751:
To have more synchronous determinism.

# Pull Request


bigopon (approved):
👍

AureliaEffect:
JS Framework Benchmark: https://58942-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

PR #1003, opened 2020-11-09
Feat/el observation extension

bigopon:
# Pull Request

## 📖 Description

- Continue the element observation refactoring
- Add ability to extend element observation back
- Merge target get accessor/observer
- Simplify setter observer

## 👩‍💻 Reviewer Notes

- There' some extra touches in i18n related to how translation binding handle its parameter: `secondaryExpression` is gone, and a real binding is used to handle parameter observation instead. This removes the need to have special cases for AST
- Merge some evaluate and connect code together in validation, resulted into simpler code for evaluating + observing validation params
- Basic doc for observation is in, mostly focus on the new API `.useConfig` & `.useGlobalConfig` to map events to props

## 📑 Test Plan

All existing tests work 

## ⏭ Next Steps

- remove redundant flags & flags properties (etc persistentFlags, ...)
- ~~remove `.connect()` (no longer has any places that use `.connect` of ASTs~~ (removed by the merged of #1004 into this branch)

### DIFF
--- a/packages/i18n/src/t/translation-binding.ts
+++ b/packages/i18n/src/t/translation-binding.ts
@@ -297,7 +297,7 @@ class ParameterBinding {
 
   private scope!: Scope;
   private hostScope: Scope | null = null;
-  
+
   public constructor(
     public readonly owner: TranslationBinding,
     public readonly expr: IsExpression,
--- a/packages/i18n/src/t/translation-binding.ts
+++ b/packages/i18n/src/t/translation-binding.ts
@@ -54,7 +54,6 @@ export class TranslationBinding implements IPartialConnectableBinding {
   public id!: number;
   public isBound: boolean = false;
   public expr!: IsExpression;
-  public parametersExpr?: IsExpression;
   private readonly i18n: I18N;
   private readonly contentAttributes: readonly string[] = contentAttributes;
   private keyExpression: string | undefined | null;
--- a/packages/i18n/src/t/translation-binding.ts
+++ b/packages/i18n/src/t/translation-binding.ts
@@ -1,28 +1,28 @@
-import { IEventAggregator, IServiceLocator, IContainer, toArray } from '@aurelia/kernel';
+import { toArray } from '@aurelia/kernel';
 import {
   BindingType,
   connectable,
   CustomElement,
   CustomExpression,
-  IBindingTargetAccessor,
-  IConnectableBinding,
-  IExpressionParser,
   Interpolation,
-  IObserverLocator,
-  IPartialConnectableBinding,
-  IsExpression,
   LifecycleFlags,
-  INode,
-  IHydratableController,
-  IsBindingBehavior,
   IPlatform,
 } from '@aurelia/runtime-html';
 import i18next from 'i18next';
 import { I18N } from '../i18n';
-import { Signals } from '../utils';
 
-import type { Scope } from '@aurelia/runtime';
-import type { CallBindingInstruction } from '@aurelia/runtime-html';
+import type { IContainer, IServiceLocator } from '@aurelia/kernel';
+import type {
+  Scope,
+  IsBindingBehavior,
+  IsExpression,
+  IBindingTargetAccessor,
+  IConnectableBinding,
+  IExpressionParser,
+  IObserverLocator,
+  IPartialConnectableBinding,
+} from '@aurelia/runtime';
+import type { CallBindingInstruction, IHydratableController, INode } from '@aurelia/runtime-html';
 
 interface TranslationBindingCreationContext {
   parser: IExpressionParser;
@@ -58,14 +58,14 @@ export class TranslationBinding implements IPartialConnectableBinding {
   private readonly i18n: I18N;
   private readonly contentAttributes: readonly string[] = contentAttributes;
   private keyExpression: string | undefined | null;
-  private translationParameters!: i18next.TOptions;
   private scope!: Scope;
   private hostScope: Scope | null = null;
-  private isInterpolatedSourceExpr!: boolean;
+  private isInterpolation!: boolean;
   private readonly targetObservers: Set<IBindingTargetAccessor>;
 
   public target: HTMLElement;
   private readonly platform: IPlatform;
+  private parameter: ParameterBinding | null = null;
 
   public constructor(
     target: INode,
@@ -75,9 +75,9 @@ export class TranslationBinding implements IPartialConnectableBinding {
     this.target = target as HTMLElement;
     this.i18n = this.locator.get(I18N);
     this.platform = this.locator.get(IPlatform);
-    const ea: IEventAggregator = this.locator.get(IEventAggregator);
-    ea.subscribe(Signals.I18N_EA_CHANNEL, this.handleLocaleChange.bind(this));
     this.targetObservers = new Set<IBindingTargetAccessor>();
+    this.i18n.subscribeLocaleChange(this);
+    connectable.assignIdTo(this);
   }
 
   public static create({
@@ -93,11 +93,11 @@ export class TranslationBinding implements IPartialConnectableBinding {
     const expr = typeof instruction.from === 'string'
       ? parser.parse(instruction.from, BindingType.BindCommand)
       : instruction.from as IsBindingBehavior;
-    if (!isParameterContext) {
+    if (isParameterContext) {
+      binding.useParameter(expr);
+    } else {
       const interpolation = expr instanceof CustomExpression ? parser.parse(expr.value, BindingType.Interpolation) : undefined;
       binding.expr = interpolation || expr;
-    } else {
-      binding.parametersExpr = expr;
     }
   }
   private static getBinding({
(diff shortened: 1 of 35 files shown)


AureliaEffect:
JS Framework Benchmark: https://59098-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/i18n/src/t/translation-binding.ts
```diff
@@ -147,33 +137,36 @@ export class TranslationBinding implements IPartialConnectableBinding {
...
     this.unobserveTargets(flags);
 
     this.scope = (void 0)!;
-    (this as unknown as IConnectableBinding).unobserve(true);
+    this.unobserve(true);
   }
 
   public handleChange(newValue: string | i18next.TOptions, _previousValue: string | i18next.TOptions, flags: LifecycleFlags): void {
-    if (flags & LifecycleFlags.secondaryExpression) {
-      this.translationParameters = this.parametersExpr!.evaluate(flags,  this.scope,  this.hostScope,  this.locator, null) as i18next.TOptions;
```
bigopon, line 161: An existing issue with translation is that it does not re-observe its bound scope for parameter changes. Which means complex expression won't be observe properly. This fixes that.

AureliaEffect:
JS Framework Benchmark: https://59136-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://59155-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://59168-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://59182-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://59209-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://59243-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/runtime-html/src/observation/style-attribute-accessor.ts
```diff
@@ -1,13 +1,11 @@
...
 import { emptyArray, kebabCase } from '@aurelia/kernel';
-import { INode } from '../dom';
 
 export class StyleAttributeAccessor implements IAccessor {
-  public readonly obj: HTMLElement;
   public currentValue: unknown = '';
   public oldValue: unknown = '';
 
-  public readonly persistentFlags: LifecycleFlags;
+  public readonly persistentFlags: LifecycleFlags = LifecycleFlags.none;
```
fkleuver, line 8: Can't we remove this one now?

packages/runtime/src/observation/observer-locator.ts
```diff
@@ -169,17 +188,12 @@ export class ObserverLocator {
         return obs;
       }
 
-      if (isNode) {
-        // TODO: use MutationObserver
-        return this.dirtyChecker.createProperty(obj, key);
-      }
```
fkleuver, line 175: I think I'd prefer to see this replaced by mutation observer or simply left alone for now, rather than being removed. Do we have any tests for this btw?

packages/runtime/src/observation/observer-locator.ts
```diff
@@ -29,18 +31,48 @@ export interface IObjectObservationAdapter {
...
+      logger.error('Using default INodeObserverLocator implementation. Will not be able to observe nodes (HTML etc...).');
+    });
+    return {
+      allowDirtyCheck: false,
+      handles: () => false,
+      useGlobalConfig: noop,
+      useConfig: noop,
+      getAccessor: () => propertyAccessor,
+      getObserver: () => propertyAccessor,
+    };
```
fkleuver, line 74: Classes instead of object literals please :)

packages/runtime/src/observation/observer-locator.ts
```diff
@@ -29,18 +31,48 @@ export interface IObjectObservationAdapter {
...
+   * Indicates the list of events can be used to observe a particular property
+   */
+  events: string[];
+  /**
+   * Indicates whether this property is readonly, so observer wont attempt to assign value
+   * example: input.files
+   */
+  readonly?: boolean;
+  /**
+   * A default value to assign to the corresponding property if the incomming value is null/undefined
```
fkleuver, line 45:
```suggestion
   * A default value to assign to the corresponding property if the incoming value is null/undefined
```

packages/runtime/src/observation/observer-locator.ts
```diff
@@ -29,18 +31,48 @@ export interface IObjectObservationAdapter {
...
+export interface INodeEventConfig {
+  /**
+   * Indicates the list of events can be used to observe a particular property
+   */
+  events: string[];
+  /**
+   * Indicates whether this property is readonly, so observer wont attempt to assign value
+   * example: input.files
+   */
+  readonly?: boolean;
```
fkleuver, line 43: This makes it html-specific again, albeit in a somewhat hidden manner. Can't we just move observer locator to runtime-html or maybe have an entirely alternative implementation there or something (bit of duplicate code is not the end of the world..)?

packages/runtime-html/src/observation/observer-locator.ts
```diff
@@ -44,160 +44,230 @@ const nsAttributes = Object.assign(
...
+  'checked',
+  'value',
+  'model',
+  // 'xlink:actuate',
+  // 'xlink:arcrole',
+  // 'xlink:href',
+  // 'xlink:role',
+  // 'xlink:show',
+  // 'xlink:title',
+  // 'xlink:type',
```
fkleuver, line 68: Why remove?

packages/runtime-html/src/observation/observer-locator.ts
```diff
@@ -44,160 +44,230 @@ const nsAttributes = Object.assign(
...
-    'xlink:show': true,
-    'xlink:title': true,
-    'xlink:type': true,
-    'xml:lang': true,
-    'xml:space': true,
-    'xmlns': true,
-    'xmlns:xlink': true,
-  }
-) as Record<string, boolean>;
+const getDefaultOverrideProps = () => [
```
fkleuver, line 55: Why not simple function declaration?

packages/runtime/src/observation/observer-locator.ts
```diff
@@ -29,18 +31,48 @@ export interface IObjectObservationAdapter {
 export interface IObserverLocator extends ObserverLocator {}
 export const IObserverLocator = DI.createInterface<IObserverLocator>('IObserverLocator').withDefault(x => x.singleton(ObserverLocator));
 
-export interface ITargetObserverLocator {
-  getObserver(flags: LifecycleFlags, observerLocator: IObserverLocator, obj: unknown, propertyName: string): IBindingTargetAccessor | IBindingTargetObserver | null;
-  overridesAccessor(flags: LifecycleFlags, obj: unknown, propertyName: string): boolean;
-  handles(flags: LifecycleFlags, obj: unknown): boolean;
+export interface INodeEventConfig {
```
fkleuver, line 34: Can we make this a class instead

packages/runtime-html/src/observation/style-attribute-accessor.ts
bigopon: We definitely can, but I'm trying to hold it off until next PR, where not only node observers but also normal accessors/observers & bindings are removed. As after this PR, none will touch/deal with this

packages/runtime/src/observation/observer-locator.ts
bigopon: It's moved to the node observer locator itself https://github.com/aurelia/aurelia/pull/1003/files#diff-24b820c4cb2f43548a6ffdeda7df9a74a89d046d7ab78431789a21ccf8d41256R220-R232

packages/runtime-html/src/observation/observer-locator.ts
bigopon: - In the existing implementation, they are treated as normal attribute, which fails to work. An example is svg `xlink:href`.
- Removing them as they are covered by special check `xlink` down below instead

packages/runtime-html/src/observation/observer-locator.ts
bigopon: 😬

packages/runtime/src/observation/observer-locator.ts
bigopon: Right nice suggestion, we can do that

AureliaEffect:
JS Framework Benchmark: https://59284-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/runtime/src/observation/observer-locator.ts
bigopon: It didn't feel as hidden to me as I think it's safe to assume there are properties that are readonly, regardless where it is? Though I think we can move this entirely to runtime html

packages/runtime/src/observation/observer-locator.ts
bigopon: Moved this to `runtime-html`, and converted into a class to provide config instead 👍

packages/runtime/src/observation/observer-locator.ts
bigopon: Moved this interface (now class) to runtime html, and it's better. I failed to realize that this shouldn't be in runtime, as they are just implementation details.

AureliaEffect:
JS Framework Benchmark: https://59390-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/runtime/src/observation/observer-locator.ts
bigopon: The existing tests are tweaked to reflect this. Also there's a few tests for observing readonly properties such as `tagName` etc.

AureliaEffect:
JS Framework Benchmark: https://59431-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

PR #1004, opened 2020-11-12
refactor: remove all ast.connect()

bigopon:
# Pull Request

## 📖 Description

The AST `.connect()` methods are no longer in use in any part, simply remove this, and tweak the tests so that they verify connecting while evaluating instead.


AureliaEffect:
JS Framework Benchmark: https://59234-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

PR #1006, opened 2020-11-15
Module loader

fkleuver:
# Pull Request

## 📖 Description

This PR introduces a ModuleLoader abstraction that can be used for efficiently dealing with dynamically imported resources. It's simpler than v1's loader due to the fact that it only accepts module objects or promises thereof, and not string paths, but its purpose is otherwise very similar.

The first consumer would probably be the router. Example usage to locate the first CustomElement (either default export or the first non-default export):
```ts
return loader.load(promiseOrObject, m => {
  let defaultExport: CustomElementDefinition | undefined = void 0;
  let firstNonDefaultExport: CustomElementDefinition | undefined = void 0;
  for (const item of m.items) {
    if (item.isConstructable) {
      const def = item.definitions.find(d => CustomElement.isType(d.Type));
      if (def !== void 0) {
        if (item.key === 'default') {
          defaultExport = def;
        } else if (firstNonDefaultExport === void 0) {
          firstNonDefaultExport = def;
        }
      }
    }
  }

  if (defaultExport === void 0) {
    if (firstNonDefaultExport === void 0) {
      throw new Error(`${promiseOrObject} does not appear to be a component or CustomElement recognizable by Aurelia`);
    }
    return firstNonDefaultExport;
  }
  return defaultExport;
});
```

## ⏭ Next Steps

Potentially add another higher-level api in `runtime-html` that offers a cleaner way to locate resources (e.g. by wrapping the snippet from this description)


AureliaEffect:
JS Framework Benchmark: https://59406-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/kernel/src/module-loader.ts
```diff
@@ -0,0 +1,190 @@
...
+
+  private loadFromPromise<TMod extends IModule, TRet>(promise: Promise<TMod>, transform: (m: AnalyzedModule<TMod>) => TRet | Promise<TRet>): TRet | Promise<TRet> {
+    if (this.promiseCache.has(promise)) {
+      return this.promiseCache.get(promise) as TRet | Promise<TRet>;
+    }
+
+    const ret = promise.then(obj => {
+      return this.loadFromObject(obj, transform);
+    });
+    this.promiseCache.set(promise, ret);
```
bigopon, line 86: caching this will prevent different transform being apply on the same module. Is this desirable?

fkleuver: I'd say the same transform should only be applied once per module at the most, but you got a point with different transforms as this would disallow different plugins from using the same modules. I'll use the transform fn as part of the cache key then.

AureliaEffect:
JS Framework Benchmark: https://59566-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

PR #1010, opened 2020-11-16
chore: update TOC.md for new component doc structure

bigopon:
Update TOC so that the doc displays correctly.


AureliaEffect:
JS Framework Benchmark: https://59537-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

PR #1013, opened 2020-11-16
refactor(all): remove persistentFlags

bigopon:
# Pull Request

## 📖 Description

Remove `persistentFlags` and its equivalence it classes as it's no longer needed for a static API set, and persistent flag can implicitly propagate without boundary.


AureliaEffect:
JS Framework Benchmark: https://59733-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

PR #1014, opened 2020-11-16
Controller / ShadowDOM fixes

fkleuver:
# Pull Request

## 📖 Description

- Let the controller automatically detect when its host comes from another controller. If that is the case, it will create its host element as a wrapper to put in-between the passed-in host and its fragment contents.

This enables shadowDOM to work seamlessly in components like `au-viewport` and `au-compose`, and also provides more consistency in non-shadowDOM situations. If this is not desired, just make the element `containerless`. However, the preferred way should be to always have a named host element as it's closer to web components.

Also:
- Make `INode`, `IEventTarget` and `IRenderLocation` easier to use with specific element types
- Improve `getVisibleText` test helper to traverse the (Shadow)DOM without relying on specific `instanceof` checks

### DIFF
--- a/packages/__tests__/3-runtime-html/binding-behaviors.spec.ts
+++ b/packages/__tests__/3-runtime-html/binding-behaviors.spec.ts
@@ -50,10 +50,7 @@ describe('binding-behaviors', function () {
     class FooAttr5 {
       @bindable({ primary: true })
       public value: any;
-      private readonly element: Element;
-      public constructor(@INode element: INode) {
-        this.element = element as Element;
-      }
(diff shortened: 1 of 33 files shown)


AureliaEffect:
JS Framework Benchmark: https://59770-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://59800-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

PR #1015, opened 2020-11-17
fix(observer-loc): new api for overriding accessors

bigopon:
checking properties to override accessor with observer should always be based on element tag names


AureliaEffect:
JS Framework Benchmark: https://59872-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

PR #1016, opened 2020-11-17
refactor: separate type import for runtime

bigopon:
Some more separation to make runtime code simpler for real values vs typings


AureliaEffect:
JS Framework Benchmark: https://59896-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://59921-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

PR #1017, opened 2020-11-17
fix(style-attribute-accessor): handle prop and url

brandonseydel:
# Pull Request

## 📖 Description

Fixing the style attribute accessor

### 🎫 Issues

CSS Properties were getting kebab cased on object/array binding which lost the property name. 
eg --background was getting turned to background

Background - When setting a base64 encoded URL on the background the getStyleTuplesFromString was splitting it incorrectly causing the URL to be chopped off at the first ;

## 📑 Test Plan

Added additional tests to cover some of the cases that failed


AureliaEffect:
JS Framework Benchmark: https://60006-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/kernel/src/platform.ts
```diff
@@ -1,5 +1,6 @@
 /* eslint-disable @typescript-eslint/no-unsafe-assignment,@typescript-eslint/no-explicit-any */
 export const emptyArray: any[] = Object.freeze<any>([]) as any;
+export const customPropertyPrefix: string = '--';
```
bigopon, line 3: I think this is a runtime-html concern, let's move it there?

brandonseydel: Yea that makes sense. I don't see it re-used anywhere else outside.

brandonseydel: just localize a const to attribute accessor you think?

brandonseydel:
https://jsben.ch/etLLK

AureliaEffect:
JS Framework Benchmark: https://60025-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

fkleuver:
Awesome, ty for fixing this 😃

PR #1023, opened 2020-11-22
Feat/templating extension

### LINKED ISSUE #975: Allow extending HtmlAttrSyntaxTransformer behaviour
# 🙋 Feature Request
Allow component authors to extend default `HtmlAttrSyntaxTransformer` behaviour to automatically transform `.bind` into `.two-way`

## 🤔 Expected Behavior
As a component author, I should be able to override default transformation algorithm so that consumers of my components can use `.bind` syntax for two way binding

## 😯 Current Behavior
Currently, consumers need to explicitly use `.two-way` for two way binding

## 💁 Possible Solution
* **Option 1** Expose a method on `IAttrSyntaxTransformer` to register additional binding mode transformations
* **Option 2** Introduce a "virtual" method instead of a local `shouldDefaultToTwoWay` function and export `HtmlAttrSyntaxTransformer`

## 🔦 Context
I am creating components and registering binding adapters for properties exposed on html elements.
For Aurelia v1 I was able to monkey-patch framework classes https://github.com/aurelia-ui-toolkits/aurelia-mdc-web/blob/4514f3d6769d8a5bfb8af72d02a3b0ce2c6f6dde/packages/base/src/mdc-component-adapters.ts#L63

bigopon:
# Pull Request

## 📖 Description

Add back the ability to extend Aurelia to transform `.bind` to `.two-way` on different combinations of elements and properties

Additionally:
- Add an explicit throw on `DirtyCheckProperty.prototype.setValue` so the error message makes more sense
- In attr syntax transformer, check two way binding from mapped target, instead of raw target

### 🎫 Issues

Closes #975
Resolves https://github.com/microsoft/fast/issues/3612

## 📑 Test Plan

Add tests for some custom element two way binding + observation

## ⏭ Next Steps

More doc on as many as possible web component library integration

### DIFF
--- a/docs/user-docs/examples/integration/ms-fast.md
+++ b/docs/user-docs/examples/integration/ms-fast.md
@@ -29,8 +29,8 @@ Aurelia.register(AppTask.with(IContainer).beforeCreate().call(container => {
     }
   });
 
-  // teaching Aurelia what events to use to observe properties of elements
-  // because fast component simply use a single change even to notify
+  // Teach Aurelia what events to use to observe properties of elements.
+  // Because FAST components all use a single change event to notify,
   // we can use a single common object
   const valuePropertyConfig = { events: ['change'] };
   nodeObserverLocator.useConfig({
--- a/docs/user-docs/examples/integration/ms-fast.md
+++ b/docs/user-docs/examples/integration/ms-fast.md
@@ -1,0 +1,63 @@
+# Integrating with Microsoft FAST Web Components
+
+If the example doesn't seem obvious, the following prerequisite reads are recommended:
+- [extending templating syntax](/app-basics/extening-templating-syntax.md)
+
+The following is a code example of how to teach Aurelia to work seamlessly with [Microsoft FAST](https://www.fast.design/):
+
+```ts
+import { AppTask, IContainer, IAttrSyntaxTransformer, INodeObserverLocator } from 'aurelia';
+
+Aurelia.register(AppTask.with(IContainer).beforeCreate().call(container => {
+  const attrSyntaxTransformer = container.get(IAttrSyntaxTransformer);
+  const nodeObserverLocator = container.get(INodeObserverLocator);
+  attrSyntaxTransformer.useTwoWay((el, property) => {
+    switch (el.tagName) {
+      case 'FAST-SLIDER':
+      case 'FAST-TEXT-FIELD':
+      case 'FAST-TEXT-AREA':
+        return property === 'value';
+      case 'FAST-CHECKBOX':
+      case 'FAST-RADIO':
+      case 'FAST-RADIO-GROUP':
+      case 'FAST-SWITCH':
+        return property === 'checked';
+      case 'FAST-TABS':
+        return property === 'activeid';
+      default:
+        return false;
+    }
+  });
+
+  // teaching Aurelia what events to use to observe properties of elements
+  // because fast component simply use a single change even to notify
+  // we can use a single common object
+  const valuePropertyConfig = { events: ['change'] };
+  nodeObserverLocator.useConfig({
+    'FAST-CHECKBOX': {
+      value: valuePropertyConfig
+    },
+    'FAST-RADIO': {
+      value: valuePropertyConfig
+    },
+    'FAST-RADIO-GROUP': {
+      value: valuePropertyConfig
+    },
+    'FAST-SLIDER': {
+      value: valuePropertyConfig
+    },
+    'FAST-SWITCH': {
+      value: valuePropertyConfig
+    },
+    'FAST-TABS': {
+      value: valuePropertyConfig
+    },
+    'FAST-TEXT-FIELD': {
+      value: valuePropertyConfig
+    },
+    'FAST-TEXT-AREA': {
+      value: valuePropertyConfig
+    }
+  });
+}))
+```
--- a/packages/__tests__/3-runtime-html/attr-syntax-extension.spec.ts
+++ b/packages/__tests__/3-runtime-html/attr-syntax-extension.spec.ts
@@ -1,5 +1,5 @@
 import { IContainer, IPlatform } from '@aurelia/kernel';
-import { AppTask, BrowserPlatform, IAttrSyntaxTransformer, NodeObserverLocator } from '@aurelia/runtime-html';
+import { AppTask, BrowserPlatform, CustomElement, IAttrSyntaxTransformer, NodeObserverLocator } from '@aurelia/runtime-html';
 import { assert, createFixture } from '@aurelia/testing';
 
 describe('3-runtime-html/attr-syntax-extension.spec.ts', function () {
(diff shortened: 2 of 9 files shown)


Sayan751:
I haven't reviewed the PR completely yet. Just skimmed through the docs written so far. The first thing that came to my mind, why not do that also for other binding modes. 

I will mention here one use-case for `from-view`. Lets say that there are 2 custom elements usually used together, as the data populated by the first CE is used by the second CE. With this in place we can do 

```html
<ce-one out-prop.from-view="data"></ce-one>
<ce-two in-prop.to-view="data"></ce-two>
```

It seems to me that with this extension, one can configure the default binding modes for `ce-one#in-prop` and `ce-two#out-prop` so that we can simply write the following instead.

```html
<ce-one out-prop.bind="data"></ce-one>
<ce-two in-prop.bind="data"></ce-two>
```

It is possible that no one need that. Finding such real life use-cases is also another objective of this comment. Personally, I find such inter-component data flow pretty cool.

bigopon:
@fkleuver raised a concern about giving too much freedom to the API could have negative effects. The scope of the effect maybe needs some clarification. Personally, I think as long as it satisfies a teachable/memoizable sentence, it's ok. And at the moment, it stands at:

> (1) .bind means one way binding, except when you tell Aurelia to be two way

Instead of

> (2) .bind most of the time means one way, some time from view, and some other time two-way

(1) is easier to grasp I believe

packages/runtime-html/src/attribute-syntax-transformer.ts
```diff
@@ -1,17 +1,77 @@
...
+  public useTwoWay(fn: ITwoWayTransformerFn): void {
+    this.fns.push(fn);
+  }
+
+  /**
+   * @internal
+   */
   public transform(node: HTMLElement, attrSyntax: AttrSyntax): void {
-    if (attrSyntax.command === 'bind' && shouldDefaultToTwoWay(node, attrSyntax)) {
+    attrSyntax.target = this.map(node.tagName, attrSyntax.target);
```
bigopon, line 64: Changing from mapping attribute to target property first, before checking if `.bind` means `.two-way` on the property feels weird. as mapping should be independent of "inferring" `bind`. For now, things still work similarly, but could use a revise later, depends on how this behaves in the wild.

AureliaEffect:
JS Framework Benchmark: https://60438-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

docs/user-docs/examples/integration/ms-fast.md
```diff
@@ -0,0 +1,63 @@
...
+      case 'FAST-SWITCH':
+        return property === 'checked';
+      case 'FAST-TABS':
+        return property === 'activeid';
+      default:
+        return false;
+    }
+  });
+
+  // teaching Aurelia what events to use to observe properties of elements
```
fkleuver, line 32:
```suggestion
  // Teach Aurelia what events to use to observe properties of elements.
```

docs/user-docs/examples/integration/ms-fast.md
```diff
@@ -0,0 +1,63 @@
...
+  // because fast component simply use a single change even to notify
```
fkleuver, line 33:
```suggestion
  // Because FAST components all use a single change event to notify,
```

packages/__tests__/3-runtime-html/attr-syntax-extension.spec.ts
```diff
@@ -0,0 +1,72 @@
...
+import { assert, createFixture } from '@aurelia/testing';
+
+describe('3-runtime-html/attr-syntax-extension.spec.ts', function () {
+  // cant deal with custom elements on nodejs
+  if (typeof process !== 'undefined') {
+    return;
+  }
+  let elementNameSeed = 0;
+  it('understands how to transform .bind on web component custom elements', async function () {
+    const elName = `my-el-${elementNameSeed++}`;
```
fkleuver, line 12:
```suggestion
    const elName = CustomElement.generateName();
```

packages/__tests__/3-runtime-html/attr-syntax-extension.spec.ts
```diff
@@ -0,0 +1,72 @@
+import { IContainer, IPlatform } from '@aurelia/kernel';
+import { AppTask, BrowserPlatform, IAttrSyntaxTransformer, NodeObserverLocator } from '@aurelia/runtime-html';
+import { assert, createFixture } from '@aurelia/testing';
+
+describe('3-runtime-html/attr-syntax-extension.spec.ts', function () {
+  // cant deal with custom elements on nodejs
+  if (typeof process !== 'undefined') {
+    return;
+  }
+  let elementNameSeed = 0;
```
fkleuver, line 10:
```suggestion
```

packages/runtime-html/src/attribute-syntax-transformer.ts
```diff
@@ -1,17 +1,77 @@
...
-import { AttrSyntax } from './resources/attribute-pattern.js';
+import type { AttrSyntax } from './resources/attribute-pattern.js';
 
 export interface IAttrSyntaxTransformer extends AttrSyntaxTransformer {}
-export const IAttrSyntaxTransformer = DI.createInterface<IAttrSyntaxTransformer>('IAttrSyntaxTransformer').withDefault(x => x.singleton(AttrSyntaxTransformer));
+export const IAttrSyntaxTransformer = DI
+  .createInterface<IAttrSyntaxTransformer>('IAttrSyntaxTransformer')
+  .withDefault(x => x.singleton(AttrSyntaxTransformer));
+
+type ITwoWayTransformerFn = (element: HTMLElement, property: PropertyKey) => boolean;
```
fkleuver, line 9:
```suggestion
type IsTwoWayPredicate = (element: HTMLElement, property: string) => boolean;
```

packages/runtime-html/src/attribute-syntax-transformer.ts
```diff
@@ -1,17 +1,77 @@
...
+    this.fns.push(fn);
+  }
+
+  /**
+   * @internal
+   */
   public transform(node: HTMLElement, attrSyntax: AttrSyntax): void {
-    if (attrSyntax.command === 'bind' && shouldDefaultToTwoWay(node, attrSyntax)) {
+    attrSyntax.target = this.map(node.tagName, attrSyntax.target);
+    if (attrSyntax.command === 'bind' && this.fns.some(fn => fn(node, attrSyntax.target))) {
```
fkleuver, line 65:
```suggestion
    if (
      attrSyntax.command === 'bind' &&
      (
        shouldDefaultToTwoWay(node, attrSyntax) ||
        this.fns.length > 0 && this.fns.some(fn => fn(node, attrSyntax.target))
      )
    ) {
```
Should only use the `fns` array if there are actually functions registered, otherwise we get an unnecessary slowdown in  the common case.

packages/runtime-html/src/attribute-syntax-transformer.ts
bigopon: Change this back so that two way interpretation is independent with attr to prop mapping, so they both can work with original attribute name

AureliaEffect:
JS Framework Benchmark: https://60456-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

PR #1032, opened 2020-11-26
Controller async optimizations

fkleuver:
# Pull Request

## 📖 Description
This PR improves the way lifecycle hook invocations are connected. Now uses a synchronous stack mechanism that completely eliminates additional overhead incurred from having async hooks. This improves performance with async hooks and also makes it much easier to deeply verify hook invocation order in complex integration tests internally (the difficulty of which is what triggered this change in the first place).


AureliaEffect:
JS Framework Benchmark: https://60879-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://60885-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

PR #1038, opened 2020-11-29
Refactor/observation cleanup

bigopon:
# Pull Request

## 📖 Description

More cleanup for a cleaner runtime, soon to help us restructure the layers better. Also removed the `.bind` from observer interface. Also add a few comments around getters/setters so that can be easier to debug

### 🎫 Issues

--

## 👩‍💻 Reviewer Notes

--

## 📑 Test Plan

--

## ⏭ Next Steps

Move all bindings to runtime html


AureliaEffect:
JS Framework Benchmark: https://61124-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://61163-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/runtime-html/src/observation/checked-observer.ts
```diff
@@ -106,12 +79,10 @@ export class CheckedObserver implements IObserver {
 
   public handleChange(newValue: unknown, previousValue: unknown, flags: LifecycleFlags): void {
     this.synchronizeElement();
-    this.callSubscribers(newValue, previousValue, flags);
```
bigopon, line 109: When source changes, the checked observer should treat this notification as a one way update, instead of trying to update back

packages/runtime-html/src/observation/checked-observer.ts
```diff
@@ -296,4 +251,18 @@ export class CheckedObserver implements IObserver {
       this.stop();
     }
   }
+
+  private observe() {
+    const obj = this.obj;
+
+    (this.valueObserver ??= obj.$observers?.model ?? obj.$observers?.value)?.subscribe(this);
```
bigopon, line 258: power of `??`??`????`???`

packages/runtime-html/src/observation/observer-locator.ts
```diff
@@ -48,12 +48,14 @@ const nsAttributes = Object.assign(
...
-// only type needs to be differentiated, so that binding queue the set call
-// doing it this way to avoid deopt in JS engines
-const elementPropertyAccessor = new PropertyAccessor();
-elementPropertyAccessor.type = AccessorType.Node | AccessorType.Layout;
+const elementPropertyAccessor = (function() {
+  // error when having .type = AccessorType.Node | AccessorType.Layout directly???
+  const node = AccessorType.Node;
+  const layout = AccessorType.Layout;
+  const accessor = new PropertyAccessor();
+  accessor.type = node | layout;
```
bigopon, line 56: got a weird error where I would get 66 instead of 2 | 8

packages/runtime/src/observation.ts
```diff
@@ -15,6 +14,18 @@ export interface IBinding {
...
 
+// todo:
+// merge collection subscription to property subscription
+// and make IWatcher simpler, so observers in static observation won't have to implement many methods
+// An alternative way is to make collection observation manual & user controllable
+// so it works even without proxy
+// todo: maybe enhance @connectable so that it provides these interfaces
+export interface IWatcher {
+  id: number;
+  observeProperty(obj: object, property: PropertyKey): void;
```
bigopon, line 25: doing this we unite the interface of observing property. Initially when doing `@watch`, `observeProperty` took an extra flags so it needed some simplification before unification

AureliaEffect:
JS Framework Benchmark: https://61296-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/runtime-html/src/observation/checked-observer.ts
fkleuver: Do we have tests verifying this in some way?

packages/runtime-html/src/observation/observer-locator.ts
fkleuver: Elaborate?

bigopon: If I do:
```ts
const elementPropertyAccessor = new PropertyAccessor();
elementPropertyAccessor.type = AccessorType.Node | AccessorType.Layout;
```
The value of `.type` is 66, which is wrong, because `AccessorType.Node` is 2, and `AccessorType.Layout` is 8

No weird behavior when doing
```ts
  const node = AccessorType.Node;
  const layout = AccessorType.Layout;
  const accessor = new PropertyAccessor();
  accessor.type = node | layout;
```
though. I'm not sure why. Maybe because I moved `Layout` as the 7th bit to 4th?

packages/runtime-html/src/observation/checked-observer.ts
bigopon: Good reminder, let me add some more

AureliaEffect:
JS Framework Benchmark: https://61390-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/runtime-html/src/observation/checked-observer.ts
bigopon: On a 2nd thought, we are doing any work, what do we need to verify here?

packages/runtime/src/binding/interpolation-binding.ts
```diff
@@ -76,20 +75,17 @@ export class InterpolationBinding implements IBinding {
...
     // Alpha: during bind a simple strategy for bind is always flush immediately
     // todo:
     //  (1). determine whether this should be the behavior
     //  (2). if not, then fix tests to reflect the changes/platform to properly yield all with aurelia.start().wait()
     const shouldQueueFlush = (flags & LifecycleFlags.fromBind) === 0 && (targetObserver.type & AccessorType.Layout) > 0;
     if (shouldQueueFlush) {
-      // an optimization: only create & queue a task when there's NOT already a task queued
-      // inside the lambda, use this.value instead of result, since it would always be the latest value
-      this.task ??= this.taskQueue.queueTask(() => {
-        targetObserver.setValue(this.value, flags, this.target, this.targetProperty);
```
bigopon, line 91: should go back to this optimization once flags is removed from `setValue`

AureliaEffect:
JS Framework Benchmark: https://61407-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/runtime-html/src/observation/checked-observer.ts
fkleuver: Imagine testing to verify that it *is* called - that test should now fail.

packages/runtime/src/binding/interpolation-binding.ts
fkleuver: I don't think we should remove flags from `setValue`. It's too useful there

packages/runtime-html/src/observation/observer-locator.ts
fkleuver: I don't get it. There is a difference between this:
```ts
  const node = AccessorType.Node;
  const layout = AccessorType.Layout;
  accessor.type = node | layout;
```

And this?
```ts
  accessor.type = AccessorType.Node | AccessorType.Layout;
```

packages/runtime-html/src/observation/checked-observer.ts
bigopon: That is what I was about to add too, but then I realized that test is a bit artificial, and will not as useful as we would hope. I think the absence of existing tests around that line, the existing functionalities & changes coverage, and the removal of it should give our "common sense" that it's non actionable.

packages/runtime/src/binding/interpolation-binding.ts
bigopon: Then probably we will need to continue with at least a rename, so that it reflects the opportunity we can have here better

packages/runtime-html/src/observation/observer-locator.ts
bigopon: Local cached and similarity between local vs CI error messages confused me. Changed it back to normal style.

AureliaEffect:
JS Framework Benchmark: https://61588-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

Issue #1040, opened 2020-12-02
Make DI.createInterface more intuitive


xenoterracide:
instead of using the builder pattern, since there is only one change, instead just use a simple callback

```ts
DI.createInterface('Foo', b => b....);
```

PR #1041, opened 2020-12-02
feat(di): remove DI.createInterface builder

xenoterracide:
change DI.create interface to use a callback interface instead of a
builder this reduces code in both the consumer and the implementation,
is more intuitive to use and usage is less error prone

closes: #1040

Signed-off-by: Caleb Cushing <xenoterracide@gmail.com>


bigopon:
The benefit of a delayed default implementation registration is that a library can export an interface and let the consumer provide the default, so the app code can be simplified. Though:
- once specified, default registration will be used for all future usages of the interface
- we are promoting to have a separate registration registry to connect implementation and interface together

So I think a lazy default registration isn't that important, or at least with the current implementation, where it doesn't resolve based on each handler.

packages/kernel/src/di.ts
```diff
@@ -1,4 +1,4 @@
-import { Metadata, isObject, applyMetadataPolyfill } from '@aurelia/metadata';
+import { Metadata, isObject, applyMetadataPolyfill, isNullOrUndefined } from '@aurelia/metadata';
```
bigopon, line 1: We are doing `== null` everywhere, I don't think we need a utility fn for this. There's historical factor, but I think we can live with it.

xenoterracide: I mean, I didn't write the utility... I just made use of it... can change it if you want. though pretty sure your eslint config is warning about `==`

bigopon: Let's change it to `==`, I'm sure we have used `==` extensively recently and it's ok. If it does complain, what we can do is to make the config aware of `== null`/`== undefined` and it should be ok (let's do against `== null`)

PR #1042, opened 2020-12-03
feat(router): add routing decorators, part 1

jwx:
# Pull Request

## 📖 Description

This is an **interim** PR that adds the routing decorators `routes` and `route`. As an **interim** PR it has a lot of code and comments that will be refactored and cleaned up in an upcoming PR.

### 🎫 Issues

## 👩‍💻 Reviewer Notes

## 📑 Test Plan

- Make sure new decorator tests pass

## ⏭ Next Steps

- Complete router refactor


AureliaEffect:
JS Framework Benchmark: https://61431-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://61628-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/router/src/decorators/routes.ts
```diff
@@ -0,0 +1,75 @@
...
+    const routes: (IRoute | Route)[] = [];
+    const metadata = Metadata.getOwn(Routes.name, Type);
+
+    // TODO: Check if they are indeed to be concatenated (and what that means
+    // for match order) or if one should replace the other
+    if (Array.isArray(metadata)) {
+      routes.push(...metadata);
+    }
+    if (Array.isArray(type.routes)) {
+      routes.push(...type.routes);
```
fkleuver, line 44: Metadata initialization is generally supposed to be a one-time operation and the returned configuration object should remain the same after the first initialization. Metadata can be changed, but only via the metadata API, not by changing static properties on a class.

Since we have already discussed this before, I know you disagree with this, and that's fair. I'm open to discussing these semantics (post-alpha) and potentially changing some of it. My agenda in this regard is to keep the router generally consistent with the runtime, and inconsistencies can either be fixed by changing the router or by changing the runtime, but they must be fixed.

packages/router/src/route.ts
```diff
@@ -0,0 +1,281 @@
...
+    }
+    if ('title' in Type) {
+      config.title = Type.title;
+    }
+
+    // if (Object.keys(config).length === 0) {
+    //   return {};
+    // }
+
+    return config instanceof Route ? config : Route.create(config, Type);
```
fkleuver, line 111: Ditto. This should only be initialized once and then reused.

fkleuver (changes requested):

PR #1046, opened 2020-12-03
Route recognizer fixes

fkleuver:
# Pull Request

## 📖 Description

- Fix empty path `''` routes (allow them to be added as well as recognized)
- Fix the negative match tests
- Add a recognize result cache for better perf (cache is invalidated whenever new routes are added, for safety)
- Remove the responsibilities w.r.t. query strings - stripping off the query string becomes the responsibility of the router.


AureliaEffect:
JS Framework Benchmark: https://61495-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

PR #1050, opened 2020-12-05
refactor: more interface reorganisation/clean up for moving bindings to platform

bigopon:
# Pull Request

## 📖 Description

Lessen the contract observer and accessor guarantees to simplify the observation interface. More refactor to prepare the move of all bindings to runtime-html


AureliaEffect:
JS Framework Benchmark: https://61692-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

PR #1051, opened 2020-12-06
refactor(connectable): make observe coll part of IConnectable, updat …

bigopon:
# Pull Request

## 📖 Description

Update `IConnectable`, and make the interface fuller with collection observation.
Also fix a bug in `ComputedObserver` where it wouldn't mark itself as not dirty after doing a computation.

## 👩‍💻 Reviewer Notes

There's a new support class for binding to record the collection observers being subscribed to. This should be soon merged with the standard record of property observers.

## 📑 Test Plan

Existing tests pass, and add a new tests to ensure getter are not invoked again, even accessed repeatedly.

## ⏭ Next Steps

Merge collection/property observation


AureliaEffect:
JS Framework Benchmark: https://61742-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

bigopon:
In this PR we have `IConnectable` changed from
```ts
interface IConnectable {
  id: number;
  observeProperty(obj: object, key: PropertyKey): void;
}
```

to
```ts
interface IConnectable {
  id: number;
  observeProperty(obj: object, key: PropertyKey): void;
  observeCollection(collection: Collection): void;
}
```

With this, by default, all bindings and watchers that use the `@connectable()` decorator will have these two observation methods enabled. This gives the observation system all the power needed to observe, and connect. Doing this allows us to move bindings/watchers to runtime-html without having and weird division, leaking abstraction, or losing power.

Though, because there's a difference in the way a collection is subscribed to, compared no a normal property subscription:
```ts
interface ICollectionSubscriber {
  handleCollectionChange(indexMap, flags);
}
```
and
```ts
interface ISubscriber {
  handleChange(newValue, oldValue, flags);
}
```

we still need to keep track of normal observers, and collection observers, and clean up the dependencies (observers) independently (for performance reasons). Moving forward, there should be a single interface, and some additional flags to signal what's the first parameter would be. whether it's a new property value, or an index map. Example names of the new flags for collection mutation:

```ts
interface BindingFlags {
  fromMutation,
  fromArray,
  fromMap,
  fromSet,
}
```

AureliaEffect:
JS Framework Benchmark: https://61797-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://61834-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/runtime/src/binding/connectable.ts
```diff
@@ -71,34 +89,68 @@ export function observeProperty(this: IConnectableBinding, obj: object, key: Pro
...
   version: number;
   count: number;
   binding: IConnectableBinding;
 } & ISubscriber & Record<string, unknown>;
 
-export interface BindingObserverRecord extends ISubscriber, ObservationRecordImplType {}
+export interface BindingObserverRecord extends ISubscriber, ObservationRecordImplType { }
 export class BindingObserverRecord implements ISubscriber {
-  public id!: number;
+  public id: number = idValue++;
```
fkleuver, line 147: Why remove `assignIdTo` here but keep it elsewhere? Can we keep the same mechanism everywhere please?

packages/runtime/src/binding/connectable.ts
```diff
@@ -168,18 +220,68 @@ export class BindingObserverRecord implements ISubscriber {
...
+  }
+
+  public clear(all?: boolean): void {
+    if (this.count === 0) {
+      return;
+    }
+    const observers = this.observers;
+    const version = this.version;
+    for (const [o, oVersion] of observers) {
+      if (all || oVersion !== version) {
```
fkleuver, line 252: Isn't this like a super hot path? Maybe not destructure and instead do normal property accessing here at least?

packages/runtime/src/observation/computed-observer.ts
```diff
@@ -295,13 +234,13 @@ export class ComputedWatcher implements IWatcher {
     this.running = true;
     this.record.version++;
     try {
-      enterWatcher(this);
+      enterConnectable(this);
```
fkleuver, line 237: Naming.. naming.. how about we name it `emitter` instead of `connectable`? Since that's kinda what it is?

packages/runtime-html/src/binding/attribute.ts
```diff
@@ -238,3 +237,5 @@ export class AttributeBinding implements IPartialConnectableBinding {
     this.isBound = false;
   }
 }
+
+connectable(AttributeBinding);
```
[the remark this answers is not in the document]
bigopon: Quite a good idea actually. When doing this, the motivations are mainly around generated code. With decorator usage, it will be something like
```ts
let MyClass = class {}

MyClass = __decorate([...])
```
instead of
```ts
class MyClass {}
decorate(...)
```
So without decorator, we get more beautiful (just class, no let), and smaller generated code (no inlined deocrate/usage of it)
---
Will do that graduate for the connectable

packages/runtime/src/binding/connectable.ts
```diff
@@ -32,34 +37,47 @@ function ensureEnoughSlotNames(currentSlot: number): void {
...
+
+  // todo:
+  // merge collection subscribable, and generally collection subscription
+  // with normal subscription
+  // there should be only unobserve()
+  /**
+   * A record storing collection observers that are currently subscribed to by this binding
+   */
+  // similar to record, this is an implementation detail
+  cRecord: BindingCollectionObserverRecord;
```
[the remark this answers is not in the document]
bigopon: It is not, but it's the same with `record` property above it. This is a temporary leak until we can merge collection/property subscription together, if we can.

packages/runtime/src/binding/connectable.ts
bigopon: 😬 just because you can

packages/runtime/src/binding/connectable.ts
bigopon: It's not super hot though. This is only called when a binding (or a connectable) does:
```ts
binding.cRecord.clear();
```
Our most common bindings: Property & Let only deal with properties, so it never come here.

Regardless, will fix, nice suggestion 👍

packages/runtime/src/observation/computed-observer.ts
bigopon: the two methods are: `observeProperty` and `observeCollection`, maybe anything will do, just not emitter 😆

packages/runtime-html/src/binding/attribute.ts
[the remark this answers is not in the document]
fkleuver: Yeah I got that with the generated code. Of course we're also working towards having native decorators, but for something as low-level as this perhaps we just shouldn't export a decorator anyway. Besides, it's technically a mixin..

packages/runtime/src/observation/computed-observer.ts
fkleuver: Doesn't it also have `callSubscribers` and stuff? Contractually it's practically an event emitter / event aggregator

fkleuver: Maybe let's call it `binding` then?

bigopon: the contract is `IConnectable`:
```ts
interface IConnectable {
  id: number;
  observeProperty(obj: object, propertyName: PropertyKey): void;
  observeCollection(obj: Collection): void;
}
```
That's all it needs to enable track on read. Binding implements this, but it's not necessarily always the case. Later, we will have API like this:

```ts
class Obj {
  @observable prop1
  @observable prop2
}

Observable.run(() => {
  var value = someObj.prop1 // track
  var value2 = someObj2.prop2 // track
});
```
So example is `Observable.run` will create an `IConnectable` under the hood, it may not necessarily a binding.

PR #1053, opened 2020-12-06
feat(plugin-conventions): rejects usage of <slot> in non-ShadowDOM mode

### LINKED ISSUE #1022: Throw or rewrite in build-time for usage of <slot> in non-shadow-dom project
If user didn't use ShadowDOM, then when `<slot>` is detected in a html template, the conventions preprocessor should 

1. either throw error to notify user that `<au-slot>` should be used instead.
2. or quietly rewrite `<slot>` to `<au-slot>`.

Probably throwing is better, the quiet rewrite might be very surprising.

3cp:
closes #1022


packages/webpack-loader/src/index.ts
```diff
@@ -1,6 +1,6 @@
 import { IOptionalPreprocessOptions, preprocess, preprocessOptions } from '@aurelia/plugin-conventions';
 import { getOptions } from 'loader-utils';
-import * as webpack from 'webpack';
+import type * as webpack from 'webpack';
```
fkleuver, line 3: Didn't know this was possible 😁

AureliaEffect:
JS Framework Benchmark: https://61767-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

PR #1057, opened 2020-12-08
Show/hide attributes attribute

fkleuver:
# Pull Request

## 📖 Description

This PR ports the `show` and `hide` attributes from v1.

The behavior is slightly different: in v1, an `au-hide` class was injected to the nearest `DOMBoundary` and applied to the to-be-hidden element.
I believe this is a somewhat unnecessary complexity.
Instead of doing that, we now directly modify the `style.display` property but backing up (and restoring) any existing value. 

Should any valid issues be reported as a result of this change (which I believe to be highly unlikely), we can then either choose to adopt v1's `au-hide` mechanism again, or those users may choose to use the new class binding in conjunction with their own "hide" class, e.g. `hide.class="someCondition"`. My reasoning is that if someone is doing some really fancy stuff with style mutations, they should probably take full control with their own class anyway.

## 👩‍💻 Reviewer Notes

Interesting difference with v1 is that instead of duplicating the class as `Show` / `Hide`, this is using the `@alias` decorator and reading from the injected instruction to determine whether it's "show" or "hide".

Had to make some minor tweaks to the rendering to accommodate for that, and also found a minor bug in DI along the way. Also took that opportunity to do a minor simplification in DI w.r.t. `Factory`, which might be a decent perf improvement actually.

## 📑 Test Plan

Combinatory integration tests added, which I believe cover everything that needs to be covered for these attributes.


AureliaEffect:
JS Framework Benchmark: https://62023-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

fkleuver:
@bigopon merging this so I can move forward with porting realworld, but if you have any concerns about the changes from v1 let me know so they can be addressed if needed

packages/kernel/src/di.ts
```diff
@@ -749,129 +736,34 @@ export class Factory<T extends Constructable = any> implements IFactory<T> {
...
-
-    if (transformers == null) {
-      return instance;
-    }
-
-    for (let i = 0, ii = transformers.length; i < ii; ++i) {
-      instance = transformers[i](instance);
+    let instance: Resolved<T>;
+    if (dynamicDependencies === void 0) {
+      instance = new this.Type(...this.dependencies.map(function (d) {
```
bigopon, line 745: if we are to use `.map()`, then we should use `.map(staticFn, container)` form, so that the `this` inside the static mapper fn `staticFn` can be used instead. It allows us to avoid having to create a lambda everytime

packages/kernel/src/di.ts
```diff
@@ -947,6 +839,8 @@ export class Container implements IContainer {
   private readonly root: Container;
 
   private readonly resolvers: Map<Key, IResolver | IDisposableResolver>;
+  // Factories are "global" per container tree
```
bigopon, line 842: Per this comment, we no longer can invoke an instance of something at a child container, is this correct understanding @fkleuver?

fkleuver: Not correct. The factories were previously cached globally at metadata level, now they are cached per container root, so they are cached *less* "aggressively" than before now.

bigopon: Right yes, ok let's get back to this later, I think ideally we should cache it that the handler level

PR #1061, opened 2020-12-11
WIP Feat/batch api

bigopon:
# Pull Request

## 📖 Description

Implement batch. The main changes are around these following methods of `SubscriberRecord`:

```ts
notify(val, old, flags) {
  if (batching) {
    batches.set(this, [val, old, flags]);
    return;
  }
}

notifyCollection(indexMap, flags) {
  if (batching) {
    batches.set(this, [indexMap, flags]);
    return;
  }
}
```

There' two ways to do batching:
- batch notifier delivering the change notification
- batch receiver handling the change notification
---
Batching notifier is the equivalent of v1, and probably the more natural way of doing it. I'm not sure about batching receiver handlers.

Notes:

- This implementation relies on subscriber record as the only way to communicate changes. This may not be true, if observers implemented in user land are done differently (e.g without a `@subscriberCollection` + `subscribe/unsubscribe` combo). Do we want to enable it? And if yes, how?
- No nested batching support for now, there doesn't seem to be a clear use case for it (concrete example, can always add it later on if needed), and it could result in simple implementation for batching

## Ergonomic values
- there should be a `@batch` decorator, so that when used to decorate a method, it will wrap the method in a `batch()` call. Prior implementation/idea: https://mobx.js.org/actions.html


AureliaEffect:
JS Framework Benchmark: https://62403-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://62407-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

bigopon:
It seems there's a need to:

- distinguish nested batch (child) from nesting batch (parent).
- nested batch will need to either:
  a. disregard nesting batch and invoke immediately at the end OR
  b. be part of the nesting batch and let the nesting batch controls when to notify the subscribers

The reason for the above is because without it, you will not have confidence in your code:
```ts
// an application code
batch(() => {
  someObj.someProp = value
  someObj.callSomeFunction()
})
```

```ts
// In library code
...
callSomeFunction() {
  batch(() => {
    obj.prop = value
  })
  // oops, invoked or not?
  if (obj2.compute_observer_that_observers_Obj_prop)
  ...
}
```

in `callSomeFunction`, after batch, it's expected that `if (obj2.....)` behaves correctly with the latest value

But doing this sort of sometimes breaks the expectation that `batch()` may not be able to always ensure everything is batched.  This can be a matter of documentation: batching third party code (mostly function calls) should take nested `batch()` call into account.

Intrigued by this dilemma, I was checking what `MobX` is doing around this, and found out they do the other way:

- always pend everything until batch depth counter reach back to 0. An example is here https://stackblitz.com/edit/mobx-example-qqtf5c?file=index.js ) created based on a mobx react example
- with an option to specific that always run immediately after a `batch()` call, as persuggested by the API here https://github.com/mobxjs/mobx/blob/8162aafaa341ea8f80b8576e8abe1a811f079efa/packages/mobx/src/core/action.ts#L58
```ts
export function executeAction(
    actionName: string,
    canRunAsDerivation: boolean,
    fn: Function,
    scope?: any,
    args?: IArguments
)
```

To test the above example:

- type something in the todo input box
- click add
- check the console, should see a log like this
image

Explanation:

- in the constructor of the file store/index.js, there's an autorun(() => ..., this can be seen as the equivalent of our binding connecting an object to an object (not node, all sync).
- This `autorun` reads `todoList.length` inside its effect function (the lambda parameter of it), which means whenever the length of `todoList` changes, it re-runs.
- the click handler of `[Add]` button will call a function, that in turn calls another function. Those 2 functions are both methods of the todo store, and both are action, one nests another.

Real behavior: 
- when adding a todo item, `todoList.length` gets changed, `autorun` will need to react. There' 2 possible scenarios:
    1. `autorun` should get notified immediately inside the child action (the method `addTodo`)
    2. `autorun` will be treated just as a normal subscriber, and thus will be only be notified at the end of a complete batch (after parent action has finished)

A 3rd possibility is an ability to configure that an action should be able to declare its own notification strategy: merge with the parent, or immediately invoke.

To start with, I think the option (1) makes more sense for us, as our binding system needs stay intuitive, deterministic and synchronous. Though there's an implication:
- if there's a existing notifier in parent batch, and child batch happens to add the the notifier, then the notifier will still be invoked in both rounds, but the value will be the latest value from the child batch.

cc @3cp

AureliaEffect:
JS Framework Benchmark: https://62603-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://62685-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://62693-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

PR #1062, opened 2020-12-11
Repo cleanup

fkleuver:
# Pull Request


bigopon:
@fkleuver it seems we are removing Krauset, or we moving it elsewhere?

fkleuver:
@bigopon I'm removing it for now, it dramatically simplifies the dependency tree and CI setup of the repo so I'm doing this primarily to make the repo lean and clean to make it easier to perform various migrations (e.g. away from lerna).
Also, we need to focus on our own benchmark. We should be able to get all measurements that krausest gave us by making a few small additions.

bigopon:
Krauset has been helped us catch a few regressions related creation/destroy perf so I'm a bit wary that it's going away. And chance is life cycle hook could have a negative effect on create/destroy, though we don't know how much (also non-functional yet important benchmarks like memory, bundle size etc).

Though the plan sounds ok, let's park it for now.

packages/__tests__/1-kernel/module-loader.spec.ts
```diff
@@ -291,8 +291,8 @@ describe('ModuleLoader', function () {
     const $res1 = await res1;
     const $res2 = await res2;
 
-    loader.load(promise, transform1);
-    loader.load(promise, transform2);
+    void loader.load(promise, transform1);
+    void loader.load(promise, transform2);
```
bigopon, line 295: these lines, and some `void Aurelia.app(...).start()` are not normal looking code. Is this from auto lint fix?

fkleuver: `eslint` is being difficult again.
It (correctly) marks these lines as "no-floating-promises" but if I `eslint-disable` them, then it complains that there were no violations of that rule. It's the only rule where it goes wrong. This was just a quick way to get it to shut up. Will try to fix this later.

PR #1065, opened 2020-12-19
Refactor/remove unused interfaces and lifecycle batching

bigopon:
# Pull Request

## 📖 Description

A cleanup for:
- removing `ILifecycle` and the existing batching mechanism, this is to prepare for static batch API
- normalize subscribing to collection, and normal property observers, so that they all use `.subscribe()`
- clean up `ISubscriberCollection` & surrounding code
- hide internals of collection observer interface


packages/__tests__/validation-html/validate-binding-behavior.spec.ts
```diff
@@ -105,8 +103,7 @@ describe('validate-binding-behavior', function () {
       if (observeCollection) {
         this.employeesMediator = new BindingMediator('handleEmployeesChange', this, observerLocator, serviceLocator);
         this.employeeObserver = new ArrayObserver(this.org.employees);
-        this.employeeObserver.lifecycle = lifecycle;
-        this.employeeObserver.getLengthObserver().addSubscriber(this.employeesMediator);
+        this.employeeObserver.getLengthObserver().subscribe(this.employeesMediator);
```
bigopon, line 106: in the previous refactor, collection length observer became a normal subscriber of collection observer. Hence it won't gets notified if it doesn't subscribe to the collection observer. Calling `.addSubscriber` on a collection length observer does not subscribe itself to the owning collection observer, hence the test below were broken last time. I misunderstood this and thought the test was broken. Fixed this

bigopon: @Sayan751 this is what I meant to ask with "whether validation auto subscribes to collection". Just checked the code, there' manual subscriptions in these tests, I missed it.

packages/runtime/src/observation/subscriber-collection.ts
```diff
@@ -13,79 +13,66 @@ import type {
...
-    ensureProto(proto, 'unsubscribe', removeSubscriber);
-  };
+/* eslint-disable @typescript-eslint/ban-types */
+export function subscriberCollection(): ClassDecorator;
+export function subscriberCollection(target: Function): void;
+export function subscriberCollection(target?: Function): ClassDecorator | void {
+  return target == null ? subscriberCollectionDeco : subscriberCollectionDeco(target);
 }
 
-export function collectionSubscriberCollection(): ClassDecorator {
```
bigopon, line 37: After subscriber record, there's no need to distinguish between a collection sub collection, or a sub collection. Removing it. Only the notification method needs to be differentiated.

packages/runtime/src/observation/observable.ts
```diff
@@ -129,47 +130,25 @@ export function observable(
...
   let notifier = lookup[key as string] as SetterNotifier;
   if (notifier == null) {
-    notifier = new SetterNotifier(set);
+    notifier = new SetterNotifier(obj, callbackKey, set, initialValue === noValue ? void 0 : initialValue);
     lookup[key as string] = notifier;
-    if (initialValue !== noValue) {
-      notifier.setValue(initialValue, LifecycleFlags.none);
-    }
-    const callback = obj[callbackKey as string];
-    if (typeof callback === 'function') {
```
bigopon, line 170: turning callback into a special knowledge of the `@observable` observer, so it will be invoked before other subscribers. This is consistent with v1, and help distinguish between callback and normal subscription

packages/runtime/src/observation.ts
```diff
@@ -162,16 +91,20 @@ export interface ICollectionSubscriber {
...
-  subscribe(subscriber: ISubscriber | ICollectionSubscriber): void;
-  unsubscribe(subscriber: ISubscriber | ICollectionSubscriber): void;
+  subscribe(subscriber: ISubscriber): void;
+  unsubscribe(subscriber: ISubscriber): void;
 }
 
 export interface ICollectionSubscribable {
-  subscribeToCollection(subscriber: ICollectionSubscriber): void;
-  unsubscribeFromCollection(subscriber: ICollectionSubscriber): void;
+  subscribe(subscriber: ICollectionSubscriber): void;
```
bigopon, line 99: I think there's no need to have special method name describing the intention of subscribing to a collection. Changed it to simply subscribe/unsubscribe. It's the interface of the subscriber that matters. In a future PR, it should be applied to validation controller add/remove subscriber methods.

packages/runtime/src/observation.ts
```diff
@@ -363,13 +298,9 @@ export interface ICollectionChangeTracker<T extends Collection> {
  */
 export interface ICollectionObserver<T extends CollectionKind> extends
   ICollectionChangeTracker<CollectionKindToType<T>>,
-  ICollectionSubscriberCollection,
```
bigopon, line 366: removing this interface from `ICollectionObserver`, to lessen the contract: only `.subscribe/.unsubscribe` methods are exposed, not `subs: ISubscriberRecord`

AureliaEffect:
JS Framework Benchmark: https://62827-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://62852-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

Issue #981, opened 2020-10-22
processContent alternative


MaximBalaganskiy:
# 🙋 Feature Request
Provide an a way to modify custom element content before everything else

## 🤔 Expected Behavior
Should work similar to `processContent` in AU1

## 😯 Current Behavior
Could not find such feature

## 🔦 Context
Sometimes it's useful to rearrange custom element content before is is processed. E.g. move different parts of the DOM to designated places. Another example is replacing `mdc-row` and `mdc-cell` tags with `tr` and `td` inside an `mdc-table` component

## 💻 Examples
```html
  <mdc-data-table>
    <mdc-data-table-header>
      <mdc-data-table-header-cell>Dessert</mdc-data-table-header-cell>
    </mdc-data-table-header>
    <mdc-data-table-content>
      <mdc-data-table-row repeat.for="dessert of desserts">
        <mdc-data-table-cell>${dessert.name}</mdc-data-table-cell>
      </mdc-data-table-row>
    </mdc-data-table-content>
  </table>
</mdc-data-table>
```
would get transformed into 
```html
<mdc-data-table>
  <table>
    <thead class="...">
      <th class="...">Dessert</th>
    </thead>
    <tbody class="...">
      <tr class="..." repeat.for="dessert of desserts">
        <td class="...">${dessert.name}</td>
      </tr>
    </tbody>
  </table>
</mdc-data-table>
```

MaximBalaganskiy:
@fkleuver I was advised you can shed some light on this feature

bigopon:
At the moment, this can be done via hooking into `define` lifecycle, and modify the content. Though it's missing:
- an official API
- a way to declare no compilation (return false in `@processContent`)

fkleuver:
Yep, this is work in progress and will be added before alpha.

Sayan751:
I think this is a duplicate of #667.

Sayan751:
During the course of development, the signature of the hook has been changed. Refer the [docs](https://docs.aurelia.io/app-basics/components-revisited#the-processcontent-hook) for details.

PR #1068, opened 2020-12-21
feat: @processContent

### LINKED ISSUE #667: `@processContent`
# 💬 Discussion

In v1, custom element has `processContent` API to help user to prepare a custom element usage in a template before the compiler start processing it. It can also optionally compile the content, as the compilation process in v1 is processing and marking by id, unlike v2, processing and marking by order.

Up until now, the view compilation is split into two steps: [html -> symbols] then [symbols -> instructions]. This is difficult to support `processContent` API as it works with html, but [html ->symbols] step is not intuitive for end user as an API, so we should consider merging the two steps into one (compile from html -> instruction)

For now, `render` hook in custom element (and soon custom attribute) will allow us to support `processContent` usage, though through a much more ceremonial code.

This issue is a reminder that we should revisit this API once other items with higher priority are in place

## 🔦 Context

N/A

## 💻 Examples

N/A

@fkleuver @EisenbergEffect

Sayan751:
# Pull Request

## 📖 Description

Closes #667, #981 
Also adds the missing docs for `enhance`.


AureliaEffect:
JS Framework Benchmark: https://63064-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

Sayan751:
I will also push the docs soon.

AureliaEffect:
JS Framework Benchmark: https://63100-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

AureliaEffect:
JS Framework Benchmark: https://63116-121594173-gh.circle-artifacts.com/0/test/js-framework-benchmark/webdriver-ts-results/table.html

packages/runtime-html/src/resources/custom-element.ts
```diff
@@ -545,3 +552,42 @@ export const CustomElement: CustomElementKind = {
...
+        (def as Writable<CustomElementDefinition<Constructable<TClass>>>).processContent = hook;
+      } else {
+        Metadata.define(pcHookMetadataProperty, hook, target);
+      }
+      return target;
+    };
+}
+
+function ensureHook<TClass>(target: Constructable<TClass>, hook: ProcessContentHookNames<TClass> | ProcessContentHook): ProcessContentHook {
+  if (typeof hook === 'string') {
```
bigopon, line 582: What do we want to support with an extra string API?

bigopon (commented):
Looking great! Left a few minor comments. For the tests, I think we should have 1 or 2 more for inheritance scenarios where:
- `processContent` is defined in parent and the child will use it
- `processContent` is defined in child, and the parent won't use it

packages/runtime-html/src/resources/custom-element.ts
Sayan751: I thought earlier that it will bring more flexibility. But now it seems to be bit superfluous, especially the associated syntax `@processContent<typeof ViewModelClass>('hookName')` is not very nice. I will ditch that class-level decorator syntax. 

However, the line you pointed out is also coming from the fact that the following syntax is also supported.

```ts
export class MyElement {
  @processContent()
  static processContent(node: INode, platform: IPlatform) { }
}
```

Sayan751: The said decorator syntax is now removed.

Sayan751:
> For the tests, I think we should have 1 or 2 more for inheritance scenarios where:
> 
> * `processContent` is defined in parent and the child will use it
> * `processContent` is defined in child, and the parent won't use it

@bigopon I am not sure if the inheritance should be supported for this. From my experience, inheritance relationship between 2 custom elements is bit problematic. If sharing the hook impl. is the main concern, then the hook method can always be promoted to a function (referencing static function will probably work as well) and can be reused many times.

From a different point of view, if we support this, then the question is whether or not the custom element definitions are inherited as well? For example, the `template` or the `dependencies` defined for base class gets inherited by the children or not. As per my understanding the only the metadata defined for the child class are considered for this. From that perspective, I think we should not make an exception for the `processContent` hook.

packages/runtime-html/src/resources/custom-element.ts
```diff
@@ -545,3 +552,40 @@ export const CustomElement: CustomElementKind = {
...
+
+type DecoratorFactoryMethod<TClass> = (target: Constructable<TClass>, propertyKey: string, descriptor: PropertyDescriptor) => void;
+type ProcessContentHook = (node: INode, platform: IPlatform) => boolean | void;
+
+const pcHookMetadataProperty = Protocol.annotation.keyFor('processContent');
+export function processContent(hook: ProcessContentHook): CustomElementDecorator;
+export function processContent<TClass>(): DecoratorFactoryMethod<TClass>;
+export function processContent<TClass>(hook?: ProcessContentHook): CustomElementDecorator | DecoratorFactoryMethod<TClass> {
+  return hook === void 0
+    ? function (target: Constructable<TClass>, propertyKey: string, _descriptor: PropertyDescriptor) {
```
bigopon, line 564: I think this is a bit error prone, where you can decorate any property, method to turn it into a `processContent` hook. Maybe give some motivation explanation why we want to have such API? For v1, we have been having `@processContent` only on the class itself and never had a complain

bigopon:
> From that perspective, I think we should not make an exception for the processContent hook.

This reminds me that we are not having inheritance for watch atm either. I think it's ok to be in this state for now, and realign our inheritance stories later.

packages/runtime-html/src/resources/custom-element.ts
Sayan751: IMO it just bring more flexibility that's all.

```ts
export class MyElement {
  @processContent()
  static processContent(node: INode, platform: IPlatform) { }
}
```

Also when used with TypeScript, it will throw compilation error if the method-level decorator is used on a non-static method or even on a static property for that matter.

However, I have not much strong opinion on this. I am even happy with the following syntax 😄 

```ts
@customElement({processContent: MyElement .processContent})
export class MyElement {
  static processContent(node: INode, platform: IPlatform) { }
}
```

But I think not everybody will use the `@customElement` deco, especially the one relying on the convention plugin. Thus the class-level `@processContent(func: (node: INode, platform: IPlatform) => boolean | void)` decorator is required.

Let me know if you find the method-level decorator too superfluous. I can kill it as well. Things stays simpler.

bigopon: I think generally, higher level API should be limited first. But I may be inconsistent with my own self here.
Probably let @fkleuver make the call

fkleuver: I like `static processContent` but I'm not sure why the decorator is even necessary then. Other decorator-level annotations are also auto-discovered when the static property exists.

Sayan751: @fkleuver Do you mean that we can kill the decorators altogether? And treat `static processContent` like a hook (convention)? That certainly makes things simpler. I was primarily focusing on restoring Au1 behavior with the decorators, but I think a conventional hook works as well.

PR #1070, opened 2020-12-23
Build/roll up

bigopon:
# Pull Request

## 📖 Description

Use rollup for all of our builds, so that it will produce a single dist file. Also simplify the circular dependencies story, if we manage to keep it under control

### DIFF
--- a/packages-cjs/aot/package.json
+++ b/packages-cjs/aot/package.json
@@ -33,7 +33,8 @@
     "dev": "tsc -b -w --preserveWatchOutput",
     "test": "node dist/cjs/test.js",
     "publish:dev": "npm publish --tag dev",
-    "publish:latest": "npm publish --tag latest"
+    "publish:latest": "npm publish --tag latest",
+    "rollup": "rollup -c"
   },
   "publishConfig": {
     "access": "public"
@@ -47,6 +48,8 @@
     "typescript": "^4.0.3"
   },
   "devDependencies": {
-    "@types/node": "^14.11.5"
+    "@rollup/plugin-typescript": "^8.1.0",
+    "@types/node": "^14.11.5",
+    "rollup": "^2.35.1"
   }
 }
--- a/packages-cjs/aot/rollup.config.js
+++ b/packages-cjs/aot/rollup.config.js
@@ -1,0 +1,15 @@
+import typescript from '@rollup/plugin-typescript'
+
+export default {
+  input: 'src/index.ts',
+  output: {
+    file: 'dist/bundle/index.js',
+    format: 'cjs',
+    sourcemap: true
+  },
+  plugins: [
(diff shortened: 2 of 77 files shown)


bigopon:
For multi module format dist (of core esm packages), we can have either:
(1) 
```
dist
    csjs
        index.js
        index.js.map
    esm
        index.js
        index.js.map
    types
        all typings here
```
or if we want something simple:
(2)
```
dist
    index.cjs.js
    index.cjs.js.map
    index.es.js
    index.es.js.map
    types
        all typings here
```

I'd prefer (2), but maybe (1) isn't too bad, as it's simpler for future changes

3cp:
`packages-cjs/` can be moved back to `packages/` again.
With rollup, we are getting rid of those build patch on tsconfig, one rollup config can build both cjs and esm dist files, so that separate `packages-cjs/` is not necessary anymore.
With rollup, the package build is controlled per package, separate `packages-cjs/` is irrelevant.

Also, as I suggested for the short term fix of the nodejs cjs/esm inter-importing issue: move back all packages to ship cjs as main.

For all packages, we can use following in package.json. It should bypass all those cjs/esm inter-importing mess (due to the lovely design of Nodejs's esm support).
```js
"main": "dist/index.cjs.js", // used by nodejs
"module": "dist/index.esm.js", // used by webpack and all other bundlers
// remove "type": "module",
```

bigopon:
I kind of like the current separation, or almost a division between main packages (runtime) and tooling packages (plugins/aot).
I think we can divide our packages that way, so that it doesn't feel too overwhelming inside `packages` folder

fkleuver:
Let's do this post-alpha when the release process has stabilized a bit more

Sayan751:
Looks good to me. A quick question: can we actually have 2 rollup configs in root; one for cjs, and another for esm? Instead of those config files for every package? Also as those files export a JS module, every package can apply its own customization if needed. Not a pressing issue, mostly a personal curiosity.

3cp:
> Looks good to me. A quick question: can we actually have 2 rollup configs in root; one for cjs, and another for esm? Instead of those config files for every package? Also as those files export a JS module, every package can apply its own customization if needed. Not a pressing issue, mostly a personal curiosity.

Rollup config is just js code. Should be able to save base config in root, then import and customise in individual package.

3cp:
Another note: there should be 4 dist files.
```
dist/index.es2020.cjs.js
dist/index.es2020.esm.js
dist/index.es2015.cjs.js // maybe even down to es5
dist/index.es2015.esm.js
```

3cp:
Note packed dist file with rollup will greatly speed up Aurelia 2 on gist.dumber.app.

bigopon:
superseded by #1170

PR #1077, opened 2021-01-04
refactor(prop-binding): remove necessity for id stamping infra

bigopon:
# Pull Request

## 📖 Description

This removes the conflation of target & source change handling, so that we can remove the flags updateTarget & updateSource, as there's no longer a need to differentiate them.

#### Debounce & Throttle changes:
- Only debounce & throttle values coming from view -> view model (v1 behavior is different: apply in 1 way binding, or only view -> view model in 2 way binding)
- discard the incoming values from view -> view model, if view model is changed while queued
- Added more tests for `debounce` & `throttle` binding behaviors

## Next steps

Remove `updateTarget` & `updateSource` flags & their references.


packages/runtime-html/src/binding-behaviors/debounce.ts
```diff
@@ -4,6 +4,14 @@ import { bindingBehavior, BindingInterceptor, LifecycleFlags } from '@aurelia/ru
 import type { ITask, QueueTaskOptions, TaskQueue } from '@aurelia/kernel';
 import type { BindingBehaviorExpression, IInterceptableBinding, IsAssign, Scope } from '@aurelia/runtime';
 
+const defaultDelay = 200;
+
+//
+// A binding behavior that prevents
+// - (v1 + v2) the view-model from being updated in two-way binding, OR
+// - (v1) the the view from being updated in to-view binding,
```
bigopon, line 12: Note that this may eventually be in v2, though we need to define clearer how it plays with the task queue. Probably to be decided before v2 finalizes for beta

PR #1083, opened 2021-01-11
fix(missing): resource.d.ts exposing incorrect

brandonseydel:
# Pull Request

## 📖 Description

Exposes partial bindable for use in the resource.d.ts file from template so it is now correct.
```ts
declare module '*.html' {
  import { IContainer, PartialBindableDefinition } from 'aurelia';
  export const name: string;
  export const template: string;
  export default template;
  export const dependencies: string[];
  export const containerless: boolean | undefined;
  export const bindables: Record<string, PartialBindableDefinition>;
  export const shadowOptions: { mode: 'open' | 'closed' } | undefined;
  export function register(container: IContainer);
}

declare module '*.css';
declare module '*.scss';
```

## ⏭ Next Steps

Update the template for aurelia to utilize

### DIFF
--- a/packages/aurelia/src/index.ts
+++ b/packages/aurelia/src/index.ts
@@ -528,7 +528,7 @@ export {
   // QueueTaskTargetOptions,
 
   bindable,
-  // PartialBindableDefinition,
+  PartialBindableDefinition,
   // BindableDefinition,
   Bindable,
 


brandonseydel:
Other repo PR waiting this if applied.
https://github.com/aurelia/new/pull/43

PR #1087, opened 2021-01-27
refactor(i18n): batch updates in changes

bigopon:
# Pull Request

## 📖 Description

At the moment, `i18n` translation binding is not batching its update, which is different with the rest of the html runtime. This PR aligns the behavior of `i18n` bindings.


packages/__tests__/i18n/t/translation-integration.spec.ts
```diff
@@ -239,8 +245,11 @@ describe('translation-integration', function () {
         this.undef = undefined;
       }
     }
-    $it('works if the keyExpression is changed to null/undefined - default value', function ({ host, app }: I18nIntegrationTestContext<App>) {
+    $it.only('works if the keyExpression is changed to null/undefined - default value', function ({ host, app, ctx }: I18nIntegrationTestContext<App>) {
```
Sayan751, line 248:
```suggestion
    $it('works if the keyExpression is changed to null/undefined - default value', function ({ host, app, ctx }: I18nIntegrationTestContext<App>) {
```

PR #1095, opened 2021-02-07
feat(inline-html): Allow inline views with convention

brandonseydel:
# Pull Request

## 📖 Description

This enables conventions to work with decorators for inlining HTML.  This will also lay some groundwork for having a default inline-view strategy which could be useful for tools like snowpack.

```ts
@template('<div>Inner Test</div>')
class InnerSimple { }

// second arg is a spread so can be used single as well
@template('<div><inner></inner></div>', [CustomElement.define('inner', InnerSimple)])
class Simple {
}
```
Or
```ts
@inlineView('<div>Inner Test</div>')
class InnerSimple { }
// second arg is a spread so can be used single as well
@inlineView('<div><inner></inner></div>', [CustomElement.define('inner', InnerSimple)])
class Simple {
}
```


Vheissu:
As discussed in Discord @brandonseydel if you want to bring your repository up to date and ensure tests are passing, we should get this one merged in after @bigopon casts his eyes over it. This change negates the need to use the `static template` property or redeclaring a component using `customElement` to change the template.

Issue #1103, opened 2021-02-14
Cannot use as-element with au-slot


MaximBalaganskiy:
# 🐛 Bug Report

Exception is thrown when `as-element` is used with `au-slot`

## 🤔 Expected Behavior

Projection should work

## 😯 Current Behavior

Exception is thrown
```
Uncaught (in promise) Error: Unsupported usage of [au-slot="default"]. It seems that projection is attempted, but not for a custom element.
    at TemplateBinder.bindAttributes (template-binder.js:330)
    at TemplateBinder.bindManifest (template-binder.js:193)
    at TemplateBinder.bindChildNodes (template-binder.js:382)
    at TemplateBinder.bindManifest (template-binder.js:184)
    at TemplateBinder.bindChildNodes (template-binder.js:382)
    at TemplateBinder.bindManifest (template-binder.js:184)
    at TemplateBinder.bindChildNodes (template-binder.js:382)
    at TemplateBinder.bindManifest (template-binder.js:184)
    at TemplateBinder.bindChildNodes (template-binder.js:382)
    at TemplateBinder.bindManifest (template-binder.js:184)
```

## 💁 Possible Solution
Replace
https://github.com/aurelia/aurelia/blob/2358948fbf2b82b3b37f993fa10bdd69b82769d9/packages/runtime-html/src/template-binder.ts#L392
with
```ts
const parentName = node.parentNode?.hasAttribute && node.parentNode?.hasAttribute('as-element') 
  ?  node.parentNode?.getAttribute('as-element')
  : node.parentNode?.nodeName.toLowerCase();
```

## 🔦 Context

There is a au-slotted element `mdc-icon-button` which needs to be used like 
```
<button as-element="mdc-icon-button">
  <template au-slot>
    <!-- content -->
  </template>
</button>
```

## 💻 Code Sample

https://gist.dumber.app/?gist=c171c022d726e1880cfa3cb7dff0885d

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Aurelia          | 0.10.0-dev.202102110248
| Language         | Typescript
| Browser          | Chrome
| Bundler          | Webpack
| Operating System | Windows 10

fkleuver:
Thanks for reporting this. I believe this should be fixed now with your fix (thanks for that btw!) right?

MaximBalaganskiy:
Yes, that's fixed

PR #1104, opened 2021-02-14
fix(template-binder): check if as-element was used

MaximBalaganskiy:
# Pull Request

## 📖 Description

Slotted elements might be used with `as-element` attribute.
The correct parent name allows for the proper usage check.

### 🎫 Issues

#1103 

## 📑 Test Plan

The test makes sure that using `as-element` with slots does not throw.


fkleuver:
Nice find, and thanks for the test + fix!

PR #1106, opened 2021-02-15
fix(observer-locator): consider as-element for events config

MaximBalaganskiy:
# Pull Request

## 📖 Description

Consider a custom element with a value observer configured to listen to `change` and `input` events.
When this element is used via `as-element` attribute the correct event value observer is not created.
This fix uses `as-element` value to get the actual node name.

## 👩‍💻 Reviewer Notes

MDC framework requires text fields to be `label` elements. Therefore the only correct usage is 
```html
<label as-element="mdc-text-field" value.bind="value"></label>
```
Without this fix value changes are not passed to bindings.

## 📑 Test Plan
The test makes sure that a legitimate use case is supported.

### DIFF
--- a/packages/__tests__/3-runtime-html/observer-locator.spec.ts
+++ b/packages/__tests__/3-runtime-html/observer-locator.spec.ts
@@ -173,6 +173,7 @@ describe('ObserverLocator', function () {
     { markup: `<div css="color:green;"></div>`, ctor: StyleAttributeAccessor },
     { markup: `<select value=""></select>`, ctor: SelectValueObserver },
     { markup: `<input value=""></input>`, ctor: ValueAttributeObserver },
+    { markup: `<div value="" as-element="input"></div>`, ctor: ValueAttributeObserver },
     { markup: `<input checked="true"></input>`, ctor: CheckedObserver },
     { markup: `<input files=""></input>`, ctor: ValueAttributeObserver },
     { markup: `<textarea value=""></textarea>`, ctor: ValueAttributeObserver },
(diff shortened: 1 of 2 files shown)


bigopon:
The fix is ok by itself, though I think we need to consider a bit, since the node observer shouldn't take into account `as-element` attribute, it should just work with just node primitives like tag name & attribute/properties. If we need to do this, maybe we can change the way config is registered, from look up into a method, or add an extra method to support super dynamic situation, additionally to look up.

MaximBalaganskiy:
This would be beyond my limited knowledge of the package I'm afraid.

fkleuver:
> the node observer shouldn't take into account `as-element` attribute,

Why not?

bigopon:
`as-element` is aurelia specific, taking it into account breaks the current contract of the node observer locator: observing ordinary elements & their properties, via the tag name and property combo.

 If `as-element` is to be changed as this, what of the following:
```html
<input value.bind="value" as-element="my-input">
```
Common expectation: everything works as is.
Uncommon expectation: needs to register how to observe a `value` property for a `MY-INPUT`

If we do that, should we handle `is` of web-components as well? since it's the same:
```html
<label is="mdc-text-field">
```

fkleuver:
> <input value.bind="value" as-element="my-input">

What would be the use case for doing this?

MaximBalaganskiy:
So, may be accept this one until you come up with a better idea? It's a real showstopper...

MaximBalaganskiy:
What about monkey patching `tagName`? Is this going to have any side effects?

fkleuver:
@bigopon Are there any major reasons to not merge this right now?

bigopon:
I think we shouldn't conflate Aurelia custom elements and normal HTML elements. Allowing `as-element` to be used for this purpose triggers that. What about we make it recognize a special attribute? `data-tagname` for example if we really wanna hack it this way?

MaximBalaganskiy:
Monkey patching tagName worked just fine. You may close this one if not keen on it

Vheissu:
@bigopon Are we still of the opinion that this PR deviates from how we want this behaviour to work? If so, it should be closed.

Issue #1111, opened 2021-02-15
Break the internationalization.md docs into sub-articles


Seebiscuit:
The internationalization.md doc is lengthy and could use better organization. I have two suggestions on that note:

1. The trivial use case of how to use internationalization in view-model comes very later in the article. This should be moved closer to the top. 
2. Date formatting should be split out into its own article.

Ideally, we'd split the document up into an 1) Intro with quick instructions on getting started 2) Translation usage in depth and advanced cases and 3) Date formatting. This implies a bit of a reworking of the translation usage which, IMO, does not clarify some fundamentals (least clear is the `t` attribute usage, and how locale file should be formatted, or at least where to look for that information).

Sayan751:
Thank you for working on the i18n docs! I wrote those long back, and never revisited :P 
Feel free to restructure that.

Vheissu:
Work on internationalisation efforts has been done, and the sections are clearer in their use.

PR #1112, opened 2021-02-15
docs: Improves grammar and small flow changes for internationalization.md

Seebiscuit:
# Pull Request

## 📖 Description

Small grammar and flow improvements to **docs/user-docs/app-basics/securing-an-app.md**.

Part of a series of first PR to aurelia, for this author.

### 🎫 Issues

Linking #1111 to be completed after this is approved. 

## 👩‍💻 Reviewer Notes

Please, see related issue. This may be only the first of a series of updates for this topic. 

And, apologies for the rather large diff. I couldn't think of an obvious way to split this up, other than arbitrarily cutting this off at the middle and splintering the review, which is probably less than optimal. So...here we are 😅 

## 📑 Test Plan

### DIFF
--- a/docs/user-docs/app-basics/internationalization.md
+++ b/docs/user-docs/app-basics/internationalization.md
@@ -4,21 +4,21 @@
 
 This documentation explains how to get up and running with Aurelia-I18N in order to provide localization \(l10n\) and internationalization \(i18n\) features for your app.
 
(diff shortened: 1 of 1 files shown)


Vheissu:
These all look really good to me. I don't think this section has been touched since it was written yonks ago.

fkleuver:
Agree, much better. Thanks a ton for taking the time to look into this!

Issue #979, opened 2020-10-17
[RFC]: async/pending/fulfilled/rejected template controllers


Sayan751:
# 💬 RFC

This RFC proposes a way to work with promises directly in the markup in a declarative way. 

## 🔦 Context

It is a common use case that before showing a view, data for that view is asynchronously fetched. While that promise can be returned from the lifecycle hooks, and Aurelia will wait for that promise, it necessarily delays (partial) rendering of the view. Sometimes this also means the delay of visual cues to the end-user. In some cases, such delay might be justified where the view is made solely to present the data being fetched. 

In other cases, it might not; and as developer I want to display the subset of data that has already been fetched while *some parts* of the view wait for some specific promises, still not settled. I am of this opinion that this pattern offers better UX.

To this end, the idea is to have a set of strongly coupled template controllers named, `async`, `pending`, `fulfilled`, and `rejected`. 'Strongly coupled' behause:
-  `pending`, `fulfilled`, or `rejected` should not work in isolation, and 
- `async` without at least one of those might not make much sense. That is possibly an edge case especially in context of working with promises in view.

Following are some examples to describe the idea.

## 💻 Examples

**Example#1:** The basic use-case

```html
<template async.bind="answerPromise">
  <span pending> Still computing... 💻 </span>
  <span fulfilled> 42 🎉 </span>
  <span rejected> What was question? 😡 </span>
</template>
```

Although it is self explanatory, for completeness here goes the short explanation. The view of the `pending` template controller will be shown while the promise is running. Once it is settled, then either the `fulfilled` or the `rejected` view will be shown depending on whether the promise was resolved or rejected, respectively.

**Example#2:** Accessing the resolved data or rejection reason

```html
<template async.bind="answerPromise">
  <span pending> Still computing... 💻 </span>
  <span fulfilled.from-view="resolvedData"> ${resolvedData} 🎉 </span>
  <span rejected.from-view="reason"> ${reason.message} 😡 </span>
</template>
```

Note that the usage of `from-view` is intentional as binding in the other direction is not very useful IMO. Consequently, the binding mode for `async` should always be `to-view`.

**Example#3:** Optional `pending`/`fulfilled`/`rejected`

These three template controllers should be optional. This means that following examples should work.

```html
<template async.bind="promise1">
  <span pending> The promise is still running, and I will vanish once it settles. Who knows how it is settled! 🤔 </span>
</template>

<template async.bind="promise2">
[...]

PR #1125, opened 2021-02-28
feat: promise tc

Sayan751:
# Pull Request

## 📖 Description

Covers basic use case of #979. The syntax is as follows.

```html
<template promise.resolve="promise">
 <template pending>The promise is not yet settled.</template>
 <template then="data">The promise is resolved with ${data}.</template>
 <template catch="err">This promise is rejected with ${err.message}.</template>
</template>
```

Please refer the docs, and tests for more examples.

Considering we would like to wait on the user feeedback, the implementation of the `promise.all`, `promise.race`, and `promise.any` is postponed. Hence the PR can close the #979.


AureliaEffect:
The benchmarking comparison can be found here: https://au2-bench-viewer.azurewebsites.net/compare?branch=master&commit=30688c25e48d259bead6621b11e3216f539cc391&branch=topic/979-promise-tc&commit=80d5b0ccec9bd07d3a3cefcee8f8aa043007981e

AureliaEffect:
The benchmarking comparison can be found here: https://au2-bench-viewer.azurewebsites.net/compare?branch=master&commit=30688c25e48d259bead6621b11e3216f539cc391&branch=topic/979-promise-tc&commit=7a1a2bf6143c8f2059aa232a78767972fc604fe3

packages/runtime-html/src/resources/template-controllers/promise.ts
```diff
@@ -0,0 +1,337 @@
...
+
+  private swap(initiator: IHydratedController | null, flags: LifecycleFlags): void {
+    const value = this.value;
+    if (!(value instanceof Promise)) { return; }
+    const q = this.platform.domWriteQueue;
+    const fulfilled = this.fulfilled;
+    const rejected = this.rejected;
+    const pending = this.pending;
+    const $controller = this.$controller;
+    const s = $controller.scope;
```
bigopon, line 77: I think ideally `promise` controller should create its own scope, so that assignment from `then` & `catch` won't leak unexpectedly when there's no corresponding properties defined. If the view inside `then`/`catch` would wants to access the parent `data`/`err`, they can, as expected, use `$parent`.

A hypothetical scenario:
- Load some user data
- upon err, retry
- When retrying, show message `Retrying...`
  - If continue to fail, show `Giving up. Error loading with ${previous_error_msg}, then ${current_err_msg}`
  - If succeed, show the data as specified in parent promise `then`

It would look like this, if we had natural scope hierarchy, with `$parent` :

```html
<template promise.resolve="getUser()">
  <div then="data">${data}</div>
  <div catch="err">
    <template promise.resolve="getUser(someValue, isRetry: true)">
      <div pending>Retrying...</div>
      <template then.bind="$parent.data"></template>
      <div catch="err">Giving up, Error loading with ${$parent.err} then ${err}</div>
    </template>
  </div>
</template>
```

packages/runtime-html/src/resources/template-controllers/promise.ts
```diff
@@ -0,0 +1,337 @@
...
+
+@attributePattern({ pattern: 'promise.resolve', symbols: '' })
+export class PromiseAttributePattern {
+  public 'promise.resolve'(name: string, value: string, _parts: string[]): AttrSyntax {
+    return new AttrSyntax(name, value, 'promise', 'bind');
+  }
+}
+
+@attributePattern({ pattern: 'then', symbols: '' })
+export class FulfilledAttributePattern {
```
bigopon, line 326: I'm a bit confused by this, and its default to `from-view`, can you help elaborate a bit? Also, do we need this if it's not `from-view`?

bigopon (commented):
Looks great! Left some comment for the first round

packages/runtime-html/src/resources/template-controllers/promise.ts
```diff
@@ -0,0 +1,337 @@
...
+  }
+
+  public valueChanged(_newValue: boolean, _oldValue: boolean, flags: LifecycleFlags): void {
+    if (!this.$controller.isActive) { return; }
+    this.swap(null, flags);
+  }
+
+  private swap(initiator: IHydratedController | null, flags: LifecycleFlags): void {
+    const value = this.value;
+    if (!(value instanceof Promise)) { return; }
```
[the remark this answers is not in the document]
Sayan751: I think we should never try to "normalize it to a promise". I thought of throwing error, but that might be too much; although IMO that should also be quite acceptable solution. I think a simple warning level log should suffice in this case. IMO it is much better than assuming something in order to normalize the value to a promise.

packages/runtime-html/src/resources/template-controllers/promise.ts
Sayan751: I sincerely think this sort of retrying is better suited for view-model. Creating a new scope from `promise` is not giving me a good feeling. The reason being that none of these template-controllers have any contextual bindable property that can be benefitted from an extra scope. My recommendation would be to use a different variable name to avoid overwrite. From scope traversal perspective, it also makes sense not to create a new scope from `promise`.

packages/runtime-html/src/resources/template-controllers/promise.ts
Sayan751: The 3 attribute patterns, `promise.resolve`, `then`, and `catch` are created to reduce verbosity in templates. As the binding modes, other than `from-view`, do not make sense in conjunction with `then` and `catch`, the `then` and `catch` attribute pattern saves the user of writing `then.from-view` and `catch.from-view` explicitly.

Note that using a bound property for `then` or `catch` is completely optional. For example, if you dealing with promises in view-model, then almost always also have the access to the resolved data or the caught error. The promise-controllers expose these values via the bindable value, because they are nice 😃

docs/user-docs/getting-started/conditional-rendering.md
```diff
@@ -405,3 +405,136 @@ This section includes few more interesting examples that you might encounter in
   ```
   {% endcode %}
 
+## promise.resolve
+
+The `promise.resolve` template controller enables us to render different content based on the status of a `promise`.
+The most basic example looks like as follows.
+
+```markup
+<template promise.resolve="promise">
```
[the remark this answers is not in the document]
Sayan751: Good point! Updated the docs, and added tests.

packages/runtime-html/src/resources/template-controllers/promise.ts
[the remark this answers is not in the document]
bigopon: I think in practice, a normal user will do this:
```ts
const value = await valueOrPromise
```
to normalize it. With that habit, that same user that `valueOrPromise` is gonna be the value of the `promise` controller, and will wonder why the UI doesn't seem to response. This can be easily hit by having a value dynamically updated by something. Though we can probably delay it

packages/runtime-html/src/resources/template-controllers/promise.ts
bigopon: In JS, we do
```ts
promise
  .then(data => {
    // do something with local data err
  })
  .catch(err => {
    // do something with local var err
  })
```
We do not do:
```ts
promise
  .then(data => {
    this.data = data;
  })
  .then(() => {
    // do something with data on this
  })
```
Having a local scope for promise controller will be sort of like the first example, I think it's also helps that people don't accidentally leak the values into "this"

packages/runtime-html/src/resources/template-controllers/promise.ts
bigopon: I see, I think we need a way to let CA participate into the compilation phase so that it can influence the usage on its own. Feels more like a complete, isolated, independent "package" to me. That way we can reduce the configuration points one need to bring in in order to have it work cc @fkleuver 
For `from-view`, I guess if we have local scope for `promise`, I got that now. Though we probably need a way to make this less confusing, I think the target of those binding should be the local scope of the promise, probably will be simpler to reason about time to time.

packages/runtime-html/src/resources/template-controllers/promise.ts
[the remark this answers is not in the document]
Sayan751: As we are not using `async`/`await` for the OOTB template controllers, I would refrain from doing that here as well. Also bringing `async`/`await` to picture might have conflicting result with current behavior. I would wait till there is any severe shortcoming with not normalizing the value to a promise.

AureliaEffect:
The benchmarking comparison can be found here: https://au2-bench-viewer.azurewebsites.net/compare?branch=master&commit=10f1fd554c7a5b44789a4b61c850bd804ac6f59d&branch=topic/979-promise-tc&commit=645fcd1a93852a8c5a80b508d88e7a49bbfa68b2

packages/runtime-html/src/resources/template-controllers/promise.ts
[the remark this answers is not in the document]
bigopon: I think i figured out whats im confused by, and not describe well. Maybe when the value is not a promise, immediately trigger the then view? Its not really async await, but not rejecting non promise values either. Win win or still no?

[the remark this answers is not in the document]
Sayan751: The problem is with the values like `null`, `undefined`, or an instance of `Error`. I am not so sure what is exactly expected in those cases. I fear that I am coming out as adamantly opposing your idea, but TBH the "assumptions" we have to make to deal with those edge cases are making me uncomfortable.

BTW, if the promise is already settled (resolved or rejected), then the associated view activation is done as fast as possible. If you look closely at the `PromiseTemplateController#swap`, it tries to do the activation as fast as possible, at the same time taking care of that the view controllers themselves do not end up in inconsistent state.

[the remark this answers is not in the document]
bigopon: Its a healthy convo. I think its also good that we start with a restricted set of input and APIs. I think always ensure its a promise is a good thing to start with.

packages/runtime-html/src/resources/template-controllers/promise.ts
Sayan751: You do have a point there. And I like that argumentation of JS scope! Having a synthetic (created from parent) scope in Promise TC will mean that we can now write this:

```diff
  <template promise.resolve="42 | promisify:true">
-   <fulfilled-host then.from-view="data1" data.bind="data1"></fulfilled-host>
-   <rejected-host catch.from-view="err1" err.bind="err1"></rejected-host>
+   <fulfilled-host then.from-view="data" data.bind="data"></fulfilled-host>
+   <rejected-host catch.from-view="err" err.bind="err"></rejected-host>
  </template>
  <template promise.resolve="'forty-two' | promisify:false">
-  <fulfilled-host then.from-view="data2" data.bind="data2"></fulfilled-host>
-  <rejected-host catch.from-view="err2" err.bind="err2"></rejected-host>
+  <fulfilled-host then.from-view="data" data.bind="data"></fulfilled-host>
+  <rejected-host catch.from-view="err" err.bind="err"></rejected-host>
  </template>
```

Now this example sort of also portrays a potential point of confusion. And that is if or not `then.from-view="data"` in both places makes it intuitive or more confusing. In the example above, we are dealing with 2 different promises, thus expecting the 2 bound `data` variables are different seems bit confusing to me, especially when this template controller does not inherently offer any contextual properties like `repeat.for`.

Contextually, I would like to elaborate one of my previous point regarding the scope traversal. IMO one of the frequent usage of this template controller will be as follows.

```html
<template promise.bind="promise">
  <template then>
    <!-- Here goes the template involving the resolved data, but manipulated/adapted from VM -->
  </template>
  <template catch>
    <!-- Here goes the template involving the error reason, but manipulated/adapted from VM -->
  </template>
</template>
```

IMO often we need to adapt the raw data resolved via promise. A prime example might be the typical use-case of `fetch`ing data. Which means that VM will have it's access on the resolved data or the rejection error, rather than grabbing those from the view. Now in that case, having another scope for the promise tc would mean that scope traversal need to go one more step before resolving those properties which are already available from "this".

Note that I am not opposing the idea of having separate scope (on the contrary, as I said before I like your JS scope argumentation, and I have even tested it 😃), I am merely pointing out potential issues with the idea.

packages/__tests__/3-runtime-html/promise.spec.ts
```diff
@@ -0,0 +1,2780 @@
...
+  assert,
+  TestContext,
+} from '@aurelia/testing';
+import {
+  createSpecFunction,
+  TestExecutionContext,
+  TestFunction,
+} from '../util.js';
+
+describe.only('promise template-controller', function () {
```
bigopon, line 48: ❌

packages/runtime-html/src/resources/template-controllers/promise.ts
bigopon: With a separate scope for the `promise` TC, we could support both local `data`:
```html
<div then.bind="data">
```
 and `this.data`:
```html
<div then.bind="$this.data">
```
 but without a scope for it, we cannot. Not to mention, this could also be one of the common scenarios:
```html
<div repeat.for="item of items">
  <template promise.bind="getSummary(item)">
    <div then.bind="data">local data of the item is: ${data}
```

packages/__tests__/3-runtime-html/promise.spec.ts
Sayan751: Removed. Thanks! 😁

packages/runtime-html/src/resources/template-controllers/promise.ts
Sayan751: That's a good point. Making the switch.

AureliaEffect:
The benchmarking comparison can be found here: https://au2-bench-viewer.azurewebsites.net/compare?branch=master&commit=28839cbc3d57dac47c884dab9e0014a12d1aa46c&branch=topic/979-promise-tc&commit=a22e8f35d682382d1d7edff355f3c8e5c45b8591

AureliaEffect:
The benchmarking comparison can be found here: https://au2-bench-viewer.azurewebsites.net/compare?branch=master&commit=0a2d77e206923bcabe43486a3d847f51569b9663&branch=topic/979-promise-tc&commit=c6660c0304fb54c889196e69ac375c8b62aa0f57

AureliaEffect:
The benchmarking comparison can be found here: https://au2-bench-viewer.azurewebsites.net/compare?branch=master&commit=0a2d77e206923bcabe43486a3d847f51569b9663&branch=topic/979-promise-tc&commit=e56aeb365358b453fa836b47b0ea9ef15338a4a4

AureliaEffect:
The benchmarking comparison can be found here: https://au2-bench-viewer.azurewebsites.net/compare?branch=master&commit=0a2d77e206923bcabe43486a3d847f51569b9663&branch=topic/979-promise-tc&commit=2b74972d4d44c5aef9441b0b5a3be3cccc6a216e

AureliaEffect:
The benchmarking comparison can be found here: https://au2-bench-viewer.azurewebsites.net/compare?branch=master&commit=0a2d77e206923bcabe43486a3d847f51569b9663&branch=topic/979-promise-tc&commit=e0e87f709fafde179449bd116d2e70b93e12c3f6

AureliaEffect:
The benchmarking comparison can be found here: https://au2-bench-viewer.azurewebsites.net/compare?branch=master&commit=0a2d77e206923bcabe43486a3d847f51569b9663&branch=topic/979-promise-tc&commit=ed81650ae0157032e4430d2a0690f1beac9afd3e

AureliaEffect:
The benchmarking comparison can be found here: https://au2-bench-viewer.azurewebsites.net/compare?branch=master&commit=567344539626a7d206f1d2f58c78415407e769bd&branch=topic/979-promise-tc&commit=25f6d8b1e56d3d1c6c6a088083844fcb8ee79612

docs/user-docs/app-basics/scope-and-binding-context.md
```diff
@@ -0,0 +1,949 @@
...
+})
+export class ShowOutput {
+  @bindable public value: number;
+}
+
+@customElement({
+  name: 'app',
+  template: `<div repeat.for="i of 3">
+  <raise-two input.bind="i+1" output.from-view="op"></raise-two>
+  <show-output value.bind="op"></show-output>
```
bigopon, line 769: Do we need a CE for this? What about the standard `<output/>`:
```html
<output value.bind="...">
```

docs/user-docs/app-basics/scope-and-binding-context.md
```diff
@@ -0,0 +1,949 @@
...
+
+## Context selection
+
+So far we have seen various aspect of scope, binding and override context.
+One thing we have not addressed so far is how the contexts are selected for expression evaluation or assignment.
+In this section we will look into that aspect.
+
+The context selection process can be summed up (simplified) as follows.
+
+1. IF `$parent` keyword is used once or several times, THEN
```
bigopon, line 591: once or more? I'm not native speaker, just feels like this reads like that it cannot be used more than several times.

packages/runtime/src/observation/binding-context.ts
```diff
@@ -121,7 +136,8 @@ function chooseContext(scope: Scope, name: string, ancestor: number): IBindingCo
   }
 
   if (overrideContext) {
-    return name in overrideContext ? overrideContext : overrideContext.bindingContext;
+    const bc = overrideContext.bindingContext;
+    return (name in overrideContext || (fromView && !(name in bc))) ? overrideContext : bc;
```
bigopon, line 140: I don't think we can afford this changes. It breaks the basic scenario like this:
```ts
<input value.from-view="message">
```
targeting override context in this example is a bit weird too: from-view binding resulting into the value coming to view (override context).
We should evaluate this in terms of intrusiveness as well.

packages/runtime-html/src/resources/template-controllers/promise.ts
```diff
@@ -0,0 +1,346 @@
...
+  ): void {
+    this.view = this.factory.create(flags, this.$controller).setLocation(this.location);
+  }
+
+  public attaching(initiator: IHydratedController, parent: IHydratedParentController, flags: LifecycleFlags): void | Promise<void> {
+    const view = this.view;
+    const $controller = this.$controller;
+
+    return onResolve(
+      view.activate(initiator, $controller, flags, this.viewScope = Scope.fromParent($controller.scope, {}), $controller.hostScope),
```
bigopon, line 65: I'm a bit confused, so are we going with a separate scope or without a separate scope for the promise? 😓

docs/user-docs/app-basics/scope-and-binding-context.md
```diff
@@ -0,0 +1,380 @@
...
+|  | overrideContext |  |
+|  |                 |  |
+|  +-----------------+  |
+|                       |
++-----------------------+
+```
+
+### Motivation
+
+If you are thinking 'Why do we need override context at all?', let me tell you that it is a great question.
```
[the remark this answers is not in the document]
Sayan751: That's a great idea! Invoking the doc wizard @Vheissu!

docs/user-docs/app-basics/scope-and-binding-context.md
```diff
@@ -0,0 +1,949 @@
...
+  }
+}
+```
+{% endcode %}
+
+This makes the binding context readily available, even when we working with override context, without necessarily traverse via the scope.
+With this information, we can again change our diagram to the following one.
+
+```text
++---------------------------------+
```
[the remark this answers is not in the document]
Sayan751: Nothing special. In my head it is natural to have the outward connection from right and the inward connection from left. That's it :)

docs/user-docs/app-basics/scope-and-binding-context.md
Sayan751: I think this can be changed to 'once or more than once' (removing the 'times').

docs/user-docs/app-basics/scope-and-binding-context.md
Sayan751: Good point! I seldom use that. Haven't thought about that. Worth a shot.

packages/runtime/src/observation/binding-context.ts
Sayan751: IMO this is alright, as long as this is a view-only property. As long as you are initializing that property in your view-model (anytime before binding) context selector will choose binding context, in this case (or mostly) the view-model. And in that case, you can read/write the value bound from view also in view-model. If you don't initialize the property in view model, then the property remains in the override context without polluting the view model. 

I wrote the scope and context doc specifically to highlight this point. Clearly, that point didn't come out from the doc as clearly as I thought. IMO this change makes sense. Thus, I will ask if we can keep the change and write better docs, explaining that?

bigopon: This will break the consistency we have in our templating: working with binding context first and foremost. Override context will only receive something special when it's predefined. that way, someone can keep in the back of their mind that if something goes missing, just look around in the binding contexts. With this change, it's look at everything: template, syntax, binding context, override context etc.
if we want to prevent the value from leaking into the owning scope in promise scope, we can add a let binding before other bindings, with the same expression it received, so that everything is defined in advanced in the promise scope:
How it looks:
```html
<div promise.resolve="...">
  <div then="data">
```
How it's actually is:
```html
<div promise.resolve="...">
  <let bla_bla_data.bind="undefined"></let>
  <div then="bla_bla_data">
```
This way, our system is not flooded with special rules in so many places, and has a chance to stay simple while it still can, also a good use for our binding primitives.

fkleuver: I agree with the idea behind the solution but I'm inclined to think that we need a new primitive to deal with this.

What you're *trying* to do here is: apply a special rule for "view-only" bindings (which is fine).
But what you're *actually* doing is: apply a special rule for "from-view" bindings, which is broader than it needs to be.

How about we introduce a new binding mode `viewOnly` which has the exact same semantics as `fromView` but then with the addition of this specific rule?
Another use case that immediately comes to mind is:
```ts
<a active.class="$home" load="route: home; active.viewOnly: $home">
```

It does kind of makes sense to me to have this..

packages/runtime-html/src/resources/template-controllers/promise.ts
fkleuver: I don't disagree with you @bigopon but I think this is fine for the time being. Let's revisit this when time comes to refactor the compiler again, otherwise we risk introducing more tech debt with quick solutions

packages/runtime/src/observation/binding-context.ts
bigopon: Currently, for a promise TC in the template:
```html
<template>
  Some other part...
  <div promise.resolve="getData()">
    <div then="data">
    <div catch="err">
```
we have 2 scopes:
```html
<template> <!-- scope 1, the custom element -->
  Some other part...
  <div promise.resolve="getData()"> <!-- scope 2, created by the promise TC -->
    <div then="data">
    <div catch="err">
```
What we want:
- `data` and `err` flows to the **binding context** of scope 2.
What we have:
- `data` and `err` **may** flow into the **binding context** of scope 1, if there's a `data` and/or `err` properties defined there.

Introducing a `view-only` binding mode won't solve this, as we want binding context.

We can try to either:
- make `data` and `err` available via synthesizing let bindings for the `then` & `catch` expressions of the `promise` TC controller.
- introduce a way to instruct the ASTs to not traverse in `.assign()`

Sayan751: @fkleuver @bigopon  Just to clarify, we don't need the change in binding context selection for promise tc specifically. The binding of `data` and `err` works as expected without the said change as well. The change fixes another general issue.

```html
<template repeat.for="i of 5">
  <el-one output.from-view="out"></el-one>
  <el-two prop.bind="out"></el-two>
</template>
```

I packed the changes in this PR, because it supports couple of interesting test cases for promise tc.

I was not too happy with the change in the context selection process, which is why I think the view-only binding mode proposal from @fkleuver is very interesting. However, I think a binding mode should not be introduce lightly, as it would have further implications.

AureliaEffect:
The benchmarking comparison can be found here: https://au2-bench-viewer.azurewebsites.net/compare?branch=master&commit=d13dbf818ff557434f9cd13c71734320852e90cc&branch=topic/979-promise-tc&commit=c9945035933874c5f62d12c909cfff0155d8e0e3

packages/runtime/src/observation/binding-context.ts
bigopon: I'm not sure where your stance is, so I'll just emphasise it again: The issue with the repeat can still be worked out, as you mentioned, by:
```html
<template repeat.for="i of 5">
  <let output.bind=""></let>
  <el-one output.from-view="out"></el-one>
  <el-two prop.bind="out"></el-two>
</template>
```
This is easy and helps keep things consistent. We really can't afford another point of complexity in that context selection, especially when it comes with a extra parameter on the AST signature.

Sayan751: We have here 2 options. 

### Option 1: @bigopon's `let`-alternative

I like the idea because it is explicit. I think we can explain the behavior in the documentation. The new doc for scope and binding context is a perfect place for that.
Along with documentation, we might need more integration tests to iron-out any oddities in this alternative approach.
And with that we are done.

Another obvious advantage is that it reduces the complexity of the binding context selection process.

The only disadvantage of the alternative is that it is bit verbose, and non-intuitive. But that can be tackled with proper documentation.

### Option 2: My proposed change in the context selection process

In defense of the change I have 2 points. Firstly the resulting syntax is intuitive, and less verbose. Secondly, I think as binding mode affects the flow of data, it is only fair that it participates in the context selection process as well.

If it creates a breaking change, then I am also in support of having a new `view-only` binding mode as suggested by @fkleuver. 

However,  the changes in the context selection process makes it more complex, which is a complex process to start with even before the change. On that I agree with @bigopon that we don't need to make it more complex.

Therefore, I would like to go for @bigopon's alternative. But I will wait for @fkleuver's take on this before I revert the changes as that's a non-trivial activity.

AureliaEffect:
The benchmarking comparison can be found here: https://au2-bench-viewer.azurewebsites.net/compare?branch=master&commit=fc1e09fdf40db3cd46395debc903d493fbf46063&branch=topic/979-promise-tc&commit=eb0effe8ecaf9c47516a6558db60b248a7e7e95c

AureliaEffect:
The benchmarking comparison can be found here: https://au2-bench-viewer.azurewebsites.net/compare?branch=master&commit=d40f53103a5325c0767920f78366711e465316df&branch=topic/979-promise-tc&commit=06a918a7bee304be3f1dbed699eacca87f77b987

AureliaEffect:
The benchmarking comparison can be found here: https://au2-bench-viewer.azurewebsites.net/compare?branch=master&commit=d40f53103a5325c0767920f78366711e465316df&branch=topic/979-promise-tc&commit=85dc410fcded91204a639da5a1f589b93c294a4a

docs/user-docs/app-basics/scope-and-binding-context.md
```diff
@@ -0,0 +1,706 @@
...
+One way to think about expression and binding context is in terms of functions and binding those functions.
+Let us consider the following example.
+
+{% code title="foo.ts" %}
+```typescript
+function foo() { return this.a ** 2; }
+```
+{% endcode %}
+
+If we execute this function we will get `NaN`.
```
bigopon, line 70: This is a bit wrong, since there's no `this`. It will throw instead. Can you help reword this part?

docs/user-docs/getting-started/conditional-rendering.md
```diff
@@ -9,7 +9,7 @@ Have you ever needed to hide or show part of a UI based on some condition? Well,
 {% hint style="success" %}
 **Here's what you'll learn...**
 
-* Adding and removing DOM with `if`/`else`.
+* Adding and removing DOM with `if`/`else`, `switch`, and `promise.resolve`.
```
bigopon, line 12: Do we have any other choices beside `promise.resolve`? If not, I guess we can go with it, it's a bit weird having plain `if`/`else`/`switch`, and then `promise.resolve`

docs/user-docs/getting-started/conditional-rendering.md
```diff
@@ -405,3 +405,316 @@ This section includes few more interesting examples that you might encounter in
...
+ <template catch="err2">Cannot get the address information</template>
+</template>
+```
+
+In the example above the `span` will be shown independently of the status of either `generalInfoPromise`, or `addressInfoPromise`.
+The section dependent on each of those promise will attach DOM Elements according to the status of each promises, at the same time being independent of each other.
+
+### Scoping
+
+As `from-view` binding mode is used for `then` and `catch`, the `promise.resolve` template controller creates its own scope.
```
bigopon, line 496: > As `from-view` binding mode is used for `then` and `catch`, the `promise.resolve` template controller creates its own scope.

The `promise.resolve` template controller creates its own scope.
This prevents ...

docs/user-docs/getting-started/conditional-rendering.md
```diff
@@ -405,3 +405,316 @@ This section includes few more interesting examples that you might encounter in
...
+ <fizz-buzz catch="$parent.err"></fizz-buzz>
+</div>
+```
+{% endtab %}
+
+{% tab title="my-app.ts" %}
+```typescript
+export class MyApp {
+  public data: any;   // the resolved data is assigned to this once and when the promise is resolved
+  public data: Error; // the rejected reason is assigned to this once and when the promise is resolved
```
bigopon, line 541: public err: Error

docs/user-docs/getting-started/conditional-rendering.md
```diff
@@ -405,3 +405,316 @@ This section includes few more interesting examples that you might encounter in
...
+In case you are interested about the details on scoping and binding context, you can refer to the [Scope and context documentation](../app-basics/scope-and-binding-context).
+{% endhint %}
+
+### Nesting
+
+The template controllers can be nested.
+
+{% code title="my-app.html" %}
+```markup
+<template promise.resolve="fetchPromise">
```
bigopon, line 580: This kinds of makes me want to add an example where we call some function based on some parameters, and return a promise, so that the entire thing is automatically reloaded/rerendered without any view model involved. Any thoughts/comments on the extra examples of html focused usage of this controller?

docs/user-docs/getting-started/conditional-rendering.md
```diff
@@ -405,3 +405,316 @@ This section includes few more interesting examples that you might encounter in
...
+In short, with `let` binding in place the output looks like as follows.
+
+```html
+<span>42</span>
+<span>foo-bar</span>
+<span>forty-two</span>
+<span>fizz-bazz</span>
+```
+
+Whereas without the `let` binding, the result looks like below that does not match the general expectation.
```
bigopon, line 645: 👍 more examples around `<let/>`. I think I should write some doc for this too

packages/runtime-html/src/binding-behaviors/throttle.ts
```diff
@@ -75,7 +75,7 @@ export class ThrottleBindingBehavior extends BindingInterceptor {
 
   public $bind(flags: LifecycleFlags, scope: Scope, hostScope: Scope | null): void {
     if (this.firstArg !== null) {
-      const delay = Number(this.firstArg.evaluate(flags, scope, hostScope, this.locator, null));
+      const delay = Number(this.firstArg.evaluate(flags,  scope,  hostScope, this.locator, null));
```
bigopon, line 78: Hmm, our linting configuration is missing these extra spaces

packages/runtime-html/src/resources/template-controllers/promise.ts
bigopon, line 65:
```ts
Scope.fromParent($controller.scope, new PromiseContext())

...

class PromiseContext {

}
```
?

packages/runtime-html/src/resources/template-controllers/promise.ts
```diff
@@ -0,0 +1,346 @@
...
+          return resolveAll(
+            fulfilled?.deactivate(initiator, flags),
+            rejected?.deactivate(initiator, flags),
+            pending?.activate(initiator, flags, s, hs)
+          );
+        }, defaultQueuingOptions)).result,
+        value
+          .then(
+            (data) => {
+              if (this.value !== value) {
```
bigopon, line 107: 👍 I guess this is good enough for handling race. If insufficient, we can use another way, maybe a counter

packages/runtime-html/src/resources/template-controllers/promise.ts
```diff
@@ -0,0 +1,346 @@
...
+      this.postSettledTask?.cancel();
+      $swap();
+    }
+  }
+
+  public detaching(initiator: IHydratedController, parent: IHydratedParentController, flags: LifecycleFlags): void | Promise<void> {
+    this.preSettledTask?.cancel();
+    this.preSettledTask = null;
+    this.postSettledTask?.cancel();
+    this.postSettledTask = null;
```
bigopon, line 159:
```ts
this.preSettledTask
  = this.postSettledTask
  = null;
```
so we can save some characters

packages/runtime/src/observation/binding-context.ts
```diff
@@ -108,6 +108,7 @@ function chooseContext(
     }
 
     overrideContext = currentScope!.overrideContext;
+    // Here we are giving benefit of doubt considering the dev has used one or more `$parent` token, and thus should know what s/he is targeting.
```
bigopon, line 111: well 😁

bigopon (commented):
Looking great, a few small comments 🚀

docs/user-docs/getting-started/conditional-rendering.md
Sayan751: In real-life use cases, view-models will be inevitably involved while dealing with promises.  The reason being that every promise unique for every instance. Thus I would refrain from adding examples which are far from the real life use cases.

However, in the tests, as well as in the docs there is an example involving a `promisify` value converter. I have used that to easily construct an example scenario (disclaimer included in the docs). IMO that will suffice.

packages/runtime-html/src/resources/template-controllers/promise.ts
Sayan751: This is an interesting idea, although I would refrain from doing so. The way binding contexts are used internally everywhere, we cannot use the prototypal information in any way (that is a method defined in that class cannot be used internally). Then the only use case for that remains, if that prototypal information is used solely by the promise TC. In that case however, I would say that we might not need that at present. If the `{}` bothers you, we can do `Object.create(null)`.

packages/runtime-html/src/resources/template-controllers/promise.ts
Sayan751: I find the verbose version bit easier to understand. If it is all the same to you, I would like to keep it as it is now.

docs/user-docs/getting-started/conditional-rendering.md
Sayan751: Done.

packages/runtime-html/src/binding-behaviors/throttle.ts
Sayan751: Done

AureliaEffect:
The benchmarking comparison can be found here: https://au2-bench-viewer.azurewebsites.net/compare?branch=master&commit=d40f53103a5325c0767920f78366711e465316df&branch=topic/979-promise-tc&commit=57280bc8ec789c7187f549ec1d99e297a8d2b0f8

packages/runtime-html/src/resources/template-controllers/promise.ts
bigopon: Yeah that's fine too. though sooner or later, we will still need to do some of the minifier job, or at least help it do its job. There's a call (`this.postSettledTask?.cancel()`) in the middle, so it won't just merge the two assignment together.

docs/user-docs/getting-started/conditional-rendering.md
bigopon: > Thus I would refrain from adding examples which are far from the real life use cases.

Thanks for the reminder

bigopon:
It seems the tests for `i18n` are failing https://github.com/aurelia/aurelia/pull/1125/checks?check_run_id=2548403460 for this branch, while they all pass in master.
I don't see any changes that touch these tests, can you help check this

docs/user-docs/app-basics/scope-and-binding-context.md
Sayan751: I still don't think that the doc should point out that calling the function will throw an error. Because in reality, as it is shown the example, won't throw any error.

What we can do is to change the wordings a bit, as you have pointed out.  For example

```diff
- One way to think about expression and binding context is in terms of functions and binding those functions.
+ One way to think about expression and binding context is in terms of functions and binding those functions with a execution context (Refer: [Function.bind](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_objects/Function/bind)).
```

```diff
- If we execute this function we will get `NaN`.
+ If we invoke this function like `foo()`, we will get `NaN`.
```

docs/user-docs/app-basics/scope-and-binding-context.md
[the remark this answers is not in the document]
Sayan751: Reworded.

docs/user-docs/app-basics/scope-and-binding-context.md
Sayan751: Rewording done :)

AureliaEffect:
The benchmarking comparison can be found here: https://au2-bench-viewer.azurewebsites.net/compare?branch=master&commit=f0ea1d851a896fde856efa6b88e42777ace67761&branch=topic/979-promise-tc&commit=6d72adeab7e907057274ad1109f980909c537f01

AureliaEffect:
The benchmarking comparison can be found here: https://au2-bench-viewer.azurewebsites.net/compare?branch=master&commit=f0ea1d851a896fde856efa6b88e42777ace67761&branch=topic/979-promise-tc&commit=48b1843c08997bf682b07c5a9433efba2b9cdfe1

AureliaEffect:
The benchmarking comparison can be found here: https://au2-bench-viewer.azurewebsites.net/compare?branch=master&commit=f0ea1d851a896fde856efa6b88e42777ace67761&branch=topic/979-promise-tc&commit=dee4b2833bd5fe1a7cd8d389eb6d5a4f8338439f

docs/user-docs/getting-started/conditional-rendering.md
```diff
@@ -405,3 +406,309 @@ This section includes few more interesting examples that you might encounter in
...
+This is useful when we need the data only in view for passing from one component to another custom element, as it does not pollute the underlying view-model.
+Note that this does not make any difference in terms of data binding or change observation.
+However, when we do need to access the settled data inside the view model, we can use the `$parent.data` or `$parent.err` as shown in the example below.
+
+
+{% tabs %}
+{% tab title="my-app.html" %}
+```markup
+<div promise.bind="promise">
+ <foo-bar then="$parent.data"></foo-bar>
```
bigopon, line 525: I'll fix this after the merge

Sayan751: Thanks!

PR #1133, opened 2021-03-12
Update navigating-with-load.md - Fixed router.load parameters text.

cluka:
There was an error with the router.load() method. The documentation stated that it should have "parameters" parameter, but it's actually "params" in the code.

# Pull Request


fkleuver:
Thanks @cluka !

PR #1142, opened 2021-03-29
Update your-first-component-part-1-the-view-model.md

arjendeblok:
If you name the custom element `HelloNameCustomElement` then it won't work with the Alpha version. It must be called `HelloName`.


Vheissu:
Thank you for this. I've been slowly getting to this repo to update it, there are a lot of chapters.

arjendeblok:
@Vheissu  The convention is without adding CustomElement to the custom element name I assume.
I use custom elements a lot and I would not like to have CustomElement to be added to **all** my custom elements.

Or do you mean it would work with and without?

Vheissu:
> @Vheissu  The convention is without adding CustomElement to the custom element name I assume.
> I use custom elements a lot and I would not like to have CustomElement to be added to **all** my custom elements.
> 
> Or do you mean it should work with and without adding CustomElement to it?

It should work with both. I checked and this is a bug. It should be resolved soon :) in the meantime omitting the CustomElement part is fine (it's what I do anyways).

3cp:
I will get the bug fixed. But this PR is still good to merge, we should promote the convention without "CustomElement" suffix.

bigopon:
Thanks @arjendeblok !

Issue #1114, opened 2021-02-16
repeat.for context is not available in au-slot


MaximBalaganskiy:
# 🐛 Bug Report

Using `repeat.for` with an au-slotted element does not render the iterated value

## 🤔 Expected Behavior

Should render regardless of slots

## 😯 Current Behavior

Using `repeat.for` with an au-slotted element does not render the iterated value.
The repeat context is probably not available inside the slot.

## 💻 Code Sample

https://gist.dumber.app/?gist=c2aa7a79faf2a940cfc97eb228e43cff

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Aurelia          | 0.10.0-dev.202102151925

bigopon:
@MaximBalaganskiy thanks for this issue.
@Sayan751 this is probably caused by the misalignment between compile time & runtime scope. I'll fix it together with the other scope issue.

Sayan751:
Generally speaking, dynamic projections with `*[template-controller] > *[au-slot]` is not supported. But that example is an interesting one. I don't think there is any misalignment so to speak. IMO it is not (yet) supported.

PR #1144, opened 2021-03-31
fix: #1114

Sayan751:
The fix involves traversing upward from host scope till the outer scope is reached.

# Pull Request


AureliaEffect:
The benchmarking comparison can be found here: https://au2-bench-viewer.azurewebsites.net/compare?branch=master&commit=b270d9d04843702d2d2b6771f4997337b0886273&branch=issue/1114-au-slot-repeater-scope&commit=1b5953f50efdb24f63e073b3289f02ce9e692635

packages/runtime/src/observation/binding-context.ts
```diff
@@ -88,7 +88,12 @@ export class BindingContext implements IBindingContext {
   }
 }
 
-function chooseContext(scope: Scope, name: string, ancestor: number): IBindingContext | undefined | null {
+function chooseContext(
+  scope: Scope,
+  name: string,
+  ancestor: number,
+  outerScope: Scope | null,
```
bigopon, line 95: `outerScope` may not be as descriptive as we'd hope. In this template:
```html
My element template
<template repeat.for="row of rows">
  <template repeat.for="col of cols">
    <my-element>
      <template au-slot>
        \${item}
      </template>
    </my-element>
  </template>
</template>`
```
It's a bit uncomfortable using outer scope. I guess it means "most" outer scope? If so, what are the cons of `boundaryScope`?

packages/__tests__/3-runtime-html/au-slot.spec.ts
```diff
@@ -512,6 +512,78 @@ describe('au-slot', function () {
...
+        {
+          'my-element': [`1`, new AuSlotsInfo(['default'])],
+          'my-element+my-element': [`2`, new AuSlotsInfo(['default'])],
+        },
+      );
+
+      yield new TestData(
+        'works with data from repeater scope - template[repeat.for]>custom-element - source: binding context in outer scope',
+        `<template>
+          <template repeat.for="person of people">
```
bigopon, line 555: while we are here, can we add another test the following template:
```html
<template repeat.for="row of rows">
  <template repeat.for="col of cols">
    <my-element>
      <template au-slot>
        \${item}
      </template>
    </my-element>
  </template>
</template>
```

packages/runtime/src/observation/binding-context.ts
Sayan751: There might be several boundaries found in the process of scope traversal, which makes me think that `boundaryScope` might not be more useful name than `outerScope`.  However, I see your point. That's why I am inclined to use the name `projectionScope`, as also used in `au-slot`.

AureliaEffect:
The benchmarking comparison can be found here: https://au2-bench-viewer.azurewebsites.net/compare?branch=master&commit=b270d9d04843702d2d2b6771f4997337b0886273&branch=issue/1114-au-slot-repeater-scope&commit=396d38e4ae257226dd481e18e1cd81f4b6fdecba

packages/runtime/src/observation/binding-context.ts
Sayan751: Done

packages/__tests__/3-runtime-html/au-slot.spec.ts
Sayan751: Done

packages/runtime/src/observation/binding-context.ts
```diff
@@ -67,14 +67,14 @@ export class BindingContext implements IBindingContext {
...
-    let context = chooseContext(scope, name, ancestor);
+    let context = chooseContext(scope, name, ancestor, null);
     if (
       context !== null
       && ((context == null ? false : name in context)
         || !hasOtherScope)
     ) { return context; }
     if (hasOtherScope) {
-      context = chooseContext(hostScope!, name, ancestor);
+      context = chooseContext(hostScope!, name, ancestor, scope);
```
bigopon, line 77: I think we may be able to skip all the logic inside `chooseContext` here. As it's the boundary of the "inner" scope (of the custom element) already. Will simplifying it to:
```ts
if (hasOtherScope) {
  if (name in hostScope.bindingContext) {
    context = hostScope.bindingContext;
  } else {
    // nothing found in the upper scope
    // and nothing found inside $host scope
    // which means it's at the projection scope
    context = scope.bindingContext;
  }
}
```

bigopon (commented):
Nice new tests! Feels a lot more complete now

packages/runtime/src/observation/binding-context.ts
Sayan751: I think that's not right. The current fix for the issue #1114 proves that we need to start from the host scope and if needed we need to traverse up the scope tree, if the property is not found. That's why I don't think this simplification will be correct.

bigopon: I think this is probably ok, as we only deal with `$hostScope` after the `projectionScope`, so missing items in all scope hierarchy will eventually be defined in `projectionScope`, and there's only 1 extra scope layer traversal 👍

PR #1151, opened 2021-04-06
fix(debounce): respect default delay

avrahamcool:
use 200ms delay for debounce and throttle when no parameter is provided

fixes aurelia#1150

### DIFF
--- a/packages/runtime-html/src/binding-behaviors/debounce.ts
+++ b/packages/runtime-html/src/binding-behaviors/debounce.ts
@@ -14,7 +14,7 @@ const defaultDelay = 200;
 //
 export class DebounceBindingBehavior extends BindingInterceptor {
   private readonly taskQueue: TaskQueue;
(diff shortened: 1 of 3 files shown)


packages/runtime-html/src/binding-behaviors/debounce.ts
```diff
@@ -14,7 +14,7 @@ const defaultDelay = 200;
 //
 export class DebounceBindingBehavior extends BindingInterceptor {
   private readonly taskQueue: TaskQueue;
-  private readonly opts: QueueTaskOptions = { delay: 0 };
+  private readonly opts: QueueTaskOptions = { delay: defaultDelay };
```
bigopon, line 17: I think this is good because if there's no timing specified, this won't ever be changed. If there's a timing parameter, then it will always be reset in `$bind`

bigopon (approved):

PR #1152, opened 2021-04-07
fix(array-obs): handle array.find for computed obs, dont cache index obs

bigopon:
# Pull Request

## 📖 Description

At the moment, `.find` method of arrays are not properly handled in computed, it's treated as a normal method/property on an object. This PR fixes this.


AureliaEffect:
The benchmarking comparison can be found here: https://au2-bench-viewer.azurewebsites.net/compare?branch=master&commit=e4b40498cab6dd9234066498f574b66acbb49ee6&branch=fix/array-observation-proxy-n-index&commit=79ba544358fc97d98632c9ebe874ca5c3fdbf714

PR #1153, opened 2021-04-08
fix(obs-record): separate count & slot number

bigopon:
# Pull Request

## 📖 Description

In a previous refactor, open observer slot number were merged into observer count, and causing inaccuracy in marking empty slot. This PR fixes this.

### 🎫 Issues

A repro of this is at https://github.com/markusrut/aurelia-example/tree/1bedcbaf2a0c9b7c72043df3e0ae578145023f77
thanks to @markusrut : Clicking reload 100 times + select an item shouldn't cause increasing time required for each selected item computation.

### Next Steps

Probably should just use a Map instead of this. Though it would lead to higher mem usage & possibly worse perf in simpler scenarios

### DIFF
--- a/packages/runtime/src/binding/connectable.ts
+++ b/packages/runtime/src/binding/connectable.ts
@@ -21,22 +21,6 @@ import type { Scope } from '../observation/binding-context.js';
 
 // TODO: add connect-queue (or something similar) back in when everything else is working, to improve startup time
 
-const slotNames: string[] = [];
-const versionSlotNames: string[] = [];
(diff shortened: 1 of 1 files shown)


AureliaEffect:
The benchmarking comparison can be found here: https://au2-bench-viewer.azurewebsites.net/compare?branch=master&commit=4e73f39cbc2e515c50615ba89e8585c0cb30bce5&branch=fix/obs-record-cleanup&commit=c824801abfe5660cd28972cbfc8015504bbbfdeb

PR #1154, opened 2021-04-11
docs(internationalization.md): fix config for copy-webpack-plugin

AntonJacobsson:
# Pull Request

## 📖 Description

Hi! I updated the docs so it works with latest version of copy-webpack-plugin (8.1.1).

example:
https://github.com/webpack-contrib/copy-webpack-plugin

### 🎫 Issues

image
image


bigopon:
@AntonJacobsson thank you!. Should we have some extra note for different version of webpack, or at least signal what's the minimum version that is compatible with the syntax?

AntonJacobsson:
@bigopon I made an update to this PR. let me know what you think.

bigopon:
@AntonJacobsson awesome, thanks for the doc change!

PR #1164, opened 2021-05-17
chore(docs): add observable properties docs

### LINKED ISSUE #1163: Observable Properties Documentation
The Aurelia v1 docs on Observable properties (https://aurelia.io/docs/binding/observable-properties) shows the convention of adding an @observable decorator and adding a propertyNameChanged(new_value, old_value) method. Aurelia v2 also appears to have this capability (https://github.com/aurelia/aurelia/pull/958) but I don't see it in the observation documentation (https://docs.aurelia.io/getting-started/observation). It would be helpful if the v1 docs for this were ported over to v2.

It also appears that you can add a static observables property to a class instead of using the decorator:
`static get observables() { return ['propertyName'] };`
Having documentation on this would also be helpful.

wesfolz:
closes #1163

# Pull Request

## 📖 Description
add observable properties docs

### 🎫 Issues
 #1163

### DIFF
--- a/docs/user-docs/getting-started/observation.md
+++ b/docs/user-docs/getting-started/observation.md
@@ -6,7 +6,7 @@ description: Observe changes in your applications.
 
 ## Observation Basic
 
-Aurelia observation system is built using observers and subscribers. By default, an observer locator is used to create observers and subscribe to them for change notification. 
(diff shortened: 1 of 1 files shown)


bigopon:
Thanks @wesfolz !

PR #1170, opened 2021-05-25
chore(dist): add a bundled dist

bigopon:
At the moment, packages are distributed in multiple file form, causing slow downloading time for online repl/IDEs. Fix this by adding additional dist format that has only a single file. Existing format will still be used, and this dist is a hidden one. First step is to add custom resolution to this new format and later make a switch.

The bundle file is at `dist/bundle/index.js`. Only the main packages for applications will be rolled this way, tooling packages (cjs) will be kept as is.

cc @fkleuver @3cp

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -56,6 +56,7 @@
       ],
       "devDependencies": {
         "@commitlint/cli": "^8.2.0",
+        "@rollup/plugin-typescript": "^8.2.1",
         "@typescript-eslint/eslint-plugin": "^4.10.0",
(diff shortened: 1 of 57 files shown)


AureliaEffect:
The benchmarking comparison can be found here: https://au2-bench-viewer.azurewebsites.net/compare?branch=master&commit=0bedb6d217158b427f924779283773f6369561c7&branch=chore/rollup&commit=c6d5db66015e5549420c01982c46c26d564f4d08

Issue #1167, opened 2021-05-21
repeat.for with child components that have slots produces unexpected behavior


michaelnero:
# 🐛 Bug Report

I have an instance where using `repeat.for` in combination with child components that have slots produces behavior that I don't expect. I'm able to work around it by cloning the component definition in `define`. Here's a gist that reproduces the issue:
https://gist.dumber.app/?gist=47b4b854b00d9dd87ab5ccd95fb655c9

## 🤔 Expected Behavior

The correct value should show for each item.

## 😯 Current Behavior

I don't see the correct value for each item.

## 💻 Code Sample

https://gist.dumber.app/?gist=47b4b854b00d9dd87ab5ccd95fb655c9

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Aurelia          | 2.0.0-alpha.3
| Language         | typescript
| Browser          | Edge 90.0.818.62
| Bundler          | webpack 5.37.1
| Operating System | Windows 10
| NPM/Node/Yarn    | npm

bigopon:
@michaelnero thank you. I happened to be looking at this currently as well. Will fix and publish.

bigopon:
The issue is related to context cache retrieval based on definition. There' two different `<listbox/>`, but both of them have the same static definition and thus got the same cached render context. A workaround is to either have a `define` lifecycle returning a different definition, or an injectable token. Will fix

michaelnero:
That's what I was assuming was happening, so you can see in my gist that I attempted to "fix" it by doing exactly what you suggested. The `assignee-fix` custom element is returning the cloned definition, and gives me the expected behavior.

I don't know enough about the code base yet to suggest a fix, but happy to help out in any other way.

bigopon:
@michaelnero the gist description and the repro helped heap, in the end, there will be another test based on the gist you provided here as well. The one I got above is based on another repro I got earlier. For the fix, I'm not sure about a fix direction yet, but will update.

PR #1173, opened 2021-05-29
fix: au-slot scoping issue

### LINKED ISSUE #1113: view-model.ref leaks from the custom element boundary
# 🐛 Bug Report
Consider the following parent element
```html
<template>
  <au-slotted-element>
    <child view-model.ref="child"></child>
  </au-slotted-element>
</template>
```
and the following usage
```html
<parent>Parent 1</parent>
<parent>Parent 2</parent>
```

## 🤔 Expected Behavior

`view-model.ref` should grab the correct value

## 😯 Current Behavior

The first parent's view model will have an incorrect child reference, pointing to the second parent's child.

## 🔦 Context

MDC text fields have quite complex hierarchy with slotted elements and floating labels residing in these elements.

## 💻 Code Sample

https://gist.dumber.app/?gist=727aed61d5d505773d9d82f33d8fb631

## 🌍 Your Environment

| Aurelia          | 0.10.0-dev.202102110248
| Language         | Typescript
| Browser          | Chrome
| Bundler          | Webpack
| Operating System | Windows 10
| NPM/Node   | 7/15

Sayan751:
# Pull Request

## 📖 Description

Closes #1113


packages/runtime-html/src/resources/custom-elements/au-slot.ts
```diff
@@ -50,6 +44,14 @@ export class ProjectionProvider {
   public getProjectionFor(instruction: IInstruction): RegisteredProjections | null {
     return projectionMap.get(instruction) ?? null;
   }
+
+  public registerScopeForAuSlot(auSlotInstruction: IInstruction, scope: Scope): void {
```
bigopon, line 48: I think we can at least remove `auSlot` out of the naming. It's also unclear what kind of resource auSlot means between element and attribute. For the idea itself, I think this is good, everything should go by instruction 👍

bigopon (commented):
I haven't been able to review & test this extensively yet, though I noticed that we are missing some failing test. Can we add the failing test I've prepared here https://github.com/aurelia/aurelia/issues/1167#issuecomment-846356537

packages/runtime-html/src/templating/render-context.ts
```diff
@@ -423,6 +425,8 @@ export class RenderContext implements IComponentFactory {
         name = this.definition.name;
       }
       factory = this.factory = new ViewFactory(name, this, contentType, projectionScope);
+    } else {
+      factory.projectionScope = projectionScope!;
```
bigopon, line 429: At the moment, 2nd and onward occurrence of an element will get the same render context, because of the cached definition. Will this change cause a inverted version of the issue we are fixing, where the 2nd and after occurrence keeps overriding the factory from the previous one.

Sayan751:
> I haven't been able to review & test this extensively yet, though I noticed that we are missing some failing test. Can we add the failing test I've prepared here [#1167 (comment)](https://github.com/aurelia/aurelia/issues/1167#issuecomment-846356537)

I was actually looking for this issue. But somehow it slipped. Thanks for pointing this out. We might be able to close both, if the test passes.

AureliaEffect:
The benchmarking comparison can be found here: https://au2-bench-viewer.azurewebsites.net/compare?branch=master&commit=3e45eadfd2e6d90390c29e10ca14fa5d2771f2c5&branch=topic/1113-au-slot-issue&commit=255b1785e259788eb1a1f5ef7643d8c71888f4af

packages/runtime-html/src/templating/render-context.ts
Sayan751: I don't think so. The `CustomElementRenderer` passes the correct scope to it. For a complied definition, we are registering the projection scope as done here: 
https://github.com/aurelia/aurelia/pull/1173/files#diff-748a8a74b12b942c71bdb1e88059c33be782e1cd8ad1268e1bee625907a0e3e3R374-R377

So I think it won't pick up the old scope.

packages/runtime-html/src/resources/custom-elements/au-slot.ts
Sayan751: Done

Sayan751:
@bigopon Added new tests.

AureliaEffect:
The benchmarking comparison can be found here: https://au2-bench-viewer.azurewebsites.net/compare?branch=master&commit=3e45eadfd2e6d90390c29e10ca14fa5d2771f2c5&branch=topic/1113-au-slot-issue&commit=e02418463c4cf5329b22cb84876cfce0b3f4a8e8

packages/runtime-html/src/templating/render-context.ts
bigopon: I checked this out and tested locally by adding the following line
```ts
console.log('assigning projection scope. Has existing:', !!factory.projectionScope);
```
before this line.
Here's what I got for the test at line 1346 (parent/tab-bar/tab)
```
(0LOG: 'assigning projection scope. Has existing:', true
(0LOG: 'assigning projection scope. Has existing:', false
```
Does the overriding of existing scope matter?

packages/runtime-html/src/templating/render-context.ts
```diff
@@ -442,6 +446,27 @@ export class RenderContext implements IComponentFactory {
...
+        if (slotInfo !== null) {
+          if (slotInfo.type === AuSlotContentType.Projection) {
+            projectionProvider.registerScopeFor(instruction, scope);
+          }
+          instructions.push(...(slotInfo.content.instructions?.flat() ?? []));
+        }
+      } else if (instruction instanceof HydrateTemplateController) {
+        instructions.push(...((instruction.def.instructions?.flat() ?? []) as IInstruction[]));
+      }
+    }
```
bigopon, line 467:
```suggestion
    let i = 0;
    while (i < instructions.length) {
      const instruction = instructions[i];
      if (instruction instanceof HydrateElementInstruction) {
        const slotInfo = instruction.slotInfo ?? null;
        if (slotInfo !== null) {
          if (slotInfo.type === AuSlotContentType.Projection) {
            projectionProvider.registerScopeFor(instruction, scope);
          }
          instructions.push(...(slotInfo.content.instructions?.flat() ?? []));
        }
      } else if (instruction instanceof HydrateTemplateController) {
        instructions.push(...((instruction.def.instructions?.flat() ?? []) as IInstruction[]));
      }
      ++i;
    }
```

Sayan751: I am not sure how does it avoid array mutation.

packages/runtime-html/src/templating/render-context.ts
Sayan751: Overriding of the previous scope was the thing that was missing before. Which caused the issues with incorrect scoping. IMO it is essential in order to make things work correctly. Also note that the `projectionScope` property of view factory is only used exclusively in `AuSlot`.

packages/runtime-html/src/templating/render-context.ts
Sayan751: Also I don't see much point keeping the old instructions in the array, even though those are no longer needed.

bigopon: It's to help us reduce the cost while iterating, since shift() will cost more than an index property access

packages/runtime-html/src/templating/render-context.ts
Sayan751: Refactored.

packages/runtime-html/src/templating/render-context.ts
Sayan751: Done

packages/runtime-html/src/templating/render-context.ts
```diff
@@ -442,6 +444,28 @@ export class RenderContext implements IComponentFactory {
...
+    if (targetedProjections === null) { return; }
+    const scope = targetedProjections.scope;
+    const projectionProvider = this.projectionProvider;
+    const instructions = this.compiledDefinition.instructions.flat();
+    let i = 0;
+    while (i < instructions.length) {
+      const instruction = instructions[i++];
+      if (instruction instanceof HydrateElementInstruction) {
+        const slotInfo = instruction.slotInfo ?? null;
+        if (slotInfo !== null) {
```
bigopon, line 457: we should have less code & generated code this way, while perf should be better too.
```suggestion
        const slotInfo = instruction.slotInfo;
        if (slotInfo != null) {
```

packages/runtime-html/src/templating/render-context.ts
```diff
@@ -442,6 +444,28 @@ export class RenderContext implements IComponentFactory {
     return this;
   }
 
+  private $registerScopeForAuSlot(targetedProjections: RegisteredProjections | null): void {
```
bigopon, line 447: This naming is different with the rest of our methods, Is the intention make this method semi private?

packages/runtime-html/src/templating/render-context.ts
```diff
@@ -70,7 +70,7 @@ export interface IRenderContext extends IContainer {
    *
    * @returns Either a new `IViewFactory` (if this is the first call), or a cached one.
    */
-  getViewFactory(name?: string, contentType?: AuSlotContentType, projectionScope?: Scope | null): IViewFactory;
```
bigopon, line 73: I like this, immediately looks cleaner and should make the refactoring easier too.

packages/runtime-html/src/templating/controller.ts
```diff
@@ -332,8 +332,6 @@ export class Controller<C extends IViewModel = IViewModel> implements IControlle
     const { projectionsMap, shadowOptions, isStrictBinding, hasSlots, containerless } = compiledContext.compiledDefinition;
```
bigopon, line 332: probably should delete `projectionMap` from the compiled definition too?

bigopon (commented):
Both the fixes seem good, though this fix looks a lot nicer than the previous. Nice one.

AureliaEffect:
The benchmarking comparison can be found here: https://au2-bench-viewer.azurewebsites.net/compare?branch=master&commit=3e45eadfd2e6d90390c29e10ca14fa5d2771f2c5&branch=topic/1113-au-slot-issue&commit=9ef966464bd2b05045a7c5e0f893375a4c43e5af

packages/runtime-html/src/templating/render-context.ts
Sayan751: Renamed.

packages/runtime-html/src/templating/controller.ts
Sayan751: Nope. That should be there.

AureliaEffect:
The benchmarking comparison can be found here: https://au2-bench-viewer.azurewebsites.net/compare?branch=master&commit=3e45eadfd2e6d90390c29e10ca14fa5d2771f2c5&branch=topic/1113-au-slot-issue&commit=ded736d8235dca912b642dc50f90cdde1314645f

PR #1179, opened 2021-06-03
fix: #779

### LINKED ISSUE #779: ValueConverter on collection not executed after adding a new item to the collection
# 🐛 Bug Report

A value converter on a collection is not executed when an item is added to the collection, see:

Is executed after adding an item to the collection in au1:
e2e test run: https://youtu.be/75rCPILsyy8 
code: https://github.com/eriklieben/mua-component-library-au1

AU2 
e2e test run: https://youtu.be/-ZwRDa8HlOQ 
code: https://github.com/eriklieben/mua-component-library

See the last e2e test "should find items that are added to the list after searching", after adding the item it is in the collection (bottom list), but the ```repeat.for``` loop with a value, binding doesn't get executed to refresh/update the result. When (by hand) I perform the same and clear the search term and apply it again, the result is correct (because it executed the value converter in between my actions).

It currently directly updates the items in the collection, because I assumed it had something to do with me cloning the collection, wanted to exclude that as a possibility. Left that in, because that might be easier in finding a solution. 
In production, one would probably clone the collection, before applying modifications, but I think/assume if it executed the value converter after a modification to the collection that scenario should also work again.

## 🤔 Expected Behavior

After a collection is modified (item added) the value converter should be executed.

## 😯 Current Behavior

The value converter is not executed after an item was added.

## 💁 Possible Solution

not sure.

## 💻 Code Sample

AU2 code https://github.com/eriklieben/mua-component-library
same logic in AU1 code: https://github.com/eriklieben/mua-component-library-au1

you can run the cypress e2e tests in both code samples to see the behavior.

## 🌍 Your Environment

| Aurelia          |dev
| Language         | ts
| Browser          | chrome 78 on windows 10
| Bundler          | webpack
| Operating System | win 10
| NPM/Node/Yarn    | npm 6.10

Let me know if I need to explain it better or more clearly. Thanks!

Sayan751:
# Pull Request

## 📖 Description

The fix involves subscribing to the collection change in the `AccessScopeExpression#evaluate`.

### 🎫 Issues

Closes #779


packages/runtime/src/binding/ast.ts
```diff
@@ -624,10 +624,16 @@ export class AccessScopeExpression {
 
   public evaluate(f: LF, s: Scope, hs: Scope | null, _l: IServiceLocator, c: IConnectable | null): IBindingContext | IOverrideContext {
     const obj = BindingContext.get(chooseScope(this.accessHostScope, s, hs), this.name, this.ancestor, f, hs) as IBindingContext;
+    const evaluatedValue = obj[this.name] as ReturnType<AccessScopeExpression['evaluate']>;
     if (c !== null) {
       c.observeProperty(obj, this.name);
+      if (Array.isArray(evaluatedValue)
```
bigopon, line 630: We should be using `instanceof Array` to avoid any hiccup with proxy.

bigopon (commented):
The idea of value converter automatically observing a collection looks interesting, but it's a risky side effect destroying the purity of value converters. Observing collection should be done per application of the collection. Example is the repeater itself should observe the inner collection in an expression that has value converter.

Also to support this, property binding is no longer a property binding, which surface even questions for maintenance. And if we are going this route, it's missing in AttrBinding.

The repeater missing the check what collection it should observe, is the issue, can you help fix it?
```html
<div repeat.for="item of items | first :5">
```
`items` is the inner collection expression, `items | first :5` is the outer collection expression, the repeater should observe the inner one instead of the outer.

Sayan751:
> The idea of value converter automatically observing a collection looks interesting

@bigopon The value converter does not observe the collection though. There was previously a property observation employed in `AccessScope#evalue`. I just extended the idea and added collection observation to it.

> the repeater should observe the inner one instead of the outer.

This is an interesting idea. However, IMO in absence of the `Expression#connect` API it might be bit difficult. Then we need to grab the inner AccessScopeExpression and `evaluate` it using the repeater as connectable.

AureliaEffect:
The benchmarking comparison can be found here: https://au2-bench-viewer.azurewebsites.net/compare?branch=master&commit=a9f422dd0610ca0e71d75d46fe612fde4d1629f6&branch=topic/779-value-converter-repeater&commit=5cdb0e328826f3ee272ad7e229f6d1c2657a1f39

AureliaEffect:
The benchmarking comparison can be found here: https://au2-bench-viewer.azurewebsites.net/compare?branch=master&commit=8eaae4facd7e39fb48e9c9664ec358ef93fc6454&branch=topic/779-value-converter-repeater&commit=a4c3429fd491a7200198f8a897b7ce25d4c1aa42

AureliaEffect:
The benchmarking comparison can be found here: https://au2-bench-viewer.azurewebsites.net/compare?branch=master&commit=8eaae4facd7e39fb48e9c9664ec358ef93fc6454&branch=topic/779-value-converter-repeater&commit=445f912364cbd16f8b7a55ed7c0995bbeeaacb31

PR #1181, opened 2021-06-04
refactor(children): make children observer a controlled observer

bigopon:
At the moment, children observation does not stop. Also, it's working using similar mechanic as other observers: start observing when there's a subscriber. This is bad related timing wise, since a subscriber can be added at any time.

Change the timing to: start observing the child list after binding, and before bound, (~$bind phase of bindings), so that changes can be accurately observed, and the owning controller will be in charge of starting/stopping the observers.

### DIFF
--- a/packages/__tests__/3-runtime-html/children-observer.spec.ts
+++ b/packages/__tests__/3-runtime-html/children-observer.spec.ts
@@ -1,131 +1,153 @@
-import { children, CustomElement, PartialChildrenDefinition, Aurelia } from '@aurelia/runtime-html';
-import { TestContext, assert } from '@aurelia/testing';
+import { children, CustomElement, PartialChildrenDefinition, Aurelia, customElement } from '@aurelia/runtime-html';
+import { TestContext, assert, createFixture } from '@aurelia/testing';
 import { IContainer } from '@aurelia/kernel';
 
 describe('ChildrenObserver', function () {
   describe('populates', function () {
-    it('children array with child view models', function () {
+    it('[without shadow DOM] static plain elements', async function () {
+      @customElement({ name: 'my-el', template: '<slot>' })
+      class MyEl {
+        @children({ filter: n => !!n, map: n => n }) public children: unknown[];
+      }
+      const { appHost, startPromise, tearDown } = createFixture(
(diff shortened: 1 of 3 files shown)


bigopon:
cc @fkleuver @Sayan751 @brandonseydel
maybe you will be interested in this change

AureliaEffect:
The benchmarking comparison can be found here: https://au2-bench-viewer.azurewebsites.net/compare?branch=master&commit=a9f422dd0610ca0e71d75d46fe612fde4d1629f6&branch=fix/children-deco&commit=673c8acfe60bd91a58e61ff5f671cf2d071888ca

Sayan751:
@bigopon Awesome change! This means that we can do something like following.

```typescript
class ViewModel {
  @children(...) children;
  
  clickDelegate() {
    // start here all this.$controller.childrenObs
    // do something that will eventually change DOM
  }

  childrenChanged() {
      // stop here all this.$controller.childrenObs
  }
}
```

The example might have some pitfalls, but you get the idea. Consider making the `Controller#childrenObs` public?

bigopon:
@Sayan751 at the moment, it's started/stopped when the corresponding controller gets activated/deactivated. With the above, you meant you want to start it before/after that?

Sayan751:
> @Sayan751 at the moment, it's started/stopped when the corresponding controller gets activated/deactivated. With the above, you meant you want to start it before/after that?

Nope. I just wanted to show that the observation can be stopped before deactivation (for example after attach), and can be started again if needed based on some event. Sorry for the confusion.

bigopon:
I was thinking of stopping before activation too (for when number of subscriber dropped to 0). There was some complication around this so the safer option seemed to be always on/off accordingly to the owning controller to avoid stale values. Maybe if you'd like to enhance it, try see if it can be independent + ensure correct value?

Sayan751:
Sorry for not making it clear. I find the timing perfectly alright and that should be acceptable and sensible choice. I am just saying that a CE view-model class can stop and start those observations at will by using the `this.$controller.childrenObs` after your change. And from that respect it would have been nice to have a test to assert that scenario. But the absence of such tests does not harm anything.

PR #1184, opened 2021-06-10
feat(compose): add v1 ability to compose from view + view model + model

### LINKED ISSUE #498: au-compose should allow parent scope inheritance
# 🙋 Feature Request

By default, dynamically composed elements are encapsulated like normal elements. However, some scenarios are greatly simplified when the composition is allowed to bind to the context in which it is declared, especially when the composed element is an HTML-only element.

## 💁 Possible Solution

We want the default behavior to result in encapsulation, ensuring that composed elements work like normal elements. So, we could add an attribute to compose to allow developers to change that behavior on a per-compose basis. We could add an attribute like `use-parent-scope` or we could do something like `scope="parent"` if we wanted to enable other possibilities in the future.

## 🔦 Context

In my case, I'm creating a set of controls where a large part of the HTML internal to the control may need to be changed based on a number of values. `au-compose` is the sensible solution for this, but the inability to pass through all available binding context from the control itself into the composed HTML complicates the authoring and maintenance of each control.

### LINKED ISSUE #953: Allow au-compose to accept view, view-model and model properties like v1
# 🙋 Feature Request

The `<au-compose>` element is missing numerous features that are present and widely used in Aurelia 1's compose functionality.

## 🤔 Expected Behavior

The ability to compose without the need to supply a subject attribute like Aurelia 1. I should be able to compose in this way `<au-compose view.bind="view" view-model.bind="viewModel" model.bind="data"></au-compose>`

## 😯 Current Behavior

The `<au-compose>` element expects a subject and does not provide the same level of functionality or flexibility as Aurelia 1.

## 💁 Possible Solution

Bring `<au-compose>` inline with `<compose>` from Aurelia 1.

## 🔦 Context

Attempting to migrate Aurelia 1 apps that make use of dynamic composition, especially in instances where the UI is driven by data.

bigopon:
# Pull Request

## 📖 Description

Compared to v1 `<compose/>`, the current `au-compose` is having a different API set aiming to take advantages of the new templating primitives in v2, though since the time it was envisioned, things have changed quite a bit and its capabilities don't address the core needs of a more familiar API set in V1. This PR adds back the v1 composition functionalities, with a few breaking changes as noted below:

- main: passing a string to either view/ view model no longer means module name. The composer only understands string as template for `view` and objects and classes for `view-model` now. Though if it's still possible to turn treat string as module name, with the help of value converter:
    ```html
    <au-compose view="https://my-server.com/views/${componentName} | loadView">
    ```
    ```ts
    class LoadViewValueConverter {
      toView(v) { return fetch(v).then(r => r.text()) }
    }
    ```
- inherit binding context is no longer the default in all cases: it will only inherit parent binding context (or scope) when it's a view only composition. User can also disable this via `scope-behavior ` bindable:
    ```html
    <au-compose scope-behavior="scoped">
    ```
  Possible values are:
      - `auto`: in view only composition: inherit the parent scope
      - `scoped`: never inherit parent scope even in view only composition
- naming change: `compose` (v1) is  changed to `au-compose` (v2), and the existing `au-compose` will be renamed to a different name. Possibly either `au-element` or `au-render`.

closes #953
closes #498

## 📑 Test Plan

Prepare tests for:
- view only
- view model only
- view + view model
- literal objects
- activate + deactivate
- shadow dom + native slot

## ⏭ Next Steps

Refactor templating to support binding instruction transferring, and au-slot. native slot should works fine with shadow dom enabled on the view model of the composition.

### DIFF
--- a/packages/__tests__/3-runtime-html/compose.spec.ts
+++ b/packages/__tests__/3-runtime-html/compose.spec.ts
@@ -8,17 +8,17 @@ import {
   Aurelia,
   RenderPlan,
   IPlatform,
+  AuCompose,
 } from '@aurelia/runtime-html';
 import {
   eachCartesianJoin,
   TestContext,
   trimFull,
   assert,
+  createFixture,
 } from '@aurelia/testing';
 
-const spec = 'compose';
-
-describe(spec, function () {
+describe('3-runtime-html/compose.spec.ts/au-render', function () {
   function createFixture(): SpecContext {
     const ctx = TestContext.create();
     const { container, platform, observerLocator } = ctx;
@@ -86,39 +86,39 @@ describe(spec, function () {
   const templateSpecs: TemplateSpec[] = [
     {
       t: '1',
-      template: `<template><au-compose subject.bind="sub"></au-compose></template>`
+      template: `<template><au-render subject.bind="sub"></au-render></template>`
     },
     {
       t: '2',
-      template: `<template><template as-element="au-compose" subject.bind="sub"></template></template>`
+      template: `<template><template as-element="au-render" subject.bind="sub"></template></template>`
     },
     {
       t: '13',
-      template: `<template><au-compose repeat.for="i of 1" subject.bind="sub"></au-compose></template>`
+      template: `<template><au-render repeat.for="i of 1" subject.bind="sub"></au-render></template>`
     },
     {
       t: '4',
-      template: `<template><au-compose if.bind="true" subject.bind="sub"></au-compose></template>`
+      template: `<template><au-render if.bind="true" subject.bind="sub"></au-render></template>`
     },
     {
       t: '5',
-      template: `<template><div if.bind="false"></div><au-compose else subject.bind="sub"></au-compose></template>`
+      template: `<template><div if.bind="false"></div><au-render else subject.bind="sub"></au-render></template>`
     },
     {
       t: '16',
-      template: `<template><au-compose if.bind="true" repeat.for="i of 1" subject.bind="sub"></au-compose></template>`
+      template: `<template><au-render if.bind="true" repeat.for="i of 1" subject.bind="sub"></au-render></template>`
     },
     {
       t: '17',
-      template: `<template><au-compose if.bind="true" repeat.for="i of 1" subject.bind="sub"></au-compose></template>`
+      template: `<template><au-render if.bind="true" repeat.for="i of 1" subject.bind="sub"></au-render></template>`
     },
     {
       t: '18',
-      template: `<template><au-compose subject.bind="sub" if.bind="true" repeat.for="i of 1"></au-compose></template>`
+      template: `<template><au-render subject.bind="sub" if.bind="true" repeat.for="i of 1"></au-render></template>`
     },
     {
       t: '19',
-      template: `<template><au-compose if.bind="true" subject.bind="sub" repeat.for="i of 1"></au-compose></template>`
+      template: `<template><au-render if.bind="true" subject.bind="sub" repeat.for="i of 1"></au-render></template>`
     },
   ];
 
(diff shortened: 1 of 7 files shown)


AureliaEffect:
The benchmarking comparison can be found here: https://au2-bench-viewer.azurewebsites.net/compare?branch=master&commit=8eaae4facd7e39fb48e9c9664ec358ef93fc6454&branch=feat/compose-v1&commit=926cb470de300b6b7195ea506ce520ab6fb6cae3

AureliaEffect:
The benchmarking comparison can be found here: https://au2-bench-viewer.azurewebsites.net/compare?branch=master&commit=8eaae4facd7e39fb48e9c9664ec358ef93fc6454&branch=feat/compose-v1&commit=b9bdc581eb1b39253173624c85946b4965132961

AureliaEffect:
The benchmarking comparison can be found here: https://au2-bench-viewer.azurewebsites.net/compare?branch=master&commit=8eaae4facd7e39fb48e9c9664ec358ef93fc6454&branch=feat/compose-v1&commit=2594944015f16b2f6135fcb28865e36b2a447202

bigopon:
Compose is now out in alpha 6

packages/__tests__/integration/app/molecules/specs-viewer/specs-viewer.html
```diff
@@ -1,6 +1,6 @@
 <template>
   <template repeat.for="pair of pairs">
-    <au-compose subject.bind="pair.vm" model="pair.thing"></au-compose>
+    <au-render subject.bind="pair.vm" model="pair.thing"></au-compose>
```
Sayan751, line 3: @bigopon This seems to be an error.

bigopon: The existing au compose was renamed to au render. Is this what you meant with the erro?

Sayan751: Nope. The end tag also needs to be changed :)

bigopon: Ahh nice thanks for spotting that

bigopon: Html parser is magical

packages/__tests__/3-runtime-html/compose.spec.ts
```diff
@@ -172,3 +172,454 @@ describe(spec, function () {
...
+      class PlainViewModelClass {
+        public constructor() {
+          constructorCallCount++;
+        }
+        public activate(model: unknown) {
+          models2.push(model);
+        }
+      }
+      const { ctx, component, startPromise, tearDown } = createFixture(
+        `<au-compose view-model.bind="vm" model.bind="model">`,
```
Sayan751, line 472: Missing the end tag here. HTML parser is truly magical 🧙

Sayan751: Ohh it is missing for couple more times in the similar instances in this file.

bigopon: Hah it's for saving keystrokes by avoiding terminating tags

Sayan751:
Awesome work! Only critical comment I have is regarding the tests. I think there should have been more tests in terms of asserting the hook timings. Let us consider this scenario: we have a list of <view-model, model> pair. The view-model and model combination can be as follows:
- VM1 - ModelShape1
- VM2 - ModelShape2 (properties are mutually exclusive from ModelShape1)
- VM3 - ModelShape1 
- VM4 - ModelShape3 (extends ModelShape1 and add new properties)

Changes (can be performed in sequence or can be performed individually, after the above mentioned initial state)

```ts
// 1
list[3].model = new ModelShape1();
list[3].ViewModel = VM1;

// 2
list[3].model = new ModelShape2();
list[3].ViewModel = VM2;

//3
list[3].model = new ModelShape3();
list[3].ViewModel = VM4;
```

Now to assert that the last change prevails. And the preceding changes were noop or does not create any `activate`on error in the composed view models caused by incompatible view-models. 

Along on these lines it will also be great to assert the sequence of the lifecycle hooks (activation and deactivation) of all the VMs composed, especially when the `activate` as well as the other lifecycle hooks takes varied time to finish, based on long running promises returned.

PR #1186, opened 2021-06-14
fix(new-instance): correctly invoke a registered interface

### LINKED ISSUE #1166: newInstanceOf & newInstanceForScope not working as expected
In this gist, https://gist.dumber.app/?gist=308d35e1bdde0a717408fb9a91765efc
try injecting `newInstanceForScope(IListboxContext)` in `Listbox` will result in an error as it tries to instantiate the interface `IListboxContext`.

bigopon:
when using @newInstanceOf and @newInstanceForScope, even with an interface that already has a registered resolver, it will still throw as as it always tries to construct the interface, instead of the  corresponding implementation.

partially fixes #1166: this commit prepares a test where the most intuitive behavior is shown: ability to invoke an interface without having to declare it, if it has a default registration. Though this is inconsistent with the core, so will have to reconsider.

At the moment, it still requires the interface to be registered in order to be correctly invoked.

## ⏭ Next Steps

Find a balance between intuitiveness & consistency for `.get()` vs `@newInstanceOf()`/ `@newInstanceForScope()`

### DIFF
--- a/packages/__tests__/1-kernel/di.spec.ts
+++ b/packages/__tests__/1-kernel/di.spec.ts
@@ -1,6 +1,8 @@
 import {
   DI,
   inject,
+  newInstanceForScope,
+  newInstanceOf,
   Registration,
   singleton,
   transient,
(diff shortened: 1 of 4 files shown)


AureliaEffect:
The benchmarking comparison can be found here: https://au2-bench-viewer.azurewebsites.net/compare?branch=master&commit=8b8b080932f5a50dd3c92fe3e4840c0adb3e7fc0&branch=fix/di-new-instanceof-gh-1166&commit=362ccdbad4ec4d22d4dda387212a8488bad76dd9

AureliaEffect:
The benchmarking comparison can be found here: https://au2-bench-viewer.azurewebsites.net/compare?branch=master&commit=8b8b080932f5a50dd3c92fe3e4840c0adb3e7fc0&branch=fix/di-new-instanceof-gh-1166&commit=8753b4e085d4ef6f5dc7aeb6b9655e85e42a1249

PR #1191, opened 2021-06-20
Fix dialog-default-impl: display flex on overlay

ben-girardet:
In order for the dialog itself to be centered with the `margin: auto`, its direct parent must be `display: flex` which is the overlay and not the container.

# Pull Request

### DIFF
--- a/packages/runtime-html/src/templating/dialog/dialog-default-impl.ts
+++ b/packages/runtime-html/src/templating/dialog/dialog-default-impl.ts
@@ -30,8 +30,8 @@ export class DefaultDialogDomRenderer implements IDialogDomRenderer {
     Registration.singleton(IDialogDomRenderer, this).register(container);
   }
 
-  private readonly wrapperCss: string = baseWrapperCss;
-  private readonly overlayCss: string = `${baseWrapperCss} display:flex;`;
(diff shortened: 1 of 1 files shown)


bigopon:
@ben-girardet ahh, I see where the issue is. These 3 lines
```ts
const wrapper = dialogHost.appendChild(h('au-dialog-container', this.wrapperCss));
const overlay = wrapper.appendChild(h('au-dialog-overlay', this.overlayCss));
const host = overlay.appendChild(h('div', this.hostCss));
```
should have been like this

```diff
const wrapper = dialogHost.appendChild(h('au-dialog-container', this.wrapperCss));
const overlay = wrapper.appendChild(h('au-dialog-overlay', this.overlayCss));
const host = wrapper.appendChild(h('div', this.hostCss));
```
Can you try again with this?

ben-girardet:
Yes this works as well. And it follows the structure you describe in <https://docs.aurelia.io/plugins/dialog#the-default-dialog-renderer> so I guess that's the right fix ;-) Thanks

ben-girardet:
@bigopon I've reflected the suggested changes in this PR

bigopon:
Thanks @ben-girardet

PR #1192, opened 2021-06-20
refactor(template-compiler): merge binder & compiler

### LINKED ISSUE #1189: Template compiler shouldn't touch the content of a template without template controller on it
# 🐛 Bug Report

At the moment, if a template looks like this
```html
<h1>${message}</h1>
<template>${message}</template>
```

It will throw an error that looks like this
image

This is unexpected. Template should only be touched if there's at least one template controller associated with it.

### LINKED ISSUE #1190: templating(compiler): simpler & more hook-able compiler implementation
- better compilation pipeline (WIP: https://github.com/aurelia/aurelia/tree/refactor/template-compiler)
  - Easier to add support for compilation pipeline hooks

bigopon:
# Pull Request

## 📖 Description

At the moment, the template compilation is done in 2 phases:
1. resolving element & attribute information in a template into an intermediate symbols
2. compiling those symbols into instructions

One of the goals of this compilation process was the ability to reuse the compilation code to target multi platforms, with little modification. It looks simpler than it's actually is, because html platform is full of gotchas & special cases that need to be handled, resulting in the common layer -- the binder -- cluttered with special logic. Beside that, having 2 layers in compilation 
 process makes the code harder to understand and as a result, less friendly for contributors.

Another motivation for this refactoring is the feature & correctness parity with v1 compiler, where there' some nuances:
- #1189 `<template/>` element is always compiled, regardless there's a template controller on it or not.
- #1190 compilation hooks isn't simple: it can be done with the `define` hook for now, but this API is quite intimidating for something like scanning & modifying a template before it's compiled.

Further motivation for this refactoring is the way binding command works. At the moment, binding commands only receive processed information, which are symbols. This makes them more of a mapper than a command, and limits them to have any intelligent behavior determining the proper instruction for a given attribute usage on an element, within a context.

A bonus of this refactoring is that the new code should be friendlier to minification and thus, a good reduction of application bundle size.

Goal:
- Easier to understand compiler process & code, perf is a bonus.
- Minimal DOM apis usage: only use:
    ```ts
    Node.prototype.nodeName
    Node.prototype.nodeType
    Node.prototype.nextSibling
    Node.prototype.parentNode
    Node.prototype.insertBefore
    Node.prototype.removeChild
    Node.prototype.textContent
    ```
  so that it is simpler to reimplement the compiler for other environment (VS Code ext & other platform).
- Improve the ergonomic of `au-slot` in scenarios where it's used in conjunction with template controller.
- Binding command abstraction should be more powerful, transfer some responsibilities from attr transformers & existing compiler to the commands.
- Better minified code.
- Simpler to implement attribute transferring

## 👩‍💻 Reviewer Notes

- Binding command: Replace `compile` method with build, that takes the an object with the following interface as a paramter:
    ```ts
    interface BindingCommandInstance {
      build(info: ICommandBuildInfo): IInstruction;
    }

    interface ICommandBuildInfo {
      node: Element;
      attr: AttrSyntax;
      expr: AnyBindingExpression;
      bindable: BindableDefinition | null;
      def: CustomAttributeDefinition | CustomElementDefinition | null;
    }
    ```
- au-slot: it is now possible to support `[au-slot]` attribute on the same element, regardless order. This reduces boilerplate in some scenarios, like the following examples:
    Before:
    ```html
    <my-element>
      <template au-slot>
        <template if.bind="someCondition">
          <li>1</li>
          <li>2</li>
        </template>
      </template>
    </my-element>
    ````
    After:
    ```html
    <my-element>
      <template au-slot if.bind="someCondition">
        <li>1</li>
        <li>2</li>
      </template>
    </my-element>
    ````

  Order doesn't matter, so it can also be:
    After:
    ```html
    <my-element>
      <template if.bind="someCondition" au-slot>
        <li>1</li>
        <li>2</li>
      </template>
    </my-element>
    ```
- compilation: this is now done in a style that is similar to v1, where compiling each node would return the next node it wants to be compiled. And a context is built upon new compilation scope (slot/template controller). This should help with the ability to understand & extend/override/patch the behavior of the compiler.
- compilation: custom attribute is removed from the template during compilation
- compilation: custom attribute in multiple binding mode now always uses attribute form of bindable. This is for consistency with bindable on custom element, and style attribute:
    For a custom attribute `square`:
    ```ts
    class SquareCustomAttribute {
      @bindable size
      @bindable backgroundColor
    }
    ```

    template syntax before:
    ```html
    <div square="size: 50; backgroundColor: red">
    <!-- not aligned -->
    <div style="size: 50; background-color: red">
    ```
    After:
    ```html
    <div square="size: 50; background-color: red">
    <!-- aligned -->
    <div style="size: 50; background-color: red">
    ```

## 📑 Test Plan

- Most of existing tests work.
- Tweak binding command tests to adapt the addition of the new method, and the removal of the old method
- New tests for au slot for the new ergonomic features
- New tests for ensuring content inside `<template/>` without TC wouldn't be compiled
- Start grouping `<input/>` tests into `__tests__/3-runtime-html/input.spec.ts`

## ⏭ Next Steps

- Further refactoring of the compiler, to reduce the repetitive code, and let `processContent` be the right primitive for `<au-slot/>`, reducing more code.
- Ability to config the removal of instruction at compile time. For now just remove everything for better rendering perf. But user should be able to retain the original template for debugging purposes.
- Start associating element definition with instruction at compile time, so that it can reduce the runtime logic of controller + binding.
- Enhance `<au-compose/>` so that it can support both slotting, and binding instruction transferring. This should now be simpler to implement.
- Implement attribute/instruction transferring. This should now be simpler to implement.

closes #1189 
closes #1190


AureliaEffect:
The benchmarking comparison can be found here: https://au2-bench-viewer.azurewebsites.net/compare?branch=master&commit=05dddd7ea8ed00ca57e6b492b0088ee3bbac4298&branch=refactor/template-compiler&commit=17cc55d9bc7e5ce5fbac10319b2dfccc23d00e4b

bigopon:
Superseded by #1194

PR #1193, opened 2021-06-21
fix(jest): fix jest v27 exports

3cp:


bigopon:
Interesting, why didn't we see it fail before. Though it can still manifest in newer PR.
Thanks @3cp

AureliaEffect:
The benchmarking comparison can be found here: https://au2-bench-viewer.azurewebsites.net/compare?branch=master&commit=89adf8f6a5ee5b8809620e89afc19785c55010cc&branch=fix-jestv27-exports&commit=520ab8efc69ef76b6d56daa2ac0eaa1b5f6f1fc2

3cp:
It did not fail because we have no test setup for jest. We got only the unit tests for the functions.

3cp:
I am also running into issue of the latest ts-jest.
It now complains about cannot find source file for my-app.html.ts which doesn't exist but synthesised at testing time.
Only babel-jest works fine in jest v27 for now.

PR #1196, opened 2021-06-23
fix(let): camel-case let target when using with interpolation/literal

bigopon:
# Pull Request

## 📖 Description

At the moment, `<let/>` does not camelising the target property if it's an interpolation or literal:
```html
<let my-prop="1" my-p="${myProp}-2" >
```
will results in instructions against `my-prop` and `my-p` respectively, which is wrong. They should be `myProp` and `myP`. This PR fixes it.

### DIFF
--- a/packages/__tests__/3-runtime-html/let.spec.ts
+++ b/packages/__tests__/3-runtime-html/let.spec.ts
@@ -1,0 +1,66 @@
+import { ILogger } from '@aurelia/kernel';
+import { assert, createFixture } from '@aurelia/testing';
+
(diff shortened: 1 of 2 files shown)


AureliaEffect:
The benchmarking comparison can be found here: https://au2-bench-viewer.azurewebsites.net/compare?branch=master&commit=948c681008344f7163d738c1002ada0222429d3a&branch=fix/let-target&commit=bee73ccaae6c3f07d9e17cfc45b14a445dea7f52

PR #1198, opened 2021-06-24
refactor(templating): change custom element own container timing

bigopon:
# Pull Request

## 📖 Description

At the moment, the container of a custom element is created after the corresponding view model has been instantiated. This creates unexpected behavior in the scope of the dependencies. Example is `@newInstanceForScope` would register the dependencies onto the parent container instead of the CE own one:
```html
app.html
<list ref="list1"></list>
<list ref="list2"></list>
```
```ts
class Model {}

class List {
  constructor(@newInstanceForScope(Model) model) {}
}
```
Would result in 2 instances of `Model` scoped to with the app component container, instead of the 2 containers of `List` components.

Other changes:

- Custom element own container now registers the constructor with the instance, allowing its children to inject based on the constructor, beside the injectable token API. This is to align the behavior of container with v1. Custom attribute container behavior is kept as is: no container is instantiated for custom attributes or template controllers. This is a nice optimization in v2.
- `getRenderContext` now uses container given as is, instead of treating it like parent container. This should allow renderer and `createElement` API to have much better control over the timing of the hydration process, and less prone to cache invalidation issues.
- tweak `au-render` & `RenderPlan` to adopt `getRenderContext` changes.

### 🎫 Issues

A related issue is at #1166 where `@newInstanceOf` and `@newInstanceForScope` of an interface is not working. It's probably should be kept this way, since we don't JIT register an interface, those 2 resolvers shouldn't do it either.

## ⏭ Next Steps

- Prepare a new injectable `IHydrationContext` so that `<au-slot/>` can inject & retrieve the scope of its bindings, instead of having to walk the controller tree. https://github.com/aurelia/aurelia/blob/6d4685a5039f9d397ddd7e268b22fd7985373e56/packages/runtime-html/src/resources/custom-elements/au-slot.ts#L54-L66
- Consider removing injectable token API of custom element definition, since this is now achievable in user land without any boilerplate.
- Tweak the compiler and hydrate CE/CA/TC instruction so that they accept the constructor instead of resource name, to avoid having to resolve the resources again during hydration. This should give a some perf improvement.

### DIFF
--- a/packages/kernel/src/di.ts
+++ b/packages/kernel/src/di.ts
@@ -200,76 +200,7 @@ export const DI = {
     }
     return annotationParamtypes;
   },
-  getDependencies(Type: Constructable | Injectable): Key[] {
-    // Note: Every detail of this getDependencies method is pretty deliberate at the moment, and probably not yet 100% tested from every possible angle,
-    // so be careful with making changes here as it can have a huge impact on complex end user apps.
-    // Preferably, only make changes to the dependency resolution process via a RFC.
-
-    const key = Protocol.annotation.keyFor('di:dependencies');
-    let dependencies = Metadata.getOwn(key, Type) as Key[] | undefined;
-    if (dependencies === void 0) {
-      // Type.length is the number of constructor parameters. If this is 0, it could mean the class has an empty constructor
-      // but it could also mean the class has no constructor at all (in which case it inherits the constructor from the prototype).
-
-      // Non-zero constructor length + no paramtypes means emitDecoratorMetadata is off, or the class has no decorator.
-      // We're not doing anything with the above right now, but it's good to keep in mind for any future issues.
-
-      const inject = (Type as Injectable).inject;
-      if (inject === void 0) {
-        // design:paramtypes is set by tsc when emitDecoratorMetadata is enabled.
-        const designParamtypes = DI.getDesignParamtypes(Type);
-        // au:annotation:di:paramtypes is set by the parameter decorator from DI.createInterface or by @inject
-        const annotationParamtypes = DI.getAnnotationParamtypes(Type);
-        if (designParamtypes === void 0) {
-          if (annotationParamtypes === void 0) {
-            // Only go up the prototype if neither static inject nor any of the paramtypes is defined, as
-            // there is no sound way to merge a type's deps with its prototype's deps
-            const Proto = Object.getPrototypeOf(Type);
-            if (typeof Proto === 'function' && Proto !== Function.prototype) {
-              dependencies = cloneArrayWithPossibleProps(DI.getDependencies(Proto));
-            } else {
-              dependencies = [];
-            }
-          } else {
-            // No design:paramtypes so just use the au:annotation:di:paramtypes
-            dependencies = cloneArrayWithPossibleProps(annotationParamtypes);
-          }
-        } else if (annotationParamtypes === void 0) {
-          // No au:annotation:di:paramtypes so just use the design:paramtypes
-          dependencies = cloneArrayWithPossibleProps(designParamtypes);
-        } else {
-          // We've got both, so merge them (in case of conflict on same index, au:annotation:di:paramtypes take precedence)
-          dependencies = cloneArrayWithPossibleProps(designParamtypes);
-          let len = annotationParamtypes.length;
-          let auAnnotationParamtype: Key;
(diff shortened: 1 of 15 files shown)


packages/runtime-html/src/templating/controller.ts
```diff
@@ -310,7 +318,7 @@ export class Controller<C extends IViewModel = IViewModel> implements IControlle
     }
 
     // todo: make projections not influential on the render context construction
-    const context = this.context = getRenderContext(definition, parentContainer, hydrationInst?.projections) as RenderContext;
+    const context = this.context = getRenderContext(definition, this.container, hydrationInst?.projections) as RenderContext;
```
bigopon, line 321: This is the main change, where it would create a new render context based on its own container, instead of the parent one. Later, we can probably remove the `parentContainer` from the parameters too.

packages/runtime-html/src/templating/render-context.ts
```diff
@@ -119,6 +128,27 @@ export interface ICompiledRenderContext extends IRenderContext {
     auSlotsInfo?: IAuSlotsInfo,
   ): IComponentFactory;
 
+  /**
+   * Prepare a new container to associate with a custom element instance
+   */
+  createElementContainer(
```
bigopon, line 134: These two methods do not have symmetry, but it should be fine for now. Later when contracting the APIs of `IRenderContext`, these probably won't stay this way.

packages/runtime-html/src/templating/render-context.ts
```diff
@@ -169,16 +199,17 @@ export interface IComponentFactory extends ICompiledRenderContext, IDisposable {
   dispose(): void;
 }
 
+let renderContextCount = 0;
 export function getRenderContext(
   partialDefinition: PartialCustomElementDefinition,
-  parentContainer: IContainer,
```
bigopon, line 174: Building & caching something based on a parent container, to then have a new child created is a bit harder to control than just treating the container itself as is, so change the behavior to this. Also renamed the parameter to make it less confusing.

packages/runtime-html/src/templating/render-context.ts
```diff
@@ -489,11 +526,144 @@ export class RenderContext implements IComponentFactory {
     if (auSlotsInfo !== void 0) {
       this.auSlotsInfoProvider.prepare(auSlotsInfo);
     }
+    if (this.resourceInvoker == null) {
```
bigopon, line 529: This is the invoker used for all custom attributes & template controllers. It works similarly like the current `IComponentFactory`

bigopon (commented):
@fkleuver @Sayan751 some small comments that may help for when you have time to review later.

AureliaEffect:
The benchmarking comparison can be found here: https://au2-bench-viewer.azurewebsites.net/compare?branch=master&commit=6d4685a5039f9d397ddd7e268b22fd7985373e56&branch=fix/ce-container-timing&commit=ac0c2902bbea4eacd3f95b56dd8e1fee145ce463

AureliaEffect:
The benchmarking comparison can be found here: https://au2-bench-viewer.azurewebsites.net/compare?branch=master&commit=6d4685a5039f9d397ddd7e268b22fd7985373e56&branch=fix/ce-container-timing&commit=02bb7968e8fdd8346fc8e86c1f3769eaec400658

packages/kernel/src/di.ts
```diff
@@ -427,6 +358,79 @@ export const DI = {
   },
 };
 
+function getDependencies(Type: Constructable | Injectable): Key[] {
```
Sayan751, line 361: Makes it harder to review the changes :(

packages/runtime-html/src/app-root.ts
```diff
@@ -78,9 +78,9 @@ export class AppRoot implements IDisposable {
     this.host = config.host;
     this.work = container.get(IWorkTracker);
     rootProvider.prepare(this);
-    if (container.has(INode, false) && container.get(INode) !== config.host) {
-      this.container = container.createChild();
-    }
+    // if (container.has(INode, false) && container.get(INode) !== config.host) {
+    //   this.container = container.createChild();
+    // }
```
Sayan751, line 83: Missed cleanup

bigopon: Thanks. This is temporarily kept for some more cleanup & refactoring around here. Will cleanup at some point soon.

PR #1202, opened 2021-06-27
feat(syntax): Nullish coalescing op ??

bigopon:
# Pull Request

## 📖 Description

Add support for `??` in the expression parser, and template. Now it's possible to do this
```html
<el value.bind="prop1 ?? prop2">
```

An issue with Aurelia expression behavior is that Aurelia automatically coerces `null`/`undefined` to empty string in non strict mode, which won't work with `??`.
We will probably need to start thinking of adding safe navigation operator `?.` and revisit strict vs non-strict mode soon.

Thanks @fkleuver for the parser work.


HamedFathi:
Can we have `?.` too? (for array and ...)

E4vIlQ-WYAgk07i

bigopon:
@HamedFathi thanks for reminding about that issue. That probably should be added too. I'm not sure whether it should be added together with this, or a separate one.
cc @fkleuver

AureliaEffect:
The benchmarking comparison can be found here: https://au2-bench-viewer.azurewebsites.net/compare?branch=master&commit=13a730fabf730bd59498ad4d84fc14361fd922b3&branch=nullish-coalescing-op&commit=ea3371707d9156d9d798b2f602449ba0921b7b4f

fkleuver:
> I'm not sure whether it should be added together with this, or a separate one.

Doesn't matter, there's no breaking changes or conflicts etc either way. I can't predict very well when I'll have time though at the moment.

milkshakeuk:
are these things not going to make it into v2?

bigopon:
> are these things not going to make it into v2?

@milkshakeuk `null`/`undefined` are often treated as an empty string, hence the mention of strict vs non strict mode (default). Applications normally prefer non strict mode I'd assume, so while it could be useful, it's not a lot more meaningful yet. Though eventually this and `?.` should be in.

PR #1206, opened 2021-06-28
refactor(au-slot): lazily determine view factory at instantiation time

### LINKED ISSUE #1203: Binding context methods are not called from au-slot
# 🐛 Bug Report

Consider the following
```html
    <mdc-tab-bar>
      <template au-slot>
        <mdc-tab label="Examples" class="viewer-tab" active="true" click.trigger="loadExamples()"></mdc-tab>
        <mdc-tab label="Api" class="viewer-tab" click.trigger="loadApi()"></mdc-tab>
      </template>
    </mdc-tab-bar>
```

## 🤔 Expected Behavior

Methods are called on the binding context

## 😯 Current Behavior
Error is thrown
```
Uncaught Error: Expected 'loadApi' to be a function
    at getFunction (ast.js:1155)
    at CallScopeExpression.evaluate (ast.js:623)
    at Listener.callSource (listener.js:28)
    at Listener.handleEvent (listener.js:36)
```
The methods are searched on `mdc-tab-bar` view model, thus not reaching higher binding context

## 🌍 Your Environment

| Aurelia          | alpha.9

### LINKED ISSUE #1205: request(au-slot): au-slot should work with `@children`, or an equivalent for slotting observation
# 🙋 Feature Request

At the moment, `<au-slot>` does not provides a way to toggle between its content coming from projection or fallback , when the content from projection has been deactivated (1), nor a way to react to whenever its content has been possibly changed (2). Consider adding an API to enable (2) as (1) isn't easy, or non-invasive to implement.

Also, there should be an event like `slotchange` event of native slot:
```html
<slot slotchange.trigger="onContentChange()">
```
so the API may look like this
```html
<au-slot slotchange.trigger="onContentChange()">
```

or a higher level decorator similar to `@children`, maybe `@auSlotChildren`

This could be implemented by either:
- Make `View` subscribe-able, so subscriber can at least know when a view is deactivated and react to.
- Use a mutation observer on the immediate parent of the au-slot comment marker location and observer changes of the nodes

bigopon:
# Pull Request

## 📖 Description

At the moment, `<au-slot/>` view factory is determined at compilation time, which is prone to cached compilation issue, also prohibits the ability to toggle between projected content and fallback content (if there's projection, do not render fallback and vice versa) to align `<au-slot/>` behavior with native `<slot/>`. Refactor this, and move factory creation to the instantiation time of `AuSlot` so that there can be more flexibility in terms on scope choosing, factory constructing, and ensure correctness in tricky situation like 3 layers of pass through slot seen in this issue #1205.
In order to support this, `IHydrationContext` has been enhanced with a `parent` property so that it can build the proper hydration context hierarchy, allowing `<au-slot/>` to pick the right scope.

### 🎫 Issues

Closes #1203

## 👩‍💻 Reviewer Notes

- In `HydrateElementRenderer`, there's no longer a need for special treatment related to au-slot factory. With this, we can probably remove quite a number of parameters across templating calls.
- Render context shouldn't be bothered by projections anymore, and we can turn `getRenderContext` back to its previous, purer form: only take a definition and a container.
- `Controller.hydrateCustomElement` now builds the `IHydrationContext` hierarchy, instead of isolated pieces.
- `InstanceProvider` now accepts a 2nd param to remove boilerplate and intermediate variables
- Remove `AuSlotInfo` and `SlotContentType`, as they are no longer needed. Fallback and projections are always compiled and included in the instructions.

## 📑 Test Plan

All existing tests works and add tests covering the scenarios at the linked issues. Also added a test for tricky combination like this
```html
<mdc-tab-bar>
    <mdc-tab au-slot repeat.for="i of 3" id="mdc-\${id}-\${i}" click.trigger="increase()">\${count + i}</mdc-tab>
```
with els:
```ts
CustomElement.define({ name: 'mdc-tab-scroller', template: '<au-slot>' }),
CustomElement.define({ name: 'mdc-tab-bar', template: '<mdc-tab-scroller><au-slot au-slot></au-slot></mdc-tab-scroller>' }),
CustomElement.define({ name: 'mdc-tab', template: '<button>Tab</button>' }),
```

## ⏭ Next Steps

- start removing unused bits related to projects caching in render context
- start removing unused bits related to projections marking in compilation
- employ root level caching for compilation, as an optimization, as now there's no need to recompile CE definition anymore.

### DIFF
--- a/packages/runtime-html/src/resources/custom-elements/au-slot.ts
+++ b/packages/runtime-html/src/resources/custom-elements/au-slot.ts
@@ -1,9 +1,10 @@
 import { DI } from '@aurelia/kernel';
 import { IRenderLocation } from '../../dom.js';
+import { customElement } from '../custom-element.js';
 import { IInstruction } from '../../renderer.js';
 import { IHydrationContext } from '../../templating/controller.js';
+import { getRenderContext } from '../../templating/render-context.js';
 import { IViewFactory } from '../../templating/view.js';
-import { customElement } from '../custom-element.js';
 
 import type { Writable } from '@aurelia/kernel';
 import type { LifecycleFlags, Scope } from '@aurelia/runtime';
@@ -29,21 +30,32 @@ export class SlotInfo {
 
 export class AuSlot implements ICustomElementViewModel {
   /** @internal */
-  public static get inject() { return [IViewFactory, IRenderLocation, IInstruction, IHydrationContext]; }
+  public static get inject() { return [/* IViewFactory, */ IRenderLocation, IInstruction, IHydrationContext]; }
 
-  public readonly view: ISyntheticView;
+  public readonly view!: ISyntheticView;
   public readonly $controller!: ICustomElementController<this>; // This is set by the controller after this instance is constructed
 
   private hostScope: Scope | null = null;
   private outerScope: Scope | null = null;
+  private hasProjection: boolean = false;
 
   public constructor(
-    factory: IViewFactory,
+    // factory: IViewFactory,
     location: IRenderLocation,
     private readonly instruction: HydrateElementInstruction,
     private readonly hdrContext: IHydrationContext,
   ) {
+    let factory: IViewFactory;
+    const slotInfo = instruction.auSlot!;
+    const projection = hdrContext.instruction?.projections?.[slotInfo.name];
+    if (projection == null) {
+      factory = getRenderContext(slotInfo.fallback, hdrContext.controller.context.container).getViewFactory()
+    } else {
+      factory = getRenderContext(projection, hdrContext.parent!.controller.context.container).getViewFactory();
+      this.hasProjection = true;
+    }
     this.view = factory.create().setLocation(location);
+    // this.view = factory.create().setLocation(location);
   }
 
   public binding(
@@ -52,9 +64,10 @@ export class AuSlot implements ICustomElementViewModel {
     _flags: LifecycleFlags,
   ): void | Promise<void> {
     this.hostScope = this.$controller.scope.parentScope!;
-    this.outerScope = this.instruction.slotInfo!.type === AuSlotContentType.Projection
-      ? this.hdrContext.controller.scope.parentScope! ?? null
-      : this.hostScope;
+    this.outerScope = /* this.instruction.slotInfo!.type === AuSlotContentType.Projection */
+      this.hasProjection
+        ? this.hdrContext.controller.scope.parentScope! ?? null
+        : this.hostScope;
   }
 
   public attaching(
--- a/packages/runtime-html/src/resources/custom-elements/au-slot.ts
+++ b/packages/runtime-html/src/resources/custom-elements/au-slot.ts
@@ -15,47 +15,32 @@ import type { HydrateElementInstruction } from '../../renderer.js';
 export type IProjections = Record<string, CustomElementDefinition>;
 export const IProjections = DI.createInterface<IProjections>("IProjections");
 
-export enum AuSlotContentType {
-  Projection,
-  Fallback,
-}
-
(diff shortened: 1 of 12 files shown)


packages/runtime-html/src/resources/custom-elements/au-slot.ts
```diff
@@ -14,35 +15,31 @@ import type { HydrateElementInstruction } from '../../renderer.js';
...
+    instruction: HydrateElementInstruction,
     private readonly hdrContext: IHydrationContext,
   ) {
+    let factory: IViewFactory;
+    const slotInfo = instruction.auSlot!;
+    const projection = hdrContext.instruction?.projections?.[slotInfo.name];
+    if (projection == null) {
+      factory = getRenderContext(slotInfo.fallback, hdrContext.controller.context.container).getViewFactory();
+    } else {
+      factory = getRenderContext(projection, hdrContext.parent!.controller.context.container).getViewFactory();
```
bigopon, line 40: if there is `<au-slot/>` in the projection, which means the template looks like this:
```html
<my-element>
  <au-slot au-slot></au-slot>
</my-element>
```
this logic will allow the projected `<au-slot/>` to keep walking up the hierarchy of hydration context and find the right scope/factory. Everything passed through to the end.

packages/runtime-html/src/template-compiler.ts
```diff
@@ -552,7 +550,7 @@ export class TemplateCompiler implements ITemplateCompiler {
             // if not a bindable, then ensure plain attribute are mapped correctly:
             // e.g: colspan -> colSpan
             //      innerhtml -> innerHTML
-            //      minlength -> minLengt etc...
+            //      minlength -> minLength etc...
```
bigopon, line 553: at least the real mapping code didn't have this typo

packages/runtime-html/src/template-compiler.ts
```diff
@@ -619,51 +617,44 @@ export class TemplateCompiler implements ITemplateCompiler {
...
-      if (isAuSlot) {
+      elementInstruction = new HydrateElementInstruction(
+        elDef.name,
+        void 0,
+        (elBindableInstructions ?? emptyArray) as IInstruction[],
+        null,
+      );
+
+      if (elName === 'au-slot') {
         const slotName = el.getAttribute('name') || /* name="" is the same with no name */'default';
```
bigopon, line 628: `fallback` content is now always built for `<au-slot/>`

packages/runtime-html/src/templating/controller.ts
```diff
@@ -211,11 +212,16 @@ export class Controller<C extends IViewModel = IViewModel> implements IControlle
...
-    hydrationContextProvider.prepare(new HydrationContext(controller, hydrationInst));
 
     ownCt.register(...definition.dependencies);
-    ownCt.registerResolver(IHydrationContext, hydrationContextProvider);
+    ownCt.registerResolver(IHydrationContext, new InstanceProvider(
+      'IHydrationContext',
+      new HydrationContext(
+        controller,
+        hydrationInst,
+        /* parent context */contextCt.get(optional(IHydrationContext)),
```
bigopon, line 222: parent hydration context to allow more powerful features, now (`<au-slot/>` and later: devtools, etc...)

bigopon (commented):
@fkleuver @Sayan751 I've added a few comments to help make the review easier for when you have time.

AureliaEffect:
The benchmarking comparison can be found here: https://au2-bench-viewer.azurewebsites.net/compare?branch=master&commit=2cbfe9240c7c1dc9a26e13d51f56dfff3d8e7736&branch=refactor/render-context_n_au-slot-scope&commit=58cc0b52d2eee401b7182749b13887e812c4bba1

PR #1208, opened 2021-06-29
chore(all): cleanup

bigopon:
# Pull Request

## 📖 Description

Currently the `blur` custom attribute isn't used much, and they way it's supposed to be used isn't always intuitive. Remove it and the corresponding doc.

Features:
- template-compiler: start recognizing `containerless` attribute on the custom element. This is a bit uncharted territory at the moment in v2, will probably have to tweak some core templating stuff to maintain v1 behavior, if we are to keep it.
- au-compose: Add an `containerless` support for this specific element as an example for discussion/API shaping.


packages/store-v1/src/decorator.ts
```diff
@@ -64,12 +64,12 @@ export function connectTo<T, R = any>(settings?: ((store: Store<T>) => Observabl
   return function (target: any) {
     const originalSetup = typeof settings === 'object' && settings.setup
       ? target.prototype[settings.setup]
-      : target.prototype.beforeBind;
+      : target.prototype.binding;
```
bigopon, line 67: This plugin is missing you @zewa666 , and the container based dispatcher too 😉

PR #1211, opened 2021-06-30
feat(di): @factory resolver

### LINKED ISSUE #1209: request(di): port Factory.of resolver from v1
# 🙋 Feature Request

Related Discourse discussion https://discourse.aurelia.io/t/will-factory-of-be-supported-in-aurelia-2/4282

bigopon:
# Pull Request

## 📖 Description

- **add**: `@factory` resolver builder. The usage can be like the following:
  1. With parameter decorator:
    ```ts
    import { IResolvedFactory } from '@aurelia/kernel';

    class MyClass {
      constructor(@factor(MyService) serviceFactory: IResolvedFactory<MyService>) {
        const service1 = serviceFactory();
        const service2 = serviceFactory(url, params, interceptors);
      }
    }
    ```
  2. with `@inject`
    ```ts
    import { inject, factory, IResolvedFactory } from '@aurelia/kernel';

    @inject(factory(MyService))
    class MyClass {
      constructor(serviceFactory: IResolvedFactory<MyService>) {
        const service1 = serviceFactory();
        const service2 = serviceFactory(url, params, interceptors);
      }
    }
    ```
  3. with static `inject`
    ```ts
    import { factory, IResolvedFactory } from '@aurelia/kernel';

    class MyClass {
      static inject = [factory(MyService)];
      constructor(serviceFactory: IResolvedFactory<MyService>) {
        const service1 = serviceFactory();
        const service2 = serviceFactory(url, params, interceptors);
      }
    }
    ```
- **remove**: interface resolution in `newInstanceOf` and `newInstanceForScope`. This is to align the interface instantiation story with the rest of the kernel. To add this behavior back in, they should be done in `ResolverBuilder`, where it would register an explicit factory for the interface instead.
- **add**: Better type inference for `@lazy`, `@newInstanceOf`, `@newInstanceForScope` when using with `IContainer.get`, should be able to infer the resolved type now.

### 🎫 Issues

Closes #1209

## ⏭ Next Steps

- Implement auto registration for interface factory in `ResolverBuilder` for strategy singleton and transient, so container can know how to instantiate interfaces.


packages/__tests__/1-kernel/di.get.spec.ts
```diff
@@ -543,4 +543,223 @@ describe('DI.get', function () {
...
+      const childContainer = container.createChild();
+      const instance = childContainer.get(newInstanceForScope(I));
+      assert.strictEqual(iCallCount, 1);
+      assert.instanceOf(instance, IImpl);
+
+      const instance2 = childContainer.get(I);
+      assert.strictEqual(iCallCount, 1);
+      assert.instanceOf(instance2, IImpl);
+      assert.strictEqual(instance, instance2);
+    });
```
Sayan751, line 682: I think here we are missing another assertion. That is `assert.strictEqual(instance, childContainer.createChild().get(I));`.

packages/kernel/src/di.ts
```diff
@@ -620,39 +631,37 @@ export function ignore(target: Injectable, property?: string | number, descripto
 }
 ignore.$isResolver = true;
 ignore.resolve = () => undefined;
+
+export const factory = createResolver((key: any, handler: IContainer, requestor: IContainer) => {
+  return (...args: unknown[]) => handler.getFactory(key).construct(requestor, args);
+}) as <K>(key: K) => IFactoryResolver<K>;
```
Sayan751, line 637: Awsome!

Sayan751 (commented):
Great addition!!

packages/__tests__/1-kernel/di.get.spec.ts
bigopon: Thanks. It's done in https://github.com/aurelia/aurelia/pull/1246

PR #1212, opened 2021-06-30
feat(template-compiler): add beforeCompile hooks

bigopon:
# Pull Request

## 📖 Description

A missing capabilities of v1 is the compilation hooks. This is a very focused hooks of the compilation that should be much more ergonomic than the existing define hooks, which is also quite intimidating. Probably the `define` hooks can be partially replaced by a `before` + `after` compilation hooks instead.

The interface of a hook is:
```ts
interface ITemplateCompilerHooks {
  beforeCompile(template: HTMLElement): void;
}
```
Note that `Hooks` in the name indicates that there may be more hooks in the future, beside `beforeCompile`.

## 👩‍💻 Reviewer Notes

- **added**: `allResources` resolver builder. This is to be used for many resources like injectable: `IAttrPattern`, `IBindingCommand` etc...

## 📑 Test Plan

- Existing tests work
- Verify the resources are retrieved with resources semantic

## ⏭ Next Steps

- Refactor the compilation process & start caching for perf improvement.

### DIFF
--- a/docs/user-docs/TOC.md
+++ b/docs/user-docs/TOC.md
@@ -85,6 +85,7 @@
   * [DOM Style Injection](developer-guides/scenarios/dom-style-injection.md)
   * [Markdown Integration](developer-guides/scenarios/markdown-integration.md)
 * [Extending templating syntax](developer-guides/extending-templating-syntax.md)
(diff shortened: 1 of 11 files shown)


bigopon:
docs is now available at https://docs.aurelia.io/developer-guides/template-compiler-hooks

PR #1215, opened 2021-07-03
fix(watch): build scope for watching attributes with expression

bigopon:
# Pull Request

## 📖 Description

Custom attribute does not have its own scope, though for watching using the ASTs, there needs to be one. Build one on the fly instead for expression watchers for custom attribute instead.

### 🎫 Issues
The follow throws an error with message: `Cannot read property bindingContext of undefined`
```ts
@customAttribute('my-attribute')
class Abc {
  items: any[];

  @watch('items.length')
  filterChanged() {}
}
```

Thanks @Vheissu

- Fixed: Ensure order of `@watch` decorator is insensitive
- Docs: rename Template Compiler doc section, added some more doc on toggling debug behavior


packages/runtime-html/src/watch.ts
```diff
@@ -76,10 +80,24 @@ export function watch<T extends object = object>(
...
+    //
+    // then @watch is called after @customAttribute
+    // which means the attribute definition won't have the watch definition
+    //
+    // temporarily works around this order sensitivity by manually add the watch def
+    // manual
+    if (CustomAttribute.isType(Type)) {
+      CustomAttribute.getDefinition(Type).watches.push(watchDef);
+    }
+    if (CustomElement.isType(Type)) {
```
Sayan751, line 98: I think we need a `else if` here. The same class cannot be a custom element and a custom attribute at the same time.

PR #1222, opened 2021-07-07
call priceUpdate and create file for value converter

marekpetak:
# Pull Request

## 📖 Description

the priceUpdate function has is not automatically called, so just added it to binding function and the currency value converter file create description
### 🎫 Issues

No issues, just added some details, on which I got stuck for a while. Thank you.


bigopon:
Thanks @marekpetak

Issue #1238, opened 2021-07-12
au-slot: require explicit declaration of what to expose on $host binding


bigopon:
# RFC

## Context

- At the moment, accessing `$host` is treated as a special case inside Aurelia expressions. It's understood by the runtime that this is a different kind of scope being accessed. This has an implication on the performance and complexity of runtime code around scope and binding context, as there are more possibilities to consider when looking up for a name. Consider the following template for a component that loads profile based on id to provide data for editor fields, that can be overridden by the consuming component

```html
profile-editor.html
<div promise.bind="loadProfile(id)">
  <div then.bind="profile">  
    <au-slot>
      <default-edtior data.bind="profile">
    </au-slot>
  </div>
</div>
```

If the `<profile-editor/>` is used like this:

```html
app.html

<profile-editor id.bind="id">
```
then the default editor will be used, and nothing special to consider.

If the `<profile-editor/>` is used like this:

```html
app.html

<profile-editor id.bind="id">
  <input value.bind="firstName">
</profile-editor>
```
There' at least 3 possible cases:
  - if there's a `firstName` property on the owning scope of `<profile-editor/>` inside `app.html`, then `value` property on the `<input/>` will be bound with this scope.
  - if `firstName` exist in `$host` scope, which is the scope created by `promise` attribute inside `<profile-editor/>`, then the binding of `value` property on `<input/>` element will be bound with this scope. If immediate scope inside `<profile-edtior/>` does not have `firstName` property, then runtime will walk up the scope hierarchy until `<profile-editor/>` own scope to find such property.
  - if neither the previous solution has matching scope, then the immediate scope surrounding `<profile-editor/>` will be used to bind with `<input/>` `value` property.

- To accommodate this, templating and the ASTs also need some special knowledge to deal with host scope, which manifests in AST signature, and hostScope property of `IController`.

The above complexity are natural if we do au slot scope the existing way. But they can be simplified, to better leverage the existing Aurelia scope & AST infra.

## Changes

- Instead of treating `$host` as a completely different kind of scope "selector", treat it as a normal property. This requires the scope created for the view controlled by `<au-slot/>` to have `$host` property assigned on it properly. This can properly be done via a binding, through a bindable on `<au-slot/>`. The template can look like this:

```html
<div>
  <au-slot scope.bind="{ propa: valueA, propB: valueB }">
```
[...]

PR #1239, opened 2021-07-13
refactor(scope): simplify au slot host scope

bigopon:
# Pull Request

## 📖 Description

Implements RFC #1238 
This greatly reduces the complexity in every thing, also opens a few opportunity to simplify our infra even more.

### DIFF
--- a/packages/__tests__/3-runtime-html/au-slot.spec.ts
+++ b/packages/__tests__/3-runtime-html/au-slot.spec.ts
@@ -85,7 +85,7 @@ describe('au-slot', function () {
       public readonly only: boolean = false,
     ) { }
   }
-  function* getTestData() {
+  function *getTestData() {
     const createMyElement = (template: string) => {
       class MyElement {
         public constructor(
@@ -373,7 +373,7 @@ describe('au-slot', function () {
           <h4>Last Name</h4>
         </au-slot>
         <template repeat.for="person of people">
-          <au-slot name="content">
+          <au-slot name="content" expose.bind="{ person, $index, $even, $odd }">
             <div>\${person.firstName}</div>
             <div>\${person.lastName}</div>
           </au-slot>
@@ -1663,6 +1663,40 @@ describe('au-slot', function () {
         {
           'div': [`content`, undefined],
         },
+      );
+    }
+
+    {
+      yield new TestData(
+        'updates expose binding on <au-slot/> dynamically',
+        `<my-element><div au-slot>\${$host.value}</div>`,
+        [createMyElement('<input value.bind="message"/><au-slot expose.bind="{ value: message }">')],
+        {
+          'my-element': ['<input class="au"><div>undefined</div>', undefined]
+        },
+        function ({ host, platform }) {
+          const input = host.querySelector('input');
+          input.value = 'hello';
+          input.dispatchEvent(new platform.CustomEvent('change'));
+          platform.domWriteQueue.flush();
+          assert.strictEqual(host.querySelector('div').textContent, 'hello');
+        }
+      );
+
+      yield new TestData(
+        'exposure of host context does not affect inner binding contexts',
+        `<my-element>`,
+        [createMyElement(`<input value.bind="message"/><au-slot expose.bind="{ value: message }">\${message}</au-slot>`)],
+        {
+          'my-element': ['<input class="au">undefined', undefined]
+        },
+        function ({ host, platform }) {
+          const input = host.querySelector('input');
+          input.value = 'hello';
+          input.dispatchEvent(new platform.CustomEvent('change'));
+          platform.domWriteQueue.flush();
+          assert.strictEqual(host.querySelector('my-element').textContent, 'hello');
+        }
       );
     }
 
--- a/packages/__tests__/validation-html/validate-binding-behavior.spec.ts
+++ b/packages/__tests__/validation-html/validate-binding-behavior.spec.ts
@@ -237,10 +237,10 @@ describe('validate-binding-behavior', function () {
   }
   @bindingBehavior('vanilla')
   class VanillaBindingBehavior implements BindingBehaviorInstance {
-    public bind(_flags: LifecycleFlags, _scope: Scope, _hostScope: Scope | null, _binding: IBinding): void {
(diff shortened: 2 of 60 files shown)


packages/runtime-html/src/resources/custom-elements/au-slot.ts
```diff
@@ -43,18 +49,37 @@ export class AuSlot implements ICustomElementViewModel {
...
+      // via overlaying the outerscope with another scope that has
+      // - binding context & override context pointing to the outer scope binding & override context respectively
+      // - override context has the $host pointing to inner scope binding context
+      outerScope = this.hdrContext.controller.scope.parentScope!;
+      overlayedOuterScope = this.outerScope = Scope.create(
+        outerScope.bindingContext,
+        OverrideContext.create(outerScope.bindingContext),
+        false
+      );
+      overlayedOuterScope.parentScope = outerScope;
```
Sayan751, line 67:
```suggestion
      overlayedOuterScope = this.outerScope = Scope.fromParent(outerScope, outerScope.bindingContext);
```

packages/runtime-html/src/resources/custom-elements/au-slot.ts
```diff
@@ -43,18 +49,37 @@ export class AuSlot implements ICustomElementViewModel {
     _parent: IHydratedParentController,
     _flags: LifecycleFlags,
   ): void | Promise<void> {
-    this.hostScope = this.$controller.scope.parentScope!;
-    this.outerScope = this.hasProjection
-      ? this.hdrContext.controller.scope.parentScope
-      : this.hostScope;
+    this.parentScope = this.$controller.scope.parentScope!;
+    let outerScope: Scope;
+    let overlayedOuterScope: Scope;
```
Sayan751, line 54: Can we simply use the name `outerScope` instead? I am not sure what you are trying to mean by "overlay".

packages/__tests__/3-runtime-html/au-slot.spec.ts
```diff
@@ -1666,6 +1666,40 @@ describe('au-slot', function () {
...
+        function ({ host, platform }) {
+          const input = host.querySelector('input');
+          input.value = 'hello';
+          input.dispatchEvent(new platform.CustomEvent('change'));
+          platform.domWriteQueue.flush();
+          assert.strictEqual(host.querySelector('my-element').textContent, 'hello');
+        }
+      );
+    }
+
```
Sayan751, line 1702: Missing test: expose rename the property, such as : `expose.bind="{foo: person}"`. then in projection use `$host.person` and `$foo`, and assert the former is `undefined`, whereas the later matches with the "exposed" `person`.

packages/__tests__/validation-html/validate-binding-behavior.spec.ts
```diff
@@ -1275,26 +1275,29 @@ describe('validate-binding-behavior', function () {
...
+
+  //     target.value = 'foo';
+  //     await assertEventHandler(target, 'change', 0, platform, app.controllerValidateBindingSpy, app.controllerValidateSpy, ctx);
+  //     await assertEventHandler(target, 'focusout', 1, platform, app.controllerValidateBindingSpy, app.controllerValidateSpy, ctx);
+  //     assert.deepStrictEqual(controller.results.filter((r) => !r.valid).map((r) => r.toString()), []);
+  //   },
+  //   {
+  //     template: `<editor><input au-slot="content" id="target" value.two-way="$host.person.name & validate"></editor>`
+  //   }
+  // );
```
Sayan751, line 1300: Here we have an issue. The test is failing not because of the non-shallow property. Check some of the preceding tests that deal with non-shallow properties. 

The validation-controller traverses till the innermost object, which in this case will be `editor` instance. But there is no rule defined on it. Previously, the `$host` keyword was treated as `this` from the expression parser which means that both of these expressions: `$host.person.name`, and `person.name` would have similar AST. 

With your change in place, we need to make some changes in the `object` selection process in the validation-controller (L163-168).

Sayan751 (commented):
Great work! The simplification of scope selection feels great!

I have made couple of comments. Other than that, the docs for `exposed.bind` is missing. Please add that part to docs.

packages/runtime/src/binding/expression-parser.ts
```diff
@@ -1166,7 +1159,7 @@ function consume(state: ParserState, token: Token): void {
  * Usage: TokenValues[token & Token.Type]
  */
 const TokenValues = [
-  $false, $true, $null, $undefined, '$this', '$host', '$parent',
+  $false, $true, $null, $undefined, '$this', null/* '$host' */, '$parent',
```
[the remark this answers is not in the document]
bigopon: We cannot actually clean this up though, because the value of the bits in the Token needs to match the index of this array. I' think we should keep things intact for later reference when we need to extend the parser as well, at least for now.

packages/runtime-html/src/resources/custom-elements/au-slot.ts
bigopon: It's to signal that the scope used to bind the view created by `<au-slot/>` isn't the scope that we expect in the template, it's wrapped inside another layer. Does this sound bad?

packages/__tests__/validation-html/validate-binding-behavior.spec.ts
bigopon: I see, combined with what I remember about v1, and the confirmation from you that it now works on nested property, I was quite confused looking at these lines. Are you interested in tweaking this? I can do it too if you prefer.

packages/runtime-html/src/resources/custom-elements/au-slot.ts
Sayan751: I just think we should refrain from accidentally introducing new terms, that's all. We already have scope, and parent scope, and every scope holds binding context and override context. I think so far those are well defined. I think this scope to should be treated as a regular scope as well (which it is), without accidentally adding any extra importance by using new terms.

bigopon: That could be true after we have settled the implementation & details around this. When I first did it, it was pretty confusing without this term. It's gone now anyway, so I think we good here 👍

PR #1243, opened 2021-07-15
refactor: remove render-context

bigopon:
# Pull Request

## 📖 Description

At the moment, render context is a mix of factory & instance based context. It also contains partial optimization & de-optimization logics. Extract & inline most of the instance based logic in render context to where they are used, and move the factory like methods of render context to a new home. Additionally, perf got better since controller is doing less work.

BREAKING CHANGES:

- remove: context property on `IController`
- remove: context container requirement on `.forCustomElement` call, it shouldn't require both context & own container to determine a controller for a custom element
- remove: `getRenderContext` & supporting infra
- remove: context parameters in renderer

CHANGES:
- lifecycle: `define` now takes `IHydrationContext | null` as 2nd parameter, which should be more useful than a container

DOC:
- lifecycle: tweak define doc t

CHORE:
- remove: all irrelevant/commented test files
- remove: old todos/stale comments


bigopon:
@jwx with this PR, we'll need to tweak a few things in the direct controller, as it no longer requires context container for `.forCustomElement` call.

PR #1254, opened 2021-07-22
refactor(compiler): rename instructions to props on CE/CA instruction

bigopon:
# Pull Request

## 📖 Description

At the moment, `instructions` property on hydrate element/attribute/template controller doesn't sound as clear as it should. If it's about properties of CE/CA/TC then should just name it so.

- doc: Add some more doc related to how to use the default template compiler elsewhere.
- add some small comments explaining the main `_compileElement` function to balance between comprehensiveness & perf

### DIFF
--- a/packages/runtime-html/src/observation/event-delegator.ts
+++ b/packages/runtime-html/src/observation/event-delegator.ts
@@ -116,8 +116,9 @@ export class EventSubscriber {
 
   public dispose(): void {
     const { target, handler } = this;
+    let event: string;
     if (target !== null && handler !== null) {
-      for (const event of this.config.events) {
+      for (event of this.config.events) {
         target.removeEventListener(event, handler);
       }
     }
--- a/packages/runtime-html/src/template-compiler.ts
+++ b/packages/runtime-html/src/template-compiler.ts
@@ -373,23 +373,52 @@ export class TemplateCompiler implements ITemplateCompiler {
   /** @internal */
   // eslint-disable-next-line
   private _compileElement(el: Element, context: CompilationContext): Node | null {
+    // overall, the template compiler does it job by compiling one node,
+    // and let that the process of compiling that node point to the next node to be compiled.
+    // ----------------------------------------
     // a summary of this 650 line long function:
     // 1. walk through all attributes to put them into their corresponding instruction groups
     //    template controllers      -> list 1
     //    custom attributes         -> list 2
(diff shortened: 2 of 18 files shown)


packages/runtime-html/src/observation/event-delegator.ts
```diff
@@ -116,8 +116,9 @@ export class EventSubscriber {
 
   public dispose(): void {
     const { target, handler } = this;
+    let event: string;
     if (target !== null && handler !== null) {
-      for (const event of this.config.events) {
+      for (event of this.config.events) {
```
Sayan751, line 121: Q: out of curiosity, is this for optimization? If yes, how does this help actually?

packages/runtime-html/src/template-compiler.ts
```diff
@@ -446,12 +475,18 @@ export class TemplateCompiler implements ITemplateCompiler {
         throw new Error(
           'Trying to enhance with a template that was probably compiled before. '
           + 'This is likely going to cause issues. '
-          + 'Consider enhancing only untouched elements.'
+          + 'Consider enhancing only untouched elements or first remove all "au" classes.'
```
Sayan751, line 478: I don't think we should add this part to the error message. As per as I know this aspect is not tested. I think the cleanest way would be to deactivate enhanced node and reactivate.

packages/runtime-html/src/observation/event-delegator.ts
bigopon: Doing it this way, only 1 object will need to be created to store the information of the variables. While keeping `const` in the for loop, it's a new object for each loop (engine job to optimize it, but we don't need to wait for its optimization to kick in)

packages/runtime-html/src/template-compiler.ts
bigopon: > deactivate enhanced node and reactivate.

It would still be unsafe to do so, since the hydration after compilation of enhance doesn't care if a node is detached, it only tries to find all the elements with "au" to apply an instruction. "Deactivated" node, would be mistakenly seen as targets of the hydration.

Sayan751: Got it. But still that message makes me bit uncomfortable. Maybe we need to test it out.

bigopon: Your tests have been awesome, maybe give some of your love here too 😉

PR #1257, opened 2021-07-23
import value converter info missing

marekpetak:
I have added import tag to import data value converter and changed the file name to date-value-converter.ts

### 🎫 Issues
I think import tag belongs to this part of the docs


bigopon:
thanks @marekpetak !

PR #1263, opened 2021-07-27
fix(plugin-conventions): fix compatibility with webpack css extraction

### LINKED ISSUE #1244: CSS extraction does not work with Webpack 5
# 🐛 Bug Report

When attempting to extract CSS using the Mini CSS Extract plugin in Webpack, I get numerous warnings which I am assuming are coming from our Webpack plugin. I am using the default convention where styles are not being manually imported and defined on a `customElement` decorator (although I did try this to no avail).

Please note I am NOT using CSS Modules or Shadow DOM for styling.

**The error I am seeing is:**

```
WARNING in ./src/components/login.html 7:57-59
export 'default' (imported as 'd0') was not found in './login.scss' (module has no exports)
```

Before creating this issue, I tried quite a few things to see if I can fix this.  Now, the thing is, I see these warnings, but my styles are being loaded in Webpack. The issue appears to be that they are not chunked. All of my CSS is in an `entry.css` file instead.

So, the CSS is being found, but it's not extracting properly. It's almost as if the conventions plugin or loader (most likely loader) is trying to import the contents of my file incorrectly.

Relevant parts of my webpack configuration file:

```
const MiniCssExtractPlugin = require("mini-css-extract-plugin");

const cssLoader = {
  loader: 'css-loader',
  options: {
    // https://github.com/webpack-contrib/css-loader#importloaders
    importLoaders: 2,
    esModule: false,
  }
};

const sassLoader = {
  loader: 'sass-loader',
  options: {
    sassOptions: {
      includePaths: ['node_modules']
    }
  }
};

const postcssLoader = {
  loader: 'postcss-loader',
  options: {
    postcssOptions: {
      plugins: ['autoprefixer']
    }
  }
};
```

```
{ test: /\.html$/i, use: '@aurelia/webpack-loader', exclude: /node_modules/ }
{ test: /\.css$/i, use: [ MiniCssExtractPlugin.loader, cssLoader, postcssLoader ] },
{ test: /\.scss$/i, use: [ MiniCssExtractPlugin.loader, cssLoader, postcssLoader, sassLoader ] },
```

## 🤔 Expected Behavior

CSS should be extracted into named chunks as per the component.

## 😯 Current Behavior

CSS is not extracted into chunks.

## 💁 Possible Solution

Check loader to ensure any parsed resources from custom elements are being properly imported.

## 🌍 Your Environment

Latest dev version of Aurelia, TypeScript and Webpack.

3cp:
close #1244

### DIFF
--- a/packages-cjs/__tests__/plugin-conventions/preprocess-html-template.spec.ts
+++ b/packages-cjs/__tests__/plugin-conventions/preprocess-html-template.spec.ts
@@ -44,12 +44,11 @@ export function register(container) {
   it('processes template with css pair', function () {
     const html = '<template></template>';
     const expected = `import { CustomElement } from '@aurelia/runtime-html';
-import { Registration } from '@aurelia/kernel';
-import d0 from "./foo-bar.css";
+import "./foo-bar.css";
 export const name = "foo-bar";
 export const template = "<template></template>";
 export default template;
-export const dependencies = [ Registration.defer('.css', d0) ];
+export const dependencies = [  ];
 let _e;
 export function register(container) {
   if (!_e) {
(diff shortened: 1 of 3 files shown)


Vheissu (approved):
All looks great to me and like this should address the issue with CSS extraction.

bigopon:
This looks good, and shadow DOM seems still properly handled. Thanks @3cp

PR #1272, opened 2021-08-01
fix(validation): #1253

### LINKED ISSUE #1253: ValidationController resolution does not work correctly for routed views
# 🐛 Bug Report

- Create a routed view.
- Apply validation rules on an object.
- Test the validation for first time. It should work.
- Navigate away from that page and come back to again.
- Test the validation, and it does not work any longer.

## 🤔 Expected Behavior

It should work.

## 😯 Current Behavior

It seems that the `ValidateBIndingBehavior#locator` is resolving an old instance of the validation controller which is different from the one created in the routed view during the 2nd navigation to the routed view.

## 💻 Code Sample

Repro: https://github.com/benbrouse/aurelia2-repro-validation
Validation is applied on the login page.

## 🌍 Your Environment

| Aurelia          |
| Language         |
| Browser          |
| Bundler          |
| Operating System |
| NPM/Node/Yarn    |

Sayan751:
# Pull Request

### 🎫 Issues

closes #1253


packages/runtime/src/binding-behavior.ts
```diff
@@ -129,13 +129,11 @@ export class BindingBehaviorFactory<T extends Constructable = Constructable> {
     const deps = this.deps;
     switch (deps.length) {
       case 0:
-      case 1:
-      case 2:
         // TODO(fkleuver): fix this cast
         return new this.Type(binding, expr) as unknown as IInterceptableBinding;
-      case 3:
+      case 1:
```
Sayan751, line 134: Reviewer note: This is not directly related to the bug fix. However, during debugging, I noticed that this behavior seems incorrect. As this will always inject the binding and expression to the binding behavior, it makes sense to apply injection token only for other preceding arguments. For example, before this change, the following will fail:

```ts
class MyBindingBehavior {
  public constructor(
    @IToken instance,              //<- binding is injected instead of IToken
    binding,                       //<- expression is injected instead of binding
    expression,                    //<- undefined for obvious reason
  ){}
}
```

Sayan751 (commented):
@bigopon Added couple of notes for you so that the review gets easier. Pushed a validation-router-integration example as well. It will make future debugging easier. It is not part of the npm7 workspace.

packages/kernel/src/di.ts
```diff
@@ -1045,7 +1050,15 @@ export class Container implements IContainer {
     } else if (result instanceof Resolver && result.strategy === ResolverStrategy.array) {
       (result.state as IResolver[]).push(resolver);
     } else {
-      resolvers.set(key, new Resolver(key, ResolverStrategy.array, [result, resolver]));
+      switch (strategy) {
```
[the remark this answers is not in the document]
Sayan751: Other alternative would be to expose the `Container#deregisterResolver` API, and perform the deregistration from `@newInstanceForScope` decorator. It can also check if the existing resolver is an array, and throws if that's the case. On a second thought that seems cleaner.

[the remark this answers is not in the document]
Sayan751: Activated the deregister API

packages/kernel/src/di.ts
```diff
@@ -686,6 +687,7 @@ export type IResolvedFactory<K> = (...args: unknown[]) => Resolved<K>;
 export const newInstanceForScope = createResolver((key: any, handler: IContainer, requestor: IContainer) => {
   const instance = createNewInstance(key, handler, requestor);
   const instanceProvider: InstanceProvider<any> = new InstanceProvider<any>(String(key), instance);
+  requestor.deregisterResolverFor(key, false);
```
bigopon, line 690: This looks quite impactful, and I'm quite unsure whether this will not have undesirable side effect. Can you help explain in which case where a completely new scope (container) is bogged by an old registration? I'm aware that we may be invoking this in an old/shared container, but if the router is issuing old container, then probably it's not doing a proper job and we should fix it there instead of here?

packages/runtime/src/binding-behavior.ts
bigopon: This is a nice fix. I was wondering about this too but haven't come to here yet to verify with tests. Thanks 👍

packages/kernel/src/di.ts
```diff
@@ -62,6 +62,7 @@ export interface IContainer extends IServiceLocator, IDisposable {
   readonly root: IContainer;
   register(...params: any[]): IContainer;
   registerResolver<K extends Key, T = K>(key: K, resolver: IResolver<T>, isDisposable?: boolean): IResolver<T>;
+  deregisterResolverFor<K extends Key>(key: K, searchAncestors: boolean): void;
```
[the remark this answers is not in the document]
Sayan751: `deregisterResolver` is also not right, as the method does not accept any resolver.

packages/kernel/src/di.ts
Sayan751: I thought about that too. I am not sure whether there is a problem there with the router. However, I think this decorator should work correctly regardless. For example, if we do `@newnewInstanceForScope(IToken)` twice, I would expect that any successive call to `Container#get(IToken)` should return the second instance of `IToken`. However, before this change, both instances were put into a new resolver with array strategy, which only returns the first registered instance, which is wrong in this case. And this is regardless of router. This change should fix that problem.

bigopon: > should return the second instance of IToken

This would make the registration order out of expectation. This makes the resolution inconsistent with how it's done mainly. At the moment, it's an append only & first in first out, unless there's a special resolution, which is the built-in `Resolver` class.
Also, if it's being done twice, the correct way to get everything is probably `.getAll()`, and get last, we shouldn't try to guarantee something harmful & irreversible like deregistering an existing registration, that is not even in its control.

For the combination of this behavior in this context, I'm a bit unsure how to proceed with the bug.

bigopon: we may be able to document it, but we may not be able to explain why it works that way with a confidence that its the right behavior. The following scenario:
```ts
container.get(newInstanceForScope(MyClass))
```
will become quite trappy and it'll be easy to make mistake.

packages/kernel/src/di.ts
```diff
@@ -683,13 +684,18 @@ export type IFactoryResolver<K = any> = IResolver<K>
...
-  requestor.registerResolver(key, instanceProvider);
+export const newInstanceForScope = createResolver(
+  (key: any, handler: IContainer, requestor: IContainer) => {
+    const instance = createNewInstance(key, handler, requestor);
+    const instanceProvider: InstanceProvider<any> = new InstanceProvider<any>(String(key), instance);
+    /**
+     * By default the new instances for scope are disposable.
+     * If need be, we can always enhance the `createNewInstance` to support a 'injection' context, to make a non/disposable registration here.
+     */
+    requestor.registerResolver(key, instanceProvider, true);
```
bigopon, line 695: When do we plan/want to dispose an instance explicitly registered via `newInstanceForScope`? Shouldn't sticky resolution is more of expected behavior?

packages/runtime-html/src/templating/controller.ts
```diff
@@ -1090,6 +1090,7 @@ export class Controller<C extends IViewModel = IViewModel> implements IControlle
     this.viewModel = null;
     this.host = null;
     this.shadowRoot = null;
+    this.container.disposeResolvers();
```
bigopon, line 1093: everything on the controller already gone to the no return point, I think we can just `.dispose()` here. But this is something I'm not sure about, it's not something the controller controls 100%. We may need to refrain from doing this.

packages/kernel/src/di.ts
Sayan751: As you have already noted, the "disposable" resolvers are disposed from `Controller#dispose`. Sticky resolutions are default. However, that does not work for the idea of `newInstanceForScope`, as that won't be a "new" instance.

packages/runtime-html/src/templating/controller.ts
Sayan751: I tried that. The problem with that is that in that case all the resolvers are disposed, including the resource resolvers.

bigopon:
@Sayan751 thanks for tackling this difficult issue. I think it's undesirable that things do not work as is (`@newInstanceForScope` not result in something sticky, even though its name suggests so), but this is a good interim step until we finish tweaking the router.
We'll keep iterating on this.

PR #1279, opened 2021-08-08
refactor(binding-command): extract binding command type out of binding type, rename binding type

bigopon:
# Pull Request

## 📖 Description

At the moment, `BindingType` is a mixed of both expression type & binding command type, making it quite difficult to understand this enum, also it's bigger than it necessary. This PR splits `BindingType` into `ExpressionType` for the expression parser, and `CommandType` for the binding commands. Also simplify `ExpressionType` to only 5 members from ~15:
```ts
export const enum ExpressionType {
                None = 0,
       Interpolation = 0b0_00001,
          IsIterator = 0b0_00010,
          IsFunction = 0b0_00100,
          IsProperty = 0b0_01000,
          IsCustom   = 0b0_10000,
}
```

Additionally, add a `name` property on `BindingCommand`, so that if needed, commands can be still distinguished.

## ⏭ Next Steps

More cleanup & preparation for attribute transferring feature.


packages/runtime-html/src/observation/event-delegator.ts
```diff
@@ -42,7 +43,7 @@ class ListenerTracker implements IDisposable {
     const lookups = this._options.capture === true ? this._captureLookups : this._bubbleLookups;
     let lookup = lookups.get(target);
     if (lookup === void 0) {
-      lookups.set(target, lookup = Object.create(null) as Record<string, EventListenerOrEventListenerObject | undefined>);
+      lookups.set(target, lookup = createLookup());
```
Sayan751, line 46: This seems bit odd. Why are you wrapping a one line code in a function, especially when the typecasting won't end up in dist?

packages/i18n/src/t/translation-parameters-renderer.ts
```diff
@@ -45,51 +48,64 @@ export class TranslationParametersBindingInstruction {
...
-    private readonly xp: IExpressionParser,
-  ) {}
+  /** @internal */ protected static inject = [IAttrMapper, IExpressionParser];
+  /** @internal */ private readonly _attrMapper: IAttrMapper;
+  /** @internal */ private readonly _exprParser: IExpressionParser;
+
+  public constructor(m: IAttrMapper, xp: IExpressionParser) {
+    this._attrMapper = m;
+    this._exprParser = xp;
+  }
```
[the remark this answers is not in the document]
bigopon: A bit of a dilemma because it makes the typings of the constructor include a `_` in the parameter name. Doing it here = do we do it everywhere, in the public API, it's kind of uncomfortable.

packages/runtime-html/src/observation/event-delegator.ts
bigopon: `Object.create(null)` when repeated a lot will cost some extra bytes, while this one will not. Another options is to leave it to minification and automatically turn `Object.createNull()` into a const function. Though I think it's better to be explicit.

bigopon:
@Sayan751 thanks for the review. I'll address any other concerns in follow up PRs.

PR #1281, opened 2021-08-11
change pre-alpha to alpha

elitastic:
"pre-alpha" is not longer correct in my opinion.

# Pull Request


bigopon:
@elitastic thanks!

PR #1290, opened 2021-08-17
feat(attr-transfer): support attributes spreading with ...$attrs

### LINKED ISSUE #319: [Feature] Transparent Wrapper Components in Aurelia

### LINKED ISSUE #422: [Bug] Compose Should Transfer Attributes to Composed Elements

### LINKED ISSUE #636: [RFC] au-host, au-bind, au-transfer-bindings, au-transfer-attrs, etc

### LINKED ISSUE #1009: [RFC]: API to capture custom element compilation metadata
# 💬 RFC

This RFC proposes introducing new hooks into the custom element compilation that adds metadata for the custom element instances (currently the "symbol"s). The idea is to make these metadata in such a manner that is can be ubiquitously used. This can be especially useful in the compilation pipeline.

## 🔦 Context

Let's discuss some use-cases, where this new infra can be used.

#### Default content projection for au-slot

Currently the usage of `[au-slot]` attribute is mandatory for projecting any content. This also includes projection to default au-slot. This can be by itself annoying. With the application of repeater/other template controllers it can get very cumbersome at the usage side.

Let's consider the following example.

```html
<!-- usage -->
<element-that-hosts-au-slot>
  <div>I want this to be projected in the default au-slot.</div>
</element-that-hosts-au-slot>
```
Currently the above example will not work as intended. because there's no `[au-slot]` attribute on the `<div/>`, as per following example:
```html
<element-that-hosts-au-slot>
  <div au-slot>I want this to be projected in the default au-slot.</div>
</element-that-hosts-au-slot>
```

With the new infra in place, we can register the fragment `<div>...</div>` as potential projection that is mapped to the `my-element` symbol (as per current terminology). And proceed as per normal course of pipeline. Then later when the *usage* is rendered and subsequently, `my-element` is compiled, in a compilation hook (generic; as part of the new infra) implemented in `AuSlot` we can read that metadata to compose the content. Also remove the associated instruction from the usage side. IMO that way, the implementation of `au-slot` can be more self contained, and the template compiler and binder can be cleaned up in a more streamlined fashion. 

Possibly with this transport mechanism in place, we don't need to abuse the DI as a 'projection content transport system'.

#### Infra for attribute transfer

The new infra can potentially be used transfer the attributes applied on the host from the usage side to the elements in the inner template of the host. Let's consider the following example.

```html
<!-- usage -->
<my-element host-attr="foo" inner-cmd-attr="bar" inner-link-attr="fizz">  
</my-element>

<!-- my-element: inner template -->
<my-cmd></my-cmd> <!-- transport the inner-cmd-attr to here -->
<my-link></my-link> <!-- transport the inner-link-attr to here -->
```

The idea is that when the `my-element` instance is compiled, the metadata injection hook can be called to add/modify the attributes applied. Also possibly remap the attributes to the children in the inner-template.

## 💻 Examples

There' two parts of compilation: compiling the usage template (1), and compiling inner template (2). This RFC is intended to have an API to with the ability to:
(a) enhance (1) to capture & store custom metadata about a usage template
(b) enhance (2) with the ability to use the metadata provided in (a) to alter the inner template before the compilation

Hypothetical scenario:
- For a custom paragraph custom element for displaying intro, body and conclusion, add corresponding header of those 3 parts if there's projection for them.

The inner template can look like this:
```html
<!-- a custom paragraph custom element -->
<template>
  <au-slot name="intro"></au-slot>
  <!-- default slot content goes here for the body -->
  <au-slot></au-slot>
  <au-slot name="conclusion"></au-slot>
</template>
```

A usage template with only body can look like this:
```html
<!-- app.html -->
<custom-paragraph>
  <div>I'm the body</div>
</custom-paragraph>
```

The final rendered html should look like this:
```html
<custom-paragraph>
  <h2>Body</h2>
  <div>I'm the body</div>
</custom-paragraph>
```

In the above example, the usage template compilation can perform these steps:
- provide information such as: has default projection (`<div/>` with text `I'm the body`).
- automatically add `[au-slot]` attribute to the above `<div/>`

The inner template compilation can be perform these steps:
- detect there's a projection for default `<au-slot/>`
- add a `<h2>Body</h2>` before the 2nd `<au-slot />` element.

bigopon:
# Pull Request

## 📖 Description

Add support for transferring attributes with `...$attrs` syntax. Example:

```html
app.html
<my-input value.bind="message">

my-input.html
<input ...$attrs>
```

Attribute transferring is a way to relay the binding(s) on a custom element to other element(s) inside it.

As an application grows, the components inside it also grow. Something that starts simple like the following component

```typescript
export class FormInput {
  @bindable label
  @bindable value
}
```

with the template

```html
<label>${label}
  <input value.bind="value">
</label>
```
can gradually grow out of hand with a number of needs for configuration: aria, type, min, max, pattern, tooltip, validation etc...

After a while, the `FormInput` component above will be come more and more like a relayer to transfer the bindings from outside, to the elements inside it. This often results in the increase of the number of `@bindable`. This is completely fine except that it's quite some boilerplate code that is not always desirable:

```typescript
export class FormInput {
  @bindable label
  @bindable value
  @bindable type
  @bindable tooltip
  @bindable arias
  @bindable etc
}
```

And the usage of such element may look like this

```html
<form-input
  label.bind="label"
  value.bind="message"
  tooltip.bind="Did you know Aurelia syntax comes from an idea of an Angular community member? We greatly appreciate Angular and its community for this."
  validation.bind="...">
```

to be repeated like this inside:

```html
<label>${label}
  <input value.bind tooltip.bind validation.bind min.bind max.bind>
</label>
```

To juggle all the relevant pieces for such relaying task isn't difficult, but somewhat tedious. With attribute transferring, which is roughly close to spreading in JavaScript, the above template should be as simple as:

```html
<label>${label}
  <input ...$attrs>
</label>
```

, which reads like this: for some bindings on `<form-input>`, change the targets of those bindings to the `<input>` element inside it.

### 🎫 Issues

close #636
close #422
close #1009
close #668
close #319

## 👩‍💻 Reviewer Notes

There' two new instructions to deal with spreading, they are mostly internal to the new `SpreadRenderer`. While doing this, I had to come up with somewhat hackish-approach: making the spread binding work as a surrogate bindings for all the bindings created from the captured attributes. In the end, it still looks fine, but later, we can see whether we can improve anything around the renderers API to avoid such hack.
The `HydrateElementInstruction` has been enhanced with a new property `captures`, to store relevant information. And the template compiler has been also made aware of `capture` property on a custom element definition.

Note of what will be captured, and what will not be:

View model:
```typescript
export class FormInput {
  @bindable label
}
```

Usage:
```html
<form-input if.bind="needsComment" label.bind="label" value.bind="extraComment" class="form-control" style="background: var(--theme-purple)" tooltip="Hello, ${tooltip}">
```

What are captured:
  - `value.bind="extraComment"`
  - `class="form-control"`
  - `style="background: var(--theme-purple)"`
  - `tooltip="Hello, ${tooltip}"`
What are not captured:
  - `if.bind="needsComment"` (`if` is a template controller)
  - `label.bind="label"` (`label` is a bindable property)

## 📑 Test Plan

- multi level transferring
- interpolation transferring
- binding command preservation

## ⏭ Next Steps

N/A

### DIFF
--- a/docs/user-docs/TOC.md
+++ b/docs/user-docs/TOC.md
@@ -34,6 +34,7 @@
   * [Template references and variables](getting-to-know-aurelia/introduction/template-references-and-variables.md)
   * [Portalling elements](getting-to-know-aurelia/introduction/portalling-elements.md)
   * [Templating - Local Template](getting-to-know-aurelia/introduction/local-templates.md)
+  * [Attributes transferring](getting-to-know-aurelia/introduction/attributes-transferring.md)
 * [Components](getting-to-know-aurelia/components/README.md)
   * [Component lifecycles](getting-to-know-aurelia/components/component-lifecycles.md)
   * [Creating components](getting-to-know-aurelia/components/creating-components/README.md)
--- a/docs/user-docs/getting-to-know-aurelia/introduction/attributes-transfering.md
+++ b/docs/user-docs/getting-to-know-aurelia/introduction/attributes-transfering.md
@@ -1,0 +1,135 @@
+# Attributes transferring (or Spread Attributes)
+
+## Introduction
+
+Attribute transferring is a way to relay the binding(s) on a custom element to other element(s) inside it.
+
+As an application grows, the components inside it also grow. Something that starts simple like the following component
+
+```typescript
+export class FormInput {
+  @bindable label
+  @bindable value
+}
+```
+
+with the template
+
+```html
+<label>${label}
+  <input value.bind="value">
+</label>
+```
+can quickly grow out of hand with a number of needs for configuration: aria, type, min, max, pattern, tooltip, validation etc...
+
+After a while, the `FormInput` component above will be come more and more like a relayer to transfer the bindings from outside, to the elements inside it. This often results in the increase of the number of `@bindable`. This is completely fine except that it's quite some boilerplate code that is not always desirable:
+
+```typescript
+export class FormInput {
+  @bindable label
+  @bindable value
+  @bindable type
+  @bindable tooltip
+  @bindable arias
+  @bindable etc
+}
+```
+
+And the usage of such element may look like this
+
+```html
+<form-input
+  label.bind="label"
+  value.bind="message"
+  tooltip.bind="Did you know Aurelia syntax comes from an idea of an Angular community member? We greatly appreciate Angular and its community for this."
+  validation.bind="...">
+```
+
+to be repeated like this inside:
+
+```html
+<label>${label}
+  <input value.bind tooltip.bind validation.bind min.bind max.bind>
+</label>
+```
+
+To juggle all the relevant pieces for such relaying task isn't difficult, but somewhat tedious. With attribute transferring, which is roughly close to spreading in JavaScript, the above template should be as simple as:
+
+```html
+<label>${label}
+  <input ...$attrs>
+</label>
+```
+
+, which reads like this: for some bindings on `<form-input>`, change the targets of those bindings to the `<input>` element inside it.
+
+## Usage
+
+To transfer attributes & bindings from a custom element, there are two steps:
+
+- Set `capture` to `true` on a custom element via `@customElement` decorator:
+
+```typescript
+@customElement({
+  ...,
+  capture: true
+})
+```
+
+As the name suggests, this is to signal the template compiler that all the bindings & attributes, with some exceptions should be captured for future usages.
+
+- Spread the captured attributes onto an element:
+
+```html
+<input ...$attrs>
+```
+
+{% hint style="warning" %}
+It's recommended that this feature should not be overused in multi level capturing & transferring. This is often known as prop-drilling in React, and could have bad effect on overall & long term maintainability of a project. It's probably healthy to limit the max level of transferring to 2.
+{% endhint %}
+
+
+## How it works
+
+### What attributes are captured
+
+Everything except template controller and custom element bindables are captured. For the following example:
+
+View model:
+```typescript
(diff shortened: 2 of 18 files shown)


bigopon:
@Sayan751 @fkleuver @brandonseydel  I'll go ahead with the merge, and address any concerns in follow up PRs.

PR #1303, opened 2021-09-08
refactor(repeat): handle array replacement & matcher better

### LINKED ISSUE #1264: Repeat.for to support key (custom resolution strategy)
# 🙋 Feature Request

While React and Vue do support additional `key` attribute to signalize strategy, which elements to create, Aurelia and Aurelia 2 (as far as I can tell from docs) relies only on indexes and array manipulation. This is an issue when the whole array is replaced.

## 🤔 Expected Behavior

```html
<div repeat.for="todo of todos & key: 'id'"> ${ todo.text } </div>
<!-- or --> 
<div repeat.for="todo of todos" key.bind="todo.id"> ${ todo.text } </div>
<!-- or similar -->
```

```js
let todos = [ { text: 'a', id: 1 }, { text: 'b', id: 2 } ];
// render
todos = [ { text: 'a', id: 1 }, { text: 'c', id: 3 }, { text: 'b', id: 2 } ];

// Produces:

// INITIAL STATE (HTML elements)
a
b

// NEW STATE (HTML elements)
a - reused
c - newly created element
b - reused
```
Every created HTML element gets an id. This id must be unique for each element in the array. When `repeat.for` renders the array again (due to a change), it searches for an existing HTML element with the given id first, and reuses it. When such HTML element cannot be found a new element is created.

## 😯 Current Behavior

```js
/* ONLY WAY HOW TO CURRENTLY ACHIEVE THE EXPECTED RESULT */

const array = [ 'a', 'b' ];
// render
array.slice( 1, 'c' );

// Produces:

// INITIAL STATE (HTML elements)
a
b

// NEW STATE (HTML elements)
a - reused
c - newly created element
b - reused
```

```js
/* CURRENT BEHAVIOR IN PROBLEMATIC SITUATION */

let array = [ 'a', 'b' ];
// render
array = [ 'a', 'c', 'b']

// Produces:

// INITIAL STATE (HTML elements)
a
b

// NEW STATE (HTML elements)
a - reused
c - reused
b - newly created element
```

## 💁 Possible Solution

Vue: https://vuejs.org/v2/guide/list.html
React: https://reactjs.org/docs/lists-and-keys.html#keys

## 🔦 Context

I fetch data periodically from a server. I have to animate newly added items (and also removed). Currently, I have to compare the new array with the existing array and "simulate" changes. Not only is that difficult, but doesn't work well on unsorted array, or in a situation when the value for the sorting key changes for a given element (for example speed has changed but the driver is known, thus the element should be way lower, but shouldn't animate).

I would like to simply replace the array.

bigopon:
# Pull Request

## 📖 Description

Better handle repeat items replacement for immutable scenarios & race condition

### 🎫 Issues

Resolves #1264 

## 📑 Test Plan

Existing repeat tests work

## ⏭ Next Steps

- More array observation refactoring

### DIFF
--- a/packages/aurelia/src/index.ts
+++ b/packages/aurelia/src/index.ts
@@ -185,7 +185,6 @@ export {
   toArray,
   // nextId,
   // resetId,
-  // compareNumber,
   // mergeDistinct,
   // isNumberOrBigInt,
   // isStringOrDate,
--- a/packages/kernel/src/functions.ts
+++ b/packages/kernel/src/functions.ts
@@ -287,14 +287,6 @@ export function resetId(context: string): void {
 }
 
 /**
- * A compare function to pass to `Array.prototype.sort` for sorting numbers.
- * This is needed for numeric sort, since the default sorts them as strings.
- */
-export function compareNumber(a: number, b: number): number {
-  return a - b;
-}
-
-/**
  * Efficiently merge and deduplicate the (primitive) values in two arrays.
  *
  * Does not deduplicate existing values in the first array.
--- a/packages/kernel/src/index.ts
+++ b/packages/kernel/src/index.ts
@@ -117,7 +117,6 @@ export {
   toArray,
   nextId,
   resetId,
-  compareNumber,
   mergeDistinct,
   isNumberOrBigInt,
   isStringOrDate,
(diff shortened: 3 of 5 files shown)


Vheissu:
@bigopon Are we still wanting this PR?

bigopon:
I'll have a newer PR for this.

PR #1304, opened 2021-09-11
feat: destructuring expression for repeater

Sayan751:
# Pull Request


docs/user-docs/getting-to-know-aurelia/introduction/built-in-template-features/repeat.for.md
```diff
@@ -61,3 +61,77 @@ Inside of the `repeat.for` these can be accessed. In the following example we di
...
+### Iterating over a map
+
+You can also iterate over a map using the `repeat.for`.
+When iterating over a map, the declaration expression contains a pair of key and value.
+
+```markup
+<p repeat.for="pair of map">key:${pair[0]} | value: ${pair[1]}</p>
+```
+
+However, you can also destructured assignment expression.
```
bigopon, line 81: You can also destructure the assignment expression

docs/user-docs/getting-to-know-aurelia/introduction/built-in-template-features/repeat.for.md
```diff
@@ -61,3 +61,77 @@ Inside of the `repeat.for` these can be accessed. In the following example we di
...
+In fact, you can use the destructured declaration for any kind of iterable collection.
+
+```markup
+<p repeat.for="{name, age} of personArray">${name} is ${age} years old</p>
+```
+
+You can also initialize any `undefined` or missing property in the input.
+
+```markup
+<p repeat.for="{name, age = 42} of personArray">${name} is ${age} years old</p>
```
bigopon, line 100: 👍 looks nice on doc

docs/user-docs/getting-to-know-aurelia/introduction/built-in-template-features/repeat.for.md
```diff
@@ -61,3 +61,77 @@ Inside of the `repeat.for` these can be accessed. In the following example we di
...
+```
+
+You can even alias your destructured properties.
```
bigopon, line 103: I think we should doc it in a more "works by default, mention if surprise" style instead of this. So aliasing should "just work", which means we can remove the "even" word.

docs/user-docs/getting-to-know-aurelia/introduction/built-in-template-features/repeat.for.md
```diff
@@ -61,3 +61,77 @@ Inside of the `repeat.for` these can be accessed. In the following example we di
...
+The destructured assignment supports wide range of complex expressions.
+However, the limitation of this is that the destructured inputs are not observed; i.e. mutations of any destructured property will not trigger change in the view.
+
+Taking the following template as example, if we change the `personArray[0].name = 'foo'` post `attach`, the change won't be reflected on the view.
+
+```markup
+<p repeat.for="{name, age} of personArray">${name} is ${age} years old</p>
+```
+
+In that sense you can think the following JavaScript analogous to the example above; that is if you change the `personArray[i].name` post assignment, that change won't be propagated to the `name`;
```
bigopon, line 120: If each binding of the destructuring was considered a let binding, wouldn't that make it more desirable to observe? The no-observation limitation seems like a big deal. If we have this, I'm afraid we may need to drop destructuring altogether, else we won't be able to handle the mis-expectation.

packages/runtime/src/binding/ast.ts
```diff
@@ -1416,6 +1466,152 @@ export class Interpolation {
...
+    public readonly list: readonly (DestructuringAssignmentExpression | DestructuringAssignmentSingleExpression | DestructuringAssignmentRestExpression)[],
+    public readonly source: AccessMemberExpression | AccessKeyedExpression | undefined,
+    public readonly initializer: IsBindingBehavior | undefined,
+  ) { }
+
+  public evaluate(_f: LF, _s: Scope, _l: IServiceLocator, _c: IConnectable | null): undefined {
+    return void 0;
+  }
+
+  public assign(f: LF, s: Scope, l: IServiceLocator, value: unknown): void {
```
bigopon, line 1484: We could have a change in the interface of `evaluate` and `assign` so that `IServiceLocator` and `IConnectable` are merged into an interface `IAstEvaluator`. Then both evaluate and assign have access to the currently connecting binding. Maybe this can help resolve the limitation?

packages/runtime/src/binding/expression-parser.ts
```diff
@@ -513,11 +517,11 @@ TPrec extends Precedence.Unary ? IsUnary :
         access = Access.Reset;
         break;
       case Token.OpenBracket:
-        result = parseArrayLiteralExpression(state, access, expressionType);
+        result = state.ip.search(/\s+of\s+/) > state.index ? parseDestructuringAssignment(state) : parseArrayLiteralExpression(state, access, expressionType);
```
bigopon, line 520: Shouldn't `indexOf(regex, state.index)` be a better choice?

packages/runtime/src/binding/expression-parser.ts
```diff
@@ -1030,6 +1231,25 @@ function parseTemplate(state: ParserState, access: Access, expressionType: Expre
   }
 }
 
+/**
+ * Returns the next token, without changing the state.
+ */
+ function lookahead(state: ParserState): Token {
```
bigopon, line 1237: nit: lookAhead?

bigopon (commented):
Nice work, left some general questions & obvious nit pick for the first round.

packages/runtime/src/binding/ast.ts
Sayan751: This would definitely help resolving the limitation. However, I am not sure what will be the impact of that.

Your comment has also made me rethink my design decisions. I am contemplating flattening out the expressions so that we can `evaluate` those expressions. However, that might be a major rework in terms of parsing.

packages/runtime/src/binding/expression-parser.ts
Sayan751: Sadly `indexOf` does not accept a regex :( https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/indexOf

docs/user-docs/getting-to-know-aurelia/introduction/built-in-template-features/repeat.for.md
Sayan751: Done

docs/user-docs/getting-to-know-aurelia/introduction/built-in-template-features/repeat.for.md
Sayan751: Done

packages/runtime/src/binding/ast.ts
Sayan751: On a second thought supporting initializer might get very difficult with that approach. I will try to add the connectable to `AST#assign` instead.

bigopon: Shouldn't it be long to `evaluate` instead of assign, for destructuring node specifically.
I realized that merging IServiceLocator and IConnectable is an issue as we will loose the ability to tell when we should/should not be connecting.

bigopon: Another note is `evaluate` is a better choice for destructuring nodes, since it's not really assigning. We have an existing example of this, which is `AssignExpression`, and it does assign when evaluate. I think it's safe to do so, then we can avoid complicating the AST any further

PR #1313, opened 2021-10-30
feat: implicit type-coercion

Sayan751:
# Pull Request

## 📖 Description

This PR adds type-coercion to Aurelia2. It leverages the `design:type` metadata (when `emitDecoratorMetadata` flag is set to `true`) for TypeScript code. For JavaScript the type of the `@bindable` needs to be specified explicitly.

```javascript
@customElement({ name:'my-el', template: 'not-important' })
export class MyEl {
  @bindable({ type: Number }) num;
}
```

Coercion of `null` and `undefined` values are skipped by default. However, there is another new `nullable` property in the bindable definition to mark a property as non-nullable.

```javascript
@customElement({ name:'my-el', template: 'not-important' })
export class MyEl {
  @bindable({ type: Number, nullable: false }) num;
}
```

There are also 2 new configuration options to modify the coercion-behavior globally.

```typescript
new Aurelia()
    .register(
      StandardConfiguration
        .customize((config) => {
          config.coercingOptions.disableCoercion = true; // disables the coercion globally; default value is false
          config.coercingOptions.coerceNullish= true;      // enable coercion for `null` and `undefined`; default value is false
        }),
      ...
    );
```

For more details refer the tests and the new docs.

### 🎫 Issues

This should be the last piece of features requested in #368; and hence that can be closed thereafter.


docs/user-docs/getting-to-know-aurelia/components/bindable-properties/bindable-coercion.md
```diff
@@ -0,0 +1,253 @@
...
+
+## Union types
+
+When using TypeScript, usages of union types are not rare.
+However, using union types for `@bindable` will deactivate the auto-coercion.
+
+```typescript
+@customElement({ name:'my-el', template: 'not-important' })
+export class MyEl {
+  @bindable public num: number | string;
```
zewa666, line 244: can a coercer handle these cases where you manually check for the exact type and return the appropriate one?

docs/user-docs/getting-to-know-aurelia/components/bindable-properties/bindable-coercion.md
```diff
@@ -0,0 +1,253 @@
+# Coercing bindable values
+
+The [bindable setter](bindable-setter.md) section shows how to adapt the value being bound to a `@bindable` property.
+One of the common usage of the setter is to coerce the values that are bound from the view.
```
zewa666, line 4: One -- common usage of the ...

docs/user-docs/getting-to-know-aurelia/components/bindable-properties/bindable-coercion.md
```diff
@@ -0,0 +1,253 @@
...
+@customElement({ name:'my-app', template: '<my-el num="42"></my-el>' })
+export class MyApp { }
+```
+{% endtab %}
+{% endtabs %}
+
+Without any setter for the `@bindable` num we might end up with the string `'42'` as the value for `num` in `MyEl`.
+You can write a setter to coerce the value.
+
+However, it is bit annoying to write a setter for every `@bindable`s.
```
zewa666, line 27: to write setters for every...

docs/user-docs/getting-to-know-aurelia/components/bindable-properties/bindable-coercion.md
```diff
@@ -0,0 +1,253 @@
...
+{% endtab %}
+{% tab title="my-app.html" %}
+```typescript
+@customElement({ name:'my-app', template: '<my-el person="john"></my-el>' })
+export class MyApp { }
+```
+{% endtab %}
+{% endtabs %}
+
+According to `Person#coercer` implementation, for the example above `MyEl#person` will be assigned an instance of `Person` that is equivalent to `new Person('john', null)`.
```
zewa666, line 125: According to the ...

docs/user-docs/getting-to-know-aurelia/components/bindable-properties/bindable-coercion.md
```diff
@@ -0,0 +1,253 @@
...
+{% endtab %}
+{% tab title="my-app.html" %}
+```typescript
+@customElement({ name:'my-app', template: '<my-el person="john"></my-el>' })
+export class MyApp { }
+```
+{% endtab %}
+{% endtabs %}
+
+With `@coercer` decorator you are free to name the static method as you like.
```
zewa666, line 183: With the ...

docs/user-docs/getting-to-know-aurelia/components/bindable-properties/bindable-coercion.md
```diff
@@ -0,0 +1,253 @@
...
+export class MyEl {
+  @bindable({ nullable: false }) public num: number;
+}
+```
+
+When `nullable` is set to `false`, Aurelia2 will try to coerce the `null` and `undefined` values.
+
+## Global coercing configuration
+
+The implicit type-coercion is the default.
```
zewa666, line 203: --- Implicit ...

docs/user-docs/getting-to-know-aurelia/components/bindable-properties/bindable-coercion.md
```diff
@@ -0,0 +1,253 @@
...
+          config.coercingOptions.disableCoercion = true;
+          config.coercingOptions.coerceNullLike = false;
+        }),
+      ...
+    );
+```
+
+There are two relevant configuration options.
+- **`disableCoercion`**: The default value for this property is `false`; that is Aurelia2 coerces the types of the `@bindable`s by default. It can be set to `true` to disable the coercion for all `@bindable`s.
+- **`coerceNullLike`**: The default value for this property is `false`; that is Aurelia2 does not coerce the `null` and `undefined` values. It can be set to `true` to coerce the `null` and `undefined` values as well. This property can be thought of as the global counterpart of the `nullable` property in the bindable definition.
```
zewa666, line 220: maybe name it after the Coalesce Nullish so it becomes `coerceNullish`?

docs/user-docs/getting-to-know-aurelia/components/bindable-properties/bindable-coercion.md
Sayan751: Can you please elaborate it bit more?

docs/user-docs/getting-to-know-aurelia/components/bindable-properties/bindable-coercion.md
Sayan751: Done.

docs/user-docs/getting-to-know-aurelia/components/bindable-properties/bindable-coercion.md
Sayan751: Done.

docs/user-docs/getting-to-know-aurelia/components/bindable-properties/bindable-coercion.md
Sayan751: Done.

docs/user-docs/getting-to-know-aurelia/components/bindable-properties/bindable-coercion.md
Sayan751: Done.

docs/user-docs/getting-to-know-aurelia/components/bindable-properties/bindable-coercion.md
Sayan751: Done.

docs/user-docs/getting-to-know-aurelia/components/bindable-properties/bindable-coercion.md
Sayan751: Definitely a better name. Renamed.

docs/user-docs/getting-to-know-aurelia/components/bindable-properties/bindable-coercion.md
Sayan751: Done.

packages/__tests__/util.ts
```diff
@@ -21,16 +21,16 @@ export function createSpecFunction<TTestContext extends TestExecutionContext<any
...
         this.timeout(setupContext.timeout);
       }
-      await wrap(testFunction, setupContext);
+      await wrap.bind(this)(testFunction.bind(this), setupContext);
     });
   }
 
   $it.only = function (title: string, testFunction: TestFunction<TTestContext>, setupContext?: TSetupContext) {
-    it.only(title, async function () { await wrap(testFunction, setupContext); });
+    it.only(title, async function () { await wrap.bind(this)(testFunction.bind(this), setupContext); });
```
bigopon, line 29: I think this is getting cryptic. it requires too much familiarity to read the code efficiently. Do we have an alternative?

packages/runtime-html/src/bindable.ts
```diff
@@ -317,3 +321,66 @@ function apiTypeCheck() {
     .add('prop').primary();
 }
 /* eslint-enable @typescript-eslint/no-unused-vars,spaced-comment */
+
+// eslint-disable-next-line @typescript-eslint/no-namespace
+declare namespace Reflect {
```
bigopon, line 326: this should be marked as `/** @internal */` so that it doesn't produce typings that will go global in consuming application

packages/runtime-html/src/bindable.ts
```diff
@@ -317,3 +321,66 @@ function apiTypeCheck() {
...
+};
+
+function getInterceptor(prop: string, target: Constructable<unknown>, def: PartialBindableDefinition = {}) {
+  if (coercionConfiguration.disableCoercion) { return noop; }
+
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+  const type: PropertyType | null = def.type ?? Reflect.getMetadata('design:type', target, prop) ?? null;
+  if (type == null) { return noop; }
+  let coercer: InterceptorFunc;
+  switch (true) {
```
bigopon, line 351: Doing it this way results in 6 x `===`, should we have
```ts
switch (type)
  case Number:
  case Boolean:
    ....
  default:
    if (typeof ...)
    else
```
instead

packages/runtime-html/src/configuration.ts
```diff
@@ -305,30 +306,49 @@ export const DefaultRenderers = [
...
+export const StandardConfiguration = createConfiguration(noop);
+
+function createConfiguration(optionsProvider: ConfigurationOptionsProvider) {
+  return {
+    optionsProvider,
+    /**
+     * Apply this configuration to the provided container.
+     */
+    register(container: IContainer): IContainer {
+      optionsProvider(runtimeConfigurationOptions);
```
bigopon, line 318: a new object should be created here so that one Aurelia application setup code won't mess another

packages/runtime-html/src/bindable.ts
```diff
@@ -317,3 +321,66 @@ function apiTypeCheck() {
...
+  define(target: Constructable<unknown>, property: string) {
+    defineMetadata(Coercer.key, ((target as any)[property] as InterceptorFunc).bind(target), target);
+  },
+  for(target: Constructable<unknown>) {
+    return getOwnMetadata(Coercer.key, target) as InterceptorFunc;
+  }
+};
+
+function getInterceptor(prop: string, target: Constructable<unknown>, def: PartialBindableDefinition = {}) {
+  if (coercionConfiguration.disableCoercion) { return noop; }
```
bigopon, line 345: I think we may not be able to afford a static configuration, since one change can affect multiple apps

packages/runtime-html/src/bindable.ts
```diff
@@ -120,17 +124,17 @@ const baseName = getAnnotationKeyFor('bindable');
 export const Bindable = Object.freeze({
   name: baseName,
   keyFrom: (name: string): string => `${baseName}:${name}`,
-  from(...bindableLists: readonly (BindableDefinition | Record<string, PartialBindableDefinition> | readonly string[] | undefined)[]): Record<string, BindableDefinition> {
+  from(type: Constructable, ...bindableLists: readonly (BindableDefinition | Record<string, PartialBindableDefinition> | readonly string[] | undefined)[]): Record<string, BindableDefinition> {
```
bigopon, line 127: I think this addition breaks the current mental model of the bindables creation. If all we know is a associated coercer of a property, can we pass a record of coercers to this function instead?

bigopon (commented):
Nice work 🚀 
I have some basic comments for the first round. Sorry for the delay in review.

docs/user-docs/getting-to-know-aurelia/components/bindable-properties/bindable-coercion.md
```diff
@@ -0,0 +1,270 @@
...
+{% tab title="my-app.ts" %}
+```typescript
+@customElement({ name:'my-app', template: '<my-el num="42"></my-el>' })
+export class MyApp { }
+```
+{% endtab %}
+{% endtabs %}
+
+Without any setter for the `@bindable` num we might end up with the string `'42'` as the value for `num` in `MyEl`.
+You can write a setter to coerce the value.
```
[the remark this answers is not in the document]
Sayan751: The bindable setter is cross referenced right at the beginning of the document. I think that should suffice.

docs/user-docs/getting-to-know-aurelia/components/bindable-properties/bindable-coercion.md
```diff
@@ -0,0 +1,270 @@
...
+```typescript
+import { coercer } from '@aurelia/runtime-html';
+
+export class Person {
+  public constructor(
+    public readonly name: string,
+    public readonly age: number,
+  ) { }
+
+  @coercer
```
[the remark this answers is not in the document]
Sayan751: I think if we use `@coerce` then it will look like we want to "coerce" a method. That is unintuitive IMO. If you think it is better to change the static method name to `coerce` I can do that. However, that might be bit inconsistent, although might be more intuitive.

packages/__tests__/util.ts
Sayan751: This is needed to make the mocha test context available to the tests function. As we are using the wrapper functions, it is necessary to bind the `this`.

packages/runtime-html/src/bindable.ts
Sayan751: Good catch!

packages/runtime-html/src/bindable.ts
Sayan751: This is a good point. However,  we have here no DI support; otherwise a singleton configuration could have been plumbed that way. However, that might also not be a very sound idea as bindable configurations are by nature static.

packages/runtime-html/src/bindable.ts
Sayan751: I think that cannot be done because we don't have a map of coercer. We extract the type information from the `design:type` metadata here: https://github.com/aurelia/aurelia/pull/1313/files/e5b16b47f3fac01d60c4c1f6dbc1a2667b16e90b#diff-61b543eff441a301b60f029ab94debdf3dda9eb91402319a2a1a978a1e8768faR348. For that we need the `type`. We can extract these information from the caller-side but then the logic is not encapsulated. I think we don't want to go that way.

docs/user-docs/getting-to-know-aurelia/components/bindable-properties/bindable-coercion.md
```diff
@@ -0,0 +1,270 @@
+# Coercing bindable values
+
+The [bindable setter](bindable-setter.md) section shows how to adapt the value being bound to a `@bindable` property.
+One common usage of the setter is to coerce the values that are bound from the view.
+Consider the following example.
+
+{% tabs %}
+{% tab title="my-el.ts" %}
+```typescript
+@customElement({ name:'my-el', template: 'not-important' })
```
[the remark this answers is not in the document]
Sayan751: done

packages/runtime-html/src/bindable.ts
Sayan751: done

packages/runtime-html/src/bindable.ts
Sayan751: Done. On related note the perf benefit is minuscule as we do end up doing almost same number of case comparisons in both cases. However, the refactoring might  make the code more readable.

packages/runtime-html/src/bindable.ts
Sayan751: This is a potential clash of mental models of static bindable and aurelia-instance-scoped DI-tree.

On a second thought I can remove the "global" configuration altogether, but that might also be frustrating for the devs who don't want type-coercion. 

Maybe we should draw attention to this oddity in our documentation?

bigopon: I think it's best that we always coerce "as is". If dev wants to handle null/undefined, using a custom coercer type is an option.

Sayan751: Agreed. Refactored.

packages/runtime-html/src/configuration.ts
Sayan751: Refactored.

packages/runtime-html/src/configuration.ts
```diff
@@ -305,30 +306,53 @@ export const DefaultRenderers = [
...
+        ...DefaultBindingSyntax,
+        ...DefaultBindingLanguage,
+        ...DefaultRenderers,
+      );
+    },
+    /**
+     * Create a new container with this configuration applied to it.
+     */
+    createContainer(): IContainer {
+      return this.register(DI.createContainer());
```
bigopon, line 347: Using `this` here forces user to be aware of the context of the `createContainer` call unnecessarily. We probably should avoid it, and prepare to remove the existing code that does so (if any)

packages/runtime-html/src/observation/bindable-observer.ts
```diff
@@ -68,6 +68,7 @@ export class BindableObserver implements IFlushable, IWithFlushQueue {
     // this observer can become more static, as in immediately available when used
     // in the form of a decorator
     public readonly $controller: IController | null,
+    private readonly coercionConfig: ICoercionConfiguration | null,
```
bigopon, line 71:
```suggestion
    private readonly _coercionConfig: ICoercionConfiguration | null,
```

packages/runtime-html/src/bindable.ts
```diff
@@ -317,3 +321,59 @@ function apiTypeCheck() {
...
+    }
+  }
+  return coercer === noop
+    ? coercer
+    : createCoercer(coercer, def.nullable);
+}
+
+function createCoercer<TInput, TOutput>(coercer: InterceptorFunc<TInput, TOutput>, nullable: boolean | undefined): InterceptorFunc<TInput, TOutput> {
+  return function (value: TInput, coercionConfiguration: ICoercionConfiguration | null): TOutput {
+    if (!(coercionConfiguration?.enableCoercion ?? false)) return value as unknown as TOutput;
```
bigopon, line 374:
```suggestion
    if (!coercionConfiguration?.enableCoercion) return value as unknown as TOutput;
```
I'm not sure which is better between ensuring boolean or just use the language feature, micro bench but if we look at code size, probably `!` is sufficient

packages/runtime-html/src/bindable.ts
```diff
@@ -317,3 +321,59 @@ function apiTypeCheck() {
...
+  },
+  for(target: Constructable<unknown>) {
+    return getOwnMetadata(Coercer.key, target) as InterceptorFunc;
+  }
+};
+
+function getInterceptor(prop: string, target: Constructable<unknown>, def: PartialBindableDefinition = {}) {
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+  const type: PropertyType | null = def.type ?? Reflect.getMetadata('design:type', target, prop) ?? null;
+  if (type == null) { return noop; }
```
bigopon, line 349: returning `noop` while there's nothing will make the flag `_hasSetter` flag always true, which is a waste in quite a big percentage of the cases. Better just return `undefined` here. Same for `coercer === noop` a several lines below

packages/__tests__/3-runtime-html/bindable-coercer.spec.ts
```diff
@@ -0,0 +1,931 @@
...
+        assert.strictEqual(Number.isNaN(ctx.app.myEl.num), true);
+      }, { app: App, template: `<my-el view-model.ref="myEl" num.bind="{}"></my-el>`, registrations: [MyEl] });
+
+      $it(`number - bound property - ${getTypeSpecification(type)}`, async function (ctx: TestExecutionContext<App>) {
+        const app = ctx.app;
+        const myEl = app.myEl;
+        assert.strictEqual(myEl.num, undefined);
+
+        app.prop = null;
+        await ctx.platform.domWriteQueue.yield();
```
bigopon, line 124: the more we can avoid this `domWriteQueue.yield()` call, the easier it'll be with the later refactoring when we merge the dom queue.
Also, there's nothing asynchronous change notification with `app.prop = null` -> `myEl.num`, we shouldn't have to be bother with this.

packages/__tests__/3-runtime-html/bindable-coercer.spec.ts
```diff
@@ -0,0 +1,931 @@
...
+        assert.strictEqual(myEl.str, '42');
+
+        app.prop = 0;
+        await ctx.platform.domWriteQueue.yield();
+        assert.strictEqual(myEl.str, '0');
+      }, { app: App, template: `<my-el view-model.ref="myEl" str.bind="prop"></my-el>`, registrations: [MyEl] });
+    }
+  }
+
+  for (const type of [undefined, Boolean]) {
```
bigopon, line 297: I think `Boolean` test results can be implied from `Number`. We can remove this. I think it's time we simplify our testing story a bit.
We can also have a basic "smoke" unit test for each of `Boolean`/`BigInt` at least something is needed.

packages/__tests__/3-runtime-html/bindable-coercer.spec.ts
```diff
@@ -0,0 +1,931 @@
...
+        public prop: any = { name: 'john', age: 42 };
+      }
+      $it(`change propagation - class`, async function (ctx: TestExecutionContext<App>) {
+        const app = ctx.app;
+        const myEl = app.myEl;
+        assert.deepStrictEqual(myEl.person, new Person2('john', 42), 'error1');
+        assert.strictEqual(Person2.coerced, 1, 'error2');
+
+        app.prop.age = 84;
+        await ctx.platform.domWriteQueue.yield();
```
bigopon, line 801: similarly to the above comment, if UI update is not involved, probably this shouldn't be involved. If the concern is it may have a side effect that results in a change next frame, then we can just do it once and assume it works fine the rest, rather than tediously trying to guarantee every step along the way.

bigopon (commented):
Thanks for the update. I have some more comments on the latest changes. Nice work 🚀

packages/runtime-html/src/bindable.ts
Sayan751: Well, surprisingly that [flag is set](https://github.com/aurelia/aurelia/blob/b901c4b86a8f7b0c36184d56a2253f3fe5c85563/packages/runtime-html/src/observation/bindable-observer.ts#L77-L77) if the setter is not `noop`. I think we need to change that.

packages/__tests__/3-runtime-html/bindable-coercer.spec.ts
Sayan751: IMO the tests always serves as additional examples. With these tests we iron out any oddities. I am keeping the tests.

packages/__tests__/3-runtime-html/bindable-coercer.spec.ts
Sayan751: Done

packages/__tests__/3-runtime-html/bindable-coercer.spec.ts
Sayan751: done

packages/runtime-html/src/observation/bindable-observer.ts
Sayan751: Done

packages/runtime-html/src/configuration.ts
Sayan751: done

packages/runtime-html/src/bindable.ts
bigopon: Oops, i misremembered it then, this should be fine I guess

docs/user-docs/getting-to-know-aurelia/components/bindable-properties/bindable-coercion.md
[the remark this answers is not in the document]
bigopon: Seems like it'll be either inconsistent or unintuitive. Maybe let's go with intuitiveness, since it's harder to make sense out of the consistency we are trying to achieve. Let's do `@coercer`/`static coerce`?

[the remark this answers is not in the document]
Sayan751: Done

bigopon:
Nice work & docs. Thanks @Sayan751 🚀

PR #1316, opened 2021-11-22
refactor(enhance): custom element as root

Sayan751:
# Pull Request

## 📖 Description

This addresses the 2nd problem described here: https://discourse.aurelia.io/t/aurelia-2-templateengine/4461/4

## 👩‍💻 Reviewer Notes

**The issue** was caused by the artifact that in order to attach a hydrated we need a "host". This caused a problem for hydratable enhance target as root. Following are couple of examples:

```ts
// disconnected enhance root
const ce = document.createElement('my-ce');
aurelia.enhance({host: ce, component:...});
```

```html
<!-- index.html -->
<app-host></app-host>
<enhance-root><!-- connected enhance root --></enhance-root>
```

These caused an issue because there are no hydratable children target found under the enhance root, and that caused mismatch of instructions and targets.

**The fix** addresses this particular problem by providing targets and thereafter manipulating the parent host for attachment. There are two cases here:

Case#1:
```html
<my-ce>
   <!-- hydrated content should be attached here-->
</my-ce>
```

For this case the node sequence generate a child "marker" element inside the root, later during attach this marker is used to grab the root as parent and attach the hydrated content in the root.

Case#2:
```html
<nativeelement attr.bind="foo"></nativeelement>
```
In this case, we should not create any "marker" element and use the native element as the target itself.


packages/runtime-html/src/templating/controller.ts
```diff
@@ -210,6 +210,11 @@ export class Controller<C extends IViewModel = IViewModel> implements IControlle
     }
 
     definition = definition ?? CustomElement.getDefinition(viewModel.constructor as Constructable);
+    if (definition.enhance && host.nodeName !== 'TEMPLATE' && !host.isConnected) {
```
bigopon, line 213: I'm a bit concerned by this, as it'll no longer be enhance, but rather "re-append" + enhance. Also I think rather than trying to reuse the existing `NodeSequence`, we should create a new one to accommodate this scenario, naming could be `EnhanceNodeSequence` (or something else)

bigopon (approved):

packages/runtime-html/src/templating/controller.ts
Sayan751: Thanks @bigopon for pointing me to node sequence. I have updated the PR description describing my new approach. Can you please review the changes?

bigopon (commented):
I've reviewed again, and I think we really need to reconsider the approach a bit. I'm looking around to see what to refactor in controller & fragment to accommodate this.
A comparison:
v1 enhance: enhance from the root node (https://gist.dumber.app/?gist=f02da39bc2d04519a78f5760b6b2e05e)
v2 enhance: enhance from the children of the root node

bigopon:
At the moment, we are allowing the enhancement of a root element even with `au` class on it, because we only enhancing child nodes of the given enhancement host. This is probably not right, as per the bug report from the discourse topic.

The current way of fixing is to check if the to-be-enhanced node is a custom element, then we treat it as a root target. This fix is probably insufficient, as it doesn't cover the case where some binding attributes are set manually on an element to enhance:
```ts
host.setAttribute('data-id.bind', 'someProperty');
au.enhance({ host })
```

In order to fix this, we should change the way enhance work: always start from the enhancement host. An issue with this is that if the host already has an `au` class, but without an binding-like attributes on it, the enhance call will fail because of target/instruction mismatching. Though, this failure is a good thing, I think we can have a documentation to remind users that make sure the target of an enhancement is "clean" from previous hydration, mostly `au` class.

bigopon:
We can close this for now. Some note about future enhancement:
- works with bindings on root enhancement node
- works with template controllers on the root enhancement node (re-activation shouldn't results in extra render location)
- proper throws on invalid conditions

PR #1317, opened 2021-12-25
docs(bindable): minor changes

Sayan751:
# Pull Request

### DIFF
--- a/docs/user-docs/getting-to-know-aurelia/components/bindable-properties/the-basics.md
+++ b/docs/user-docs/getting-to-know-aurelia/components/bindable-properties/the-basics.md
@@ -11,7 +11,7 @@ Bindable properties on components are denoted by the `@bindable` decorator on th
 {% tabs %}
 {% tab title="loader.ts" %}
 ```typescript
-import { bindable } from 'aurelia'; 
+import { bindable } from 'aurelia';
 
 export class Loader {
     @bindable loading = false;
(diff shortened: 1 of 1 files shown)


docs/user-docs/getting-to-know-aurelia/components/bindable-properties/the-basics.md
```diff
@@ -22,16 +22,47 @@ export class Loader {
...
 <loader loading.bind="loadingVal"></loader>
 ```
 
+You can also bind values without the `.bind` part, as can be seen the following example.
+
+```html
+<loader loading="true"></loader>
+```
+
+However, there is a subtle difference. In this case, Aurelia consider the attribute value as a string. Thus, instead of a boolean `true` value the string `'true'` gets bound to the `loading` property. This might cause some unexpected issues. However, you can apply coercion using a [bindable setter](bindable-setter.md) or [defining the bindable datatype explicitly](bindable-coercion.md).
```
CollinHerber, line 43: > Aurelia consider the attribute value as a string.
-> Aurelia `considers` the attribute value as a string.

docs/user-docs/getting-to-know-aurelia/components/bindable-properties/the-basics.md
```diff
@@ -22,16 +22,47 @@ export class Loader {
...
+
+## Attribute naming
+
+As can be seen in the examples above, the `@bindable`s defined inside a custom element can be used a HTML attribute while using the custom element. For example, the `@bindable loading` property is used as the attribute `loading.bind` in the view.
```
CollinHerber, line 47: > can be used a HTML attribute while using the custom element  

-> can be used as an HTML attribute while using the custom element

CollinHerber (commented):
Just some minor English adjustments

docs/user-docs/getting-to-know-aurelia/components/bindable-properties/the-basics.md
Sayan751: Thanks for the review!

docs/user-docs/getting-to-know-aurelia/components/bindable-properties/the-basics.md
Sayan751: Thanks for the review!

docs/user-docs/getting-to-know-aurelia/components/bindable-properties/the-basics.md
```diff
@@ -22,16 +22,47 @@ export class Loader {
 
 This will allow our component to be passed in values. Our specified bindable property here is called `loading` and can be used like this:
 
-```text
-<loader loading="true"></loader>
+```html
```
bigopon, line 25: `html` will be changed to `markup` by gitbook, it's quite an annoying one, I;m not sure how we can handle it to make it more natural to markdown

bigopon (approved):
Nice doc work 👍

PR #1318, opened 2021-12-25
feat: complex bindables from processContent

Sayan751:
# Pull Request

## 📖 Description

This adds support for adding complex bindables from the processContent hook. Additionally the processContent hook now also accpts an instance of logger.

### 🎫 Issues

As of now, it is only possible to bind primitive typed values from processContent. This PR tries to improve that situtation.


docs/user-docs/developer-guides/components-revisited.md
```diff
@@ -695,10 +695,10 @@ The signature of the hook function is as follows.
 
 ```typescript
 // pseudo-code; `typeof TCustomElement` doesn't work in Generics form.
-<TCustomElement>(this: typeof TCustomElement, node: INode, platform: IPlatform) => boolean | void;
+<TCustomElement>(this: typeof TCustomElement, node: INode, platform: IPlatform, logger: ILogger, bindableMap: Map<string, unknown>) => boolean | void;
 ```
 
-There are two important things to note here.
+There are couple of things to note here.
```
CollinHerber, line 701: > There are couple of things to note here.  

-> There `are a` couple of things to note here.

CollinHerber (commented):
Minor English adjustment

docs/user-docs/developer-guides/components-revisited.md
Sayan751: Thanks for the review!

packages/runtime-html/src/template-compiler.ts
```diff
@@ -1906,3 +1924,11 @@ export const templateCompilerHooks = (target?: Function) => {
 /* eslint-enable */
 
 const _generateElementName = CustomElement.generateName;
+
+const SyntheticPropertyName = {
```
bigopon, line 1928: this should just be 2 local variables + 1 loose function if grouping effect is all needed for readability. We can use either naming prefix or source code position to express their relationship

bigopon (commented):
I think we need to discuss the idea a bit before this PR/changes inside it. `processContent` is a static call, why should we introduce scope into the call? This removes the boundary that was quite hard to refactor to create.

Sayan751:
> I think we need to discuss the idea a bit before this PR/changes inside it. `processContent` is a static call, why should we introduce scope into the call? This removes the boundary that was quite hard to refactor to create.

I am a bit hesitant about this as well. I think an alternative would be to create and pass the bindable map from the controller itself. And iterate over the map twice: once in the template compiler to add the HTML attributes, and once again in the controller to register the values to the scope.

bigopon:
I think we need to understand clearer the issues is this feature supposed to solve. So far, the implementation looks complex in the moving parts and the wiring of them.

Sayan751:
> I think we need to understand clearer the issues is this feature supposed to solve. So far, the implementation looks complex in the moving parts and the wiring of them.

@bigopon There are 2 new arguments added to the processContent hook.

The first one is an instance of logger. IMO this is nice to have, so that the devs can log error/warning in a streamlined fashion without resorting to `console.xx`.

The second argument is a "container" (not to confuse with DI) to register complex bindable values. Primitive bindable values are easy to add by adding attributes with primitive values directly on the element from inside processContent. However, when your need is to add an instance of a specific class to one of the bindable property, then the aforementioned way does not work (naturally when you serialize and bind that value to the attribute, there is potential loss of information).  For such cases, the dev can register the bindable-property-name and the complex value tuple to this "container" (for now it's a map) and the templte compiler and the controller do the rest of the work of adding the attributes to the element and registering those values to overrideContext respectively. 

Hopefully this makes the intention clearer.

bigopon:
I think I can see a bit clearer what this PR is trying to add. I'd say there' other ways to do the same thing at the moment, in user land, without adding this API/extras to the framework core. We should explore that option first. An example is to stamp the host node with an ID for the template extracted and then retrieve it in the instance.

Sayan751:
Closing this as discussed in our discord chat.

PR #1319, opened 2022-01-06
chore(docs): add registering a custom element documentation

CollinHerber:
# Pull Request

## 📖 Description

In order to reduce the amount of confusion around create a new custom element and having it be consumable I believe a dedicated page around registering the component would be useful.

### 🎫 Issues

N/A


bigopon (approved):
lgtm

PR #1322, opened 2022-01-14
fix(promise): suppressed TaskAbortError on cancellation

Sayan751:
The promise TC cancels any pre-settlement task if the promise is found to be settled and the pre-settlement task is not yet running. However, as the promise TC acquires the "task-promise" by doing `preSettledTask.result`, cancelling the task results in the rejection of the promise. This rejection on the users' end can be bit confusing. Thus, this error is suppressed in this PR.

# Pull Request


jwx (approved):
lgtm

Issue #1339, opened 2022-01-21
Router: Inability to set a base title


Vheissu:
I tried specifying a title property on the `customize` method on `RouterConfiguration` and there is no title property. There is a title property on the routes themselves, but they completely override the title value that is specified inside of `index.html`

Inside of the provided example app do the following to see the behaviour.

- Run the app
- Take note of the title value, "Routing Extravaganza"
- Navigate to the about page and notice the title now becomes "About"

I would have expected that the title would have used a separator format and looked like this: About | Routing Extravaganza — this behaviour deviates from v1. We need to be able to set the application title, preferably from within the router configuration `customize` call.

Issue #1338, opened 2022-01-21
Router: Binding using params.bind is broken on the load attribute


Vheissu:
The load attribute supports supplying a component/path as well as any parameters to be bound using `params.bind` in the router, `params.bind` is broken. In the example app, you can see that the URL for the "View Profile" tab resolves to `http://localhost:9000/profile(name=rob)` which appears to be using a form of direct routing syntax. This URL should match that of the defined path in `my-app.ts` and look like this: `http://localhost:9000/profile/rob`

Sayan751:
@Vheissu I think there is a configuration mistake for the routes.

The routes for the `profile` is defined as follows:

```ts
    {
      path: 'profile/:name',
      component: ViewProfile,
      title: 'View Profile'
    },
```

Note that there is a `id` property in the routing configuration option, and when that is absent, the `path` is used as fallback for `id`. When you use the `load` custom attribute like this: `<a load="route:profile; params.bind:{name: 'rob'}">View Profile</a>`, the value bound to the `route` property is used as the `id` to search among the configured routes. When none found, the direct routing variant is used as fallback. So with this situation can be corrected by adding a `id` property to the routing configuration:

```ts
    {
      id: 'profile',
      path: 'profile/:name',
      component: ViewProfile,
      title: 'View Profile'
    },
```

Contextually, it can be noted that with the previous configuration in place, `<a load="route:profile/:name; params.bind:{name: 'rob'}">View Profile</a>` will also work. However, adding a `id` property might be more ergonomic.

I think we can update the docs accordingly, and this issue can be closed.

PR #1357, opened 2022-03-03
fix(router): #1338, #1339

Sayan751:
# Pull Request


Vheissu (approved):

docs/user-docs/developer-guides/routing/navigating.md
```diff
@@ -105,10 +105,11 @@ And where things really start to get interesting is when you want to pass parame
 <a load="route:profile; params.bind:{name: 'rob'}">View Profile</a>
 ```
 
-In the above example, we provide the route value. But, then also provide an object of parameters. These parameter values correspond to any parameters configured in your route definition. In our case, our route looks like this:
+In the above example, we provide the route (`id`) value. But, then also provide an object of parameters. These parameter values correspond to any parameters configured in your route definition. In our case, our route looks like this:
```
bigopon, line 108:
```suggestion
In the above example, we provide the route (`id`) value via `route: profile`. But, then also provide an object of parameters via `params.bind: { name: 'rob' }`. These parameter values correspond to any parameters configured in your route definition. In our case, our route looks like this:
```

packages/__tests__/router/smoke-tests.spec.ts
```diff
@@ -988,5 +991,32 @@ describe('router (smoke tests)', function () {
 
       await au.stop();
     });
+
+    it('shows title correctly', async function () {
```
bigopon, line 995: 👍

packages/router/src/router.ts
```diff
@@ -106,6 +106,13 @@ export class RouterOptions {
      * Default: `ignore`
      */
     public readonly sameUrlStrategy: ValueOrFunc<SameUrlStrategy>,
+
+    /**
+     * An optional handler to build the tile.
+     * When configured, the work of building the title string is completely handed over to this function.
+     * If this function returns `null`, the title is not updated.
+     */
+    public readonly buildTitle: ((transition: Transition) => string | null) | null,
```
bigopon, line 115: 👍

packages/router/src/router.ts
```diff
@@ -426,6 +436,9 @@ export class Router {
   private nextTr: Transition | null = null;
   private locationChangeSubscription: IDisposable | null = null;
 
+  /** @internal */
+  public readonly hasTitleBuilder: boolean = false;
```
bigopon, line 440: shouldn't this be a private prop? And start with `_` to minify?

bigopon (approved):
Great find! 👍

packages/router/src/router.ts
Sayan751: You are right. I wanted this to be `private` too. Moreover, I also wanted to express that this is a `readonly` property. However, this property cannot be set in constructor, rather inside the `Router#start` method as follows:

```typescript
(this as Writable<Router>)._hasTitleBuilder = typeof this.options.buildTitle === 'function';
```

And the `Writable`/TS doesn't pick up the private properties. However, I have prefixed the prop with `_` and that does the trick of minifying the property name.

PR #1361, opened 2022-03-05
chore: upgrade ts to 4.6.2

bigopon:
# Pull Request

## 📖 Description

To leverage some new goodies!


bigopon:
Hitting an issue with TS 4.5+ https://github.com/microsoft/TypeScript/issues/48130
Will wait a bit to resolve or change the types to go ahead.

bigopon:
it seems we are relying on a currently broken - but anticipated to work behavior of TS, so we should refactor our typings to work around it for now.

PR #1372, opened 2022-03-20
refactor(switch+promise): deferred view instantiation

Sayan751:
# Pull Request

## 📖 Description

This PR addresses the reported issue on discourse (https://discourse.aurelia.io/t/switch-case-statements-in-templates/4702)
of the CE VM are being instantiated even if the cases are not matched. Previously, the controller was created in the `Case#ctor`. This commit defers it to the `Case#activate`; that is, when the case matches the switch condition and about to be activated.

It also defers the instantiations for promise-tc's VMs.

### DIFF
--- a/packages/__tests__/3-runtime-html/switch.spec.ts
+++ b/packages/__tests__/3-runtime-html/switch.spec.ts
@@ -1069,7 +1069,6 @@ describe('3-runtime-html/switch.spec.ts', function () {
         null,
         getDeactivationSequenceFor(['case-host-1', 'case-host-6']),
         (ctx) => {
-          // console.log(ctx.host.outerHTML);
           const switches = (ctx.controller.children[0] as Controller<Repeat>).viewModel
             .views
             .map((v) => v.children[0].viewModel as Switch);
--- a/packages/__tests__/3-runtime-html/switch.spec.ts
+++ b/packages/__tests__/3-runtime-html/switch.spec.ts
@@ -86,12 +86,12 @@ describe('3-runtime-html/switch.spec.ts', function () {
       public constructor(
         private readonly config: Config,
         @ILogger private readonly $logger: ILogger,
-        @INode node: HTMLElement,
+        @INode node: INode,
       ) {
-        const ceId = node.dataset.ceId;
+        const ceId = (node as HTMLElement).dataset.ceId;
         if (ceId) {
           (this.logger = $logger.scopeTo(`${name}-${ceId}`)).debug('ctor');
-          delete node.dataset.ceId;
+          delete (node as HTMLElement).dataset.ceId;
         }
       }
 
--- a/packages/__tests__/3-runtime-html/switch.spec.ts
+++ b/packages/__tests__/3-runtime-html/switch.spec.ts
@@ -30,6 +30,7 @@ import {
   ICustomElementViewModel,
   ICustomElementController,
   bindable,
+  INode,
 } from '@aurelia/runtime-html';
 import {
   assert,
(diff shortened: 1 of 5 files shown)


packages/runtime-html/src/resources/template-controllers/switch.ts
```diff
@@ -258,21 +258,20 @@ export class Case implements ICustomAttributeViewModel {
...
-  public view: ISyntheticView;
+  public view: ISyntheticView | undefined = void 0;
   private $switch!: Switch;
   /** @internal */ private readonly _debug: boolean;
   /** @internal */ private readonly _logger: ILogger;
   /** @internal */ private _observer: ICollectionObserver<CollectionKind.array> | undefined;
 
   public constructor(
-    factory: IViewFactory,
+    private readonly factory: IViewFactory,
```
bigopon, line 268: if these are private readonly, can we prefix them with `_` for minifcation yet? If the tests are poking these props, maybe we can tweak them so that they test public stuff only.

packages/runtime-html/src/resources/template-controllers/promise.ts
```diff
@@ -234,14 +231,14 @@ export class FulfilledTemplateController implements ICustomAttributeViewModel {
 
   public activate(initiator: IHydratedController | null, flags: LifecycleFlags, scope: Scope, resolvedValue: unknown): void | Promise<void> {
     this.value = resolvedValue;
-    const view = this.view;
+    const view = this.view ??= this.factory.create().setLocation(this.location);
```
bigopon, line 234: same with the other comment, probably shouldn't do this.

packages/__tests__/karma.conf.cjs
```diff
@@ -34,7 +34,7 @@ const testDirs = [
   'validation-i18n',
 ];
 
-const baseKarmaArgs = 'karma start karma.conf.cjs  --browsers=ChromeDebugging --browsers=ChromeHeadlessOpt --browsers=FirefoxHeadless --single-run --coverage --watch-extensions js,html'.split(' ');
+const baseKarmaArgs = 'karma start karma.conf.cjs  --browsers=ChromeDebugging --browsers=ChromeHeadlessOpt --browsers=FirefoxHeadless --single-run --coverage --watch-extensions js,html --bail'.split(' ');
```
bigopon, line 37: can we adjust the test script to make `bail` configurable? I don't think we want to bail on the CI, but locally, this should be great.

packages/__tests__/3-runtime-html/switch.spec.ts
```diff
@@ -1045,6 +1069,7 @@ describe('3-runtime-html/switch.spec.ts', function () {
         null,
         getDeactivationSequenceFor(['case-host-1', 'case-host-6']),
         (ctx) => {
+          // console.log(ctx.host.outerHTML);
```
bigopon, line 1072: maybe just remove it?

bigopon (commented):
Nice work 👍

packages/__tests__/karma.conf.cjs
Sayan751: Note that the `bail` only applies if you invoke the test script like `npm run test-chrome -- --bail`. As we aren't doing that from the CI steps, in CI, the tests won't bail out on first failure. Additional info: the args listed here are ignored so that rest of the unknown/unlisted args can be treated as file pattern (I think though that you already know this part).

packages/runtime-html/src/resources/template-controllers/promise.ts
```diff
@@ -282,14 +277,14 @@ export class RejectedTemplateController implements ICustomAttributeViewModel {
 
   public activate(initiator: IHydratedController | null, flags: LifecycleFlags, scope: Scope, error: unknown): void | Promise<void> {
     this.value = error;
-    const view = this.view;
+    const view = this.view ??= this.factory.create().setLocation(this.location);
```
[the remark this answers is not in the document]
Sayan751: Good point! Changed.

packages/runtime-html/src/resources/template-controllers/switch.ts
Sayan751: Agreed. Changes done!

Sayan751:
@bigopon it should now again be ready for review/merge.

packages/__tests__/karma.conf.cjs
bigopon: yes, I forgot this, thanks for the reminder.

PR #1375, opened 2022-03-29
fix(runtime-html): local dependencies for local element

### LINKED ISSUE #1373: Custom elements do not appear in local templates.
# 🐛 Bug Report

I'm not able to use custom elements in local templates. This is easily reproducible; you can check out the gist here: https://gist.dumber.app/?gist=018dd79b7b043b8f99403c735dcaa04f

Note in the linked gist that using `<import>` inside the local template doesn't seems to have any affect.

## 🤔 Expected Behavior

Custom elements should work in local templates as well.

Sayan751:
# Pull Request

## 📖 Description

It adds a flag for local element and for local elements falls back to parent container if the current container cannot find the resource.

### 🎫 Issues

Closes #1373


packages/runtime-html/src/template-compiler.ts
```diff
@@ -41,6 +41,11 @@ import type { IProjections } from './resources/slot-injectables.js';
 import type { BindingCommandInstance, ICommandBuildInfo } from './resources/binding-command.js';
 import type { ICompliationInstruction, IInstruction, } from './renderer.js';
 
+declare module '@aurelia/kernel' {
```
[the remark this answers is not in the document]
Sayan751: Good point. Done.

bigopon:
I think the fix looks appropriate, though it's a bit uncomfortable just to poke the parent container "randomly" like this. And the PR is lacking tests for spreading to local element, I just realized. I'll be having a look at this for a bit.

bigopon:
I think there's another interesting case: local element using owning element:
```html
// my-element
<template as-custom-element="my-local-element">
  <my-element>
</template>
```
This is not covered at the moment. Do we support this?

Sayan751:
> I think the fix looks appropriate, though it's a bit uncomfortable just to poke the parent container "randomly" like this. And the PR is lacking tests for spreading to local element, I just realized. I'll be having a look at this for a bit.



> I think there's another interesting case: local element using owning element:
> 
> ```
> // my-element
> <template as-custom-element="my-local-element">
>   <my-element>
> </template>
> ```
> 
> This is not covered at the moment. Do we support this?

Great catch! Need to avoid that. Thanks.

Sayan751:
> I think there's another interesting case: local element using owning element:
> 
> ```
> // my-element
> <template as-custom-element="my-local-element">
>   <my-element>
> </template>
> ```
> 
> This is not covered at the moment. Do we support this?

@bigopon pushed a fix for this issue.

> bit uncomfortable just to poke the parent container "randomly" like this

It is not "random" though. Either you register your custom elements globally (in which case the first attempt at finding the custom element will work) or you register the custom elements locally as dependencies of individual custom elements (I am referring to `CustomElementDefinition#dependencies`).  So when dealing with local elements, it makes sense to fall back on the parent container to resolve/find the custom elements, because in that the parent container would that one from the parent custom element.

> And the PR is lacking tests for spreading to local element

Can you please elaborate this part?

bigopon:
The "random" word might have been off. I wanted to express a concern related to relying on container hierarchy where it's not supposed to. Resources has a different semantic in the context of a container hierarchy. So walking to the parent, while being correct in this case, could set a trap that we may walk into later.

> > And the PR is lacking tests for spreading to local element

> Can you please elaborate this part?

I was thinking that if it can recognizes custom element, then it should be able to compile a spread instruction against that custom element usages. Though the API is quite isolated and doesn't walk, so it's probably fine either way.

With regards to the issue, I think there could be another way to fix that may not require us to deal with uncertainty (as in in applications, there may be a custom template controller that adds a container into its hierarchy).

bigopon:
@Sayan751 I've pushed a commit, in which we build proper dependency for local elements during the construction of its definition/class instead of delaying it until future. This is quite safe because the local element is in complete control of the template compiler, and not exposed anywhere.

packages/runtime-html/src/template-compiler.ts
```diff
@@ -1514,11 +1508,21 @@ export class TemplateCompiler implements ITemplateCompiler {
...
-      LocalElement.define(LocalTemplateType);
+      // LocalElement.define(LocalTemplateType);
 
       root.removeChild(localTemplate);
     }
+
+    let i = 0;
+    const ii = localElTypes.length;
+    for (; ii > i; ++i) {
+      (CustomElement.getDefinition(localElTypes[i]).dependencies as Key[]).push(
```
Sayan751, line 1521: Nice! I have not though about that. Great solution. I thought that Container will complain about the duplicate registrations.

It makes sense to put/compute the deps array outside the loop as it is not dependent on the iterations of the loop.

packages/__tests__/3-runtime-html/template-compiler.spec.ts
```diff
@@ -2213,6 +2213,43 @@ describe('TemplateCompiler - local templates', function () {
...
+    <my-le prop.bind></my-le>
+    <template as-custom-element="my-le">
+      <bindable property="prop"></bindable>
+      my-le-content
+      <my-app if.bind="prop"></my-app>
+    </template>
+    `.trim();
+    @customElement({ name: 'my-app', template })
+    class App {
+      public prop = false;
```
Sayan751, line 2228: I think when you set this to `true` we will have a stack overflow. From that sense, it makes more sense for me to put the previous restriction in place.

bigopon: there's an if to prevent that in the `<my-le/>`. I think it's an inherent "rule" that if there is a recursive call, it should be constrained.

packages/runtime-html/src/template-compiler.ts
bigopon: when we have more than 1 local element, it's ideal that all local elements can "see" each other. In order to have that, the local elements should have their dependencies built after the owning element has collected all the local elements. Though great poke, I didn't have a test for this case.

bigopon:
Thanks @Sayan751 , the fix is in.

Sayan751:
> Thanks @Sayan751 , the fix is in.

Thank you @bigopon for jumping in!!

packages/__tests__/3-runtime-html/template-compiler.spec.ts
Sayan751: Exactly! That's why IMO we should never support this case and throw a compilation error, because `my-le my-app` will always fail.

packages/runtime-html/src/template-compiler.ts
Sayan751: I was referring to this part:

```ts
 ...context.def.dependencies ?? emptyArray,
 ...context.deps ?? emptyArray,
```

This can be extracted right before this loop (L1523), right?

packages/__tests__/3-runtime-html/template-compiler.spec.ts
bigopon: It'll be quite unnatural if normal custom element can do that, and local element cannot. I think we can add a warning though, in dev mode.

packages/runtime-html/src/template-compiler.ts
bigopon: ahh yes thanks, I thought there was only one loop, while I actually wrote this 2nd loop lol

PR #1377, opened 2022-04-03
chore(rollup): fix __DEV__ build var

hiaux0:
# Pull Request

## 📖 Description
I ran npm run build:test successfully, but the /dist index.js file (both esm and cjs) show non-dev code
img

Steps to repro
npm run build:test
check any dist folder
Issue: non-dev branch

@Sayan751 for the actual fix

## 👩‍💻 Reviewer Notes

After fix, correctly shows expected error message
image


Sayan751:
I think we should also leverage the flag to produce a non-minified versions as well. With that from userland, we can alias (I am currently thinking webpack) the `/aurelia|@aurelia/` imports to the dev versions during development. To his end, we can also adjust the skeleton used by the CLI.

bigopon:
@Sayan751 at the moment, we are having 2 dist forms per format: minified and non-minified (dev). Can you elaborate what non-minified in your comment means?

Sayan751:
@bigopon I might have misspoke there. I thought we are always producing minified versions where the variables are obfuscated. If there is a non-minified dev build is produced, then it can also be used in the user land to debugging purpose. My comment is more of a suggestion/wish rather, as I have not checked the details myself yet.

bigopon:
@Sayan751 thanks. I thought you might have been onto something 😄 Nvm me then, I'll get this fix going.

bigopon (approved):

PR #1379, opened 2022-04-06
fix(parcel-transformer): fix html->js transformer path

3cp:
I only tested ts project locally, missed this issue on js project.


packages-cjs/parcel-transformer/src/index.ts
```diff
@@ -17,13 +17,20 @@ export default new Transformer({
     // parcel conventions puts app's index.html inside src/ folder.
     if (asset.filePath.endsWith('src/index.html')) return [asset];
 
+    const auOptions = preprocessOptions(config as IOptionalPreprocessOptions);
+    // after html template is compiled to js, parcel will apply full js transformers chain,
+    // we need to skip them here, then parcel will apply the rest standard js chain.
```
bigopon, line 22: 👍

Issue #1374, opened 2022-03-25
Aurelia 2 does not work with file names index.html/index.ts and unnamed classes


hoangcuongvn:
# 🐛 Bug Report
In Aurelia 1, I can use file names `index.html/index.ts` for custom elements and routes. E.g
```
resources
  custom-elements
    my-element
      index.html
      index.ts
routes
  my-route
    index.html
    index.ts
```
But in Aurelia 2, the app crashes if I use file names `index.html/index.ts` (as in the file structure above) for custom elements and routes. It's strange that Aurelia 2 accepts any file names e.g `foo.html/foo.ts` but `index.html/index.ts`.

In Aurelia 1, I can use unnamed for route classes. E.g
```
// routes/my-route.ts
export default class {
}
```
Aurelia 2 does not accept the unnamed class above even through the route configuration does not require a class name. E.g
```
@route({
  routes: [
    {
      path: "my-path",
      component: import("routes/my-route"),
    },
  ]
})
```
I think Aurelia 2 should support unnamed classes as in Aurelia 1.
## 🌍 Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Aurelia          | Latest
| Language         | Typescript
| Browser          | Chrome/Firefox
| Bundler          | Webpack
| Operating System | Win10
| NPM/Node/Yarn    | NPM 1.17.0, Node 16.3.0, Yarn 1.22.15

3cp:
I did use `index.js` for router component in Aurelia v1.
Aurelia v1 figures out every details at runtime, while Aurelia 2 does it in compile time with static analysis. In other words, Aurelia 2 tooling modifies your source code.

For Aurelia v1, as long as the module loading works, it can get the element. For example `<require from="./my-element"></require>` will try `my-element.js/ts`, and fallback to `my-element/index.js/ts` (this is part of Nodejs module resolution).

For Aurelia 2, right now, we assign custom-element name at compile name based on file name. So it will assign name "index" to the element `@customElement({name: 'index',` for that `index.js` file. We could update the logic to decide custom-element name based on the whole file path, not just the file name.

As for router component, the story is little bit different. Aurelia v1 router does NOT case about the custom-element name (because it didn't need to use the name). As long as you have only one export in the component, no matter what name you gave, Aurelia v1 will just load that component. The default export is almost same as a named "default" export (although strictly it should be default export from a namespace, it actually is a "default" property on an object due to transpiling.)

[...]

PR #1383, opened 2022-04-08
feat(plugin-conventions): support foo-bar/index.js Nodejs convention

3cp:
Partially addresses #1374


packages-cjs/__tests__/plugin-conventions/preprocess-html-template.spec.ts
```diff
@@ -732,4 +732,62 @@ export function register(container) {
...
+    _e = CustomElement.define({ name, template, dependencies });
+  }
+  container.register(_e);
+}
+`;
+      const result = preprocessHtmlTemplate({ path: path.join('lo', 'FooBar', 'index.html'), contents: html }, preprocessOptions());
+      assert.equal(result.code, expected);
+    });
+
+    it('processes template with css pair in index file', function () {
```
bigopon, line 774: I think later we should have a test with shadow DOM for css too.

bigopon (approved):

packages-cjs/__tests__/plugin-conventions/preprocess-html-template.spec.ts
3cp: We should have e2e test for parcel later, just like webpack. No even parcel doesn't have unit tests example, we are still covered.

PR #1387, opened 2022-04-15
BREAKING CHANGES(router-lite): removed direct routing infra; configured routing only

Sayan751:
# Pull Request

## 📖 Description

This PR primarily removes the direct routing infra from route-lite as direct routing is now supported via the router (formerly direct-router) package. That is, the router-lite package will support configured routing only. Additionally, it removes a couple of other unused, confusing routing configurations.

It also adds better support for fallback routes or unrecognized/unknown routes.

## 📑 Test Plan

Most of the tests written for direct routing are converted to configured routing variant. Also, some new tests are added.

## ⏭ Next Steps

I will continue working on the router-lite  towards fixing the reported issues.


packages/router-lite/src/viewport-agent.ts
```diff
@@ -549,67 +543,26 @@ export class ViewportAgent {
         const deactivateFlags = this.viewport.stateful ? LifecycleFlags.none : LifecycleFlags.dispose;
         const activateFlags = LifecycleFlags.none;
         b.push();
-        switch (tr.options.swapStrategy) {
```
bigopon, line 552: I think it's worth adding a comment here explaining the intentionally omitted swap order feature.

Sayan751: You are right. As it is difficult to find a proper place in the source code to add a comment, as now it is removed, I will post the explanation below.

The originally planned `SwapStrategy` is the strategy to deactivate the current component and activate the next one. There are following possible values:

- `sequential-add-first`: the next component is activated before deactivating the current component; this happens sequentially.
- `sequential-remove-first`: the current component is deactivated before activating the next component; this happens sequentially (default).
- `parallel-remove-first`: the deactivation of the current component and the activation of the next component happens in parallel.

In my opinion, the default is quite alright and there is no need to implement the other strategies for 90% of the use-cases. Therefore, I have removed the other 2 options for the sake of simplicity and reduced effort towards testing and maintenance. 

If we do see valid use-cases, then we can consider bringing those back or simply make the Router API more inheritance/extension friendly.

bigopon (approved):
Changes look good to me. Let's go with this. We will fix it forward if there's any issues.

PR #1395, opened 2022-04-22
docs: HttpClient.configure returns the config, not void

dfroger:
Following the docs before the fix, I get this error:

TS2345: Argument of type '(config: HttpClientConfiguration) => void' is not assignable to parameter of type 'RequestInit | ((config: HttpClientConfiguration) => HttpClientConfiguration) | HttpClientConfiguration'. Type '(config: HttpClientConfiguration) => void' is not assignable to type '(config: HttpClientConfiguration) => HttpClientConfiguration'. Type 'void' is not assignable to type 'HttpClientConfiguration'.


bigopon:
I think both void / config is fine. Because void is the intention, while config is the simpler way to write it. Though let's get this in first.

PR #1400, opened 2022-04-25
feat(hmr): add in hmr capabilities

brandonseydel:
# Pull Request


packages-cjs/plugin-conventions/src/preprocess-resource.ts
```diff
@@ -152,16 +192,71 @@ function modifyResource(unit: IFileUnit, options: IModifyResourceOptions) {
     conventionalDecorators.forEach(([pos, str]) => m.insert(pos, str));
   }
 
+  if (exportedClassName && process.env.NODE_ENV !== 'production') {
```
bigopon, line 195: This case we probably have to either parse the `@customElement` deco call, or walk up the folder the derive the name from there, we can do it at a later point.

packages-cjs/plugin-conventions/src/preprocess-resource.ts
```diff
@@ -152,16 +192,71 @@ function modifyResource(unit: IFileUnit, options: IModifyResourceOptions) {
...
+    const hmr = `
```
bigopon, line 196: I think maybe the entire block here can be in some form of a callback that takes exportedClassName as its parameter.
So different bundler can output different code, the options can have an extra prop:
```ts
hmrCode: (className: string, type /* Element | Attribute | etc... */) => {
  return 'some hmr code here ...'
}
```

bigopon: cc @3cp

bigopon (commented):
Looks great! Can we have some demo gif pls 😄

3cp:
I am against to add HMR inside conventions.

1) HMR sounds like has nothing to do with conventions.

2) If user manually writes customElement decorator, it will bypass conventions logic. But we still want to support HMR in that use case. I believe the implementation here didn't cover the use case.

User can also write Aurelia 2 app without using our conventions in the toolchain at all. Aka, totally manual and verbose code.

For the above two arguments, I suggest to invest the possibility to add HMR support in other place (or a standalone npm package like we did for Aurelia v1).

Sayan751:
I think not everyone would like to use the convention, though. I think it would be better to create multiple plugin (if not multiple packages) that focuses on a single aspect, and then an umbrella plugin to aggregate the config or whatever for the ease of use (although it might be postponed if super complicated or if the individual plugins can be easily used). IMO, it is important to ensure that HMR works individually without relying on convention.

brandonseydel:
> I think not everyone would like to use the convention, though. I think it would be better to create multiple plugin (if not multiple packages) that focuses on a single aspect, and then an umbrella plugin to aggregate the config or whatever for the ease of use (although it might be postponed if super complicated or if the individual plugins can be easily used). IMO, it is important to ensure that HMR works individually without relying on convention.

It doesn't rely on convention.  HMR operates by itself entirely in the plugin-convention package.  That is why I am thinking it is named improperly as it takes care of more than just one thing.  We can seperate out into many packages, but there should be one entry into code manipulation I feel.

3cp:
plugin-conventions is NOT used when user does not used any of our conventions plugins.

In our doc, never forced user to use webpack-loader. It's an optional helper.

This still didn't answer my question on that use case.

brandonseydel:
@3cp - hmr and conventions are controlled via options

3cp:
Oh sorry. I missed that. Although I would like hmr has its own package, as long as hmr is decoupled with conventions, that meets the expectation.

Yes please rename the package to something else to reflect the feature set.

3cp:
Question:
Correct me if I was wrong, this implementation injects some code to deal with hmr into every single component in user app.
Why is this needed? Cannot we support hmr inside the core implementation of CustomElement itself (without duplicating hmr code in every single instances of CustomElement)?

brandonseydel:
@3cp - I don't think putting HMR code in the runtime-html is good in any way.

brandonseydel:
I do see what you are saying though.  I don't know what @bigopon thinks of that, or if any limitations come to mind?

3cp:
It's just that I don't think any other frameworks are using duplicated code injections to support hmr.

3cp:
Putting in CustomElement does not necessarily mean put it in runtime-html package.

JavaScript is a language allowing money patching which lots of dev and testing tools abuse :-)

brandonseydel:
> It's just that I don't think any other frameworks are using duplicated code injections to support hmr.
> 
> It feels like we could work on making this one cleaner.

Vue/ember/angular/svelte all do

brandonseydel:
https://github.com/vuejs/vue-loader/blob/b391b04ed3335a2d2c9010101d98c2d9977626e8/src/hotReload.ts
https://github.com/sveltejs/svelte-hmr/blob/master/packages/svelte-hmr/lib/make-hot.js

brandonseydel:
> Putting in CustomElement does not necessarily mean put it in runtime-html package.
> 
> JavaScript is a language allowing money patching which lots of dev and testing tools abuse :-)

Yea we are using that to our advantage to get this to work.  I think we can inch towards something in the future trying to patch CustomElement from outside.

brandonseydel:
Actually HMR needs to be granular to the specific component that is changing and must reside in the file we are swapping.

3cp:
Thx for showing up the code examples, and sorry for my inexperienced assertion :-)

I didn't know they do that in webpack loader, never passed skin deep into hmr.

brandonseydel:
https://gyazo.com/83fa4963c094103847fe921d0a5ac182
https://gyazo.com/0458a641653339485aa3f1bc63291cc9
image
https://gyazo.com/5f3bae5bbce974d548391af490a8336e
https://gyazo.com/4bfc99edd29556000f69f91af06d7470

3cp:
Very cool!

BTW, I had a look again at au1 hmr.

au1 did hmr in aurelia-loader-webpack at runtime. Because au1 has an abstracted module loader, it can actually inject those logic at runtime which is cool.

bigopon:
I guess our next task is to find a name that is appropriate for the package. Any suggestion on that? Some immediate candidates:
- component-processor
- resource-processor
maybe add `pre` prefix to `processor`. Any suggestions?

3cp:
I recall, quite some time ago, we talked about moving all package-cjs into individual repos.

Those packages are not core packages. Putting them into monorepo actually hurts development more than helping, especially the cjs/esm mode build switching for local setup.

We should revisit the idea before production release.

bigopon:
> cjs/esm mode build switching for local setup

With rollup, we no longer need to switch to build. Though we are still switching the `main` and `module` fields to run the tests of these packages. If we are to split it out, I'd imagine it'd be more difficult.

When running test with mocha, we still need to do this because some reasons I'm not sure 100% yet. Running `mocha --loader esm` or `mocha --register ts-node/esm` didn't seem to work. I'll need to investigate that.

3cp:
If you don't use `"type": "module"` in package.json, you can still ship both cjs and esm with explicit .cjs and .mjs files. And mocha should work better. At least my ava setup is happier with default commonjs type in package.json.

```js
"main": "dist/index.cjs",
  "module": "dist/index.mjs",
  "types": "dist/index.d.ts",
  "exports": {
// works with require(...)
    "require": "dist/index.cjs",
// works with import.
    "import": "dist/index.mjs"
  },
```

Ref: https://github.com/gorbjs/shared/blob/main/package.json

bigopon:
@3cp thanks, I'll be applying that.

bigopon:
I've created a discussion for our rename here https://github.com/aurelia/aurelia/discussions/1404
I think we can go ahead first, with a rename later.

3cp:
@brandonseydel BTW, I could not get my head around how to support vite prod build.

Vite uses different path for dev and prod build. The prod build tries to parse our html as html, but we already transpiled html into js code. The whole rollup chain does not have concept of asset type, nor allowing changing extname in id (the file name).

When you are on Vite, see if you can come up with any workaround, thx.

For reference: the vite au2 plugin which works (kind of) in dev mode:
https://github.com/3cp/au2-vite/blob/main/vite.config.ts

brandonseydel:
> @brandonseydel BTW, I could not get my head around how to support vite prod build.
> 
> Vite uses different path for dev and prod build. The prod build tries to parse our html as html, but we already transpiled html into js code. The whole rollup chain does not have concept of asset type, nor allowing changing extname in id (the file name).
> 
> When you are on Vite, see if you can come up with any workaround, thx.
> 
> For reference: the vite au2 plugin which works (kind of) in dev mode: https://github.com/3cp/au2-vite/blob/main/vite.config.ts

Yea been fighting with vite some on the mangler side of things.  It doesn't seem to want to listen when using esbuild.....

packages-cjs/plugin-conventions/src/preprocess.ts
```diff
@@ -27,7 +27,12 @@ export function preprocess(
         unit.filePair = path.basename(filePair);
       }
     }
-    return preprocessHtmlTemplate(unit, allOptions);
+
+    const hasViewModel = Boolean(allOptions.jsExtensions.map(e =>
```
3cp, line 31: @brandonseydel I might be confused, but I don't think this ever worked. We don't unit test hmr setup, so might missed this bug.

Boolean([false, false]) yields true

cc @bigopon

3cp: @brandonseydel @bigopon my bad, I miss understood the code :(

PR #1403, opened 2022-04-28
build: change dist extension to .cjs and .mjs, optimize CI caching

bigopon:
- rename all dist extensions to `.mjs` and `.cjs`
- change `main` and `module` fields in `package.json` for all packages
- remove field `"type": "module"` in all packages
- Optimize our CircleCI caching, and remove unnecessary ci steps for so pipeline gets faster and uses fewer credits.

cc @3cp

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -643,12 +643,13 @@ workflows:
           <<: *filter_ignore_develop_release
       - checkout_install_cache:
           <<: *filter_ignore_develop_release
+          name: checkout_install
           requires:
               - check_doc_abort
       - build_dev:
           <<: *filter_ignore_develop_release
           requires:
-              - checkout_install_cache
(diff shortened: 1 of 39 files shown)


3cp:
The scripts to switch between cjs and esm can be removed.
Dev doc needs update.

This also means we don't need separate folder for packages-cjs, they can all go back to packages folder.

Local dev and testing should be very straight now :-)

bigopon:
> This also means we don't need separate folder for packages-cjs, they can all go back to packages folder.

Thanks. They may not need to be separate, but I think tooling packages could still use a different folder paths so we can stay focused in local development.

What about renaming it to `packages-tooling`?

> Dev doc needs update.

Thanks for this, I forgot this bit.

3cp:
Packages-tooling does sound better.

Issue #1341, opened 2022-01-21
router-lite: Once router error is triggered, the router does not recover


Vheissu:
If, for whatever reason an error is triggered in the console, the router is never able to recover. A good example of this is the repository I created is attempting to click the "View Profile" menu item, seeing the console error and then attempting to navigate to the welcome or about page, the router does nothing and no further errors are triggered.

Furthermore, you'll notice more weird behaviour. Go to the "About" tab first, see the about page load and then click "View Profile" an error appears in the console, the URL remains the same `http://localhost:9000/about` however, it appears to have gone back to the default route as we see `Welcome to Aurelia 2!` — what should happen if a route fails is that it falls back to the prior working route (looks like maybe it attempts to do that considering the URL remains the same as the current route).

Vheissu:
Closing this as a fix appears to have been merged.

Sayan751:
@Vheissu The issue is still there for router-lite. I have renamed this and planning a fix. Keeping this assigned to myself.

PR #1406, opened 2022-04-29
fix(router-lite): downstream propagation of fallback

Sayan751:
# Pull Request

### 🎫 Issues

This PR partially addresses the issue #1341. Whenever, the router gets an unconfigured route it throws error and could not recover from that. This PR resets the transition in such cases and also facilitates propagation of fallback in downstream of the configured route-tree.

## 👩‍💻 Reviewer Notes

ATM the router-lite guards against erroneous transition. Although I think it might be a bit too strict, it is not essentially a bad thing. 

A complete solution would be to implement an "undo" of erroneous transitions and there by resetting the error status of the transition. This is not be done for every error but for case-by-case basis at first. However, that is a bit involved, and I would like to address that later.

The fallback propagation would help in terms of recovery from erroneous transition due to configured route or in other words preventing such erroneous transition.


packages/router-lite/src/route-definition.ts
```diff
@@ -57,33 +58,33 @@ export class RouteDefinition {
...
-      const config = isPartialChildRouteConfig(routeable)
-        ? Route.isConfigured(type)
-          ? Route.getConfig(type).applyChildRouteConfig(routeable)
-          : RouteConfig.create(routeable, type)
-        : Route.getConfig(def.Type);
-
       let routeDefinition = $RouteDefinition.get(def);
       if (routeDefinition === null) {
-        routeDefinition = new RouteDefinition(config, def);
+        const type = def.Type;
```
bigopon, line 80: 👍

packages/router-lite/src/route-tree.ts
```diff
@@ -761,5 +757,8 @@ function createFallbackNode(
       emptyObject
     ),
     null);
+  // Do not pass on any residue. That is if the current path is unconfigured/what/ever ignore the rest after we hit an unconfigured route.
+  // If need be later a special parameter can be created for this.
+  vi.children.length = 0;
```
bigopon, line 762: 👍 nice one, though we probably will need to refactor this soon, since this viewport instruction mutation likely to causes confusion in the future.

packages/__tests__/router-lite/hook-tests.spec.ts
```diff
@@ -2072,4 +2073,1006 @@ describe('router hooks', function () {
...
+      await tearDown();
+      verifyInvocationsEqual(mgr.fullNotifyHistory, [
+        ...$(phase, ['gc12', 'c1', 'gc21', 'c2', 'root'], ticks, 'detaching'),
+        ...$(phase, ['gc12', 'c1', 'gc21', 'c2', 'root'], ticks, 'unbinding'),
+        ...$(phase, ['root', 'c1', 'gc12', 'c2', 'gc21'], ticks, 'dispose'),
+      ]);
+      mgr.$dispose();
+    });
+
+    for (const [name, fallback] of [['fallback same as CE name', 'nf'], ['fallback different as CE name', 'not-found']]) {
```
bigopon, line 2920: 👍

bigopon (approved):
Nice fix, thanks! Though I start to think we may need to add some overview of the route structure to the start of each test to help with imagining what to be asserted. Took me a bit to go through the cases.

PR #1412, opened 2022-05-07
feat: aurelia state plugin

bigopon:
# Pull Request

## 📖 Description

Enable global store connect/dispatch pattern directly in templating. Example template:

```html
<input value.state="text">

to dispatch from event
<input value.state="text" input.dispatch="$event.target.value">
```

Registration to enable the state plugin:

```ts
import { StandardStateConfiguration } from '@aurelia/state';
...

.register(StandardStateConfiguration)
// or with a custom initial state & reducers:
.register(StandardStateConfiguration.init(
  myInitialState,
  /* reducer 1 */ (state, action) => {
    switch (action.type) {
      case 'event':
        return { ...state, newProp: action.payload.value }
    }
  },
  /* reducer 2 */ (state, action) => {
      return { ...state, lastModified: Date.now() };
  }
)
```


bigopon:
@zewa666 I'll go ahead with this current implementation, it's a private package so it won't be published to npm, we can continue the iteration after this.

PR #1424, opened 2022-05-17
fix(plugin-conventions): upgrade deep-dep source-map to fix nodejs v18 issue

3cp:
https://github.com/mozilla/source-map/issues/454

Upgraded modify-code which upgraded source-map to 0.8.0-beta.0


package-lock.json
```diff
@@ -340,7 +340,7 @@
     },
     "examples/ui-virtualization": {
       "name": "@examples/app-ui-virtualization",
-      "version": "2.0.0-alpha.31",
+      "version": "2.0.0-alpha.30",
```
bigopon, line 343:
```suggestion
      "version": "2.0.0-alpha.31",
```

PR #1438, opened 2022-05-28
refactor(compiler): remove no-action mode for custom element content

### LINKED ISSUE #1431: Have a default compilation behavior for slotting content
# 💬 RFC

Assume we have a custom element `<toggle-button />` with the following template
```html
<button>
  <au-slot></au-slot>
</button>
```

At the moment, when the default template compiler sees the following template:
```html
<toggle-button>
  Switch to ${viewType === 'grid' ? 'list' : 'grid' } view
</toggle-button>
```
It won't know that `Switch to ${viewType === 'grid' ? 'list' : 'grid' } view` is meant for the default `<au-slot/>`, since it's not safe to make such assumption.

Consider adding a configuration in the template compiler to have a default way of compiling slot this way, bringing the behavior inline with v1 light dom slotting by default.

bigopon:
At the moment, template compiler has a "default" mode beside shadow DOM and `au-slot`, where it doesn't act on the content, this results into unintuitive situation where it doesn't work like either shadow DOM or v1. Remove this mode, and make the template compiler automatically pick `au-slot` mode, unless `shadowOptions` is given for a custom element.
Pseudo logic check:
```ts
if (elementDefinition.shadowOptions != null) // shadow dom mode
else // au-slot mode
```
Closes #1431


Vheissu:
I think whatever makes it easier for those wanting to port over to v2 from v1, default slot behaviour that developers are already familiar with is a great thing. This is an A+ from me.

Alexander-Taran:
Agree with @Vheissu 
still mass replace in files will be in order..

bigopon:
With this PR going in, the `leave things intact` will be removed, making the following usages not possible:
```html
<case-host switch.bind="day">
  Expected to be delivered
  <case-host case.bind="1" >tomorrow.</case-host>
  <case-host case.bind="2">in 2 days.</case-host>
  <case-host default-case>in few days.</case-host>
</case-host>
```
It'll have to be re-written this way
```html
<case-host>
  <template switch.bind="day">
    Expected to be delivered
    <case-host case.bind="1">tomorrow.</case-host>
    <case-host case.bind="2">in 2 days.</case-host>
    <case-host default-case >in few days.</case-host>
  </template>
</case-host>
```

Similarly, this will stop working:
```html
<my-report promise.bind="fetchReports()">
  <report then="reportsData" reports.bind="reportsData.items"></report>
  <no-report catch=$error>${$error}</no-report>
</my-ce>
```
a more boilerplate-y form is required:
```html
<my-report>
  <template promise.bind="fetchReports()">
    <report then="reportsData" reports.bind="reportsData.items"></report>
    <no-report catch=$error>${$error}</no-report>
  </template>
</my-ce>
```

Basically hierarchically reliant template controllers will need to have their parent-child declared next to each other. This could be a good thing since it'll be harder to make a weird/hard to understand template.

PR #1443, opened 2022-06-02
feat(router): add default route to child viewports

jwx:
# Pull Request

## 📖 Description

This PR activates the default configured route for viewports below the root viewport.

### 🎫 Issues

Related to https://github.com/aurelia/aurelia/issues/307, https://github.com/aurelia/aurelia/issues/369, https://github.com/aurelia/aurelia/issues/367 and https://github.com/jwx/aurelia-direct-router/issues/35. 

## 👩‍💻 Reviewer Notes

## 📑 Test Plan

## ⏭ Next Steps

- Resolve remaining issues/requests from Aurelia Direct Router
- Ensure v1 feature parity
- Continue with comments
- Move relevant parts to `Metadata`
- Add tests for `au-viewport-scope`
- Review the extension points


packages/router/src/router.ts
```diff
@@ -368,7 +371,7 @@ export class Router implements IRouter {
     // If there are instructions appended between/before any navigation,
     // append them to this navigation. (This happens with viewport defaults
     // during startup.)
-    coordinator.appendedInstructions.push(...this.appendedInstructions);
+    coordinator.appendedInstructions.push(...this.appendedInstructions.splice(0));
```
bigopon, line 374: should this be `slice`? if `splice` is intended, then let's use `.shift()` so it's less confusing.

packages/router/src/routing-scope.ts
```diff
@@ -793,22 +792,29 @@ export class RoutingScope {
     if (path.startsWith('/') || path.startsWith('+')) {
       path = path.slice(1);
     }
-    const recognizer = new RouteRecognizer();
 
-    recognizer.add(cRoutes);
-    const result = recognizer.recognize(path);
-    if (result !== null) {
+    const idRoute = routes.find(route => route.id === path);
```
bigopon, line 796: 👍 I think the router lite also does this, this'll be good to have the parity here.
cc @Sayan751

packages/router/src/routing-scope.ts
```diff
@@ -466,6 +459,12 @@ export class RoutingScope {
           matchedInstructions = clearEndpoints.map(endpoint => RoutingInstruction.createClear(router, endpoint));
         }
       }
+      // If there are any unresolved components (functions or promises) to be appended, resolve them
+      const unresolved = matchedInstructions.filter(instr => instr.component.isFunction() || instr.component.isPromise());
```
bigopon, line 463: Probably we can have a `type` property on `ComponentAppellation` to describe it instead
```ts
const unresolved = matchedInstructions.filter(instr => instr.component.type & (ComponentType.Function | ComponentType.Promise));
```

packages/router/src/router.ts
```diff
@@ -826,6 +829,10 @@ export class Router implements IRouter {
       this.configuration.options.useUrlFragmentHash) {
       basePath = '';
     }
+    // if (basePath === null || (state !== '' && state[0] === '/') /* ||
+    //   this.configuration.options.useUrlFragmentHash */) {
+    //   basePath = '';
+    // }
```
bigopon, line 835: if it's not needed, let's delete it?

packages/router/src/routing-scope.ts
jwx: I think the `lite` requires that you specify the id as the `route` part of a load. At least that's the restriction Fred wanted to put on it when I talked about this the first time. That's a long time ago, though, so it might've been implemented differently. Or changed since then.

packages/router/src/routing-scope.ts
jwx: The `router`'s had function resolution of components since before there even were two routers. The addition here is that _appended_ instructions also get function resolution to the components (something I skipped earlier for, in hindsight, not a good enough reason). It's also moved slightly to make sure all are resolved as late as possible.

packages/router/src/router.ts
jwx: Yeah, decided to move this to a later PR and missed this when I was cleaning up. Will address.

PR #1446, opened 2022-06-03
feat(router-lite): navigation model

Sayan751:
# Pull Request

## 📖 Description

This is parallel to v1's `router.navigation`.


3cp:
Is there any difficulty for router to automatically create and dispose the navModel? (I assume router has some visibility of the lifecycle of the view-model.

The current user land code looks bit unnecessarily verbose, at least comparing to au1's router.

Sayan751:
@3cp The API can be made a bit simpler; for example, instead of `this.navModel = await this.router.getNavigationModel(this.routeCtx);` it can be `this.navModel = await this.routeCtx.getNavigationModel();`. I can make this change.

However, I think disposing the navigation model automatically is a bit tricky ATM. And since such navigation models are expected to be used in a nav-bar-like CE, which is not typically limited to a handful of instances in an app, it may not be that boiler-platey as you suggested.

3cp:
The api design implies getNavigationModel is expensive:
1. Only create if user wants it. 
2. Dispose after use. 

The au1 router.navigation is always available, which is much easier to use.

Sayan751:
@3cp @bigopon I have refactored the implementation. Now the API should be more user-friendly. I have updated the example as well in the original post.

3cp:
It looks nicer!

Question, why the async components need to be resolved for navigation? au1 didn't need that.

I thought navigation only generates urls without the need of the component class definition?

packages/router-lite/src/route-context.ts
```diff
@@ -475,3 +494,90 @@ export class $RecognizedRoute {
...
+    }
+    const index = routes.length;
+    routes.push((void 0)!); // reserve the slot
+    const promise = this.promise;
+    this.promise = onResolve(promise, () => {
+      return onResolve(routeDef, $routeDef => {
+        routes[index] = NavigationRoute.create($routeDef);
+        if(this.promise === promise) {
+          this.promise = void 0;
+        }
```
bigopon, line 542: we may not need these 3 lines:
- if it's sync, then this'll be true, and the return of the `onResolve` callback (`undefined`) will be assigned to `this.promise`
- if it's async, then we probably don't want this either, since 2 promise will never be the same object. I think we want this instead:
```ts
const promise = this.promise = onResolve(this.promise, () => {
  ...
  if (promise === this.promise) {
    this.promise = void 0;
  }
})
```

Sayan751: You are right!. Don't know what I thought writing that. Thanks!

Sayan751:
> Question, why the async components need to be resolved for navigation? au1 didn't need that.
>
> I thought navigation only generates urls without the need of the component class definition?

@3cp In `router-lite` I am trying to normalize the route-configurations at the end in form of route-definitions and use that everywhere. The process of resolving route-definition is dependent on the component, tied to the route. Consider the follow example, which is currently supported, to understand why this is needed.

```ts
@route('foo') //<- Foo is reachble via '/foo'
class Foo {}

@route({
  routes: [
    { component: [Foo] /*this can also be `import('./foo')`*/ }
  ]
})
class SomeRoot {}


```

Another obvious way is to take the route configuration and translate that to nav-model. In that case, we will end up having duplicated code for some parts, which I would like to avoid as that can be a potential source of surprise. I find it cleaner this way.

Also thank you both @3cp and @bigopon for helping me to shape this API!

3cp:
I see.
But the usage `@route('some-part')` defeats the purpose of dynamic loading `component: () => fetch(...)`.

It sounds to me you need to resolve all components in order to support possible usage of `@route('some-part')`. That means the dynamic loading actually didn't delay anything at runtime.

I think this is something we need to think about. I am not convinced that it's a useful feature to let component to define its own route url part.

Sayan751:
> That means the dynamic loading actually didn't delay anything at runtime.

You are right, if such components are direct children of the root. However, if the children resides in lower levels of the hierarchy, then it is resolved lazily.

> I am not convinced that it's a useful feature to let component to define its own route url part.

Dealing with ASP.NET WebApis, it certainly makes sense to me. But on the other hand, I am inclined towards removing the deco in future. But for that I think it is better to open RFC to gather the user-voices.

3cp:
Is this reasonable to make "path" mandatory when "component" is dynamic loaded? There is definitely some TypeScript dark magic to describe the typing.

I expect user should not define the route path in component itself when planning to use dynamic loading on that component.

bigopon:
With this, we are expecting probably 99% of applications to call `.resolve()` before navigation can be available. I think it's not a reasonable default behavior. If it's annoying, we may as well just remove component information out of the nav model, and let the nav model purely be its configuration form, and that'be more usable.

Sayan751:
> With this, we are expecting probably 99% of applications to call `.resolve()` before navigation can be available. I think it's not a reasonable default behavior. If it's annoying, we may as well just remove component information out of the nav model, and let the nav model purely be its configuration form, and that'be more usable.

As long as we are supporting distributed configuration (as I have given example here: https://github.com/aurelia/aurelia/pull/1446#issuecomment-1147048221), I think we cannot build the navigation model purely out of route configuration, ignoring the component information. Note that it is not about the `@route` decorator either, as we can also have similar stuffs with the instance-hook to get the route configuration.

IMO, it might not be that of an issue, as people are likely to create CE for menu where it can be efficiently encapsulated. In a real app, the number of such menu CEs might be rather limited. I think we should keep the current behaviour and revisit if we get user feedbacks on the suggested API to be very cumbersome.

bigopon:
> I think we should keep the current behaviour and revisit if we get user feedbacks on the suggested API to be very cumbersome.

Yeah agree. A possible way to do this in the future is to have an additional right before after routing end, the router walks through the contexts and just resolve it.

Sayan751:
Just for the completeness of the discussion regarding the async nature of the nav model, I would like to point out that the route-configurations can also be async. Updated an existing test to reflect that in the PR #1452.

https://github.com/aurelia/aurelia/blob/d67504d35aea552fb57c2f8a75abebee49c8e02c/packages/__tests__/router-lite/smoke-tests.spec.ts#L1668-L1679

That is, even if we don't want to wait for the route-definition resolution, we need to at least wait for such async route-configuration, if we decide later to build the nav model from route-configuration (mentioning this for purely academic discussion).

PR #1450, opened 2022-06-06
build(dev): tooling dev script enhancement & doc

bigopon:


3cp:
Why you moved some aot's responsibility to plugin-conventions? Are we removing aot?

bigopon:
aot = full aot
the work im moving is partial AOT only, and it'll make a lot of assumptions, not sure if it's gonna go anywhere, and also I'm not sure where to house it  yet.

3cp:
Once the template is pre-compiled, assume you (are going to) pre-compile those expressions in template too, we could ship au2 prod runtime without the template compiler and the expression parser.

I mean the default dist file pointed by package.json.

That will reduce au2 size quite a bit.

bigopon:
> Once the template is pre-compiled, assume you (are going to) pre-compile those expressions in template too, we could ship au2 prod runtime without the template compiler and the expression parser.
> 
> I mean the default dist file pointed by package.json.
> 
> That will reduce au2 size quite a bit.

yes that is the plan 👍

PR #1457, opened 2022-06-07
refactor(router-lite): isNavigating flag

### LINKED ISSUE #1340: Router: Missing isNavigating property
In many valid scenarios, you want to know when the router is doing somethnig. In v1 we had a property that allowed us to check for this, we also had access to events emitted throughout the routing lifecycle via the Event Aggregator. In v2, we have no such property nor events to determine what the router is doing. The closest we have to this are lifecycle hooks. Where lifecycle hooks do not work are for loading indicators especially. In Aurelia v1 router we had numerous router states, including a isNavigating property.

Sayan751:
# Pull Request

## 📖 Description

Closes #1340

### DIFF
--- a/packages/router-lite/src/router.ts
+++ b/packages/router-lite/src/router.ts
@@ -377,6 +377,11 @@ export class Router {
   /** @internal */
   public readonly _hasTitleBuilder: boolean = false;
 
+  private _isNavigating: boolean = false;
+  public get isNavigating(): boolean {
+    return this._isNavigating;
+  }
+
   public constructor(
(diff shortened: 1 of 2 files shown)


packages/router-lite/src/router.ts
```diff
@@ -377,6 +377,11 @@ export class Router {
   /** @internal */
   public readonly _hasTitleBuilder: boolean = false;
 
+  private _isNavigating: boolean = false;
```
bigopon, line 380: Not required in this PR, but all privates should be marked internal the next time we do the cleanup.

PR #1463, opened 2022-06-12
feat(router-lite): custom root provider

Sayan751:
# Pull Request

## 📖 Description

This is a parallel to the v1's router's `options.root`. This PR makes it possible to set a custom routing root. By default, the routing root is derived from the `<base>` element's `href`. Although the default should work for most of the case, for some cases, it might not. 

For example, in an app `foo-bar` you may have the following `<base>` element.

```html
<base href="/foo-bar/"></base>
```

If your hosting service is also serving the static files (images, JS files etc.) from under the `foo-bar` directory (naive assumption here is that the `base#href` maps to a physical directory), then it works fine.

However, there are cases, where you want to change the location URL in such a  way that it changes the routing root. A typical example is to add tenant/customer name to the location URL; that is, instead of `https://example.com/foo-bar` we want to have `https://example.com/tenant-one/foo-bar` in runtime (typically done via replacing the location history). However, in that case we also want this change to be reflected in every other places such as navigation or `href` generation for anchors, etc.

This PR facilitates that by adding an option to the routing configuration so that a custom base-href-provider implementation or instance can be provided via the router options.

Additionally, it removes the `BaseHref` in favour of native `URL`

### DIFF
--- a/packages/__tests__/router-lite/smoke-tests.spec.ts
+++ b/packages/__tests__/router-lite/smoke-tests.spec.ts
@@ -1,6 +1,6 @@
-import { LogLevel, Constructable, kebabCase, ILogConfig } from '@aurelia/kernel';
+import { LogLevel, Constructable, kebabCase, ILogConfig, DI } from '@aurelia/kernel';
 import { assert, TestContext } from '@aurelia/testing';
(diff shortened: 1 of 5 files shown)


bigopon:
I'm not sure if requiring applications to go to such low level to specify base href is a good idea. Wouldn't a `root` option config be better?

Sayan751:
> I'm not sure if requiring applications to go to such low level to specify base href is a good idea. Wouldn't a `root` option config be better?

You are right. Will make the change.

Sayan751:
@bigopon It should now be ready for the next round of review.

bigopon:
Looks good, thanks @Sayan751

Issue #1347, opened 2022-01-21
Router: Inability to use router load method to navigate to a configured route path without using string


Vheissu:
Due to the fact the current core router supports configured routing but also seems to offer direct router features, the `load` method on the IRouter instance is confusing.

To navigate to a route with parameters the only way to do that is using load like this:

```
this.router.load(`/profile/rob`)
```

Or if you want the properties to be dynamic, you have to do this:

```
this.router.load(`/profile/${name}`)
```

We need the ability to route in this way, which would make it more consistent with v1 and also be less flaky.

```
this.router.load('profile', {
    params: {
        name: 'rob'
    }
});
```

Furthermore, the load method has a lot going on. It seems to be able to support a lot of different compositional features like providing a component to the method and so on. I think this method should only support navigating to paths that have been specified by the user in their route definitions.

PR #1467, opened 2022-06-17
fix(router-lite): better handling of parameters and querystring

Sayan751:
# Pull Request

## 📖 Description

This closes the issue #1347. That is, the router now supports `router.load('route/:id/as-configured', {params:{id:'foo'}})`. Any additional un-configured parameters are added to the query string.
This same behaviour is now also applicable for the `load` CA.

Finally, it removes the inclusion of parent node's parameters to the child node's parameter collection, in a hierarchical routing configuration.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -153,6 +153,31 @@
     {
       "type": "node",
       "request": "launch",
+      "name": "Launch mocha router-lite tests",
+      "program": "${workspaceRoot}/node_modules/mocha/bin/_mocha",
+      "env": {
+        "TS_NODE_PROJECT": "${workspaceRoot}/packages/__tests__/tsconfig.json"
+      },
+      "args": [
+        "--ui",
+        "bdd",
+        "--reporter",
+        "min",
+        "--colors",
+        "--recursive",
+        "--timeout",
+        "5000",
+        "--watch-extensions",
+        "js",
+        "${workspaceRoot}/packages/__tests__/dist/esm/__tests__/setup-node.js",
+        "${workspaceRoot}/packages/__tests__/dist/esm/__tests__/router-lite/**/*.spec.js"
+      ],
+      "cwd": "${workspaceRoot}",
+      "internalConsoleOptions": "openOnSessionStart"
+    },
+    {
+      "type": "node",
+      "request": "launch",
       "name": "Launch mocha scheduler tests",
       "program": "${workspaceRoot}/node_modules/mocha/bin/_mocha",
       "env": {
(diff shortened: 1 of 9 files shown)


.vscode/launch.json
```diff
@@ -150,6 +150,31 @@
...
+        "min",
+        "--colors",
+        "--recursive",
+        "--timeout",
+        "5000",
+        "--watch-extensions",
+        "js",
+        "${workspaceRoot}/packages/__tests__/dist/esm/__tests__/setup-node.js",
+        "${workspaceRoot}/packages/__tests__/dist/esm/__tests__/router-lite/**/*.spec.js"
+      ],
```
bigopon, line 174: If it's preferred this way, maybe you can do
```suggestion
      "args": [
        "-r ts-node/register",
        "--ui",
        "bdd",
        "--reporter",
        "min",
        "--colors",
        "--recursive",
        "--timeout",
        "5000",
        "--watch-extensions",
        "js",
        "${workspaceRoot}/packages/__tests__/dist/esm/__tests__/setup-node.ts",
        "${workspaceRoot}/packages/__tests__/dist/esm/__tests__/router-lite/**/*.spec.ts"
      ],
```

packages/__tests__/router-lite/resources/load.spec.ts
```diff
@@ -49,6 +49,7 @@ describe('load custom-attribute', function () {
       template: `
     <a load="route:foo; params.bind:{id: 1}; active.bind:active1" active.class="active1"></a>
     <a load="route:foo/2; active.bind:active2" active.class="active2"></a>
+    <a load="route:foo; params.bind:{id: 3, a: 2}; active.bind:active3" active.class="active3"></a>
```
bigopon, line 52: can we add an extra test, but simple to assert the new work? I think it's getting cozy in here. Just need a single assertion and that'd be it.

packages/router-lite/src/route-context.ts
```diff
@@ -455,6 +456,69 @@ export class RouteContext {
     });
   }
 
+  // this is separate method might be helpful if we need to create a public path generation utility function in future.
+  private _generatePathInternal(id: string, params?: Params | null): PathGenerationResult | null {
```
bigopon, line 460: `_` private should be marked `@internal`

packages/__tests__/router-lite/resources/load.spec.ts
Sayan751: Done

packages/router-lite/src/route-context.ts
Sayan751: Done

bigopon (approved):

Issue #1474, opened 2022-06-29
The containerless decorator has no effect after alpha v29


bsrdjan:
The `containerless()` decorator has no effect after alpha v29 and still works in v29.

Sayan751:
@bsrdjan Are you using this with the conventions? Or are you explicitly using the `@cutomElement` decorator? If it is the latter, then you can simply set the `containerless` property in the decorator option.

In case you are using both `@cutomElement` and `@containerless` decorators, ensure to apply the `@cutomElement` decorator on top of `@containerless` decorator, and not the other way around. A correct usage will be as follows:

```typescript

// @containerless() // <-- this won't work

@customElement({
  name: 'ce-foo',
  template: 'ce-foo'
})
@containerless()
export class CeFoo { }
```

This is an artefact of the order of execution of the decorators. The `@containerless()` decorator must mark (store the relevant metadata) the CE as containerless first, which is later utilized by the `@customElement`. And that's why the order is important.

bsrdjan:
I am using `containerless()` decorator with conventions, on regular page, like welcome page from au2 bootstrap app. It works in au1 and worked in au2 v29

Best regards / Viele Grüße,

Srdjan

~ sent from handheld ~
________________________________

PR #1475, opened 2022-07-03
fix(runtime-html): containerless #1474

Sayan751:
# Pull Request

### DIFF
--- a/packages/runtime-html/src/resources/custom-element.ts
+++ b/packages/runtime-html/src/resources/custom-element.ts
@@ -181,11 +181,21 @@ export function containerless(): (target: Constructable) => void;
 export function containerless(target?: Constructable): void | ((target: Constructable) => void) {
   if (target === void 0) {
     return function ($target: Constructable) {
-      annotateElementMetadata($target, 'containerless', true);
+      markContainerless($target);
     };
   }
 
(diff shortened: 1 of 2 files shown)


packages/runtime-html/src/resources/custom-element.ts
```diff
@@ -181,11 +181,21 @@ export function containerless(): (target: Constructable) => void;
...
+}
+
+/** Manipulates the `containerless` property of the custom element definition for the type, when present else it annotates the type. */
+function markContainerless(target: Constructable) {
+  const def = getOwnMetadata(ceBaseName, target) as CustomElementDefinition;
+  if(def === void 0) {
+    annotateElementMetadata(target, 'containerless', true);
+    return;
+  }
+  (def as Writable<CustomElementDefinition>).containerless = true;
```
bigopon, line 198: I think it should do both: metadata annotation and override, otherwise we may lose the information where the containerless comes from. Though maybe let's go with this for now.

bigopon (approved):
Thanks @Sayan751 
At somepoint, we need to either do this for more decos or think of a way to automatically do it for all the decos.

bsrdjan:
Thanks 👍

PR #1479, opened 2022-07-18
docs(repeat): adds repeator documentation

roemhildtg:
# Pull Request

## 📖 Description

adds repeator `repeat.for` documentation that I have copied and updated from the previous documentation site.


Vheissu (commented):
@bigopon  This all looks good. I like how it's a separate section. Should we remove the other repeat documentation that we have here? https://docs.aurelia.io/getting-to-know-aurelia/templates/built-in-template-features#working-with-collections-of-data-using-repeat.for or is there anything in that other section we can also copy over?

roemhildtg:
I can take a look at that tomorrow and try and consolidate it :)

roemhildtg:
@Vheissu I've merged the two docs, and tried to keep the best of both. Let me know if you see any issues. 

I think it makes sense to have repeat as a separate page, there's a lot of content there.

bigopon (approved):
Thanks @roemhildtg

Issue #1370, opened 2022-03-16
Error when using multiple path for redirect with router-lite


bigopon:
# 🐛 Bug Report

At the moment, when doing

```ts
    {
      path: ['', 'login'],
      redirectTo: 'auth/login'
    },
```

it'll throw
![image](https://user-images.githubusercontent.com/9994529/158518659-52cc8e9e-aa4c-4ed2-bf41-a4a17d573dea.png)

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Aurelia          | alpha-26 |

PR #1482, opened 2022-07-23
fix(router-lite): #1370

Sayan751:
# Pull Request


packages/router-lite/src/route-tree.ts
```diff
@@ -505,7 +504,7 @@ export function createAndAppendNodes(
           new Endpoint(
             // TODO(sayan): probably need to do parameter matching and select the "most-matched" path instead of picking the first
             new ConfigurableRoute(rd.path[0], rd.caseSensitive, rd),
-            toArray(Object.values(params))
```
bigopon, line 508: why are we having this change?

Sayan751: This was incorrect. We need here only the parameter names. BTW in my next PR, I am going to change this part quite a lot, in order to address the TODO, as seen here. Just a heads-up. Till that change lands, this change is needed. Also, as per my current findings, we are not using the `endpoint` of a recognized route, hence this issue was undetected. Just for information and context, my next change caches the endpoint for a path, and that saves us creating the endpoint manually here.

PR #1483, opened 2022-07-26
fix(hmr): add some more ignores for strict mode TS

brandonseydel:
# Pull Request


bigopon:
Thanks @brandonseydel , I'll be adding some e2e setup for the hmr part so we can test all the features easier.

PR #1484, opened 2022-07-28
build(tests): add e2e test for hmr

bigopon:
Add tests for hmr so it'll be easier to develop & verify.


bigopon:
@brandonseydel  Merging this without integrating it in the pipeline. Somehow local runs fine but circleCI doesn't. You can test new HMR features with this in the meantime.

PR #1493, opened 2022-08-07
fix(plugin-conventions): fix wrong options reference

3cp:


Vheissu (approved):

Vheissu:
@bigopon Looks like a good simple fix we should merge.

PR #1494, opened 2022-08-07
docs(attributes): update docs to include explicit bindable call

tomtomau:
Per Discord discussions:
https://discord.com/channels/448698263508615178/448699158052864001/1005682444839682100

# Pull Request

## 📖 Description

```typescript
import { customAttribute, inject, INode } from 'aurelia';

@inject(Element)
@customAttribute('my-cool-attr')
export class MyCoolAttribute {
  private value: string;

  constructor(@INode private element: Element) {
    console.log('Attaching to an element'); // ✅
  }

  public bound() {
    console.log('Initial value set', this.value) // ✅;
  }

  public valueChanged(): void {
    console.log('Value just changed', this.value) // ❌;
  }
}
```

I had this simple example of a custom attribute, but "valueChanged" was not being called when the binding was updated.

@Sayan751 suggested trying `@bindable()` which fixed the issue and so I was encouraged to update docs. When I've come through to update the docs, I'm now wondering whether this is actually a bug - and the code above was expected to be correct?

### 🎫 Issues

n/a

## 👩‍💻 Reviewer Notes

* refer to above example

## 📑 Test Plan

n/a

## ⏭ Next Steps

* work out whether this is a bug (close this PR), or the expected behaviour (merge this PR)

### DIFF
--- a/docs/user-docs/getting-to-know-aurelia/custom-attributes.md
+++ b/docs/user-docs/getting-to-know-aurelia/custom-attributes.md
@@ -102,16 +102,16 @@ In some instances, you want a custom attribute that only has one bindable proper
   }
 ```
 
-The `value` property is automatically populated if a value is supplied to a custom attribute, without the need to explicitly define the value property as a bindable.
+The `value` property is automatically populated if a value is supplied to a custom attribute, however requires you to explicitly define the value property as a bindable.
 
 When the value is changed, we can access it like this:
 
 ```
-  import { customAttribute, INode } from 'aurelia-framework';
+  import { bindable, customAttribute, INode } from 'aurelia-framework';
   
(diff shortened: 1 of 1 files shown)


bigopon:
We normally go out of our way to make things that are of common expectation work without requiring extra work from apps. And its probably expected to stay this way. I think if it doesnt work then its a bug, but it seems weird that this doesnt work. Ill have a check.

Vheissu:
@bigopon Going to merge this one as others might consult the docs and attempt to omit the `@bindable`, which will result in the `value` change handler not firing. If we decide that value shouldn't require a bindable, we can further modify the documentation to mention this again. But this PR is great for now. Thanks @tomtomau

PR #1497, opened 2022-08-15
Check full and mangled method names in HMR

netcitylife:
# Pull Request

## 📖 Description

When using dev app build one must configure aliases to include dev Aurelia build for HMR to work. This add checks for original and mangled private method name to avoid this requirement.


brandonseydel:
Nice looks good thanks!

Issue #1492, opened 2022-08-05
Aurelia Router - Infinite loop with deep invalid path


aegenet:
# 🐛 Bug Report

With `@aurelia/router`, if we specify a deep path in a configured route ({ path: '/page/x' }), and a user goes to /page/y, the missing page appears, but with an infinite loop.

## 🤔 Expected Behavior

Just the missing page.

## 😯 Current Behavior

Missing page and infinite loop.

## 🔦 Context

I am using `@aurelia/router` and configured routes with sub paths.

## 💻 Code Sample

Here: [Project](https://gist.dumber.app/?gist=34ca8e1c17beaeda594e4a673bd78554)

1. Go to #/page/3
2. Then, #/page/8

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Aurelia          | 2.0.0-alpha.37
| Language         | TypeScript
| Browser          | Chrome 104.0.5112.81, Firefox 103.0.1
| Bundler          | Dumber / Webpack
| Operating System | Windows 11
| NPM/Node/Yarn    |

aegenet:
I tried the .39 version as it was supposed to fix the above issues.
Unfortunately, the router on the alpha.39 version does not seem to work (on all our projects).

Before (alpha.38): https://gist.dumber.app/?gist=34ca8e1c17beaeda594e4a673bd78554
After (alpha.39): https://gist.dumber.app/?gist=93ba7c8415e501935dd6411f10ef23b8

I don't know if it's related to #1532 (router lite).

jwx:
The original issue is fixed in alpha.39 and a PR is merged (https://github.com/aurelia/aurelia/pull/1534) for the new issue mentioned in the comment above. As soon as alpha.40 is out both of these should be resolved.

PR #1507, opened 2022-08-19
feat(router): configurable viewport fallback behavior

jwx:
# Pull Request

## 📖 Description

This PR improves the fallback feature for viewports by adding a `fallback-action` to control the behavior with multiple levels of instructions:
- `process-children` - replace the unloadable component with the fallback component and continue processing the child instructions. The unloadable component name is passed as first argument to the fallback component.
- `abort` - don't process the child instructions of the unloadable instruction. The unparsed string, if any, of the unloadable component and everything after it is passed as first argument to the fallback component.

In addition, `fallback` and `fallbackAction` (default: `process-children`) properties are added to router configuration to control behavior of all viewports that doesn't explicitly set their own attributes.

These new features will make it easier to avoid unwanted behavior such as https://github.com/aurelia/aurelia/issues/1492.

### 🎫 Issues

Related to https://github.com/aurelia/aurelia/issues/307, https://github.com/aurelia/aurelia/issues/369, https://github.com/aurelia/aurelia/issues/367 and https://github.com/jwx/aurelia-direct-router/issues/35. 

## 👩‍💻 Reviewer Notes

## 📑 Test Plan

- Make tests for the different fallback combinations

## ⏭ Next Steps

- Resolve remaining issues/requests from Aurelia Direct Router
- Ensure v1 feature parity
- Continue with comments
- Move relevant parts to `Metadata`
- Add tests for `au-viewport-scope`
- Review the extension points


packages/router/src/endpoints/viewport-content.ts
```diff
@@ -209,10 +213,18 @@ export class ViewportContent extends EndpointContent {
...
         if ((fallback ?? '') !== '') {
-          // ...set the failed component as the first parameter (0)...
-          this.instruction.parameters.set([this.instruction.component.name]);
-          // ...fallback is component...
+          if (fallbackAction === 'process-children') {
+            // ...set the failed component as the first parameter (0)...
+            this.instruction.parameters.set([this.instruction.component.name]);
+          } else { // 'abort'
+            // ...set the unparsed string of the failed component as the first parameter (0)...
+            this.instruction.parameters.set([this.instruction.unparsed ?? this.instruction.component.name]);
```
bigopon, line 221: doesn't it look weird passing the component name in the same way with the parameters? If something goes wrong and there' some others config because someone was trying to add some new stuff to their routing structure, what the data/error actually is will be unclear.

packages/router/src/endpoints/viewport-options.ts
```diff
@@ -37,6 +38,14 @@ export class ViewportOptions implements IEndpointOptions {
      */
     public fallback: string = '',
 
+    /**
+     * Whether the fallback action is to load the fallback component in
+     * place of the unloadable component and continue with any child
+     * instructions or if the fallback is to be called and the processing
+     * of the children to be aborted.
+     */
+    public fallbackAction: FallbackAction | '' = '',
```
bigopon, line 47: why are we giving it an empty string as default? Shouldnt this be `undefined` with `?:` ?

bigopon (commented):
Looking good, just some nit picks and questions for round 1. I'll have a look at the actual functionality soon.

packages/router/src/endpoints/viewport-options.ts
jwx: This is comes from the `fallback-action` attribute of the `au-viewport` custom element and all the rest of them are turned into their "target types" (string in this case) so I did it like that to be consistent. We could change all of them, though (but then preferably in a new PR).

packages/router/src/endpoints/viewport-content.ts
jwx: Route configuration and route matching parameters is preserved in the routing instruction so there's no information lost there. My intention was however to unshift the component/unparsed to the parameters, but I can see now that I forgot to do that so thanks for pointing that out. With the unshift, all data should be preserved even with direct routing.

bigopon: Thanks for the clarification. Can we add at least ab w2e test for this feature?

jwx: Sure, I'll look into adding something. (But I'm not sure I'll find a combination that's not already covered by the 74 tests that are there now.)

bigopon:
The default behavior process children seems like a footgun more than it otherwise intended. Routing configuration can get really brittle while being changed, we should making the errors loud and clear and as early as possible, rather than try figuring out what's maybe ok to proceed with.
If we are to have this feature, `abort` sounds like a better choice.

jwx:
> we should be making the errors loud and clear and as early as possible

To be clear, someone would configure a `fallback` to prevent the router from making loud errors and themselves control the behavior when there's no matching component to load. Regardless of whether behavior is `process-children` or `abort`, the router will load the fallback component instead of the failing one and then consider the navigation as completed. I might need to change the name of the `abort` option to make it less confusing.

But even with the right meaning, I'm not against changing the default to `abort`. I simply picked `process-children` because it was the only behavior before this option was added (so this addition didn't cause a change in behavior.

bigopon:
Imo, abort could be a good "middle ground" for dev to have controllable ramification when changing their routing configuration during the evolution of an app. Though maybe we can keep process children for now, as you mentioned it's been the only behavior for sometimes. Probably first step is to have better logging around this fallback to communicate the potential issue to devs.

bigopon:
Nice work, thanks @jwx 🚀

Issue #1119, opened 2021-02-19
repeat.for throws error when wrapped with if.bind


bravius:
# 🐛 Bug Report

When a repeat.for element is wrapped with an element containing if.bind, it throws "Uncaught Error".

```
<div if.bind="items.length>0">
  <div repeat.for="item of items">
    ...
  </div>
</div>

```
## 🤔 Expected Behavior

Shouldn't throw error... lol.

## 😯 Current Behavior

Throws the following error:

Uncaught Error: viewsLen=2, mapLen=1
    at Repeat.createAndActivateAndSortViewsByKey (VM157 entry-bundle.js:23859)
    at VM157 entry-bundle.js:23681
    at onResolve (VM157 entry-bundle.js:3175)
    at Repeat.handleCollectionChange (VM157 entry-bundle.js:23679)
    at SubscriberRecord.notifyCollection (VM157 entry-bundle.js:37853)
    at ArrayObserver.notify (VM157 entry-bundle.js:34993)
    at Array.splice (VM157 entry-bundle.js:34873)
    at App.remove (VM157 entry-bundle.js:195)
    at CallScopeExpression.evaluate (VM157 entry-bundle.js:31010)
    at Listener.callSource (VM157 entry-bundle.js:15425)

## 🔦 Context

Using latest aurelia@dev (0.10.0-dev.202102161145)

## 💻 Code Sample

See gist to replicate the bug.  

Click "add" to add some list items, then "remove" on each one to empty the list. 

Repeating the add/remove procedure will throw "Uncaught Error: viewsLen=2, mapLen=1"

https://gist.dumber.app/?gist=a29951a42030975f2fdcce78b42de6b1

Note: if the wrapping element uses show.bind instead of if.bind, no error is thrown

bigopon:
@bravius thank you for creating this issue! This is because `[if]` running before `[repeat]` causing `[repeat]` to skip removing existing views when the value of `[if]` evaluates to false. A fix should be in soon.

bigopon:
Closed by #1511

PR #1511, opened 2022-08-22
fix(repeater): re attaching when using the same array with if

bigopon:
# Pull Request

## 📖 Description

Currently, the repeater does not unsubscribe the collection when getting detached, resulting in double work when:
1. it becomes attached again and
2. the collection getting updated at the same time

The template for the 2 above conditions to happen at the same time looks like this:
```html
<div if.bind="items.length">
  <p repeat.for="i of items">...
```
Reproduction:
1. start with an array of 1 item
2. remove the only item in the array, this will trigger detaching lifecycle on the repeat without unsubscribing to the array `items`
3. add a new item into the array `items` -> 💥

This PR fixes this.

Thanks to @bravius for reporting this issue in #1119

### DIFF
--- a/packages/__tests__/3-runtime-html/repeater-if-else.spec.ts
+++ b/packages/__tests__/3-runtime-html/repeater-if-else.spec.ts
@@ -9,6 +9,7 @@ import {
   TestConfiguration,
   trimFull,
   assert,
+  createFixture,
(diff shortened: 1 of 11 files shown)


bigopon:
This one reducing coverage by ~10% 😵 there's something weird with coverage reporting.

PR #1514, opened 2022-08-23
fix(router-lite): navigation to routes configured in ancestor nodes

Sayan751:
# Pull Request

### 🎫 Issues

This fixes the issue reported here: https://discourse.aurelia.io/t/the-aurelia-2-book/4874
This support navigating to the routes defined in the ancestor view-models. The support for `../`-prefix was partially present, but was not tested. In this the tests are added and the bug is also fixed. That is, the following use-case is now supported:

```ts
@customElement({ 
  name: 'pro-duct', 
  template: `product \${id} <a load="../products"></a>`  //<-- navigation to the route defined in parent context
})
class Product {
  private id: unknown;
  public canLoad(params: Params, _next: RouteNode, _current: RouteNode): boolean {
    this.id = params.id;
    return true;
  }
}

@customElement({ name: 'pro-ducts', template: `<a load="../product/1"></a><a load="../product/2"></a> products` })
class Products { }

@route({
  routes: [
    { id: 'products', path: ['', 'products'], component: Products },
    { id: 'product', path: 'product/:id', component: Product },
  ]
})
@customElement({ name: 'ro-ot', template: '<au-viewport></au-viewport>' })
class Root { }
```

Moreover, it adds support for binding a routing context explicitly for more fine-grained control.

The support is added for both `load` and `href` attributes.

## ⏭ Next Steps

---

cc: @Vheissu just in case you are in the process of maintaining the docs for router-lite as well. No stress, if not; I can later later fill up the gaps.

### DIFF
--- a/packages/__tests__/router-lite/_shared/create-fixture.ts
+++ b/packages/__tests__/router-lite/_shared/create-fixture.ts
@@ -1,4 +1,4 @@
-import { Constructable, LogLevel, Registration, ILogConfig, DI, LoggerConfiguration, ConsoleSink, IContainer, Resolved, IPlatform } from '@aurelia/kernel';
+import { Constructable, LogLevel, Registration, ILogConfig, DI, LoggerConfiguration, ConsoleSink, IContainer, Resolved, IPlatform, Class } from '@aurelia/kernel';
 import { Aurelia } from '@aurelia/runtime-html';
 import { IRouterOptions, RouterConfiguration, IRouter } from '@aurelia/router-lite';
 import { TestContext } from '@aurelia/testing';
@@ -88,4 +88,25 @@ export async function createFixture<T extends Constructable>(
       await au.stop(true);
     },
   };
+}
+
+/**
+ * Simpler fixture creation.
+ */
+export async function start<TAppRoot>(appRoot: Class<TAppRoot>, ...registrations: any[]) {
+  const ctx = TestContext.create();
+  const { container } = ctx;
+
+  container.register(
+    TestRouterConfiguration.for(LogLevel.warn),
+    RouterConfiguration,
+    ...registrations,
+  );
+
+  const au = new Aurelia(container);
+  const host = ctx.createElement('div');
+
+  await au.app({ component: appRoot, host }).start();
+  const rootVm = au.root.controller.viewModel as TAppRoot;
+  return { host, au, container, rootVm };
 }
--- a/packages/__tests__/router-lite/resources/href.spec.ts
+++ b/packages/__tests__/router-lite/resources/href.spec.ts
@@ -115,4 +115,110 @@ describe('href custom-attribute', function () {
 
     await au.stop();
   });
+
+  it('allow navigating to route defined in grand-parent context using ../../ prefix', async function () {
+    @customElement({ name: 'l-21', template: `l21 \${id} <a href="../../l12"></a>` })
+    class L21 { }
+    @customElement({ name: 'l-22', template: `l22 \${id} <a href="../../l11"></a>` })
+    class L22 { }
+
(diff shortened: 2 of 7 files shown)


bigopon:
> adds support for binding a routing context explicitly

I'm a bit concerned that this may destroys the what you see is what you get of the routing configuration for our routing. At the moment, with `../` it's already feeling somewhat unclear sometimes. If we enable this, it may get worse.

What's the main high level motivation for this feature?

Sayan751:
@bigopon As you already know, the routes are configured at the `RouteContext` level and the route contexts are hierarchical in nature. Given this, when we issue a `load` request to the `Router`, it always starts from the root `RouteContext` unless a specific route-context is specified (a side note: the route-context here is polymorphic, it can be a real `RouteContext` instance or a routed view model instance, or a host element etc.) and then the navigation occurs relative to the given route-context. 

Now, the problem with `href` and `load` was that there was an injected instance of `RouteContext` with no way to override that. This absence of the most basic option was annoying. Hence, I added this support. Also, as that might be boilerplate-y, I have fixed the `../`-prefix-support. In my last commit, I have also added the support to bind a `null` context to perform navigation relative to the root `RouteContext`. Note that without these fixes, ATM it is not possible to navigate to a route, defined in parent, from a child view (template; you could always use the `Router.load` for the same).

You may also ask, "what about using structured instruction with route-id and parameters etc.? Can't we simply traverse upward till we find the route?". Sure we can. But whether that might a solution for `load`, but not for `href`, which deals with raw paths. Which is why I have not tried to implement that. Also, it is easier if the routing-context is known before we start resolving the recognized routes etc.; it gets difficult to change in downstream. We can rework that, but that bit time-consuming and might want to postpone that for now. 

I am not sure whether I have answered your question or created new confusions, but feel free to let me know in either way! 😄

bigopon:
My main issue is around this template
```html
<a href="value:product/1; context.bind:rtCtx.parent">
```
Prior to the `context` exposure, it'll be simple with whatever the value of `href`, now it'll have to be read based on the context. This will make the href different with standard href and the following simple pattern no longer works:
```html
<a href="https://bla.bla.com/bla">
```
If we proceed with this configurable context, at least it'll be only `load`

Sayan751:
> My main issue is around this template
> 
> ```
> <a href="value:product/1; context.bind:rtCtx.parent">
> ```
> 
> Prior to the `context` exposure, it'll be simple with whatever the value of `href`, now it'll have to be read based on the context. This will make the href different with standard href and the following simple pattern no longer works:
> 
> ```
> <a href="https://bla.bla.com/bla">
> ```
> 
> If we proceed with this configurable context, at least it'll be only `load`

@bigopon Awesome find; an artefact `href` of not being a non-multiBinding any longer. I would like to say that it is a shortcoming of the template compiler. However, I also agree that it might be a bit of much effort to change the template compiler to handle such edge-case (although I do wonder how much of an edge case that is). For now, I can remove the bindable context from `href`, considering the `../` prefix support is already present.

Sayan751:
@bigopon I have updated the PR. It should now be ready to merge.

bigopon:
> ... a shortcoming of the template compiler

When there's ambiguity in an attribute, the template compiler may have choices:
- throw
- try to guess and continue
- warn & ignore

For throwing is the easiest way for both compiler and app developers since it's clear where an error comes from.
For trying to guess and continue, it'll be a bit messy because of scenarios like this, where you are not really sure what to guess. Unless we can have a clear definition of what to do and what to expect, this behavior should be avoided.
For ignoring the attribute altogether, it's not really useful so maybe there's no need for it.

bigopon:
Thanks @Sayan751

PR #1515, opened 2022-08-24
fix(hmr): add ts-ignore to fix hmr in strict TS

Lakerfield:
# Pull Request

## 📖 Description

Fix for "TS7006: Parameter 'data' implicitly has an 'any' type. " on every compontent.ts file when using HMR with typescript in strict mode

## ⏭ Next Steps


bigopon:
thanks @Lakerfield

PR #1522, opened 2022-08-29
refactor(router-lite): lifecycle hook invocation

Sayan751:
# Pull Request

## 📖 Description

This PR changes the order of the `canLoad` and `canUnload` lifecycle hook invocation for router-lite. Previously, all the `canLoad` hooks were invoked in parallel. It was also the same for the `canUnload` hooks. This can sometimes be confusing when we are considering pre-emption; that is, the hooks are either returning `false` or a new navigation instruction. In the case of the pre-emption, the pre-empting hook that finishes its execution earliest, won. That was inherently indeterministic.

This PR changes that behaviour. It ensures that all the `canLoad` hooks are invoked serially. If any hook pre-empts in the process, it always causes skipping execution of the succeeding hooks. The same is the case for the `canUnload` hooks as well.

## 👩‍💻 Reviewer Notes

Take a look at the tests to understand the changes better.


bigopon:
Looks good, nice work 🚀 
If we have a hierarchy like this:
```
root
    + route 1
    |    + child 1.1
    |    + child 1.2
    |
    + route 2
        + child 2.1
```
When root loads, will `canLoad` of `route 1` and `route 2` be loaded in parallel or sequence? Similar question for `route 1` and `child 1.1` + `child 1.2`.

Sayan751:
@bigopon those will be parallel

bigopon:
Thanks I've reviewed the code again 👍

PR #1523, opened 2022-08-31
optional chaining & nullish coalescing

fkleuver:
Supersedes & closes #1202

### DIFF
--- a/packages/runtime/src/binding/expression-parser.ts
+++ b/packages/runtime/src/binding/expression-parser.ts
@@ -665,13 +665,13 @@ TPrec extends Precedence.Unary ? IsUnary :
           break;
         case Token.TemplateTail:
           if (state._optional) {
-            throw invalidTaggedTemplateOnOptionalChain(state)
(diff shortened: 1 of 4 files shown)


packages/runtime/src/binding/expression-parser.ts
```diff
@@ -501,48 +505,43 @@ TPrec extends Precedence.Unary ? IsUnary :
         if (expressionType & ExpressionType.IsIterator) {
           result = new BindingIdentifier(state._tokenValue as string);
         } else {
-          result = new AccessScopeExpression(state._tokenValue as string, access & Access.Ancestor);
-          access = Access.Scope;
+          result = new AccessScopeExpression(state._tokenValue as string, ancestor);
         }
-        state._assignable = true;
+        state._assignable = !state._optional;
```
bigopon, line 510: we probably can remove this `_assignable` as well

fkleuver: Remove it from where? The state?

PR #1536, opened 2022-09-03
fix(router-lite): specific module import

### LINKED ISSUE #1530: Router lite does not work with lazily imported classes
# 🐛 Bug Report

Currently, router lite recognizes this pattern
```ts
@route({
  routes: [
    { path: '', component: () => import('./pages/home')},
  ],
})
```
but it doesn't work with
```ts
@route({
  routes: [
    { path: '', component: () => import('./pages/home').then((m) => m.Home) },
  ],
})
```
because it assumes the component will be a whole module, instead of a class in that module.

Repro link
https://stackblitz.com/edit/au2-conventions-j5xjkh?file=src%2Fmy-app.ts

## 🤔 Expected Behavior

It should work with any kind of lazily resolved classes:
```ts
Promise.resolve(class {})
import('...').then(m => m.Class)
```

Sayan751:
# Pull Request

### 🎫 Issues

Closes #1530

### DIFF
--- a/packages/router-lite/src/route-context.ts
+++ b/packages/router-lite/src/route-context.ts
@@ -1,4 +1,4 @@
-import { DI, IContainer, ILogger, IModule, IModuleLoader, InstanceProvider, noop, onResolve, Registration, ResourceDefinition } from '@aurelia/kernel';
+import { Constructable, DI, IContainer, ILogger, IModule, IModuleLoader, InstanceProvider, noop, onResolve, Protocol, Registration, ResourceDefinition } from '@aurelia/kernel';
 import { Endpoint, RecognizedRoute, RouteRecognizer } from '@aurelia/route-recognizer';
 import { CustomElement, CustomElementDefinition, IAppRoot, IController, ICustomElementController, IPlatform, isCustomElementController, isCustomElementViewModel, PartialCustomElementDefinition } from '@aurelia/runtime-html';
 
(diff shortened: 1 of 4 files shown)


Sayan751:
@bigopon I am merging this. In case you see any issue, please let me know.

packages/router-lite/src/route-context.ts
```diff
@@ -439,6 +439,13 @@ export class RouteContext {
 
   public resolveLazy(promise: Promise<IModule>): Promise<CustomElementDefinition> | CustomElementDefinition {
     return this.moduleLoader.load(promise, m => {
```
bigopon, line 441: it feels weird we are sending the promise towards module loader regardless. A bit wasteful, inappropriate and side-effect-ful. Shouldn't we do the check here before analyzing the module?

Sayan751: I wouldn't like to create a branching here, to be honest. The module loader is at the end a promise resolver in a way. Also, technically, you can export a module from another module. In terms of promise, that might look something like this:

```ts
Promise.resolve({exportedModule: Promise.resolve({CE})}).then(x => x.exportedModule)
```

In this case, it is expected the `CE` is used as the routed component. I think creating a check here would eventually cause duplicated effort. IMO, it is best to avoid that in favour of supporting broader use cases and non-duplicated code.

bigopon: I see, thanks. I think a small correction for the example would be
```ts
Promise.resolve({exportedModule: {CE} }).then(x => x.exportedModule)
```
since that would be close to how the export statements are transformed into.

PR #1539, opened 2022-09-03
Expression parser cleanup / reduce bundle size

fkleuver:


packages/runtime/src/binding/expression-parser.ts
```diff
@@ -464,153 +430,153 @@ TPrec extends Precedence.Unary ? IsUnary :
...
-          } else if ((state._currentToken as Token) === Token.OpenBracket) {
-            result = parseKeyedExpression(state, result, true);
+          $optional = true;
+          $assignable = false;
+          nextToken();
+          ensureOptionalSuffix();
+
+          if ($currentToken & Token.IdentifierName) {
+            result = new AccessScopeExpression($tokenValue as string, result.ancestor);
+            nextToken();
```
bigopon, line 553: is codecov correct? How is this line not covered?

packages/runtime/src/binding/expression-parser.ts
```diff
@@ -641,39 +607,39 @@ TPrec extends Precedence.Unary ? IsUnary :
...
           break;
         case Token.OpenBracket:
-          result = parseKeyedExpression(state, result as IsLeftHandSide, false);
+          result = parseKeyedExpression(result as IsLeftHandSide, false);
           break;
         case Token.TemplateTail:
-          if (state._optional) {
-            throw invalidTaggedTemplateOnOptionalChain(state);
+          if ($optional) {
+            throw invalidTaggedTemplateOnOptionalChain();
```
bigopon, line 634: Should we add a simple test that throws here?

packages/runtime/src/binding/expression-parser.ts
```diff
@@ -641,39 +607,39 @@ TPrec extends Precedence.Unary ? IsUnary :
...
           }
-          result = createTemplateTail(state, result as IsLeftHandSide);
+          result = createTemplateTail(result as IsLeftHandSide);
           break;
         case Token.TemplateContinuation:
-          if (state._optional) {
-            throw invalidTaggedTemplateOnOptionalChain(state);
+          if ($optional) {
+            throw invalidTaggedTemplateOnOptionalChain();
```
bigopon, line 640: Same for this, should we add a simple test that throws here?

packages/runtime/src/binding/expression-parser.ts
```diff
@@ -1448,18 +1416,19 @@ const TokenValues = [
...
-KeywordLookup.false = Token.FalseKeyword;
-KeywordLookup.undefined = Token.UndefinedKeyword;
-KeywordLookup.$this = Token.ThisScope;
-KeywordLookup.$parent = Token.ParentScope;
-KeywordLookup.in = Token.InKeyword;
-KeywordLookup.instanceof = Token.InstanceOfKeyword;
-KeywordLookup.typeof = Token.TypeofKeyword;
-KeywordLookup.void = Token.VoidKeyword;
-KeywordLookup.of = Token.OfKeyword;
+const KeywordLookup: Record<string, Token> = Object.assign(Object.create(null), {
```
bigopon, line 1419: we have a local fn `createRecord` for this, so that we can avoid multiple `Object.etc`

bigopon (approved):

bigopon:
thanks @fkleuver, nice work 🚀

PR #1541, opened 2022-09-04
implement arrow function syntax

### LINKED ISSUE #1503: Allow arrow functions inside HTML
# 🙋 Feature Request

The ability to allow arrow functions to reside inside views.

## 🤔 Expected Behavior

It works when using something like a filter arrow function on array

## 😯 Current Behavior

Blows up with parsing error

## 💁 Possible Solution

Allow the parser to correctly extract and run the JS

## 🔦 Context

When using an HTML only view it is impossible for this to work without brining in a backing store.  Array filter expression is one example which is widely used.

## 💻 Examples

```html
<div repeat.for="users of users.filter(y=>y.online)" >
```

fkleuver:
## Context:

One of the commonly requested features in Aurelia is the ability to use arrow function in templates. It can be handy in many ways: filtering items for a repeat, passing a callback to a child component, handling an event, etc... Additionally, it is a good way to avoid having to create a view model to accommodate some code that can only be expressed via a function.

With the arrow function, the following code
```html
<my-input change.call="updateValue($event)">
<my-button click.trigger="handleClick">
```
 can be rewritten with simpler, closer to JavaScript syntax:
```html
<my-input change.bind="v => updateValue(v)">
<my-button click.trigger="e => handleClick(e)">
```

It also enables us to have an easier time dealing with ad-hoc collection filtering/sorting:
```html
<div repeat.for="i of list | specialFilterBy">
```
can now be rewritten as
```html
<div repeat.for="i of list.filter(item => isGood(item))">
```

Recently, it's been requested again at #1503 , we decided to take this opportunity to finally add the support, albeit not full range of syntax, for the arrow function/lambda.

The subset of the Arrow Function syntax in Aurelia template support simple and rest parameters list, and simple concise function body expression. Which means if the arrow function body starts with a `{`, it'll not be supported.

### Valid usages:
```ts
() => 42
```
```ts
(a) => a
```
```ts
(...a) => a[0]
```
```ts
a => a
```
```ts
(a, b) => `${a}${b}`
```
```ts
(a, ...rest) => `${a}${rest.join('')}`
```
```ts
a => b => a + b
```

### Invalid usages:
```ts
() => {} // no function body
```
```ts
(a = 42) => a // no default parameters
```
```ts
([a]) => a // no destructuring parameters
```
```ts
({a}) => a // no destructuring parameters
```

### Practical examples:
```html
<div repeat.for="item of items.filter(x => x.selected).sort((a, b) => a.pos - b.pos)">
  ${item.name}
</div>
```
Observation wise, Aurelia knows to only observe `selected` property of every item in `items`, as well as `pos` property of every **selected item**. This means changing the value of `selected` property of any item will result in the re-evaluation of the above expression. Changing the value of `pos` property of any **selected item** will also trigger the re-evaluation.

```html
<p>Total: ${items.reduce((sum, item) => sum + item.count, 0)}</p>
```
Similar to the above practical example, the `count` property of every item will be observed and changing the value of any `count` will result in re-evaluation.

## Next step

Since now we have a better way to express a function in the template, the following syntaxes are going to be deprecated:
- `.call`

Close #1503

### DIFF
--- a/docs/user-docs/reference/examples/binding-and-templating/README.md
+++ b/docs/user-docs/reference/examples/binding-and-templating/README.md
@@ -171,6 +171,25 @@ Exponentiation (`a**b`) and bitwise operators are not supported.
 * `foo = bar`
 * `foo = bar = baz`
 
+### Lambda Expressions
+
+Lambda expressions can be used to call methods like `Array#filter`, `Array#sort`, but can also be passed in as an argument to a method on your ViewModel or even invoked as an IIFE.
+
+* `() => 42`
+* `x => x.name` (e.g. `${items.map(x => x.name).join(', ')}`)
+* `(...args) => args.join(', ')`
+* `(a => b => c => a + b + c)(1)(2)(3)` (renders '6')
+* `items.reduce((sum, x) => sum + x, 0)`
+
+{% hint style="warning" %}
+**Warning**
+
+- Wrapping the expression body in braces (`(a, b) => { ... }`) is not supported.
+- Default parameters (`(a = 42) => ...`) are not supported.
+- Destructuring parameters (`({a}) => ...` or `([a]) => ...`) is not supported.
+
+{% endhint %}
+
 ### Member and Call Expressions
 
 Member expressions with special meaning in Aurelia:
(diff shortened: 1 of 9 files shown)


bigopon:
Thanks @fkleuver , nice work 🚀

Issue #1509, opened 2022-08-20
[RFC] Rename routing lifecycle hooks


jwx:
# 💬 RFC

## Context

Currently, the component lifecycle hooks have a naming convention that specifies whether the lifecycle step is happen*ing* (`attaching`) or has happen*ed* (`attached`). The router component hooks, however, does not follow this naming convention and are instead named `canLoad`, `load`, `canUnload` and `unload`. 

The hooks `canLoad` and `canUnload` are for a functionality that doesn't exist in the component lifecycle hooks so they can decide the naming convention themselves. (And it makes sense. To the degree that if the functionality were ever to appear in lifecycle hooks the naming convention would work well for that as well (`canAttach`)). 

The hooks `load` and `unload` do have clear equivalents among lifecycle hooks but don't follow the naming convention of those. By the names themselves, it's also unclear whether the step is happening or has happened when the hooks are called. 

## Proposal

Rename the `load` and `unload` hooks to `loading` and `unloading` respectively. That way they'd have the expected names of component hooks and they'd also clearly indicate that they are at the start of the load/unload step.

## Benefits

Consistent naming among the component hooks will make them easier to understand and remember. It will also open up for the possibility to add routing hooks at the end of the load/unload process; `loaded` and `unloaded`. (There are use cases for both `loading/unloading` and `loaded/unloaded` component hooks when using multiple viewports.)

## Potential issues

It would be a breaking change for existing apps. But the fix is hopefully quite quick to do. And these kind of changes are in a way not to be unexpected as long as we're in alpha.

## Alternative solution

We could support both names, `load/loading` and `unload/unloading`.

Sayan751:
>  It will also open up for the possibility to add routing hooks at the end of the load/unload process; loaded and unloaded. (There are use cases for both loading/unloading and loaded/unloaded component hooks.)

Can you please give some example of such use-cases? I am intrigued as such hooks were not there in V1, and sincerely I never faced a need to have that. To me, the `attached` is practically that `loaded` hook.

As a side remark, we should be cautious about introducing too many hooks or even renaming the existing ones, as that will bring (probably needless) breaking changes.

Vheissu:
I think the way the Lifecycle methods are currently named makes more sense than those presented in this RFC. I understand the intent, but I like how the router isn't the same as core methods. Hooks like the current canLoad just make sense.

jwx:
> Hooks like the current canLoad just make sense.

[...]

PR #1546, opened 2022-09-08
refactor(router): rename load, unload to loading, unloading

jwx:
# Pull Request

## 📖 Description

This PR changes the names of the routing hooks `load` and `unload` to `loading` and `unloading` respectively in order to follow the naming convention of the other lifecycle hooks.

**BREAKING CHANGES**: After a short grace period, hooks named `load` and `unload` will no longer work. Instead, their new names `loading` and `unloading` will have to be used.

### 🎫 Issues

Resolves https://github.com/aurelia/aurelia/issues/1509. 

## 👩‍💻 Reviewer Notes

There are a couple of instances in `/examples/validation-with-router` and `/test/benchmarking-apps` that I have not updated since I'm unfamiliar with those apps. I have also not updated `router-lite` for the same reason.

## 📑 Test Plan

## ⏭ Next Steps

- Update the remaining instances mentioned above

### DIFF
--- a/packages/router/src/endpoints/viewport-content.ts
+++ b/packages/router/src/endpoints/viewport-content.ts
@@ -398,21 +398,36 @@ export class ViewportContent extends EndpointContent {
         const parameters = this.instruction.typeParameters(this.router);
         const merged = { ...this.navigation.parameters, ...parentParameters, ...parameters };
 
-        const hooks = this.getLifecycleHooks(instance, 'load').map(hook =>
+        const hooks = this.getLifecycleHooks(instance, 'loading').map(hook =>
           () => hook(instance, merged, this.instruction, this.navigation));
+
+        hooks.push(...this.getLifecycleHooks(instance, 'load').map(hook =>
+          () => {
+            console.warn(`[Deprecated] Found deprecated hook name "load" in ${this.instruction.component.name}. Please use the new name "loading" instead.`);
+            return hook(instance, merged, this.instruction, this.navigation);
+          }));
 
         if (hooks.length !== 0) {
           // Add hook in component
-          if (instance.load != null) {
-            hooks.push(() => instance.load!(merged, this.instruction, this.navigation));
+          if (instance.loading != null) {
+            hooks.push(() => instance.loading!(merged, this.instruction, this.navigation));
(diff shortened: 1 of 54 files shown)


packages/router/src/endpoints/viewport-content.ts
```diff
@@ -398,21 +398,36 @@ export class ViewportContent extends EndpointContent {
         const parameters = this.instruction.typeParameters(this.router);
         const merged = { ...this.navigation.parameters, ...parentParameters, ...parameters };
 
-        const hooks = this.getLifecycleHooks(instance, 'load').map(hook =>
+        const hooks = this.getLifecycleHooks(instance, 'loading').map(hook =>
           () => hook(instance, merged, this.instruction, this.navigation));
 
+        hooks.push(...this.getLifecycleHooks(instance, 'load').map(hook =>
+          () => {
+            console.warn(`[Deprecated] Found deprecated hook name "load" in ${this.instruction.component.name}. Please use the new name "loading" instead.`);
```
bigopon, line 406: the warnings probably should be wrapped in `__DEV__` or removed before we go beta. Also, don't we need to avoid calling it twice, to prevent confusion when someone is migrating their app from `load` to `loading`?

jwx: Yeah, I was thinking the warnings and the calling of the old names should be removed after a grace period. I've got both old and new names in there so that apps don't suddenly stop working. (But, as I said, I intended it only for a short period of time.) I don't think they will be called twice cause that would only happen if someone were to have both old names and new names in the same components (which feels unlikely). Do you want me to wrap them in `__DEV__` even though they won't remain?

bigopon: > I don't think they will be called twice cause that would only happen if someone were to have both old names

I was thinking we just need to add a local variable so that if it's already called `loading`, `load` won't be called, just a log. This should help one migrate with easy if we intent to have both at the same time. Though maybe if someone does, it can be quickly fixed anyway.

For the log, as long as we remove before going beta, that's fine too.

bigopon:
cc @Sayan751 for router lite.

Issue #808, opened 2020-01-20
Expression parser invalid syntax doesn't throw


bigopon:
# 🐛 Bug Report

Expression parser does not throw on invalid expression

## 🤔 Expected Behavior
For template:
```html
    <div repeat.for="entry of table">${entry.hour}:${entry:minute}</div>
```
An error would be thrown

## 😯 Current Behavior

It still works, but it treat the template as if it's only `${entry}`

## 💁 Possible Solution

N/A

## 💻 Code Sample

N/A

Vheissu:
@bigopon Is this still an issue or can this be closed?

bigopon:
resolved by #1549

PR #1549, opened 2022-09-10
Add regression test for #808

fkleuver:
Seems that https://github.com/aurelia/aurelia/issues/808 was already fixed at some point @bigopon


bigopon:
@fkleuver the bug is not easy to see without a running app. It doesn't seem to throw, though it doesn't terminate the expression properly, as demonstrated by this repro https://stackblitz.com/edit/au2-conventions-vjrl41?file=src%2Fmy-app.html

fkleuver:
@bigopon Yep, you're right. It's fixed and properly tested now. Thanks for the repro

bigopon:
Nice work 🚀 , thanks @fkleuver

PR #1550, opened 2022-09-10
Fix bindables reset after HMR component reload

netcitylife:
# Pull Request

## 📖 Description

Wrong filtering of bindables (comparison was with template bindings, not bindable properties) caused values loss, so component state was wrong after hot reload

### DIFF
--- a/packages-tooling/plugin-conventions/src/hmr.ts
+++ b/packages-tooling/plugin-conventions/src/hmr.ts
@@ -73,17 +73,14 @@ export const getHmrCode = (className: string, moduleText: string = 'module'): st
 
       // @ts-ignore
       previousControllers.forEach(controller => {
-        const values = { ...controller.viewModel };
         const hydrationContext = controller.container.get(IHydrationContext)
         const hydrationInst = hydrationContext.instruction;
 
-        // @ts-ignore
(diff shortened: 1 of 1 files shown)


bigopon:
Thanks @netcitylife , the existing behavior is problematic:
1. when there's no binding to target a bindable property. then the property value wouldn't be transferred.
2. Though this behavior got the internal state persisted right if there' bindings using those internal properties

The new behavior proposed in this fix would  fix (1), and remove (2) by:
3. always retain bindable values, while
4. always discarding other internal states.

I don't think this is ideal since it would defeat the purpose of hmr.

It's not clear to me when to keep and when to discard the existing value of a bindable value, because we don't track whether it has been modified. Maybe there's should be some property to indicate whether a bindable has been changed.
I think a reasonable fix would be in the mean time would be:
5. always discard old bindable property values
6. transferred other internal state if they are used

Does this make sense? I'll have a fix push up for this.

netcitylife:
Ok, it's up to you to decide. I just warry about internal state tranferring because it may lead to a broken state and immediate exceptions after component reload, if it's internal logic was somehow changed. I'll update after you fix this and open a new issue when i'll meet a real case that i describe (theoretically for now).

PR #1554, opened 2022-09-13
feat(router): multiple bindable props for load attribute

jwx:
# Pull Request

## 📖 Description

This PR makes it possible to use multiple bindable properties for the `load` attribute as well as adds configured route `id` as an option to select configured route with. So given a configured route
```ts
  { id: 'thing-params', path: 'thing/:id', component: MyThing, title: 'The Thing!' },
```
it will be possible to use the following options to `load`
```html
  <a load="thing/123">
  <a load="id: thing-params; parameters.bind: { id: 123 };">
  <a load.bind="{ id: 'thing-params', parameters: { id: 123 } }">
```

## 📑 Test Plan

- Make tests for the new options

### DIFF
--- a/packages/router/src/instructions/routing-instruction.ts
+++ b/packages/router/src/instructions/routing-instruction.ts
@@ -69,7 +69,7 @@ export class RoutingInstruction {
   /**
    * The configured route, if any, that the routing instruction is part of.
    */
-  public route: FoundRoute | null = null;
+  public route: FoundRoute | string | null = null;
 
   /**
    * The instruction is the start/first instruction of a configured route.
@@ -97,9 +97,9 @@ export class RoutingInstruction {
   /**
    * Whether the routing instruction has been cancelled (aborted) for some reason
    */
-   public cancelled: boolean = false;
+  public cancelled: boolean = false;
 
-   public constructor(
+  public constructor(
     component?: ComponentAppellation | Promise<ComponentAppellation>,
     endpoint?: EndpointHandle,
     parameters?: ComponentParameters,
@@ -158,9 +158,10 @@ export class RoutingInstruction {
         instructions.push(RoutingInstruction.create(instruction) as RoutingInstruction);
       } else if (InstructionComponent.isDefinition(instruction)) {
         instructions.push(RoutingInstruction.create(instruction.Type) as RoutingInstruction);
-      } else if ('component' in instruction) {
+      } else if ('component' in instruction || 'id' in instruction) {
         const viewportComponent = instruction;
         const newInstruction = RoutingInstruction.create(viewportComponent.component, viewportComponent.viewport, viewportComponent.parameters) as RoutingInstruction;
+        newInstruction.route = instruction.id ?? null;
         if (viewportComponent.children !== void 0 && viewportComponent.children !== null) {
           newInstruction.nextScopeInstructions = RoutingInstruction.from(context, viewportComponent.children);
         }
@@ -231,6 +232,21 @@ export class RoutingInstruction {
         .map(instruction => instruction.stringify(context, excludeEndpoint, endpointContext))
         .filter(instruction => instruction.length > 0)
         .join(Separators.for(context).sibling);
+  }
+
+  /**
+   * Resolve a list of routing instructions, returning a promise that should be awaited if needed.
+   *
+   * @param instructions - The instructions to resolve
+   */
+  public static resolve(instructions: RoutingInstruction[]): void | Promise<void | ComponentAppellation[]> {
+    const resolvePromises = instructions
+      .filter(instr => instr.isUnresolved)
+      .map(instr => instr.resolve())
+      .filter(result => result instanceof Promise);
+    if (resolvePromises.length > 0) {
+      return Promise.all(resolvePromises) as Promise<void | ComponentAppellation[]>;
+    }
   }
 
   /**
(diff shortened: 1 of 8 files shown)


packages/router/src/resources/load.ts
```diff
@@ -1,15 +1,29 @@
...
+import { RoutingScope } from '../routing-scope';
+import { IRoute } from '../route';
 
 @customAttribute('load')
 export class LoadCustomAttribute implements ICustomAttributeViewModel {
-  @bindable({ mode: BindingMode.toView })
-  public value: unknown;
+  @bindable({ mode: BindingMode.toView }) public value: unknown;
+  @bindable public component?: string;
+  @bindable public parameters?: Parameters;
```
bigopon, line 18: What's the motivation of changing params to parameters?

packages/router/src/resources/load.ts
```diff
@@ -40,31 +57,71 @@ export class LoadCustomAttribute implements ICustomAttributeViewModel {
...
 
+    const unresolvedPromise = RoutingInstruction.resolve(instructions);
+    if (unresolvedPromise instanceof Promise) {
+      await unresolvedPromise;
+    }
+
     element.classList.toggle(this.activeClass, this.router.checkActive(instructions, { context: controller }));
   }
+
+  private findRoute(value: string | IRoute): FoundRoute {
```
bigopon, line 113: We can reduce the effort required for the clean up later if all the new code can be prefixed with `_` on the god, it'll be fewer changes to go through. Can we start doing this?

packages/router/src/routing-scope.ts
```diff
@@ -760,35 +758,36 @@ export class RoutingScope {
...
 
     return found;
   }
 
-  private findMatchingRouteInRoutes(path: string, routes: Route[]): FoundRoute | null {
+  private findMatchingRouteInRoutes(path: string, routes: Route[], parameters: Parameters): FoundRoute {
+    const found = new FoundRoute();
     if (routes.length === 0) {
-      return null;
+      return found;
```
bigopon, line 790: It seems like it'll be more prone to bug not distinguishing carefully between not found and found routes this way. Maybe add a `type` property on `FoundRoute` with value `none` to indicate it, if we want to avoid null?

packages/router/src/instructions/routing-instruction.ts
```diff
@@ -158,9 +158,10 @@ export class RoutingInstruction {
         instructions.push(RoutingInstruction.create(instruction) as RoutingInstruction);
       } else if (InstructionComponent.isDefinition(instruction)) {
         instructions.push(RoutingInstruction.create(instruction.Type) as RoutingInstruction);
-      } else if ('component' in instruction) {
+      } else if ('component' in instruction || 'id' in instruction) {
```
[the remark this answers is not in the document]
jwx: A configured route with a redirect doesn't get this far.

packages/router/src/resources/load.ts
jwx: It's no change. This, parameters for routing instruction, has been `parameters` from the start.

packages/router/src/routing-scope.ts
jwx: `FoundRoute` already has a `foundConfiguration` for this.

packages/router/src/resources/load.ts
bigopon: where is that `parameters`? I can see here that we are only adding that bindable in this PR. Also, this is a public API, which means we have a mismatch in terms on public API:
v1: params.bind
v2: parameters.bind

While this is minor, it's unnecessary, and it comes with downsides: longer names to type (boilerplate-y). Can we change it to `params`?

jwx:
```html
 <a load.bind="{ component: 'my-thing', parameters: { id: 123 } }">
 ```
 has literally been available for years in the v2 router as a public API. I'd rather not have a mismatch between the `load.bind` api and the multiple bindable properties api.

Comment (bigopon) [reply packages/router/src/resources/load.ts:18]: with regards to bindable, we can do this
```ts
@bindable({ attribute: 'params' }) parameters: unknown
```
so it can be `params` in template and `parameters` in the view model. Does this resolve your concern?

jwx: Not really, because the mismatch between 
```html
 <a load.bind="{ component: 'my-thing', parameters: { id: 123 } }">
 ```
 and
 ```html
  <a load="component: 'my-thing'; params.bind: { id: 123 } }">
  ```
would be the most confusing part in the api.

bigopon: Right, I see the previous comment clearer now. Then my first question still applies: why would we want to change from params to parameters at the start of v2. I can't find a good reason for this, though if you really insist, we can go with it and ask folks to rename while migrating.

jwx: There were, and still are, reasons to use `parameters`. (And it's technically not a change from v1 to v2, since what this originally was, parameters for direct routing instructions, doesn't really exist in v1.) The reasons might however not be good enough so I think we can discuss it. But preferably outside of this PR, since if we are to rename it it will require a lot of renaming throughout the router and I'd prefer that to be done in a separate PR. If you want, I can create an RFC in a couple of days with the alternatives and the pros and cons I can come up with and we can take it from there.

bigopon: Cool, thanks. We probably can't have parameters in load and params in href so whatever the choice, it should be for both. renaming/discussion of it in a separate PR sounds good 👍

PR #1556, opened 2022-09-14
refactor(router): add route compare to active check

jwx:
# Pull Request

## 📖 Description

This PR adds a compare route step to the active link check, removing the need to resolve instructions in `load`, as well as addresses some feedback on https://github.com/aurelia/aurelia/pull/1554.

## 📑 Test Plan

### DIFF
--- a/packages/router/src/instructions/routing-instruction.ts
+++ b/packages/router/src/instructions/routing-instruction.ts
@@ -406,10 +406,10 @@ export class RoutingInstruction {
       return false;
     }
     if (typeof thisRoute === 'string' || typeof otherRoute === 'string') {
-      return (thisRoute ?? '') === (otherRoute ?? '');
+      return thisRoute === otherRoute;
     }
 
-    return ((thisRoute as Route).id ?? '') === ((otherRoute as Route).id ?? '');
+    return (thisRoute as Route).id === (otherRoute as Route).id;
   }
 
   /**
--- a/packages/router/src/instructions/routing-instruction.ts
+++ b/packages/router/src/instructions/routing-instruction.ts
@@ -8,7 +8,7 @@ import { FoundRoute } from '../found-route';
 import { Endpoint, EndpointType } from '../endpoints/endpoint';
 import { Viewport } from '../endpoints/viewport';
 import { CustomElement } from '@aurelia/runtime-html';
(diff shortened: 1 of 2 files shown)


packages/router/src/instructions/routing-instruction.ts
```diff
@@ -393,6 +393,25 @@ export class RoutingInstruction {
...
+   * @param other - The routing instruction to compare to
+   */
+  public sameRoute(other: RoutingInstruction): boolean {
+    const thisRoute = this.route?.match;
+    const otherRoute = other.route?.match;
+    if (thisRoute == null || otherRoute == null) {
+      return false;
+    }
+    if (typeof thisRoute === 'string' || typeof otherRoute === 'string') {
+      return (thisRoute ?? '') === (otherRoute ?? '');
```
bigopon, line 409: why do we need to normalize to an empty string here?

packages/router/src/instructions/routing-instruction.ts
```diff
@@ -393,6 +393,25 @@ export class RoutingInstruction {
...
+    }
+
+    return ((thisRoute as Route).id ?? '') === ((otherRoute as Route).id ?? '');
```
bigopon, line 412: same with this for the normalization to a string

packages/router/src/resources/load.ts
```diff
@@ -22,7 +22,7 @@ export class LoadCustomAttribute implements ICustomAttributeViewModel {
    */
   @bindable public id?: string;
 
-  private separateProperties = false;
+  private _separateProperties = false;
```
bigopon, line 25: beside prefixing private property with `_`, we should also annotate `/** @internal */` so it wont be in the final dist typings

bigopon (approved):

packages/router/src/instructions/routing-instruction.ts
bigopon: also, because of line 405, `thisRoute`/`otherRoute` cannot be null here. This is from the deepscan report.

jwx: Yeah, added that check afterwards but didn't remove the normalization. Fixed now.

packages/router/src/instructions/routing-instruction.ts
jwx: Fixed.

PR #1564, opened 2022-09-16
fix(router): fix path issue when redirecting

jwx:
# Pull Request

## 📖 Description

This PR fixes an occasional issue with location path not being updated when redirecting.

## 📑 Test Plan

- Make tests for location path check

### DIFF
--- a/packages/__tests__/router/router.spec.ts
+++ b/packages/__tests__/router/router.spec.ts
@@ -1832,54 +1832,70 @@ describe('Router', function () {
     const Four = CustomElement.define({ name: 'four', template: '!four!', }, class { public canLoad() { return 'zero-id'; } });
 
     const tests = [
-      { load: '/route-two', result: '!zero!', },
-      { load: '/route-one/route-two', result: '!one!!zero!', },
-      { load: '/route-one/route-one/route-two', result: '!one!!one!!zero!', },
+      { load: '/route-two', result: '!zero!', path: '/route-zero', },
+      { load: '/route-one/route-two', result: '!one!!zero!', path: '/route-one/route-zero', },
+      { load: '/route-one/route-one/route-two', result: '!one!!one!!zero!', path: '/route-one/route-one/route-zero', },
 
-      { load: '/route-two/route-one', result: '!zero!', },
-      { load: '/route-one/route-two/route-one', result: '!one!!zero!', },
-      { load: '/route-one/route-one/route-two/route-one', result: '!one!!one!!zero!', },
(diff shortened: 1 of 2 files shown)


packages/router/src/routing-scope.ts
```diff
@@ -466,14 +466,12 @@ export class RoutingScope {
...
-      // Remove cancelled endpoints from changed endpoints
-      earlierMatchedInstructions.filter(instruction => instruction.cancelled).forEach(instruction => {
-        const lastIndex = earlierMatchedInstructions.lastIndexOf(instruction);
-        const lastInstruction = earlierMatchedInstructions[lastIndex];
-        if (lastInstruction.cancelled) {
-          allChangedEndpoints = allChangedEndpoints.filter(endpoint => endpoint !== lastInstruction.endpoint.instance);
-        }
-      });
+      // Remove cancelled endpoints from changed endpoints (last instruction is cancelled)
+      allChangedEndpoints = allChangedEndpoints.filter(endpoint => !(earlierMatchedInstructions
```
bigopon, line 470: this will keep mutating the `earlierMatchedInstructions` for every item of `allChangedEndPoints`, is this intentional?

packages/__tests__/router/router.spec.ts
```diff
@@ -1832,54 +1832,70 @@ describe('Router', function () {
...
+      { load: '/route-two/route-one', result: '!zero!', path: '/route-zero', },
+      { load: '/route-one/route-two/route-one', result: '!one!!zero!', path: '/route-one/route-zero', },
+      { load: '/route-one/route-one/route-two/route-one', result: '!one!!one!!zero!', path: '/route-one/route-one/route-zero', },
 
-      { load: '/three', result: '!zero!', },
-      { load: '/route-one/three', result: '!one!!zero!', },
-      { load: '/route-one/route-one/three', result: '!one!!one!!zero!', },
+      { load: '/three', result: '!zero!', path: '/zero', },
+      { load: '/route-one/three', result: '!one!!zero!', path: '/route-one/zero', },
+      { load: '/route-one/route-one/three', result: '!one!!one!!zero!', path: '/route-one/route-one/zero', },
```
bigopon, line 1845: aren't these lines the same with the first 3 tests?

jwx: No, since `route-two` is a configured route while `three` is just a component.

packages/router/src/routing-scope.ts
jwx: No, it's not. Missed that `reverse` mutates. Good catch!

PR #1569, opened 2022-09-19
fix(router-lite): recovery from unconfigured route

Sayan751:
# Pull Request

### 🎫 Issues

Closes #1341 by handling the errors triggered due to the encounter of un-configured routes, and enqueuing the previous instruction tree.
Note that due to that current design, re-invocation of some hooks is unavoidable. This can be avoided, if the whole path is recognized eagerly, and cancelling the transition as a whole in a more clean way. However, I am planning to push the eager recognition of the paths to later.

## ⏭ Next Steps

Documentation.

PR #1578, opened 2022-09-26
fix(router-lite): handling hash in load, href CAs

Sayan751:
# Pull Request

## 📖 Description

The configuration options supported was supported to be a function
that starts the Router. This was undocumented, also probably contrived.
This is removed, and the options are initialized in the `hydrated` phase
rather than the `activated` phase which was the case previously.
This hindered the availability of the customization of options to the
CAs at the right time.

Note that this is a BREAKING CHANGE, although it might be relatively
low-impacting, as the "feature" was undocumented in the first place.

Additionally, it contains some residual cleanups from the PR #1569 as pointed out by @bigopon.

### DIFF
--- a/docs/engineering-notes/router-architecture-notes.md
+++ b/docs/engineering-notes/router-architecture-notes.md
@@ -138,10 +138,9 @@ This process of updating a given route node with a given viewport instruction tr
 ## Compiling viewport instruction and appending to node
 
 This process starts with a route node and a viewport instruction.
-Quite evidently, in this step the viewport instruction is compiled to a route node, and later down the line the new route node is added as a children to the given route node.
+In this step the viewport instruction is compiled to a route node, and later down the line the new route node is added as a children to the given route node.
 
-Once the child node is created, the process of **appending** the node is very straight-forward.
-It involves appending the child node as one of the children of the given route node, and scheduling update for the viewport-agent associated to the routing context of the child node.
+Once the child node is created, the node (child-node) is appended as one of the children of the given route node, and scheduling update for the viewport-agent associated to the routing context of the child node.
 
 Depending on the type of the viewport instruction, the route node can be **compiled** differently.
 In this phase the viewport instruction can be categorized broadly into string instruction and routed view-model or custom element definition instruction.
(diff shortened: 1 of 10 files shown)


packages/__tests__/router-lite/resources/href.spec.ts
```diff
@@ -195,4 +195,56 @@ describe('href custom-attribute', function () {
...
+      ]
+    })
+    class Root { }
+
+    const { au, host, container } = await start(Root, true, CeOne, CeTwo);
+    const queue = container.get(IPlatform).domWriteQueue;
+    await queue.yield();
+
+    const anchors = Array.from(host.querySelectorAll('a'));
+    assert.deepStrictEqual(anchors.map(a => a.getAttribute('href')), ['#ce-one', '#ce-two', '#ce-two']);
```
bigopon, line 234: shouldn't this require us to have a tests for deep URLs? It seems the template can have 2 forms: with and without `#`, this means they convey different meanings at a nested level.

docs/engineering-notes/router-architecture-notes.md
```diff
@@ -152,7 +151,7 @@ It involves
 - generating the viewport instruction using the routing context, from the route definition; [details](#generating-viewport-instruction)
 - creating a configured node from the viewport instruction; the details can be found [here](#creating-configured-route-node).
 
-Compiling **from the sting instruction** is inherently bit more involved.
+Compiling **from the string instruction** is inherently bit more involved.
```
bigopon, line 154: the `-` is gone, if it was unintentional

packages/__tests__/router-lite/resources/href.spec.ts
Sayan751: Thanks! That skipped me. I will make the changes as required.

docs/engineering-notes/router-architecture-notes.md
Sayan751: Probably you are confusing the `-` (Git deletion) with the list above?

bigopon: yes, thanks for pointing that out.

packages/__tests__/router-lite/resources/href.spec.ts
Sayan751: @bigopon I have pushed the changes. You may take another look.

Sayan751:
@bigopon Thank you for the review. I have merged this one. In case you note any other issues, please let me know.

PR #1583, opened 2022-10-01
feat(repeat): add keyed mode

### LINKED ISSUE #1264: Repeat.for to support key (custom resolution strategy)

fkleuver:
In v1 you could specify how to determine the identity of repeated items via `matcher.bind` and it would reuse existing views in the case of a match. In v2 there is currently no support for this - it'll only recycle the view based on the index. Furthermore, in the case of the array being replaced, the item identity is not taken into account and all views are replaced as well. This is problematic as pointed out in #1264.

This PR addresses both issues: it adds "keyed mode" to `repeat.for` to enable reusing views based on a specific item property or an arbitrary expression. A new syntax has been added to support this. It also takes into account item identity when the whole collection is replaced (this happens automatically).

The syntax is similar to the existing semicolon-separated syntax for multiple bindings in custom attributes.

Examples:

1. `key` (without `.bind`). The value following the colon is taken as a literal string. Use this if your key is simply a property on each item.
  ```html
  <div repeat.for="item of list; key: id">
  ```
2. `key.bind`. The value following the colon is taken as an expression and supports the full Aurelia template expression syntax. Useful when you need composite keys. This specific example is functionally equivalent to the above example
  ```html
  <div repeat.for="item of list; key.bind: item.id">
  ```
3. Collection replacement with item identity remaining the same (without explicit keyed mode)
  ```html
  <div repeat.for="item of list">
  ```
```ts
this.list = this.list.slice(); // item identity and positions remain the same, so nothing will update
```

Keys and key expressions are read/evaluated once per item while the repeater is active and will be re-evaluated only when the repeater itself is deactivated and activated again. In other words, changing the key of an already rendered item from within your viewmodel will not trigger any updates in the repeater.

For `Map` and `Set` it works the same way with one important caveat: when using a `Map` you must use `key.bind`. The reason being that the repeater won't make an assumption about whether to read the provided property from the key or the value out of each map entry.
```html
  <div repeat.for="[key, value] of map; key: id">
                                        ^^^^ not working, unclear
                                             `id` from key or value?
  <div repeat.for="[key, value] of map; key.bind: value.id">
```

Resolves #1264


packages/__tests__/3-runtime-html/repeat.keyed.spec.ts
```diff
@@ -0,0 +1,2232 @@
...
+          assert.strictEqual(mutations[2].removedNodes.length, 1, 'mutations[2].removedNodes.length');
+          assert.strictEqual(mutations[3].addedNodes.length, 1, 'mutations[3].addedNodes.length');
+          assert.strictEqual(mutations[4].addedNodes.length, 1, 'mutations[4].addedNodes.length');
+          assert.strictEqual(mutations[5].addedNodes.length, 1, 'mutations[5].addedNodes.length');
+          assert.strictEqual(mutations[0].removedNodes[0].textContent, '0', 'mutations[0].removedNodes[0].textContent');
+          assert.strictEqual(mutations[1].removedNodes[0].textContent, '2', 'mutations[1].removedNodes[0].textContent');
+          assert.strictEqual(mutations[2].removedNodes[0].textContent, '4', 'mutations[2].removedNodes[0].textContent');
+          assert.strictEqual(mutations[3].addedNodes[0].textContent, '9', 'mutations[3].addedNodes[0].textContent');
+          assert.strictEqual(mutations[4].addedNodes[0].textContent, '7', 'mutations[4].addedNodes[0].textContent');
+          assert.strictEqual(mutations[5].addedNodes[0].textContent, '5', 'mutations[5].addedNodes[0].textContent');
```
bigopon, line 1921: should build an array to present added and removed text content from the first node of added & removed of each mutation record, then can assert the length and text simpler
```ts
assert.strictEqual(removedNodes.length, 3)
assert.deepStrictEqual(removedNodeText, ['0', '2', '4'])
```
similar to the other tests.

packages/runtime-html/src/resources/template-controllers/repeat.ts
```diff
@@ -131,6 +132,11 @@ export class Repeat<C extends Collection = unknown[]> implements ICustomAttribut
 
   // called by SetterObserver
   public itemsChanged(): void {
+    if (typeof this.key === 'string') {
```
bigopon, line 135: without a key specified, it should enter "identity" as key mode

packages/runtime-html/src/resources/template-controllers/repeat.ts
```diff
@@ -175,22 +325,21 @@ export class Repeat<C extends Collection = unknown[]> implements ICustomAttribut
...
-      const oldLength = this.views.length;
       const $indexMap = applyMutationsToIndices(indexMap);
       // first detach+unbind+(remove from array) the deleted view indices
       if ($indexMap.deletedIndices.length > 0) {
         const ret = onResolve(
           this._deactivateAndRemoveViewsByKey($indexMap),
           () => {
             // TODO(fkleuver): add logic to the controller that ensures correct handling of race conditions and add a variety of `if` integration tests
-            return this._createAndActivateAndSortViewsByKey(oldLength, $indexMap);
+            return this._createAndActivateAndSortViewsByKey(oldLen, $indexMap);
```
bigopon, line 335: a question for the internal of tthis method here: why do we still need to apply LIS algo, when we have the exact indexmap of where things are moved?

fkleuver: The LIS algo uses the indexMap to determine the least amount of moves required to achieve the desired state. If you use the raw indexMap you can get the same end result but will often require many more DOM mutations

packages/__tests__/3-runtime-html/repeat.keyed.spec.ts
fkleuver: Yeah that's fair. Hadn't spent time on it yet, but now might be a good moment

packages/runtime-html/src/resources/template-controllers/repeat.ts
fkleuver: Done, see: https://github.com/aurelia/aurelia/pull/1583/files#diff-cd0edd0ad8cf1b4c6968caa4e7882f8394cf4f1c63debd7b153512f956ea47dfR145

bigopon (commented):
At the moment, in the array keyed mode tests, we have tests like this
```
1 moved from pos 1 to 8
1 moved from pos 1 to 8
2 moved from pos 2 to 8
2 moved from pos 2 to 9
```
8 & 9 means nothing without context. If the intention is to assert it moves to the last one and 2nd last one, then maybe we should rename the test title to reflect that.

Also for the single item movement, it's asserting for every position movement in an array of 10 items to the 2nd last and last, is this necessary? While the multiple item movements, which could be much more common in the wild, receive only a handful of cases. I think we can just reduce the number of single movement assertions to the first middle last and that'll be fine.

fkleuver:
> At the moment, in the array keyed mode tests, we have tests like this
> 
> ```
> 1 moved from pos 1 to 8
> 1 moved from pos 1 to 8
> 2 moved from pos 2 to 8
> 2 moved from pos 2 to 9
> ```
> 
> 8 & 9 means nothing without context. If the intention is to assert it moves to the last one and 2nd last one, then maybe we should rename the test title to reflect that.
> 
> Also for the single item movement, it's asserting for every position movement in an array of 10 items to the 2nd last and last, is this necessary? While the multiple item movements, which could be much more common in the wild, receive only a handful of cases. I think we can just reduce the number of single movement assertions to the first middle last and that'll be fine.

Perhaps they're not all necessary but it is through this kind of redundancy in tests that I tend to discover edge cases that would otherwise go unnoticed. Some interactions between the diffing and the lis algorithm yielded surprising results to me before, so I added a bigger set of tests like that for more confidence.

Normally I would have done this with a double loop (one loop on the number of items to move, and another loop on the number of positions to move them). Still, there are more complex scenarios I can think of that combine different kinds of add/replace/remove/move operations, batched and in sequence.

Per your request I avoided test loops entirely but perhaps adding some loops in addition to this set of ordinary tests wouldn't be such a bad idea? I mean, pick your lesser of the evils:
- Minimalist test suite without loops -> potential bugs lingering
- Thorough test suite without loops -> huge spec file that's hard to navigate
- Thorough test suite with loops -> harder to grok what exactly is being tested

bigopon:
Thanks for the explanation of the abundant of the tests.
For looping vs non looping, I don't think we want to turn back to that, even if we have to. If something can only be tested by looping 20 30 scenarios, we better not have it or find another way to have it, because there won't be anyone able to maintain it after awhile. I'm finding myself hesitant, and sometimes struggle to refactor the tests, even for the ones that are only not too complex.

What about my suggestion of renaming 8 & 9 to 2nd last & last? I'm not able to guess why 8 and 9 are picked to tests, or why they are more important than, say 3 and 6.

> The LIS algo uses the indexMap to determine the least amount of moves required to achieve the desired state. If you use the raw indexMap you can get the same end result but will often require many more DOM mutations

Maybe this is right, but we may want to do better picking the cases when and when NOT to enter LIS. Imagine a scenario where a single item is deleted from 200 item collection. I don't think we want to go through all the trouble to figure out that we just want to remove a view, which is quite obvious from the indexmap.

fkleuver:
> Thanks for the explanation of the abundant of the tests. For looping vs non looping, I don't think we want to turn back to that, even if we have to. If something can only be tested by looping 20 30 scenarios, we better not have it or find another way to have it, because there won't be anyone able to maintain it after awhile. I'm finding myself hesitant, and sometimes struggle to refactor the tests, even for the ones that are only not too complex.

What about a middle ground, setting a limit on one loop (so no nested loops) and keep the looped variables/specs simple? Certainly not all looped tests are equal in that sense. I think the route recognizer tests are not so bad for example.

> What about my suggestion of renaming 8 & 9 to 2nd last & last? I'm not able to guess why 8 and 9 are picked to tests, or why they are more important than, say 3 and 6.

I'm fine with any kind of renaming for better clarity. As for the choices of positions themselves, here's what I wanted to cover:
- Moving something to the first and last position is generally important for eliminating potential off-by-one issues
- Moving to 2nd and 2nd last position for the same reason, because there's a specific piece of logic that tries to narrow the loop boundaries so this will cover both the "true" boundaries and the "virtual" boundaries
- Moving to some other position in the middle (in addition to moving to positions away from the middle) to verify that the more complex part of the diff works properly too (not just the obvious examples with linear sequences of mutations)

The tests can be renamed/restructured according to those goals, that wouldn't be a bad idea. I'll take care of it.

> > The LIS algo uses the indexMap to determine the least amount of moves required to achieve the desired state. If you use the raw indexMap you can get the same end result but will often require many more DOM mutations
> 
> Maybe this is right, but we may want to do better picking the cases when and when NOT to enter LIS. Imagine a scenario where a single item is deleted from 200 item collection. I don't think we want to go through all the trouble to figure out that we just want to remove a view, which is quite obvious from the indexmap.

LIS is a very cheap and fast algorithm, and I personally don't see a worthwhile benefit in conditionally skipping it for e.g. a single add or delete. I'm only skipping it (and the mapping) right now when it's all adds or all deletes.

If we go down this path there are plenty more scenarios imaginable where we can skip work that are much more costly than going into LIS. You could also say that if the list of items is small, a simple loop can be used so no key mapping is needed. But these are all micro optimizations that will add code and complexity. How far do we want to go with that?

packages/runtime-html/src/resources/template-controllers/repeat.ts
```diff
@@ -164,6 +156,161 @@ export class Repeat<C extends Collection = unknown[]> implements ICustomAttribut
...
+        const newIndices = new Map<unknown, number>();
+
+        i = 0;
+        // Step 1: narrow down the loop range as much as possible by checking the start and end for key equality
+        outer: {
+          // views with same key at start
+          // eslint-disable-next-line no-constant-condition
+          while (true) {
+            if (isMap) {/* empty */}
+            oldItem = isMap ? oldItems[i][1] as IIndexable : oldItems[i];
```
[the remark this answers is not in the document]
fkleuver: Yeah I mentioned this earlier to you as well in chat. I personally think this whole use case is a bit odd. I spent some time thinking about it and just couldn't come up with a very sensible reason to use keyed mode in conjunction with a map. I would be more inclined to disable map+keyed support and wait for a feature request so that we can understand the use cases. If you have any other ideas, I'm all ears

bigopon:
One problem with loop I find is its ability to get us on the lazy mode for meaningful test title, but maybe it's the problem with the existing test not trying to do a better job at the title, stating what it asserts rather than why it asserts.
Though if it can be avoided, I'd still prefer to read case by case so tests can be easily isolated with `grep` / `.only` / `.skip` without having to modify a bunch of code. Without this ability, evolving any feature further has a significant mental overload I don't think is worth it. Unless it is determined it will never be changed again.

> LIS is a very cheap and fast algorithm, and I personally don't see a worthwhile benefit in conditionally skipping it for e.g. a single add or delete. I'm only skipping it (and the mapping) right now when it's all adds or all deletes.

Thanks, maybe it is such. For fun: this will ensure that we lose in bench for the extreme simple case like "Who drops a row fastest" in Krauset, but I guess that's fine if it's cheap enough.

PR #1603, opened 2022-10-13
refactor(router-lite): same URL handling

Sayan751:
# Pull Request

## 📖 Description

This PR drops the `sameUrlStrategy` router-configuration option in favour of consolidating the behaviour to the `transitionPlan` route-configuration option. This also enables the inheritance of the `transitionPlan` from the parent configuration. See the tests for example usage.

### DIFF
--- a/packages/__tests__/router-lite/smoke-tests.spec.ts
+++ b/packages/__tests__/router-lite/smoke-tests.spec.ts
@@ -4,6 +4,7 @@ import { RouterConfiguration, IRouter, NavigationInstruction, IRouteContext, Rou
 import { LifecycleFlags, Aurelia, valueConverter, customElement, CustomElement, ICustomElementViewModel, IHistory, IHydratedController, ILocation, INode, IPlatform, IWindow, StandardConfiguration, watch } from '@aurelia/runtime-html';
 
 import { TestRouterConfiguration } from './_shared/configuration.js';
+import { start } from './_shared/create-fixture.js';
 
 function vp(count: number): string {
   return '<au-viewport></au-viewport>'.repeat(count);
@@ -2425,6 +2426,49 @@ describe('router (smoke tests)', function () {
       assert.strictEqual(await router.load({ component: CeL23, params: { id: '5', a: '6' } }, { context: ce12 }), true);
       assert.match(location.path, /12\/1\/23\/5\/tt\/6$/);
       BaseRouteViewModel.assertAndClear('params5', ['cel23', [{ id: '5', a: '6' }, new URLSearchParams()]]);
+
+      await au.stop();
+    });
+  });
+
+  describe('transition plan', function () {
+    it('replace', async function () {
+      @customElement({ name: 'ce-one', template: 'ce1 ${id1} ${id2}' })
+      class CeOne implements IRouteViewModel {
+        private static id1: number = 0;
+        private static id2: number = 0;
+        private readonly id1: number = ++CeOne.id1;
+        public id2: number;
+        public canLoad(): boolean {
+          this.id2 = ++CeOne.id2;
+          return true;
+        }
+      }
+
+      @route({
+        transitionPlan: 'replace',
+        routes: [
+          {
+            id: 'ce1',
+            path: 'ce1',
+            component: CeOne,
+            transitionPlan: 'replace',
+          },
+        ]
+      })
+      @customElement({ name: 'ro-ot', template: '<a load="ce1"></a><au-viewport></au-viewport>' })
+      class Root { }
+
+      const { au, container, host } = await start(Root, false, CeOne);
+      const queue = container.get(IPlatform).domWriteQueue;
+
+      host.querySelector('a').click();
+      await queue.yield();
+      assert.html.textContent(host, 'ce1 1 1', 'round#1');
+
+      host.querySelector('a').click();
+      await queue.yield();
+      assert.html.textContent(host, 'ce1 2 2', 'round#2');
 
       await au.stop();
     });
--- a/packages/__tests__/router-lite/smoke-tests.spec.ts
+++ b/packages/__tests__/router-lite/smoke-tests.spec.ts
@@ -2438,7 +2438,7 @@ describe('router (smoke tests)', function () {
         private static id1: number = 0;
         private static id2: number = 0;
         private readonly id1: number = ++CeOne.id1;
-        public id2: number;
+        private id2: number;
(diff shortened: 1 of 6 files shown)


packages/__tests__/router-lite/smoke-tests.spec.ts
```diff
@@ -2429,4 +2430,290 @@ describe('router (smoke tests)', function () {
...
+
+  describe('transition plan', function () {
+    it('replace - inherited', async function () {
+      @customElement({ name: 'ce-one', template: 'ce1 ${id1} ${id2}' })
+      class CeOne implements IRouteViewModel {
+        private static id1: number = 0;
+        private static id2: number = 0;
+        private readonly id1: number = ++CeOne.id1;
+        public id2: number;
+        public canLoad(): boolean {
```
bigopon, line 2442: nit: in tests, we better leave `public` out so there's not too much noise

packages/__tests__/router-lite/smoke-tests.spec.ts
```diff
@@ -2429,4 +2430,290 @@ describe('router (smoke tests)', function () {
...
+      })
+      @customElement({ name: 'ro-ot', template: '<a load="ce1"></a><au-viewport></au-viewport>' })
+      class Root { }
+
+      const { au, container, host } = await start(Root, false, CeOne);
+      const queue = container.get(IPlatform).domWriteQueue;
+      const router = container.get<Router>(IRouter);
+
+      host.querySelector('a').click();
+      await router['currentTr'].promise;
```
bigopon, line 2508: nit: we shouldn't be poking private properties this way. Either make it public and assert or use some other different observation mechanism. If we want it to be readonly, can turn it into `_currentTr` + a getter `currentTr`.

Also, in tests, we are allowed to "know" how long it takes to resolve the transition, so can just do `await waitFor(10ms)` for example.

Having private accessed in tests makes it quite difficult to refactor later on.

packages/__tests__/router-lite/smoke-tests.spec.ts
```diff
@@ -2429,4 +2430,290 @@ describe('router (smoke tests)', function () {
...
+        private readonly id1: number = ++CeOne.id1;
+        public id2: number;
+        public canLoad(): boolean {
+          this.id2 = ++CeOne.id2;
+          return true;
+        }
+      }
+
+      @route({
+        transitionPlan: 'replace',
```
bigopon, line 2449: 👍

packages/__tests__/router-lite/smoke-tests.spec.ts
```diff
@@ -2429,4 +2430,290 @@ describe('router (smoke tests)', function () {
...
+
+      host.querySelector('a').click();
+      await router['currentTr'].promise;
+      await queue.yield();
+      assert.html.textContent(host, 'ce1 1 2', 'round#2');
+
+      await au.stop();
+    });
+
+    it('transitionPlan function #2 - sibling', async function () {
```
bigopon, line 2520: 👍

packages/__tests__/router-lite/smoke-tests.spec.ts
```diff
@@ -2429,4 +2430,290 @@ describe('router (smoke tests)', function () {
...
+
+      host.querySelector('a').click();
+      await router['currentTr'].promise;
+      await queue.yield();
+      assert.html.textContent(host, 'ce1 2 2 ce2 1 2', 'round#2');
+
+      await au.stop();
+    });
+
+    it('transitionPlan function #3 - parent-child - #1', async function () {
```
bigopon, line 2583: rather than `parent - child - #1`, would `parent - child - replace parent` be better? similar to the test after this, should be `parent - child - invoke lifecycle parent`. `#1` and `#2` mean nothing and would require reading the whole tests to find out what is asserted, when this test fail because of some refactoring.

bigopon (commented):
nice work @Sayan751, some nits about noises, refactoring guards and test titles

packages/__tests__/router-lite/smoke-tests.spec.ts
Sayan751: Done.

packages/__tests__/router-lite/smoke-tests.spec.ts
Sayan751: Converted the getter to `public`. 

However, I don't agree to adding a `waitFor` certain ms. From my experience, that simply the first step towards creating flaky tests.

packages/__tests__/router-lite/smoke-tests.spec.ts
Sayan751: You are right. This was me being lazy. Thank you for pointing it out.

Sayan751:
Thanks @bigopon for the review. I have made the changes as pointed out by you. I have merged this PR. In case you see any further issues, let me know.

packages/__tests__/router-lite/smoke-tests.spec.ts
bigopon: it may sound flaky, but in my experience, it's actually nicer than waiting for the promise from the transaction itself, which translates to:
- `await`: just ensure you run to the end, whatever side effects you may bring. Changes in the internal can make a subtle timing difference here and it'll still be fine.
- `wait for ms`: the execution of this is known exactly, and it should be this way until there's a change in the timing of the router lite, then this number will catch it.

Sayan751: > wait for ms: the execution of this is known exactly, and it should be this way until there's a change in the timing of the router lite, then this number will catch it.

Unfortunately, cannot agree with you here. According to my experience, that's simply unsustainable in the long run. Nobody has the time to analyse later why a test must wait more milliseconds when the tests start failing randomly triggered by some changes, either framework level or app level (speaking generally from devs' perspective). As I see it, it is always better to wait for something definitive. From a theoretical perspective, you might be right, however, my practical experience sounds alarm, when I see something like that.

bigopon: I missed this. thanks for the reply. In practice, I've dealt with both ways, and both feel wrong in different scenarios.
In cases where the timing is opaque, this is fine. 
In other cases where it's simpler and more transparent what's going on, awaiting for it to resolves doesn't feel as good.

My original comment is a suggestion for an environment where we know everything exactly, but if the router is deemed to have too complicated internal timing, then it's fine where it is.

PR #1606, opened 2022-10-16
fix(router-lite): history state change timing

Sayan751:
# Pull Request

## 📖 Description

Previously, the history state was changed after the navigation end event is raised. Any subscriber listening to the navigation-end event might end up with wrong history state. This PR makes the history adjustment before the navigation-end event is raised.


bigopon (approved):
This seems too obvious to miss, so it probably was intentional. Maybe @fkleuver can comment on this for the original implementation.

Sayan751:
@bigopon I think the original implementation was based on the assumption that in post-navigation phase, the order of the history state change and the navigation-end event does not matter. If you aren't too picky, then the previous order or the changed one, won't make much of a difference for you. I discovered it while preparing docs for router-lite and now, IMO, this is how the order should be.

@fkleuver I am going ahead and merging the PR. In case you have any feedback/comment, please let me know.

PR #1609, opened 2022-10-20
feat(router): improve href creation

jwx:
# Pull Request

## 📖 Description

This PR improves the creation of the `href` attribute by the `load` attribute by adding scope path. It also lets the router avoid to catch the wrong errors when a component can't be created.

## 📑 Test Plan

- Add an e2e test with several `href` checks

### DIFF
--- a/packages/router/src/routing-scope.ts
+++ b/packages/router/src/routing-scope.ts
@@ -17,6 +17,7 @@ import { EndpointContent, Navigation, Router, RoutingHook, ViewportCustomElement
 import { IContainer } from '@aurelia/kernel';
 import { arrayRemove, arrayUnique } from './utilities/utils';
 import { Parameters } from './instructions/instruction-parameters';
+import { Separators } from './router-options';
 
 export type TransitionAction = 'skip' | 'reload' | 'swap' | '';
 
@@ -71,7 +72,7 @@ export class RoutingScope {
    */
   public children: RoutingScope[] = [];
 
-  public path: string | null = null;
+  // public path: string | null = null;
 
   public constructor(
     public readonly router: IRouter,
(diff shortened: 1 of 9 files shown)


packages/router/src/routing-scope.ts
```diff
@@ -95,12 +96,12 @@ export class RoutingScope {
     // console.log('Created RoutingScope', this.id, this);
   }
 
-  public static for(origin: Element | ICustomElementViewModel | Viewport | ViewportScope | RoutingScope | ICustomElementController | IContainer | null): RoutingScope | null {
+  public static for(origin: Element | ICustomElementViewModel | Viewport | ViewportScope | RoutingScope | ICustomElementController | IContainer | null, instruction?: string): { scope: RoutingScope | null; instruction: string | undefined } {
```
bigopon, line 99: Does the type `{scope: RoutingScope; instruction: string | undefined }` have a name? if not, we probably can try to come back later to give it a name for better API comprehensibility, since this is a public API.

packages/router/src/routing-scope.ts
```diff
@@ -121,12 +122,44 @@ export class RoutingScope {
...
+    if (instruction.startsWith('/')) {
+      return { scope: null, instruction: instruction.slice(1) };
+    }
+    // Instruction specifies scope traversals
+    while (instruction.startsWith('.')) {
+      // The same as no scope modification
+      if (instruction.startsWith('./')) {
+        instruction = instruction.slice(2);
+      } else if (instruction.startsWith('../')) { // Traverse upwards
+        scope = scope.parent ?? scope;
```
bigopon, line 151: this `??` seems surprising, as we are traversing upward but the scope may remain the same. Is this expected?

bigopon (commented):
👍

packages/router/src/routing-scope.ts
jwx: Yes. The intention is that no matter how many steps you traverse upwards, you're not traversing above the top (trying to traverse above the root means you end up in the root).

packages/router/src/routing-scope.ts
jwx: Yeah... I had a name blockage but didn't want to get stuck there. I'll take a look at it with some fresh eyes soon and give it a name.

Issue #1618, opened 2022-11-22
Binding to value of select element throws error in macOS Safari


akesser:
# 🐛 Bug Report

When I try to bind to the selected value of a `<select>` element via something like `<select value.bind="selectedValue">` Safari throws an error:

```
Unhandled Promise Rejection: TypeError: null is not an object (evaluating 'new this._obj.ownerDocument.defaultView.MutationObserver')
```

The `defaultView` is null. I am using Safari version 16.1 on macOS Monterey 12.5.1. In brave the binding works without an error.

The `ownerDocument` in Safari looks like a reference to a blank page, as `URL`  is `about:blank` whereas in Brave, `URL` is the correct `http://localhost:9000/` 

## 🤔 Expected Behavior

Safari should not throw an error and should bind the variable to the selected value of the `<select>` element as Brave does

## 😯 Current Behavior

Error Message in Safari:
Unhandled Promise Rejection: TypeError: null is not an object (evaluating 'new this._obj.ownerDocument.defaultView.MutationObserver')

Stacktrace in Safari:
_start (index.dev.mjs:8003)
subscribe (index.dev.mjs:8042)
$bind (index.dev.mjs:1874)
bind (index.dev.mjs:3685)
activate (index.dev.mjs:3665)
(anonyme Funktion) (index.dev.mjs:8829)
onResolve (index.dev.mjs:263)
attaching (index.dev.mjs:8810)
_attach (index.dev.mjs:3775)
bind (index.dev.mjs:3712)
activate (index.dev.mjs:3665)
_attach (index.dev.mjs:3788)
bind (index.dev.mjs:3712)
activate (index.dev.mjs:3665)
(anonyme Funktion) (index.dev.mjs:1933)
process (index.dev.mjs:997)
(anonyme Funktion) (index.dev.mjs:1011)

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Aurelia          | 2.0.0-alpha.41
| Language         | typescript
| Browser          | safari 16.1
| Bundler          | web pack
| Operating System | macOS Monterey 12.5.1
| NPM   |  8.19.2
| node | v18.12.1

Vheissu:
This is the bug I mentioned to you in Discord @bigopon it effectively breaks apps in Safari and is a bit of a show stopper. I think we need to address this one with high priority (if not already).

bigopon:
Thanks @Vheissu
The bug boils down to the following test
```ts
[...]

PR #1622, opened 2022-12-02
Fix `MutationObserver` from platform (#1618)

aegenet:
# Pull Request

## 📖 Description

Fixed **Safari 16.1** issue on `select` and others components.

Due to recent webkit breaking changes, `adoptNode` does not fill `ownerDocument` field correctly and therefore `defaultView` is `undefined`.

In our case, `MutationObserver` is used in `runtime-html` through `ownerDocument.defaultView`.
This PR change this behavior, we get the `MutationObserver` from `platform`.

### 🎫 Issues

* #1618

## 👩‍💻 Reviewer Notes

I'm not familiar with the code base, so I browsed through the history to figure out the "why": why we get the MutationObserver from `ownerDocument.defaultView.MutationObserver`.

The reason seems to be:

> "make runtime-html fully work in jsdom/nodejs"

At first MutationObserver was retrieved by `dom.createNodeObserver`, then `dom.window.MutationObserver`, then `platform.MutationObserver` and finally `obj.ownerDocument.defaultView!.MutationObserver`.

So going back to `platform` seems correct (with my limited view of the project).

### DIFF
--- a/packages/runtime-html/src/binding/attribute.ts
+++ b/packages/runtime-html/src/binding/attribute.ts
@@ -156,6 +156,7 @@ export class AttributeBinding implements IBinding {
       this.target as HTMLElement,
       this.targetProperty,
       this.targetAttribute,
+      this.l,
     );
 
     if (this.mode & (BindingMode.toView | BindingMode.oneTime)) {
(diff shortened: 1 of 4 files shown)


bigopon:
Thanks for the PR. The reason why `MutationObserver` is retrieved directly from the `window` that owns the `document` that owns the node is because of a scenario where an application from the main tab observing some element from an iframe. We won't be able to achieve the same thing if the `MutationObserver` is always coming from `IPlatform`.

aegenet:
Thanks for the clarification, I understand better.

I may be wrong, but once the `adoptNode` was done, the `ownerDocument` became the one of the current document anyway, right? (https://developer.mozilla.org/en-US/docs/Web/API/Document/adoptNode)

PR #1636, opened 2023-01-09
chore(create-aurelia): holder for create-aurelia

brandonseydel:
# Pull Request


3cp:
I think the trend is to use name "create-aurelia-app"? (Following create-react-app and create-vue-app).

It will be used as "npm create aurelia-app", instead of "npm create aurelia" in this PR.

brandonseydel:
It's the name on npm which don't have app in them.  

npm create vue|svelte|solid etc

fkleuver:
Closing this placeholder because Aurelia already provides application scaffolding through `npx makes aurelia`. If a separate `create-aurelia` package becomes necessary, it can return as a concrete proposal with a defined purpose beyond the existing workflow.

PR #1641, opened 2023-01-12
chore: fix the circleci issues

Sayan751:
# Pull Request


.circleci/config.yml
```diff
@@ -101,6 +101,7 @@ commands:
         type: boolean
         default: true
     steps:
+      - checkout
```
bigopon, line 104: @Sayan751 why do we need to add this step? I think no `checkout` is intentional, to save on time and unnecessary work.

Sayan751: Hi @bigopon! It was already there, I just moved it to the first step. Otherwise, git complained in CI that it cannot checkout in a non-empty directory.

PR #1647, opened 2023-01-19
fix(hmr): no invalidate on parcel

brandonseydel:
# Pull Request

### DIFF
--- a/packages-tooling/plugin-conventions/src/hmr.ts
+++ b/packages-tooling/plugin-conventions/src/hmr.ts
@@ -59,50 +59,49 @@ export const getHmrCode = (className: string, moduleText: string = 'module'): st
       $$M.define(newDefinition.name, newDefinition, newDefinition);
       hot.data.aurelia.container.res[$$CE.keyFrom(newDefinition.name)] = newDefinition;
 
-      const previousControllers = hot.data.controllers;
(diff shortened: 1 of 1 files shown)


packages-tooling/plugin-conventions/src/hmr.ts
```diff
@@ -62,46 +62,47 @@ export const getHmrCode = (className: string, moduleText: string = 'module'): st
...
       }
-
-      // @ts-ignore
-      previousControllers.forEach(controller => {
-        const values = { ...controller.viewModel };
-        const hydrationContext = controller.container.get($$IHC)
-        const hydrationInst = hydrationContext.instruction;
-
-        const bindableNames = Object.keys(controller.definition.bindables);
+      else{
```
bigopon, line 67: hmm good else, in case `previousControllers` is `null`, the next line will throw. Though it adds quite a bit of indentation to this code block string. Maybe on the line 62
```ts
      const previousControllers = hot.data.controllers;
```
we should do
```ts
      const previousControllers = hot.data.controllers ?? []
```
instead?

brandonseydel: @bigopon - yea makes sense then i can only check length next line

PR #1653, opened 2023-01-23
Support TypeScript moduleResolution: "nodenext"/"node16"

### LINKED ISSUE #1652: Support TypeScript moduleResolution "nodenext"/"node16"
# 🐛 Bug Report
When ECMAScript module support was finally added to Node, the package.json `"exports"` section was introduced.
When resolving packages in an ESM project, the `"exports"` section is used over of `"main"` (CommonJS).
Aurelia2 packages already are built as hybrid ESM/CommonJS packages, with conditional exports in package.json for CommonJS (`require`) and ESM (`import`).

Both Node and also f.ex. **WebPack** use the package.json `"exports"` section when resolving modules.
To ensure the same behavior with the TypeScript compiler at compile-time, the compilerOption `"moduleResolution"` can be set to `"nodenext"` or `"node16"` (introduced in TypeScript 4.7).

It is desirable to use the `nodenext` (or `node16`) moduleResolution compiler option, because it ensures modules are resolve the same way with TypeScript at compile time as with f.ex. WebPack at "bundle"-time. Consistent module resolution provides more predictable/correct artifacts.

For TypeScript to support the `"nodenext"`/`"node16"` `moduleResolution` compiler option, consumed packages need to include _TypeScript types_ in the `"exports"` section of package.json. (This way you can potentially have different TypeScript types for different conditional exports, but you may also share the same types across different conditional exports).
PS! One should still include the `"types"`/`"typings"` property in package.json as a fallback.

[package.json Exports, Imports, and Self-Referencing](https://devblogs.microsoft.com/typescript/announcing-typescript-4-7/#package-json-exports-imports-and-self-referencing)

In Aurelia2, packages are released with both ESM (import) and CommonJS (require) exports, and the TypeScript types are shared between for both. Please see the PR linked below to how package.json files should be updated.

FYI: When including "types" in "exports", TypeScript requires that "types" is the first property (on whichever level of condition objects).

I have created PR #1653 that fixes this.

PS! I have already tested these package.json changes for the aurelia packages directly in the `node_modules` folder, and verified that it works 👍

## 🤔 Expected Behavior

Aurelia2 should support TypeScript >= 4.7 moduleResolution "nodenext"/"node16".

## 😯 Current Behavior

Aurelia2 does not support TypeScript >= 4.7 moduleResolution "nodenext"/"node16".

## 💁 Possible Solution

PR #1653

## 🌍 Your Environment

| Aurelia          | 2.0.0-beta.1
| Language         | TypeScript 4.9
| Browser          | N/A
| Bundler          | WebPack
| Operating System | N/A
| NPM/Node/Yarn    | LTS

josundt:
# Pull Request

## 📖 Description

Fixes #1652

I have also changed the name of the `"typings"` property in package.json files to `"types"` (alias), because this seem to have become the industry standard.

### 🎫 Issues

#1652

### DIFF
--- a/packages-tooling/aot/package.json
+++ b/packages-tooling/aot/package.json
@@ -4,10 +4,11 @@
   "main": "dist/cjs/index.cjs",
   "module": "dist/esm/index.mjs",
   "exports": {
+    "types": "./dist/types/index.d.ts",
     "require": "./dist/cjs/index.cjs",
     "import": "./dist/esm/index.mjs"
   },
-  "typings": "dist/types/index.d.ts",
+  "types": "dist/types/index.d.ts",
   "license": "MIT",
   "homepage": "https://aurelia.io",
   "repository": {
--- a/packages-tooling/au/package.json
+++ b/packages-tooling/au/package.json
@@ -4,10 +4,11 @@
   "main": "dist/cjs/index.cjs",
   "module": "dist/esm/index.mjs",
   "exports": {
+    "types": "./dist/types/index.d.ts",
     "require": "./dist/cjs/index.cjs",
     "import": "./dist/esm/index.mjs"
   },
-  "typings": "dist/types/index.d.ts",
+  "types": "dist/types/index.d.ts",
   "license": "MIT",
   "homepage": "https://aurelia.io",
   "repository": {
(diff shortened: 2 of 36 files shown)


bigopon:
Thanks @josundt

PR #1661, opened 2023-01-25
fix(router-lite): routeTree adjustment - canUnload

Sayan751:
# Pull Request

## 📖 Description

This PR fixes an edge case (however, maybe not so edge case).

Previously, it was such that when a `canUnload` returned `false` for an attempt to load `some/path`, another repeat attempt to load the same `some/path` was no-operational. This was due to the fact that even before the `canUnload` is invoked, the route tree was updated. Therefore, on the next attempt, the children node for the current tree and the nodes from the (retry) instruction matched, which instructs the router not to process the instruction.

This commit fixes that by marking unsuccessful unload and thereby enabling processing of the same routing instruction.

### DIFF
--- a/packages/router-lite/src/router.ts
+++ b/packages/router-lite/src/router.ts
@@ -330,6 +330,8 @@ export class Router {
     return this._isNavigating;
   }
 
+  private _cannotBeUnloaded: boolean = false;
+
   public constructor(
     @IContainer private readonly container: IContainer,
     @IPlatform private readonly p: IPlatform,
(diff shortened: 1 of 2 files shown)


packages/router-lite/src/router.ts
```diff
@@ -330,6 +330,8 @@ export class Router {
     return this._isNavigating;
   }
 
+  private _cannotBeUnloaded: boolean = false;
```
bigopon, line 333:
```suggestion
  /** @internal */
  private _cannotBeUnloaded: boolean = false;
```

bigopon (approved):

PR #1664, opened 2023-02-01
fix(router-lite): lifecycle hooks were invoked twice

Sayan751:
# Pull Request

## 📖 Description

This PR fixes the issue of the lifecycle hooks being called twice when registered as dependencies of custom elements.
Previously, the component dependencies were registered to the container when creating route context. This has caused the issue of same hooks being called twice because the hooks were registered twice (the one from the `Controller.$el` should happen normally). This PR removes the 2nd registration of dependencies.

### DIFF
--- a/packages/kernel/src/di.container.ts
+++ b/packages/kernel/src/di.container.ts
@@ -373,27 +373,13 @@ export class Container implements IContainer {
     disposableResolvers.clear();
   }
 
-  public find<TType extends ResourceType, TDef extends ResourceDefinition>(kind: IResourceKind<TType, TDef>, name: string, searchAncestor: boolean): TDef | null {
+  public find<TType extends ResourceType, TDef extends ResourceDefinition>(kind: IResourceKind<TType, TDef>, name: string): TDef | null {
     const key = kind.keyFrom(name);
-    let resolver: IResolver | IDisposableResolver | undefined;
-    if (searchAncestor) {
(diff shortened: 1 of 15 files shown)


packages/router-lite/src/route-context.ts
```diff
@@ -166,7 +166,6 @@ export class RouteContext {
     );
 
     container.register(definition);
-    container.register(...component.dependencies);
```
bigopon, line 169: 👍

packages/kernel/src/di.container.ts
```diff
@@ -375,7 +375,7 @@ export class Container implements IContainer {
 
   public find<TType extends ResourceType, TDef extends ResourceDefinition>(kind: IResourceKind<TType, TDef>, name: string): TDef | null {
     const key = kind.keyFrom(name);
-    let resolver = this.res[key];
+    let  resolver = this.res[key];
```
bigopon, line 378: quite random addition :)

packages/router-lite/src/router.ts
```diff
@@ -330,6 +330,7 @@ export class Router {
     return this._isNavigating;
   }
 
+  /** @internal */
```
bigopon, line 333: 👍

packages/kernel/src/di.container.ts
Sayan751: Lol! Sorry. I changed that file earlier, and later decided against that reverted that changes. Missed it somehow.

PR #1668, opened 2023-02-04
We don't need to use the '.bind' juste set the value to the attribute…

altus34:
# Pull Request

## 📖 Description

Documentation seems not be in sync with feature.

### 🎫 Issues

The code sample in the doc seems not valid.
It not working for me until I removed the .bind on my attribute and I feel it's the desired feature for attribute binding.


Vheissu:
Thank you for your contribution. However, the documentation is correct. You need to use the bind attribute if you're binding a variable value (like in this example). You are correct that bind is not needed if it's a static value (number, string, etc). In this example, myEmail is a property defined inside a view model being bound.

altus34:
:thumbsup:

PR #1669, opened 2023-02-05
docs: router-lite

Sayan751:
# Pull Request

## 📖 Description

Documentation for router-lite.


docs/user-docs/router-lite/navigation-model.md
```diff
@@ -0,0 +1,129 @@
+---
+description: Create a navigation menu using navigation model in Router-Lite.
+---
+
+# Navigation model
+
+The navigation model can be thought of a "view" of the configured routes, in that sense that it lists the configured routes, without providing all the of the same information, as of route configuration, again and rather adding some additional information.
```
bigopon, line 7: If we are providing analogy, maybe it should be done clearer? What maps to view and what map to controller, maybe?

docs/user-docs/router-lite/router-hooks.md
```diff
@@ -0,0 +1,408 @@
...
+    AuthenticationHook,
+    // register the second lifecycle hook
+    AuthorizationHook
+  );
+  au.app({ host, component });
+  await au.start();
+})().catch(console.error);
+```
+
+Note that the authentication hook is registered first and then the authorization hook.
```
bigopon, line 225: 👍 Registration order doc is good, though if there's a way to emphasize this more, it'd probably be even better.

bigopon (approved):
Good work 👍

docs/user-docs/router-lite/router-hooks.md
Sayan751: I have rephrased and added links to the respective section. That should be better.

docs/user-docs/router-lite/navigation-model.md
Sayan751: You are right, the analogy was unclear. I was referring "SQL view" 😋. But I think that analogy is completely confusing in the context. I have rephrased.

PR #1670, opened 2023-02-06
Fix Promise binding sample code.

altus34:
# Pull Request

## 📖 Description

Find an issue in the code sample : Promise binding using functions

### 🎫 Issues

The code not working as expected

### DIFF
--- a/docs/user-docs/templates/template-syntax.md
+++ b/docs/user-docs/templates/template-syntax.md
@@ -384,8 +384,8 @@ Ignore the `i` variable being incremented, this is only there to make Aurelia fi
 {% tab title="my-app.ts" %}
 ```
 export class MyApp {
-  fetchAdvice() {
-    return fetch("https://api.adviceslip.com/advice")
+  fetchAdvice(id: number) {
+    return fetch("https://api.adviceslip.com/advice/" + id)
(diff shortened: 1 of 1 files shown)


altus34:
Sorry I see the note to have a pure function after but when I click on the button there is no advice change.

The method is well called but nothing change maybe an issue in the binding.

Sayan751:
@altus34 Thank you for your contribution. However, the docs are correct. Please refer this live example, created from the example code from the docs: https://stackblitz.com/edit/au2-promise-gh1670?file=src/my-app.ts

Sayan751:
I am not sure what you have meant, as I can see it being updated.

GH1670

altus34:
Yes you're right.
Working on Chrome but not on my Firefox 109.0.1  on Ubuntu  Linux 22.04

Regards

Sayan751:
Actually, the problem with FF is something else. The FF caches the response and uses the cached response every time. You can verify that by looking at the network tab in dev tools every time you click the try again button.

You can work around this issue by using a `{ cache: 'no-store' }` option for the `fetch`. Here is the working example also in FF: https://stackblitz.com/edit/au2-promise-gh1670?file=src%2Fmy-app.ts.

If you want, you can reopen this PR with relevant changes.

PR #1671, opened 2023-02-10
WIP: refactore dialog DOM creation

ekzobrain:
# Pull Request

## 📖 Description

The goal of this change is to refactore DialogService and DialogController to be more DOM-agnostic and allow flexible DefaultDialogDomRenderer overriding (for example create BootstrapDialogDomRenderer which renders dialog using Bootstrap's markup and native JS).
Changes include:

- DialogService does not listen and react to dialog's wrapper keydown event, moved to DefaultDialogDomRenderer
- DialogController does not listen and react to dialog's overlay click event, moved to DefaultDialogDomRenderer
- DialogDom is removed, it's functionality is moved to DefaultDialogDomRenderer
- keydown event is now listened on dialog's wrapper elements instead of Window (tabindex="-1" added to wrapper element to make it focusable and thus dispatch this event type). This allows not to warry about dialog hierarhy progmatically - the last open dialog (top) will always receive this event.

### DIFF
--- a/packages/dialog/src/plugins/dialog/dialog-controller.ts
+++ b/packages/dialog/src/plugins/dialog/dialog-controller.ts
@@ -4,7 +4,6 @@ import {
   DialogDeactivationStatuses,
   IDialogController,
   IDialogDomRenderer,
-  IDialogDom,
   DialogOpenResult,
   DialogCloseResult,
   DialogCancelError,
@@ -47,9 +46,9 @@ export class DialogController implements IDialogController {
   public readonly closed: Promise<DialogCloseResult>;
 
   /**
-   * The dom structure created to support the dialog associated with this controller
+   * The renderer used to create dialog DOM
    */
-  private dom!: IDialogDom;
+  private renderer!: IDialogDomRenderer;
 
   /**
    * The component controller associated with this dialog controller
@@ -76,14 +75,14 @@ export class DialogController implements IDialogController {
   public activate(settings: IDialogLoadedSettings): Promise<DialogOpenResult> {
     const container = this.ctn.createChild();
     const { model, template, rejectOnCancel } = settings;
-    const hostRenderer: IDialogDomRenderer = container.get(IDialogDomRenderer);
     const dialogTargetHost = settings.host ?? this.p.document.body;
-    const dom = this.dom = hostRenderer.render(dialogTargetHost, settings);
+    const renderer = container.get(IDialogDomRenderer);
+    const contentHost = renderer.render(dialogTargetHost, settings, this);
     const rootEventTarget = container.has(IEventTarget, true)
       ? container.get(IEventTarget) as Element
       : null;
-    const contentHost = dom.contentHost;
 
+    this.renderer = renderer;
     this.settings = settings;
     // application root host may be a different element with the dialog root host
     // example:
@@ -97,7 +96,7 @@ export class DialogController implements IDialogController {
 
     container.register(
       instanceRegistration(INode, contentHost),
-      instanceRegistration(IDialogDom, dom),
+      instanceRegistration(IDialogDomRenderer, renderer),
     );
 
     return new Promise(r => {
@@ -106,7 +105,7 @@ export class DialogController implements IDialogController {
       })
       .then(canActivate => {
         if (canActivate !== true) {
-          dom.dispose();
+          renderer.dispose();
           if (rejectOnCancel) {
             throw createDialogCancelError(null, 'Dialog activation rejected');
           }
@@ -126,12 +125,11 @@ export class DialogController implements IDialogController {
             )
           ) as ICustomElementController;
           return onResolve(ctrlr.activate(ctrlr, null, LifecycleFlags.fromBind), () => {
-            dom.overlay.addEventListener(settings.mouseEvent ?? 'click', this);
             return DialogOpenResult.create(false, this);
           });
         });
       }, e => {
-        dom.dispose();
+        renderer.dispose();
         throw e;
       });
   }
@@ -143,7 +141,7 @@ export class DialogController implements IDialogController {
     }
 
     let deactivating = true;
-    const { controller, dom, cmp, settings: { mouseEvent, rejectOnCancel }} = this;
+    const { controller, renderer, cmp, settings: { rejectOnCancel }} = this;
     const dialogResult = DialogCloseResult.create(status, value);
 
     const promise: Promise<DialogCloseResult<T>> = new Promise<DialogCloseResult<T>>(r => {
@@ -162,8 +160,7 @@ export class DialogController implements IDialogController {
           return onResolve(cmp.deactivate?.(dialogResult),
             () => onResolve(controller.deactivate(controller, null, LifecycleFlags.fromUnbind),
               () => {
-                dom.dispose();
-                dom.overlay.removeEventListener(mouseEvent ?? 'click', this);
+                renderer.dispose();
                 if (!rejectOnCancel && status !== DialogDeactivationStatuses.Error) {
                   this._resolve(dialogResult);
                 } else {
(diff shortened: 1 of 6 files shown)


Vheissu:
This is looking good so far @netcitylife can't wait to see this evolve. I think an important thing to note is for such a change we will require:

- Tests
- Doc changes

But, you're probably already planning on those. Just thought I would mention.

packages/dialog/src/plugins/dialog/dialog-interfaces.ts
```diff
@@ -45,17 +45,8 @@ export interface IDialogController {
...
-
-/**
- * An interface describing the DOM structure of a dialog
- */
-export const IDialogDom = createInterface<IDialogDom>('IDialogDom');
-export interface IDialogDom extends IDisposable {
-  readonly overlay: HTMLElement;
-  readonly contentHost: HTMLElement;
+export interface IDialogDomRenderer extends IDisposable {
+  render(dialogHost: Element, controller: IDialogController): HTMLElement;
```
bigopon, line 49: I'm a bit concerned by this interface changes. This change implies states are being brought into the renderer. Doing it this way make it questionable what's the relationship between rendered elements and the renderer. Are they tied to each other until disposed?
Previously renderer is just renderer, after rendering, nothing is needed from the renderer.

netcitylife: Yes, renderer is now transient, not singleton. It is merged together with DialogDom, which is now gone, because it reflected a concrete DOM structure

bigopon: > it reflected a concrete DOM structure

This DOM structure is quite standard for dialog though. Overlay + content host are commonly expected. Was this structure the issue leading to this PR?

netcitylife: Sure it is quite common, but most of the times we will not need access to this structure. Previously, as I understand, DialogDom was created for keeping a statefull reference to DOM elements for these purposes:
1. Allow DialogController to access DOM nodes to handle their events. This is not the case any more because it is now renderer responsibility.
2. Allow their styling (https://docs.aurelia.io/aurelia-packages/dialog#centering-uncentering-dialog-position) in default renderer implementation. This is a very specific approach to set container element styles directly from a dialog component, but it is still available for default implementation, you just need to inject IDialogDomRenderer instead of IDialogDom in dialog content component.

Now that renderer subscribes and unsubscribes from DOM events - it needs to be statefull, so I made it transient and merged with DialogDom.

Now consider that we need to implement a new renderer for Bootstrap modal (https://getbootstrap.com/docs/5.3/components/modal/#via-javascript). We need to create Bootstrap DOM structure and instantiate it's JS modal component (or use a wrapper custom element around it). So we need to keep some state between render and disposal, but that state would be much different than DialogDom was. And if we implement a renderer for Material Web Components dialog - it would be the third different state. So state depends heavily on underlying renderer implementation (library API) and keeping separate DialogDom with this concrete structure (state) does not make much sense, except for current default implementation.

netcitylife:
> This is looking good so far @netcitylife can't wait to see this evolve. I think an important thing to note is for such a change we will require:
> 
> 
> 
> - Tests
> 
> - Doc changes
> 
> 
> 
> But, you're probably already planning on those. Just thought I would mention.

Sure, when finished

packages/dialog/src/plugins/dialog/dialog-default-impl.ts
```diff
@@ -20,45 +16,93 @@ export class DefaultDialogGlobalSettings implements IDialogGlobalSettings {
...
-    singletonRegistration(IDialogDomRenderer, this).register(container);
+    transientRegistration(IDialogDomRenderer, this).register(container);
   }
 
-  private readonly wrapperCss: string = `${baseWrapperCss} display:flex;`;
-  private readonly overlayCss: string = baseWrapperCss;
-  private readonly hostCss: string = 'position:relative;margin:auto;';
-
-  public render(dialogHost: HTMLElement): IDialogDom {
+  public render(dialogHost: HTMLElement, controller: IDialogController): HTMLElement {
```
bigopon, line 42: Associating a parameter to some state in a private property this way is not appropriate. It's brittle in a way that this render call is now supposed to be called only once, but the interface doesn't communicate that. Overall, this is where my concerns are: it's unclear at what point what states are available if we make something named "renderer" a "lazily-stateful" service.

bigopon:
@netcitylife can you elaborate what would be the challenge extending the existing way of doing things for something like a bootstrap modal? `IDialogRenderer` accepts `controller` in its stateless `render` method, so there's no reason a custom renderer cannot create a custom `IDialogDom` that does more than the default one.
```ts
class MyRenderer {
  render(_, controller) {
    const dom = new BootstrapDialogDom(controller);
  }
}

class BootstrapDialogDom {
  constructor(private controller) {
    // attach listener here and close controller when necessary.
  } 
}
```

packages/dialog/src/plugins/dialog/dialog-default-impl.ts
netcitylife: I agree that renderer interface should be improved. This change was just the first step, showing the main goal - incapsulation of DOM manipulation inside renderer or some DialogDom instance, cleanup of DialogService and dialogController. Now I plan to create several POC renderers to see what states and renderer capabilities would they need to integrate seemlessly into current dialog lifecycle.

netcitylife:
> @netcitylife can you elaborate what would be the challenge extending the existing way of doing things for something like a bootstrap modal? `IDialogRenderer` accepts `controller` in its stateless `render` method, so there's no reason a custom renderer cannot create a custom `IDialogDom` that does more than the default one.
> 
> ```ts
> class MyRenderer {
>   render(_, controller) {
>     const dom = new BootstrapDialogDom(controller);
>   }
> }
> 
> class BootstrapDialogDom {
>   constructor(private controller) {
>     // attach listener here and close controller when necessary.
>   } 
> }
> ```

In this approach I see no need in a renderer at all, it does nothing but instantiating BootstrapDialogDom, which will create DOM in constructor() and then destroy it in dispose(). So it may be just called directly from controller. I see your confusion about current renderer interface, see my comment above.

bigopon:
> ...no need in a renderer at all, it does nothing but instantiating...

We are doing registration for plug & play so that custom renderer can be used. It's certainly a useful layer to have for customization.

Can you also help tackle the other parts in my comment, about (a) the challenges of extending the existing one, and (b) why wouldn't you be able to do it with just different `IDialogDom` above?

ekzobrain:
> > ...no need in a renderer at all, it does nothing but instantiating...
> 
> We are doing registration for plug & play so that custom renderer can be used. It's certainly a useful layer to have for customization.
> 
> Can you also help tackle the other parts in my comment, about (a) the challenges of extending the existing one, and (b) why wouldn't you be able to do it with just different `IDialogDom` above?

See my comment in Discord, please

packages/dialog/src/plugins/dialog/dialog-controller.ts
```diff
@@ -125,14 +128,12 @@ export class DialogController implements IDialogController {
               this.getDefinition(cmp) ?? { name: CustomElement.generateName(), template }
             )
           ) as ICustomElementController;
-          return onResolve(ctrlr.activate(ctrlr, null, LifecycleFlags.fromBind), () => {
-            dom.overlay.addEventListener(settings.mouseEvent ?? 'click', this);
-            return DialogOpenResult.create(false, this);
+          return onResolve(renderer.render(controller), () => {
```
bigopon, line 131: if it's desirable, we can add custom lifecycle hooks the way we have for router. Just need to figure out an acceptable API. We shouldn't  be changing the renderer just to accommodate this. Rendering process should be simple, idempotent and synchronous.

PR #1682, opened 2023-02-23
chore(router-lite): tests and docs for historyStrategy as nav opt

Sayan751:
# Pull Request


bigopon (approved):
looks good to me, though maybe also describe what happens in the case of multiple viewports

Sayan751:
@bigopon Thank you for the review! 

I think the history strategy is quite unrelated to multiple viewports. If there is a history/location events like `popstate` or `hashchange`, the event will be handled, irrespective of how many viewports are there, or if the viewports are arranged hierarchically or as siblings. The URL associated with the history state is enough to reload the components into correct viewports, because the URL should contain the viewport information/instruction. Let me know if I am missing something obvious, or if you intend to point to something specifically.

I am merging this PR. In case you have follow-up comments, I will take that into account for the next PR.

PR #1705, opened 2023-03-08
fix(router-lite): removed pre-mature optimization

### LINKED ISSUE #1701: navigate to parent route from not default child route
# 🐛 Bug Report

```
export class Parent
{
  getRouteConfig() {
    return {
      routes: [
        { path: '', component: DefaultChild },
        { path: 'other-child', component: OtherChild },
      ],
    };
  }
}
```

If `other-child` is the active route, routing to `parent` is not working. I would expect the empty route `default-child` to get loaded again. If this is by design (I think this would be different from the behaviour in Au1), is there a way to force this behaviour? 

repro: https://stackblitz.com/edit/au2-conventions-fdgtod

Sayan751:
# Pull Request

## 📖 Description

The `shouldProcess`-logic caused more problem than solving. Hence, it is removed completely, instead of adding one more condition to the ladder and thereby making it even more flimsy.

### 🎫 Issues

Fixes #1701

### DIFF
--- a/packages/__tests__/router-lite/smoke-tests.spec.ts
+++ b/packages/__tests__/router-lite/smoke-tests.spec.ts
@@ -5012,4 +5012,64 @@ describe('router-lite/smoke-tests.spec.ts', function () {
       await au.stop();
     });
   });
+
+  it('navigate repeatedly to parent route from child route works - GH 1701', async function () {
+    @customElement({ name: 'c-1', template: 'c1 <a load="../c2"></a>' })
+    class C1 { }
+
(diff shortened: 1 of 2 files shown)


bigopon (approved):
Looks good. While the optimization failed, the idea might not. I think better leave the note there what can be optimized around that areas and come back for another fight later.

Sayan751:
> Looks good. While the optimization failed, the idea might not. I think better leave the note there what can be optimized around that areas and come back for another fight later.

@bigopon Thanks for the review! Added a comment there.

PR #1706, opened 2023-03-10
Update navigating.md

ivanbacher:
`queryParams` should be `parameters` in [Specifying load options](https://docs.aurelia.io/routing/navigating#specifying-load-options) section.


bigopon:
Thanks @ivanbacher

PR #1707, opened 2023-03-11
refactor: remove lifecycle flags

bigopon:
Lifecycle flags have been gradually removed from many places. This is the last step of removing them completely. This involves removing the dispose flag, which is used during deactivation. Previously deactivation can also signal dispose, but now, the dispose call should come separately and manually rather than packed inside deactivation:
Before
```ts
controller.deactivate(..., flags: LifecycleFlags.dispose)
```
After
```ts
controller.deactivate(...)
controller.dispose()
```


packages/router-lite/src/viewport-agent.ts
```diff
@@ -565,14 +564,17 @@ export class ViewportAgent {
         Batch.start(b1 => {
           tr.run(() => {
             b1.push();
-            return curCA.deactivate(null, controller, LifecycleFlags.dispose);
+            // return curCA.deactivate(null, controller, true);
+            return onResolve(curCA.deactivate(null, controller), () => {
+              curCA.dispose();
+            });
           }, () => {
```
bigopon, line 571: @Sayan751 can you help verify this change?

Sayan751: Seems logical

packages/router-lite/src/viewport-agent.ts
```diff
@@ -444,7 +444,7 @@ export class ViewportAgent {
           case 'replace': {
             const controller = this.hostController;
             tr.run(() => {
-              return this.curCA!.deactivate(initiator, controller, LifecycleFlags.dispose);
+              return this.curCA!.deactivate(initiator, controller);
```
Sayan751, line 447: We probably also want to dispose here.

packages/router-lite/src/viewport-agent.ts
```diff
@@ -114,7 +114,7 @@ export class ViewportAgent {
         return;
       case State.currIsActive:
         this.logger.trace(`deactivateFromViewport() - deactivating existing componentAgent at %s`, this);
-        return this.curCA!.deactivate(initiator, parent, flags);
+        return this.curCA!.deactivate(initiator, parent);
```
Sayan751, line 117: And here as well.

PR #1708, opened 2023-03-15
feat(proxy): add nowrap decorator

### LINKED ISSUE #1614: i18n interpolation causes infinite getter calls because of proxy on getter
# 🐛 Bug Report

When i18n has a interpolation and used in getters it calls resetRegex forcing the proxy to invoke the getters on every call.

## 🌍 Your Environment

| Aurelia          |
| Language         |
| Browser          |
| Bundler          |
| Operating System |
| NPM/Node/Yarn    |

bigopon:
feat(i18n): do not wrap i18next

resolves #1614

When there's a large 3rd party library object, together with many of its internal properties, being part of the dependencies of a getter observer, it can cause a lot of change triggers because whatever the internal working way of the 3rd party lib.
Add a decorator `@nowrap` so that we can have a way to draw a boundary where we don't want to go into to collect dependencies.

The decorator is also allowed to work with inheritance, means a property decorated with `@nowrap` on the parent class will also have the same effect in the child class.

@brandonseydel @Sayan751 @fkleuver

### DIFF
--- a/packages/__tests__/2-runtime/proxy-observable.spec.ts
+++ b/packages/__tests__/2-runtime/proxy-observable.spec.ts
@@ -71,7 +71,7 @@ describe('2-runtime/proxy-observable.spec.ts', function () {
     // eslint-disable-next-line @typescript-eslint/no-unused-expressions
     m.i18n;
     ConnectableSwitcher.exit(connectable);
-    assert.strictEqual(count, 1);
+    assert.strictEqual(count, 0);
   });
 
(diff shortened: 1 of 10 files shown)


bigopon:
We'll go ahead with this `@nowrap` name for now. We can consider other names later if we can see one fits better. Example list
```
@unobserved
@noProxy
@unProxy
@wrap(false)
@proxy(false)
```

PR #1714, opened 2023-03-29
fix(router-lite): multiple configurations for same component

Sayan751:
# Pull Request

### 🎫 Issues

Previously, the support for configuration such as the following was broken.

```ts
@route({
  routes: [
    { path: '', component: C1, title: 't1', data: { foo: 'bar' } },
    { path: 'c1', component: C1, title: 't2', data: { awesome: 'possum' } },
  ],
})
```

This PR fixes that. 

Additionally, it fixes the component disposal issue, caused by the PR #1707

## 👩‍💻 Reviewer Notes

Check the smoke tests for more examples.

## ⏭ Next Steps

- Release new version.
- Add new examples to the docs.

PR #1715, opened 2023-03-30
chore(router-lite): residual changes from #1714

Sayan751:
# Pull Request

### DIFF
--- a/packages/router-lite/src/route.ts
+++ b/packages/router-lite/src/route.ts
@@ -1,5 +1,5 @@
 import { Metadata } from '@aurelia/metadata';
-import { Constructable, emptyArray, onResolve, Protocol, ResourceType, Writable } from '@aurelia/kernel';
+import { Constructable, emptyArray, onResolve, Protocol, ResourceType, Writable, IContainer } from '@aurelia/kernel';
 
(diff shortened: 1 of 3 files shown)


packages/router-lite/src/route.ts
```diff
@@ -281,6 +281,12 @@ export class RouteConfig implements IRouteConfig, IChildRouteConfig {
       ? fallback(viewportInstruction, routeNode, context)
       : fallback;
   }
+
+  public register(container: IContainer): void {
+    const component= this.component;
+    if(component == null) return;
+    container.register(component);
```
bigopon, line 288: is this safe? if component is an vm instance, for example. The container will do unexpected thing in this case.

Sayan751: It is safe. Because when a RouteConfig is instantiated, the class is always resolved, before instantiating it. Only for a redirection configuration it is null. Which is ignored here anyway.

bigopon: That's a lot of implication to carry around when one reads the code here. Maybe we should put a note or something, at least to express how internal this `register` is.

Sayan751: Good point 👍 Will put a note.

PR #1718, opened 2023-03-31
feat(router-lite): transitionplan as nav opt

Sayan751:
# Pull Request

## 📖 Description

This is reported here: https://discourse.aurelia.io/t/router-lite-navigate-to-optional-param-within-same-route/5115

cc: @elitastic 

## ⏭ Next Steps

Docs and demo.


Sayan751:
Docs have been added in #1719.

PR #1729, opened 2023-04-05
feat(templating): allow deactivate when activating

jwx:
# Pull Request

## 📖 Description

This PR makes it possible to call deactivate on a controller while it's activating. The controller will complete the activation step it's on and then deactivate through the steps it had done so far while activating.

## 📑 Test Plan

- Create passing tests for the issue

## ⏭ Next Steps

- Remove the errors in repeaters and similar that now works


Sayan751:
For such a large change in a core component like the controller, we need some supporting tests.

Sayan751:
@bigopon @jwx Can you guys please hold on this PR? There might be an easier way to deal with this. I will try to submit a PR in the next few days.

bigopon:
Thanks @jwx @Sayan751 nice work!

PR #1735, opened 2023-04-09
feat(slotted): add slotted decorator WIP

### LINKED ISSUE #1694: @children decorator is not described in docs
There is not even a mention of it

bigopon:
# Pull Request

## 📖 Description

At the moment, there's not a way to listen to changes of the projection within an `<au-slot/>` element. This leads to the inability to have similar pattern like the `@children` deco for `<au-slot/>`, and sometimes one needs to resolve to some complex query with `@children` to achieve the same outcome.

This PR:
- Adds the capability for `<au-slot/>` to watch, capture and notify changes of its projection.
- Adds doc for `@children` decorator
- Adds doc for `@slotted` decorator

Resolves #1694

## 👩‍💻 Reviewer Notes

- `@slotted` decorator will add a `ILifecycleHooks` dependency to the metadata associated with a class, and upon resolving, will provide the infrastructure to hook with `<au-slot/>` for change observation and notifcation

## 📑 Test Plan

- initial rendering
- initial rendering of multiple slots that have elements matching the `@slotted` query
- mutation of a single slot
- mutation of multiple slots that have elements matching the `@slotted` query
- mutiple `@slotted` working together

@Sayan751 @fkleuver @brandonseydel 

Doc can be temporarily viewed here

https://github.com/aurelia/aurelia/blob/feat/slotted-deco/docs/user-docs/components/shadow-dom-and-slots.md#listening-to-slot-change

### DIFF
--- a/packages/__tests__/3-runtime-html/au-slot.slotted.spec.ts
+++ b/packages/__tests__/3-runtime-html/au-slot.slotted.spec.ts
@@ -1,0 +1,73 @@
+import { customElement, slotted } from '@aurelia/runtime-html';
+import { assert, createFixture } from '@aurelia/testing';
+
+describe('3-runtime-html/au-slot.slotted.spec.ts', function () {
+  describe('intitial rendering', function () {
+    it('assigns value', async function () {
+      @customElement({
+        name: 'el',
+        template: '<au-slot>'
+      })
+      class El {
+        @slotted('div') divs;
+
+      }
+
+      const { component: { el } } = createFixture(
+        '<el view-model.ref=el><div></div><div></div>',
+        class App {
+          el: El;
+        },
+        [El,]
+      );
+
+      assert.strictEqual(el.divs.length, 2);
+    });
+
+    it('calls change handler', async function () {
+      let call = 0;
+      @customElement({
+        name: 'el',
+        template: '<au-slot>'
+      })
+      class El {
+        @slotted('div') divs;
+
+        divsChanged() {
+          call = 1;
+        }
+      }
+
+      createFixture(
+        '<el view-model.ref=el><div></div>',
+        class App { },
+        [El,]
+      );
+
+      assert.strictEqual(call, 1);
+    });
+
+    it('does not call change handler there are no matching nodes', function () {
+      let call = 0;
+      @customElement({
+        name: 'el',
+        template: '<au-slot>'
+      })
+      class El {
+        @slotted('div') divs;
+
+        divsChanged() {
+          call = 1;
+        }
+      }
+
+      createFixture(
+        '<el view-model.ref=el><input>',
+        class App { },
+        [El,]
+      );
+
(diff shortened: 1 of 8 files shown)


packages/runtime-html/src/resources/custom-elements/au-slot.ts
```diff
@@ -53,9 +59,38 @@ export class AuSlot implements ICustomElementViewModel {
...
+      if (curr.nodeType !== /* comment */8) {
+        nodes.push(curr);
+      }
+      curr = curr.nextSibling;
+    }
+    return nodes;
+  }
+
+  /** @internal */
+  private readonly _subs = new Set<ISlotSubscriber>();
```
Sayan751, line 84: This can be `WeakSet`.

packages/runtime-html/src/resources/custom-elements/au-slot.ts
```diff
@@ -53,9 +59,38 @@ export class AuSlot implements ICustomElementViewModel {
...
+
+  public subscribe(subscriber: ISlotSubscriber): void {
+    if (!this._subs.has(subscriber) && this._subs.add(subscriber).size === 1) {
```
Sayan751, line 87: Why not put the `.add` inside the if block?

packages/__tests__/3-runtime-html/au-slot.slotted.spec.ts
```diff
@@ -0,0 +1,73 @@
+import { customElement, slotted } from '@aurelia/runtime-html';
+import { assert, createFixture } from '@aurelia/testing';
+
+describe('3-runtime-html/au-slot.slotted.spec.ts', function () {
+  describe('intitial rendering', function () {
```
Sayan751, line 5: I think we need further tests
- CE with multiple slots
- CE with multiple `@slotted` decorators; with and without mutually exclusive or overlapping query.

Sayan751 (commented):
@bigopon Great changes! Left some comments regarding some small stuffs and tests.

packages/runtime-html/src/resources/custom-elements/au-slot.ts
bigopon: WIP, left over of some undecided implementation

packages/runtime-html/src/resources/custom-elements/au-slot.ts
bigopon: we need to iterate the subscribers to notify later, so it cannot be weakset

packages/__tests__/3-runtime-html/au-slot.slotted.spec.ts
bigopon: Thanks, for the tests, we can probably omit "CE", since it must be a CE anyway. For the overlapping, nice suggestion.

PR #1739, opened 2023-04-12
feat(debounce-throttle): flush via signals

### LINKED ISSUE #1145: add override option for `debounce` and `throttle`
[the example references `debounce` - but everything is true for `throttle` as well]

consider the following scenario:

a search input, that is bounded to some property in the VM.

typing into the input need to initiate some kind of search from the backend in "real time".
hitting "Enter" in this same input initiate another method that also need the value of the input.

we typically don't want to initiate the search immediately after each keystroke, so we actually use the `debounce` binding behavior (`bb` from now on).
this will cause the binding to refresh the VM only after some amount of time has passed without the user typing anything new.
barely visible to the human eye - but helps a lot with performance.

but now we have a problem: if a user types really fast and hits "Enter" - before the binding had a chance to update the value.
the value of the input in the VM is still `undefined` (or whatever the last old value was) instead of the "real" text the user typed.

we need to allow some kind of override for the `debounce` bb, that will flush the value immediately.
for AU1, I've created a copy of the `debounce` bb, that "flushes" the value whenever "Enter" was pressed.
and I was thinking of making the same fix for AU2.

but this time, I think this should be baked in original bb itself (and controlled via some kind of smart option), instead of being another copy and almost similar bb.

# 🙋 Feature Request
enable some sort of override for `debounce` and `throttle` binding behaviors - that will flush the value immidiattly.

## 🤔 Expected Behavior

* default: default timeout. no overrides.
```<input value.bind="myValue & debounce" >```

* provided timeout. no overrides.
```<input value.bind="myValue & debounce:200" >```

* provided timeout. with one override.
```<input value.bind="myValue & debounce:200:'Enter'" >```

* provided timeout. with more overrides.
```<input value.bind="myValue & debounce:200:'Enter':'Space'" >```

this is not a good syntax. I would love to hear other thoughts and ideas regarding this.

**EDIT: new syntax based on @zewa666 comments.**

* default: default delay. no flush overrides.
```<input value.bind="myValue & debounce" >```

* provided delay. no fulsh overrides.
```<input value.bind="myValue & debounce:200" >```

* default delay. with a single flush override.
```<input value.bind="myValue & debounce:'Enter'" >```

* default delay. with an array of flush overrides.
```<input value.bind="myValue & debounce:['Enter', 'Space']" >```

* complex option. (where you can specify delay, overrides (single or array), and more if needed..)
```<input value.bind="myValue & debounce:{delay: 200, flushOn: 'Enter'}" >```
```<input value.bind="myValue & debounce:{delay: 200, flushOn: ['Enter', 'Space']}" >```
```<input value.bind="myValue & debounce:{flushOn: ['Enter', 'Space']}" >```
```<input value.bind="myValue & debounce:{delay: 200}" >```
etc..

## 😯 Current Behavior
there is no way to express this kind of behavior.

## 💁 Possible Solution

1. reading the value from the DOM instead of the binding if we want to be sure to have the "latest" value.
2. waiting the amount of time specified by the `debounce` before each operation regarding the value. (not really a solution: because the user might keep on typing - further delaying the binding flush)

### PR DESCRIPTION
**Pull request by bigopon**
# Pull Request

## 📖 Description

Support a valid use case described in #1145 , where it's desirable to flush throttled/debounced update upon some condition.

- `throttle` & `debounce` now accept a 2nd parameter with type `string | string[]` as signals, to listen to and flush whenever there's a dispatch. Example:

```html
<input value.bind="value & throttle :200 :`finishTyping`" blur.trigger="signaler.dispatchSignal('finishTyping')">
<!-- or it can be a list of signals -->
<input value.bind="value & throttle :200 :[`finishTyping`, `newUpdate`]">
```

Resolves #1145

Thanks @avrahamcool

### DIFF
--- a/docs/user-docs/components/shadow-dom-and-slots.md
+++ b/docs/user-docs/components/shadow-dom-and-slots.md
@@ -85,6 +85,7 @@ The `<slot>` element comes with an event based way to listen to its changes. Thi
 <slot slotchange.trigger="handleSlotChange($event.target.assignedNodes())"></slot>
 ```
 {% endcode %}
+
 {% code title="my-app.ts overflow="wrap" lineNumbers="true" %}
 ```typescript
 class MyApp {
@@ -126,8 +127,9 @@ the property being decorated, example: `divs` -> `divsChanged`
 | `@children('div') prop` | Observe mutation, and select only `div` elements |
 
 {% hint style="info" %}
+
 Note: the `@children` decorator wont update if the children of a slotted node change — only if you change (e.g. add or delete) the actual nodes themselves.
-{% %}
+{% endhint %}
 
 ## Au-slot
 
--- a/docs/user-docs/templates/binding-behaviors.md
+++ b/docs/user-docs/templates/binding-behaviors.md
@@ -42,6 +42,19 @@ The throttle behavior is particularly useful when binding events to methods on y
 <div mousemove.delegate="mouseMove($event) & throttle"></div>
 ```
 
+### Flush pending throttled calls
+
+Sometimes it's desirable to forcefully run the throttled update, so that the application syncs the latest values. This can happen in a form, when a user previously was typing into a throttled form field, and hit tab key to go to the next field, as an example.
+The `throttle` binding behavior supports this scenario via signal. These signals can be added via the 2nd parameter, like the following example:
+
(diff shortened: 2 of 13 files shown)


bigopon:
cc @zewa666

PR #1745, opened 2023-04-18
chore: add pure annotation for better tree shakability

bigopon:
## 📖 Description

Add `/*@__PURE__*/` annotation to all non native function calls so that application terser/minifier can drop unused imports from Aurelia modules.


brandonseydel:
Any savings observed?

bigopon:
Hello world saved another 4KB minified. The resources are not shaked fully due to side effectful calls, and helloworld is still not minimal yet. Woll need to tune more.

PR #1748, opened 2023-04-22
feat(di): property injection with `resolve`

### LINKED ISSUE #1742: RFC - Field level injection via added get function to the inject
# 💬 RFC

While injecting via constructor is great, there is a problem when base classing or trying to use in generic functions to construct a result.

## 🔦 Context

Ctor is getting messy when injecting lots of parameters, hard to get an instance from the container in a generic function, and not having to call super with the parent injections would be beneficial for lib auths.

## 💻 Examples

```ts
class AbcNew {
  a = inject.get(A);
  b = inject.get(B);
}

class AbcOld {
  constructor(private readonly a:A, private readonly b:B,}{}
}


// new
class Def extends AbcNew {
 c = inject.get(C);
}

// old
class Def extends AbcOld {
  constructor(a:A, b:B, private readonly c:C}{
  super(a,b);
  }
}

class CrazyStuff {
  a = doSomethingCrazy();
}


doSomethingCrazy(){
const service= inject.get(Service);
const result = service.findAKey('sdfsd');
return inject.get(result);
}
```

@zewa666 @bigopon

bigopon:
## 📖 Description

Resolves #1742 

### 🎫 Issues

Add an export `resolve` to support the use case described in #1742. So now it's simpler to do this

Before:
```ts
abstract class Base {
  constructor(obj1: SomeType, obj2: SomeType2) {}
}

@inject(SomeType, SomeType2)
class MyElement extends Base {
  constructor(obj1: SomeType, obj2: SomeType2) {
    super(obj1, obj2)
  }
}
```

After:
```ts
import { resolve } from 'aurelia';

abstract class Base {
  obj1 = resolve(SomeType);
  obj2 = resolve(SomeType2);
}

class MyElement extends Base {}
```

It's also possible to call `resolve` with multiple keys at once:
```ts
const [abc, all_def_instances, someOther] = resolve(Abc, all(Def), optional(SomeOtherClass));
```

### DIFF
--- a/packages/kernel/src/di.container.ts
+++ b/packages/kernel/src/di.container.ts
@@ -1,16 +1,43 @@
+/* eslint-disable @typescript-eslint/no-this-alias */
 /* eslint-disable @typescript-eslint/strict-boolean-expressions, @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-explicit-any */
 import { isObject } from '@aurelia/metadata';
 import { isNativeFunction } from './functions';
-import { Class, Constructable, IDisposable } from './interfaces';
+import { type Class, type Constructable, type IDisposable } from './interfaces';
 import { emptyArray } from './platform';
-import { IResourceKind, ResourceDefinition, ResourceType, getAllResources, hasResources } from './resource';
+import { type IResourceKind, type ResourceDefinition, type ResourceType, getAllResources, hasResources } from './resource';
 import { createError, createObject, getOwnMetadata, isFunction, isString, safeString } from './utilities';
-import { IContainer, type Key, type IResolver, type IDisposableResolver, Factory, ContainerConfiguration, type IRegistry, Registration, Resolver, ResolverStrategy, Transformer, type RegisterSelf, type Resolved, getDependencies, containerGetKey, IFactory, IContainerConfiguration } from './di';
+import {
+  IContainer,
+  type Key,
+  type IResolver,
+  type IDisposableResolver,
+  ContainerConfiguration,
+  type IRegistry,
+  Registration,
+  Resolver,
+  ResolverStrategy,
+  type Transformer,
+  type RegisterSelf,
+  type Resolved,
+  getDependencies,
+  type IFactory,
+  type IContainerConfiguration,
+  type IFactoryResolver,
+  type ILazyResolver,
+  type INewInstanceResolver,
+  type IResolvedFactory,
+  type IResolvedLazy,
+  type IAllResolver,
+  type IOptionalResolver
+} from './di';
 
 const InstrinsicTypeNames = new Set<string>('Array ArrayBuffer Boolean DataView Date Error EvalError Float32Array Float64Array Function Int8Array Int16Array Int32Array Map Number Object Promise RangeError ReferenceError RegExp Set SharedArrayBuffer String SyntaxError TypeError Uint8Array Uint8ClampedArray Uint16Array Uint32Array URIError WeakMap WeakSet'.split(' '));
 // const factoryKey = 'di:factory';
 // const factoryAnnotationKey = Protocol.annotation.keyFor(factoryKey);
 let containerId = 0;
+
+let currentContainer: IContainer | null = null;
+
 /** @internal */
 export class Container implements IContainer {
   public readonly id: number = ++containerId;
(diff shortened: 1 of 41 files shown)


packages/kernel/src/di.container.ts
```diff
@@ -481,13 +575,69 @@ function validateKey(key: any): void {
...
+  if (currentContainer == null) {
+    throw createInvalidResolveCallError();
+  }
+  return keys.length === 1
+    ? currentContainer.get(keys[0])
+    : keys.map(containerGetKey, currentContainer);
+}
+
+/* eslint-disable @typescript-eslint/no-unused-vars, @typescript-eslint/ban-ts-comment, prefer-const */
+function testResolve() {
```
Sayan751, line 616: Is this required? Or just an overlooked cleanup? If this is really a test, consider moving it to the tests.

packages/kernel/src/di.ts
```diff
@@ -557,38 +559,45 @@ export function singleton<T extends Constructable>(targetOrOptions?: (T & Partia
...
 };
 
-export const all = /*@__PURE__*/createAllResolver(
-  (key: any,
-    handler: IContainer,
-    requestor: IContainer,
-    searchAncestors: boolean
-  ) => requestor.getAll(key, searchAncestors)
-);
+function testAll() {
```
Sayan751, line 582: Is this required? Or just an overlooked cleanup? If this is really a test, consider moving it to the tests.

packages/kernel/src/di.ts
```diff
@@ -619,14 +628,36 @@ export const all = /*@__PURE__*/createAllResolver(
...
-export type ILazyResolver<K = any> = IResolver<K>
+}) as <K extends Key>(key: K) => ILazyResolver<K>;
+export type ILazyResolver<K extends Key = Key> = IResolver<() => K>
   // type only hack
   & { __isLazy: undefined }
   // any is needed for decorator usages
   & ((...args: unknown[]) => any);
 export type IResolvedLazy<K> = () => Resolved<K>;
 
+function testLazy() {
```
Sayan751, line 639: Is this required? Or just an overlooked cleanup? If this is really a test, consider moving it to the tests.

packages/kernel/src/di.ts
```diff
@@ -655,7 +686,40 @@ export const optional = /*@__PURE__*/createResolver((key: Key, handler: IContain
...
-});
+}) as <K extends Key>(key: K) => IOptionalResolver<K>;
+export type IOptionalResolver<K extends Key = Key> = IResolver<K | undefined> & {
+  // type only hack
+  __isOptional: undefined;
+  // any is needed for decorator usages
+  (...args: unknown[]): any;
+};
+
+function testOptional() {
```
Sayan751, line 697: Is this required? Or just an overlooked cleanup? If this is really a test, consider moving it to the tests.

packages/runtime-html/src/dom.ts
```diff
@@ -509,3 +509,17 @@ export interface IHistory extends History {
    */
   replaceState(state: {} | null, title: string, url?: string | null): void;
 }
+
```
Sayan751, line 512: Good cleanup!

Sayan751 (commented):
Great work @bigopon! Just a bit curious about all those `test*` functions.

packages/kernel/src/di.container.ts
bigopon: This is to test the typing of our exports without having to have a dedicated place for it. Example is around the built-in resolver: `all`/`lazy`/`factory`. The reason is because they need to give the right resolved types when called with `container.get`, and `resolve`, as well as not yield any errors when used as decorators. These functions will be eliminated during the build as well. This is a good and simple way to have TS check on these complex types.

Sayan751: That's alright. I am not objecting to it. Rather, suggesting pushing it to tests, if the existing tests do not sufficiently cover this. And if they do cover, then these `test*` methods can be removed.

bigopon: Its a bit hard to test the api in the test project. Since this is just API design test, not runtime behavior test, having the test code next to the code is really beneficial. We have this in multiple decorator files. Though if we want to organise differently, one way we could do is to have separate test files, but right inside this project rather than test, so it doesnt pollute the normal code.

PR #1751, opened 2023-04-24
fix(plugin-conventions): ensure esm cjs compat

bigopon:
## 🎫 Issues

At the moment, the dist of the `plugin-conventions` package does not work with mjs in node, as it relies on some package that does not have an esm dist (e.g `modify-code`). This results in runtime error that looks like this
image

One way of working around this is to rename the config file to either:
```
.ts -> .cts
.js -> .cjs
```
But it's inconvenience, and not satisfying.

Tweak the import code of plugin conventions so that it works with both mode of import in node without having to worry about changing extension in application config files.

## 📑 Test Plan

All existing hmr tests work with changing `.config.js` to either `.config.mjs` or `.config.cjs` or

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,3 +34,4 @@ test/benchmarking-apps/runner/cosmos.config.json
 **/webdriver-ts/traces
 
 .turbo
+.parcel-cache
--- a/package-lock.json
+++ b/package-lock.json
@@ -2256,10 +2256,14 @@
     },
     "node_modules/@lezer/common": {
       "version": "0.15.12",
+      "resolved": "https://registry.npmjs.org/@lezer/common/-/common-0.15.12.tgz",
+      "integrity": "sha512-edfwCxNLnzq5pBA/yaIhwJ3U3Kz8VAUOTRg0hhxaizaI1N+qxV7EXDv/kLCkLeq2RzSFvxexlaj5Mzfn2kY0Ig==",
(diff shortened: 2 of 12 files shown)


3cp:
Finally enable esModuleInterop :)

Easier path is to update modify-code to provide both esm and cjs dist files. So you don't need some strange patch here.

I will update modify-code.

bigopon:
@3cp oh well, cant be stubborn forever if you dont have the energy.

For `modify-code`, maybe we can try `createRequire` too?
```ts
import { createRequire } from 'module';
import $modifyCode from 'modify-code'

const modifyCode = createRequire(import.meta.url)('modify-code') as typeof $modifyCode;
```

3cp:
modify-code v2.1.2 is released with a simple esm wrapper (following nodejs doc recommandation).

However I had to make small adjustment because nodejs ignores the long established convention `__esModule`.

https://github.com/dumberjs/modify-code/commit/a32ed8eed63878e0591cc5202683db88ab9c9795

I tried tsc setup but tsc doesn't emit `import ... from "./local-file.mjs"` with the file extension I wanted. It also doesn't compile to cjs/mjs dual files easily. I gave it up.

The wrapper didn't affect cjs consumption as expected. For esm consumption, I only ran some simple test, it seems fine. You can try this new version. It should give you correct default export.

PR #1760, opened 2023-05-03
feat(runtime-html): deactivate while activating

Sayan751:
# Pull Request

## 📖 Description

This should supersede the PR #1729.

## 👩‍💻 Reviewer Notes

- Take a look at the tests for the use-cases.
- The fix seems to work for the router-lite error recovery in my branch.


bigopon (commented):
Nice one @Sayan751. This covers case where we are guaranteed to exit during a phase. Will it work when there' no such guarantee? Example is when there's a long running promise returned in attaching and deactivate is call while that promise is still running? If it doesn't do that, we probably cant supersede #1729.

packages/__tests__/3-runtime-html/controller.error-recovery.spec.ts
```diff
@@ -0,0 +1,330 @@
...
+    }
+    const rootVm = au.root.controller.viewModel as TAppRoot;
+    return { host, au, container, rootVm, startUpError };
+  }
+
+  for (const hook of ['binding', 'bound', 'attaching', 'attached'] as const) {
+    it(`Aurelia instance with error on ${hook} can be deactivated  - root`, async function () {
+      @customElement({ name: 'ro-ot', template: '' })
+      class Root extends TestVM {
+        public [`$${hook}`](_initiator: IHydratedController, _parent: IHydratedController): void | Promise<void> {
```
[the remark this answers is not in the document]
Sayan751: No. Because I am here overriding the `$[HOOK]` method.

packages/runtime-html/src/templating/controller.ts
```diff
@@ -812,7 +812,7 @@ export class Controller<C extends IViewModel = IViewModel> implements IControlle
       ret = resolveAll(...this._lifecycleHooks!.detaching.map(callDetachingHook, this));
     }
 
-    if (this._hooks.hasDetaching) {
+    if (this._hooks.hasDetaching && this.isBound) {
```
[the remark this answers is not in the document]
Sayan751: Not necessarily, but I have missed tests involving the lifecycle hooks. I will try to add those.

Sayan751:
> Example is when there's a long running promise returned in attaching and deactivate is call while that promise is still running? 

Good point @bigopon! I will add the tests. Thanks for the review.

jwx:
> Nice one @Sayan751. This covers case where we are guaranteed to exit during a phase. Will it work when there' no such guarantee? Example is when there's a long running promise returned in attaching and deactivate is call while that promise is still running? If it doesn't do that, we probably cant supersede https://github.com/aurelia/aurelia/pull/1729.

Yeah, I don't think this covers all use cases that https://github.com/aurelia/aurelia/pull/1729 covers. I'd be surprised if there's a so much easier way to solve those that it's worth searching for a different enough take than that in https://github.com/aurelia/aurelia/pull/1729. 🤷

packages/runtime-html/src/templating/controller.ts
```diff
@@ -762,11 +773,15 @@ export class Controller<C extends IViewModel = IViewModel> implements IControlle
...
   ): void | Promise<void> {
+    let prevActivation: void | Promise<void> = void 0;
     switch ((this.state & ~State.released)) {
       case State.activated:
-        // We're fully activated, so proceed with normal deactivation.
         this.state = State.deactivating;
         break;
+      case State.activating:
+        this.state = State.deactivating;
+        prevActivation = this.$promise?.catch(noop); /** justification: because we are about to deactivate, the error from activation can be ignored */
```
bigopon, line 783: I think we can afford to add quite a bit of helping text in a dev build around here.
```suggestion
        // we are about to deactivate, the error from activation can be ignored
        prevActivation = this.$promise?.catch(__DEV__
          ? ex => {
            // ...
          }
          : noop
        )
```

packages/runtime-html/src/templating/controller.ts
```diff
@@ -1028,14 +1056,17 @@ export class Controller<C extends IViewModel = IViewModel> implements IControlle
...
 
-        if (cur._hooks.hasUnbinding) {
-          if (cur.debug) { cur.logger!.trace('unbinding()'); }
+          if (cur._hooks.hasUnbinding) {
+            if (cur.debug) { cur.logger!.trace('unbinding()'); }
 
-          ret = resolveAll(ret, cur.viewModel!.unbinding(cur.$initiator, cur.parent));
+            // console.log('calling vm.unbinding');
+            ret = resolveAll(ret, cur.viewModel!.unbinding(cur.$initiator, cur.parent));
+          }
```
bigopon, line 1069: Using existing `isBound` is indeed simple to handle the deactivation, as we have only 2 hooks during deactivation. Nice details ✅
Here we are calling unbinding only when it's already been bound, but it feels quite weird. Since we cant determine whether `binding` lifecycle has been called or not. An example issue is there's setup work done in `binding` (which is asyncable) won't be undone during activation if deactivation comes in while binding promise is still running. I think we will need more state.
If we had synchronous `binding`, this probably already suffices.

packages/runtime-html/src/templating/controller.ts
```diff
@@ -656,8 +662,13 @@ export class Controller<C extends IViewModel = IViewModel> implements IControlle
     if (isPromise(ret)) {
       this._ensurePromise();
       ret.then(() => {
-        this.isBound = true;
-        this._attach();
+        // because controller can be deactivated, during a long running promise in the bound phase
+        if (this.state !== State.activating) {
+          this._leaveActivating();
+        } else {
```
bigopon, line 668: Do we also need to leave attaching in case of ret throwing?

packages/runtime-html/src/templating/controller.ts
```diff
@@ -602,7 +603,12 @@ export class Controller<C extends IViewModel = IViewModel> implements IControlle
     if (isPromise(ret)) {
       this._ensurePromise();
       ret.then(() => {
-        this.bind();
+        if (this.state !== State.activating) {
+          // because controller can be deactivated, during a long running promise in the binding phase
+          this._leaveActivating();
+        } else {
```
bigopon, line 609: do we also need to leave attaching in case ret throws?

packages/runtime-html/src/templating/controller.ts
```diff
@@ -762,11 +773,15 @@ export class Controller<C extends IViewModel = IViewModel> implements IControlle
...
     _parent: IHydratedController | null,
   ): void | Promise<void> {
+    let prevActivation: void | Promise<void> = void 0;
     switch ((this.state & ~State.released)) {
       case State.activated:
-        // We're fully activated, so proceed with normal deactivation.
         this.state = State.deactivating;
         break;
+      case State.activating:
+        this.state = State.deactivating;
```
bigopon, line 782: Is activating while deactivating allowed? If yes then we'll need to add similar work to `activate` call as well.

packages/__tests__/3-runtime-html/controller.deactivation.partially-activated.spec.ts
```diff
@@ -0,0 +1,2677 @@
...
+      this._mode = value;
+      this._resolve = null;
+      this._reject = null;
+    }
+  }
+
+  abstract class TestVM implements ICustomElementViewModel {
+    public constructor(
+      @INotifierManager public readonly mgr: INotifierManager,
+    ) {
```
bigopon, line 125: For simpler code, I think we can start using `resolve`
```suggestion
    public readonly mgr = resolve(INotifierManager);
    public constructor() {
```

bigopon (commented):
Nice work @Sayan751 A few more details to iron out for the implementation then I'll start reviewing the test 👍

packages/runtime-html/src/templating/controller.ts
```diff
@@ -1028,14 +1056,17 @@ export class Controller<C extends IViewModel = IViewModel> implements IControlle
...
+            ret = resolveAll(...cur._lifecycleHooks!.unbinding.map(callUnbindingHook, cur));
+          }
 
-        if (cur._hooks.hasUnbinding) {
-          if (cur.debug) { cur.logger!.trace('unbinding()'); }
+          if (cur._hooks.hasUnbinding) {
+            if (cur.debug) { cur.logger!.trace('unbinding()'); }
 
-          ret = resolveAll(ret, cur.viewModel!.unbinding(cur.$initiator, cur.parent));
+            // console.log('calling vm.unbinding');
```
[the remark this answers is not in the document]
Sayan751: Thanks! Cleaned up.

packages/testing/src/startup.ts
```diff
@@ -251,6 +253,16 @@ export function createFixture<T extends object>(
         return dispose();
     }
 
+    public async stop(dispose: boolean): Promise<void> {
+      try {
+        await au.stop(dispose);
+      } finally {
+        if (dispose) {
+          root.remove();
```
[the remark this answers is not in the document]
Sayan751: Done.

packages/runtime-html/src/templating/controller.ts
Sayan751: Good idea! Added some logging.

packages/runtime-html/src/templating/controller.ts
Sayan751: I agree what you are saying. I think instead of a "state", which is always a bit cumbersome to incorporate and has full potential to be overloaded, we can simply use a flag like `isBindingDone`.

packages/runtime-html/src/templating/controller.ts
Sayan751: Nope. Because the .catch in this case will reject the activation promise, which will in turn be logged and ignored in the .catch in the `Controller#deactivate`. I think we are good here also without a `_leaveActivating`.

packages/runtime-html/src/templating/controller.ts
Sayan751: Same argumentation as the previous comment.

packages/runtime-html/src/templating/controller.ts
Sayan751: Uuhhh! Nice idea, but I will wait on that till a real life use case arrives. As I see it, the deactivating while activating has a real use case for routers, consider error etc. I am not sure about the reverse use-cases.

packages/runtime-html/src/templating/controller.ts
bigopon: Yeah nice ideal, actually let's go for the simplest "state".

Sayan751: Done.

packages/runtime-html/src/templating/controller.ts
bigopon: Thanks. I just realized we may need to handle more than just this to enable error recovery reliably everywhere. Here's we are only catching error coming from lifecycle hooks/vm hooks, but there could be ones that come from binding as well.
We don't have to handle that now, phase 1 could be enough with just lifecycle.

packages/testing/src/startup.ts
```diff
@@ -251,6 +253,17 @@ export function createFixture<T extends object>(
         return dispose();
     }
 
+    public async stop(dispose: boolean = false): Promise<void> {
+      try {
+        await au.stop(dispose);
+      } finally {
+        if (dispose) {
+          root.remove();
+          au.dispose();
```
bigopon, line 262: We will need to incrase tornCount by 1 here too, so that auto teardown dont try to tear this down again later after each tests.

packages/__tests__/3-runtime-html/controller.deactivation.partially-activated.spec.ts
```diff
@@ -0,0 +1,2684 @@
...
+          'start.Global.if.attached.leave',
+
+          'start.Global.else.binding.enter',
+          'start.Global.else.binding.leave',
+          'start.Global.else.bound.enter',
+          'start.Global.else.bound.leave',
+          'start.Global.else.attaching.enter',
+          'start.Global.else.attaching.leave',
+          'start.Global.else.attached.enter',
+          'start.Global.else.attached.leave',
```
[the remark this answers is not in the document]
bigopon: Similarly, I think we can ignore all the if/else hooks, they are not relevant enough in this tests. We can just focus on the vm hooks.
If we want to test if/else, we can test it separately rather than combining them here.

packages/__tests__/3-runtime-html/controller.deactivation.partially-activated.spec.ts
```diff
@@ -0,0 +1,2684 @@
...
+        const expectedLog = getPendingActivationLog(false);
+        mgr.assertLog(expectedLog, 'phase#1 - pre-resolve');
+        assert.html.textContent(appHost, hook === 'attached' || hook === 'attaching' ? 'c2' : '', 'phase#1.textContent');
+
+        // trigger deactivation then resolve the promise and wait for everything
+        /**
+         * Note on manual deactivation instead of setting the flag to false:
+         * The `if`-TC is not preemptive.
+         * That is it waits always for the previous change.
+         * Thus, setting the flag to false would only queue the deactivation rather than a pre-emptive action.
```
bigopon, line 771: Suggestion: maybe avoid flag as it's unclear.

> ... thus setting the value of the if binding to false ...

packages/__tests__/3-runtime-html/controller.deactivation.partially-activated.spec.ts
```diff
@@ -0,0 +1,2684 @@
...
+         * We might want to change the `if`-TC to be pre-emptive, but that requires more discussion among team and community.
+         */
+        const deactivationPromise = ifVm.elseView.deactivate(ifVm.elseView, ifCtrl);
+        promiseManager.resolve();
+        queue.queueTask(() => Promise.resolve());
+        await Promise.all([promiseManager.currentPromise, deactivationPromise, queue.yield(), ifVm['pending']]);
```
bigopon, line 777: what's the purpose of `queue.yield` here?

bigopon (commented):
Nice work on the tests @sayan751 I have some minor comments for now, will continue soon. Currently I don't agree with the form of the tests, loops are still employed and I anticipate that it's not gonna help us move forward. I'll think of a way to remove the loop and propose that in a separate PR targeting this branch.

packages/__tests__/3-runtime-html/controller.deactivation.partially-activated.spec.ts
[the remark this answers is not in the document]
Sayan751: The if/else is there because it is important to check the activation of a different component and/or reactivation of the same component after the previous activation is interrupted. Any TC could have sufficed for this, but the if/else offers the least complexity in terms of tests, and thereby least cognitive load for the devs coming back to these tests. I wouldn't advise removing this test aspect. 

> This is just noise imo.

Well so is the case for global hooks. I don't think it is either wise or necessary to suppress/ignore those logs.

packages/__tests__/3-runtime-html/controller.deactivation.partially-activated.spec.ts
Sayan751: Good catch! It was an old artefact. Now removed.

Sayan751:
> Currently I don't agree with the form of the tests, loops are still employed and I anticipate that it's not gonna help us move forward. I'll think of a way to remove the loop and propose that in a separate PR targeting this branch.

@bigopon Thanks for the review. I think if we remove the loop, the test suite will explode in size and will soon be unmanageable. I have tried to keep the loop to minimum, not to create much cognitive load.

PR #1762, opened 2023-05-14
chore: add tests for partial deactivation

bigopon:
@Sayan751


bigopon:
Closing this for now since we'll need to refactor the deactivation/action process in the Aurelia instance a bit, the tests will be reference for what future tests should be.

PR #1767, opened 2023-05-18
feat(di): ability to use newInstance() with interface

### LINKED ISSUE #1765: @newInstanceOf decorator doesn't work properly on IHttpClient
In the following snippet of code, `config` here seems to be a singleton so whichever service registers last overwrites the config on any previous ones. It could just be that a new instance of `IHttpClient` is never created thus the entire client instance is a singleton.  I've discussed this with @bigopon on discord and even though the examples at https://docs.aurelia.io/aurelia-packages/fetch-client#configuring-the-fetch-client show this as being the proper usage, it simply doesn't work.  If I switch to use `HttpClient` instead, all is well. 

image

You can see in the following screenshot that upon service 2 being registered, it has service 1's base URL.  As soon as service 2 sets its base URL, service 1's will be incorrect.  

image
        
I have created a repo to demonstrate this located at https://stackblitz.com/edit/au2-conventions-pphynj?file=src%2Fmy-app.ts.

bigopon:
resolves #1765

At the moment, `@newInstanceOf` doesn't work with interface because it expects an implementation. This creates an inconsistency where all other resolvers/APIs of the container works well with both interface and implementation. Add the ability to support `@newInstanceOf`/`@newInstanceForScope` for interface.

Thanks @mt-mshorrosh.

### DIFF
--- a/packages/__tests__/1-kernel/di.get.spec.ts
+++ b/packages/__tests__/1-kernel/di.get.spec.ts
@@ -554,15 +554,31 @@ describe('1-kernel/di.get.spec.ts', function () {
       assert.throws(() => container.get(newInstanceOf(I)), `No registration for interface: 'I'`);
     });
 
-    it('throws when there is NOT factory registration for an instance', function () {
+    it('instantiates when there is an constructable registration for an interface', function () {
       const container = DI.createContainer();
       const I = DI.createInterface('I');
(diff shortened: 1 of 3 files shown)


mt-mshorrosh:
I appreciate your swift efforts here @bigopon!

PR #1770, opened 2023-06-02
feat(build): add a development entry point

bigopon:
## 📖 Description

Add a "development" condition in exports to support some platforms like https://esm.sh/#docs
Those platforms that do support it will be able to pick the right dev dist.


3cp:
Interesting Node is still adding more and more fields into package.json.
Those fields affect module resolution. This is one of the complexities that Deno decided to get ride of from day 1.

PR #1778, opened 2023-06-14
fix(router-lite): URL generation in child component

### LINKED ISSUE #1776: [router-lite] Open in new tab does not grab correct url
# 🐛 Bug Report

On Aurelia 2 opening a link in a new tab results in the url only containing what in the load area.
IE if I am on the route /something/foo
and I have a <a load="bar">
If I just click on it I am properly navigated to /something/foo/bar
However if you open it in a new tab you are navigated to /bar

Repro link https://stackblitz.com/edit/au2-conventions?file=src%2Fmy-app.html

## 🤔 Expected Behavior

I am properly navigated to /something/foo/bar

## 😯 Current Behavior

if you open it in a new tab you are navigated to /bar

## 🌍 Your Environment

| Aurelia          |
| Language         |
| Browser          |
| Bundler          |
| Operating System |
| NPM/Node/Yarn    |

Sayan751:
# Pull Request

### 🎫 Issues

Fixes #1776

### DIFF
--- a/packages/router-lite/src/instructions.ts
+++ b/packages/router-lite/src/instructions.ts
@@ -16,7 +16,7 @@ import {
 } from '@aurelia/runtime-html';
 
 import { IRouteViewModel } from './component-agent';
-import { type RouteConfig, RouteType } from './route';
+import { RouteType } from './route';
 import { type $RecognizedRoute, IRouteContext, RouteContext } from './route-context';
 import { expectType, isPartialViewportInstruction, shallowEquals } from './validation';
 import { INavigationOptions, NavigationOptions, type RouterOptions } from './options';
--- a/packages/router-lite/src/instructions.ts
+++ b/packages/router-lite/src/instructions.ts
@@ -288,10 +288,6 @@ export class ViewportInstructionTree {
     }
 
     if (typeof instructionOrInstructions === 'string') {
-      if (hasContext) {
-        const config = (context.childRoutes as RouteConfig[]).find(x => x.id === instructionOrInstructions);
-        instructionOrInstructions = config?.path.find(x => x) ?? instructionOrInstructions;
-      }
       const expr = RouteExpression.parse(instructionOrInstructions as string, routerOptions.useUrlFragmentHash);
       return expr.toInstructionTree($options);
     }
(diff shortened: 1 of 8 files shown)


packages/router-lite/src/instructions.ts
```diff
@@ -286,7 +288,7 @@ export class ViewportInstructionTree {
     }
 
     if (typeof instructionOrInstructions === 'string') {
-      const expr = RouteExpression.parse(instructionOrInstructions, routerOptions.useUrlFragmentHash);
+      const expr = RouteExpression.parse(instructionOrInstructions as string, routerOptions.useUrlFragmentHash);
```
bigopon, line 291: we probably dont need the cast here

packages/router-lite/src/viewport-agent.ts
```diff
@@ -860,7 +860,7 @@ function ensureTransitionHasNotErrored(tr: Transition): void {
   if (tr.error !== void 0 && !tr.erredWithUnknownRoute) throw tr.error;
 }
 
-const enum State {
+export const enum State {
```
bigopon, line 863: if this is only internal then we'll need to hide & remove this export too, similar like other const enums in our runtime.

packages/router-lite/src/instructions.ts
```diff
@@ -330,14 +332,36 @@ export class ViewportInstructionTree {
...
 
-  public toUrl(useUrlFragmentHash: boolean = false): string {
+  public toUrl(isFinalInstruction: boolean, useUrlFragmentHash: boolean): string {
     let pathname: string;
     let hash: string;
+    let parentPath = '';
+
+    if (!isFinalInstruction) {
+      const parentPaths: string[] = [];
+      let ctx: IRouteContext | null = this.options.context as IRouteContext | null;
```
bigopon, line 342: 👍

packages/router-lite/src/viewport-agent.ts
Sayan751: Fixed it in #1779

packages/router-lite/src/instructions.ts
Sayan751: Thanks for noticing this. Missed cleanup. Fixed it in #1779.

PR #1792, opened 2023-07-08
feat(compose): passthrough bindings + support containerless

### LINKED ISSUE #497: au-compose should pass through bindings and content

bigopon:
## 📖 Description

At the moment, `<au-compose/>` does not "pass through" bindings declared on the `<au-compose>` element. This creates undesirable workaround and sometimes, limitation. This is also a frequently requested feature of v1. This PR adds the ability to pass through all bindings declared on `<au-compose/>` to the composed custom element, so that it is as if the bindings, even bindable properties, were declared on the custom element itself.

- Also enable containerless custom element composition. At the moment, an attempt of compositing a containerless custom element as component for the `<au-compose/>` will throw. Remove this constraint and containerless composition will just work as normal. This effectively make `<au-compose/>` as natural as a normal custom element usage, except it's more dynamic.

cc @Sayan751 @fkleuver

Resolves #497


docs/user-docs/getting-to-know-aurelia/dynamic-composition.md
```diff
@@ -89,30 +89,41 @@ In some scenarios, you may want to access the view model of the component being
 <au-compose view-model.ref="myCompose"></au-compose>
 ```
 
-This will add a property to the host class called `myCompose`
+This will works as if it was a `view-model.ref` binding on a normal custom element
```
Vheissu, line 92:
```suggestion
This will work as though it were a `view-model.ref` binding on a standard custom element.
```

docs/user-docs/getting-to-know-aurelia/dynamic-composition.md
```diff
@@ -89,30 +89,41 @@ In some scenarios, you may want to access the view model of the component being
...
 
-```typescript
-export class MyApp {
-    readonly myCompose;
-}
-```
+## Passing props through
 
-However, one pitfall you will encounter is the view model that gets passed to the `ref` binding is a constructible component and not the instance itself. If you worked with Aurelia 1, you might expect the passed `view-model` instance to be the instance itself, not the class definition.
+The `<au-compose>` will pass all the bindings, except the ones that target its bindable properties (`model`/`component`/`template`) declared on it to the composed view model, if it's a custom element.
```
Vheissu, line 96:
```suggestion
The `<au-compose>` will pass all bindings, except those targeting its bindable properties (`model`/`component`/`template`) declared on it, to the composed view model, assuming it is a custom element.
```

docs/user-docs/getting-to-know-aurelia/dynamic-composition.md
```diff
@@ -144,6 +155,9 @@ If you still want a view supporting a dynamically loaded module, you can create
 
 The above value converter will load the URL and return the text response. For view models, something similar can be achieved where an object or class can be returned.
 
+3. In Aurelia 2, all bindings are passed through to the underlying custom element
+composition, so `view-model.ref` no longer means getting a reference to the composer, but the composed view model instead.
+
```
Vheissu, line 160:
```suggestion
3. In Aurelia 2, all bindings are transferred to the underlying custom element composition. Therefore, `view-model.ref` no longer signifies obtaining a reference to the composer, but rather to the composed view model.
```

Vheissu (commented):
Looks awesome to me. I made a few brief suggestions for the docs. Nothing overly dramatic!

PR #1794, opened 2023-07-11
refactor(router-lite): query in fragment when using useUrlFragmentHash option

Sayan751:
# Pull Request

## 📖 Description

This fixes the issue reported on [Discourse](https://discourse.aurelia.io/t/router-lite-href-generation-incorrect/5183).

## ⏭ Next Steps

cc: @ivanbacher

### DIFF
--- a/packages/__tests__/router-lite/smoke-tests.spec.ts
+++ b/packages/__tests__/router-lite/smoke-tests.spec.ts
@@ -3654,6 +3654,80 @@ describe('router-lite/smoke-tests.spec.ts', function () {
       await au.stop(true);
     });
   }
+
+  it('querystring is added to the fragment when hash-based routing is used', async function () {
+    @customElement({ name: 'c-1', template: `c1 params: \${id} query: \${query} fragment: \${fragment}` })
+    class C1 implements IRouteViewModel {
+
+      private id: string;
+      private query: string;
+      private fragment: string;
+
+      public loading(params: Params, node: RouteNode): void | Promise<void> {
+        this.id = params.id;
+        this.query = node.queryParams.toString();
+        this.fragment = node.fragment;
+      }
+    }
+
+    @route({ routes: [{ id: 'c1', path: 'c1/:id', component: C1 }] })
+    @customElement({ name: 'app', template: '<a load="route: c1; params.bind: {id: 42, foo: \'bar\'}"></a><au-viewport></au-viewport>' })
+    class App { }
+
+    const { host, container } = await start({ appRoot: App, useHash: true });
+
+    const anchor = host.querySelector('a');
+    assert.match(anchor.href, /#\/c1\/42\?foo=bar$/);
+
+    anchor.click();
+    await container.get(IPlatform).taskQueue.yield();
+
+    assert.html.textContent(host, 'c1 params: 42 query: foo=bar fragment:');
+    const path = (container.get(ILocation) as unknown as MockBrowserHistoryLocation).path;
+    assert.match(path, /#\/c1\/42\?foo=bar$/);
+
+    // assert the different parts of the url
+    const url = new URL(path);
+    assert.match(url.pathname, /\/$/);
+    assert.strictEqual(url.search, '');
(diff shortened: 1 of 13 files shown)


packages/router-lite/src/url-parser.ts
```diff
@@ -0,0 +1,105 @@
+import { emptyQuery } from './router';
+
+export class ParsedUrl {
```
bigopon, line 3: doesn't this feel like it belongs to route-recognizer level? It's more of an utility that can be reused by both routers.
cc @jwx

packages/router-lite/src/route-expression.ts
```diff
@@ -570,22 +515,17 @@ export class ActionExpression {
...
+    state._discard();
+    return new ActionExpression(name, parameterList);
   }
 }
 
 export class ViewportExpression {
   public get kind(): ExpressionKind.Viewport { return ExpressionKind.Viewport; }
 
-  public static get EMPTY(): ViewportExpression { return new ViewportExpression('', ''); }
+  public static get EMPTY(): ViewportExpression { return new ViewportExpression(''); }
```
bigopon, line 526: we don't upper case anywhere in our packages, this should be `Empty`/or `$Empty` if you want it to get special attention.

packages/router-lite/src/url-parser.ts
```diff
@@ -0,0 +1,105 @@
...
+
+  private readonly id: string;
+
+  public constructor(
+    public readonly path: string,
+    public readonly query: Readonly<URLSearchParams>,
+    public readonly fragment: string | null,
+  ) {
+    Object.freeze(query);
```
bigopon, line 12: even though this class is more internal than public, freezing an object that it doesn't create doesn't feel right. Maybe clone before freeze?

packages/router-lite/src/instructions.ts
```diff
@@ -359,19 +358,11 @@ export class ViewportInstructionTree {
...
-
-    let search = this.queryParams.toString();
-    search = search === '' ? '' : `?${search}`;
-
-    return `${pathname}${search}${hash}`;
+    return parser.stringify(
+      parentPath.length > 0 ? `${parentPath}/${currentPath}` : currentPath,
+      this.queryParams,
+      this.fragment
+    );
```
bigopon, line 365: clean 👍

packages/__tests__/router-lite/smoke-tests.spec.ts
```diff
@@ -3654,6 +3654,152 @@ describe('router-lite/smoke-tests.spec.ts', function () {
...
+    assert.match(path, /#\/c1\/42\?foo=bar#for-whatever-reason$/);
+
+    // assert the different parts of the url
+    const url = new URL(path);
+    assert.match(url.pathname, /\/$/);
+    assert.strictEqual(url.search, '');
+    assert.strictEqual(url.hash, '#/c1/42?foo=bar#for-whatever-reason');
+  });
+
+  it('fragment is added to fragment (nested fragment JFF) when using hash-based routing', async function () {
```
bigopon, line 3766: JFF?

bigopon (approved):

packages/__tests__/router-lite/smoke-tests.spec.ts
Sayan751: Just for fun 😛

packages/router-lite/src/route-expression.ts
Sayan751: Done!

packages/router-lite/src/url-parser.ts
Sayan751: That's a great point. Changes done.

packages/router-lite/src/url-parser.ts
Sayan751: As we are using this parser only in router-lite, my vote is for leaving it just there. If @jwx decides later to use it, then it can be moved to route-recognizer.

PR #1805, opened 2023-07-23
fix(router-lite): handling slash in parameter value

### LINKED ISSUE #1786: route parameter includig slash breaks router-lite
# 🐛 Bug Report

If a route parameter contains a slash (`product/1`), it is interpreted as a route despite URL encoding:

https://stackblitz.com/edit/au2-conventions-9x2nsu?file=src%2Fmy-app.html

I am using router-lite in Aurelia 2 dev version.

// UPDATE:
Even when using load attribute, created url is wrong:

`load="route: test; params.bind: { id: 'product/1' }"`

will be resolved to `test/product/1`.

## 🤔 Expected Behavior

Encoded route param should not affect the route itself.

## 😯 Current Behavior

Encoded route param (product%2F1) is decoded first and `product/1` will be translated to routes `product` and `1`.

## 💻 Code Sample

https://stackblitz.com/edit/au2-conventions-9x2nsu?file=src%2Fmy-app.html

## 🌍 Your Environment

Aurelia `dev` version

Sayan751:
# Pull Request

### 🎫 Issues

Fixes #1786

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2.1
 
 orbs:
   common: aurelia/common@volatile
-  browser-tools: circleci/browser-tools@1.1.0
+  browser-tools: circleci/browser-tools@1.4.3
 
 # # # # # # # # # # # # # # # #
 # - Scalar variables -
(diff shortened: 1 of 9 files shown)


bigopon (approved):
we should have a test to load the params the way it is used in the issue report, since that's how it's seen in applications, though it can come later.

PR #1809, opened 2023-08-07
fix(dialog): use startingZIndex

tomtomau:
# Pull Request

## 📖 Description

This is kinda a bug report and a PR in one...ish

Whilst building app, I've noticed that startingZIndex doesn't appear to be used in the `@aurelia/dialog` package.

I've had a crack at trying to use it, but not entirely sure if I'm holding it correctly.

## ⏭ Next Steps

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,6 +21,7 @@ packages/**/*.txt
 # to make it easy for application, probably add this
 # so in the future, when we move to a higher level of npm, it won't break
 package-lock.json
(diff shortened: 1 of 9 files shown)


Vheissu:
@bigopon Assigned this for you to review. Looks like this was just overlooked in the implementation coming from v1.

bigopon:
Thanks @tomtomau , I've pushed the fix that will take care of zindex, will merge this soon.

PR #1816, opened 2023-09-27
fix(router-lite): excluded redirectTo from nav-model

### LINKED ISSUE #1815: Bug: redirectTo in navigationModel
# 🐛 Bug Report

`redirectTo` route should be excluded from navigationModel in router-lite:

https://stackblitz.com/edit/au2-conventions-gt6vkx

I also cannot exclude the redirectTo route from navigationModel as `nav: false` ist not a valid property of a `redirectTo` route.

## 🤔 Expected Behavior

A `redirect` route should be excluded from navigationModel

## 😯 Current Behavior

A `redirect` route is included in navigationModel

Sayan751:
# Pull Request

### 🎫 Issues

Closes #1815

### DIFF
--- a/package.json
+++ b/package.json
@@ -113,7 +113,7 @@
     "@typescript-eslint/eslint-plugin": "5.51.0",
     "@typescript-eslint/parser": "5.51.0",
     "chalk": "4.1.2",
-    "chromedriver": "114.0.3",
+    "chromedriver": "117.0.3",
     "codecov": "^3.8.3",
     "concurrently": "7.6.0",
     "cross-env": "7.0.3",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -113,7 +113,7 @@
     "@typescript-eslint/eslint-plugin": "5.51.0",
     "@typescript-eslint/parser": "5.51.0",
     "chalk": "4.1.2",
-    "chromedriver": "114.0.3",
+    "chromedriver": "117.0.3",
```
bigopon, line 116: 👍

bigopon (approved):

PR #1819, opened 2023-10-13
Feat/current router

brandonseydel:
# Pull Request

### DIFF
--- a/packages/router-lite/src/route-callback.ts
+++ b/packages/router-lite/src/route-callback.ts
@@ -11,16 +11,17 @@ export type IRoute<TParams = Record<string, string | undefined>, TQueryParams = 
     url: string;
     params: TParams;
     queryParams: TQueryParams;
-    event: Partial<NavigationEndEvent>
+    event: Partial<NavigationEndEvent>;
   };
   export const IRoute = DI.createInterface<IRoute>();
-  
-const flattenParams = (arr:ViewportInstructionTree['children']):Readonly<Params> => {
-  return arr.reduce(function(done,curr){
+
+const flattenParams = (arr: ViewportInstructionTree['children']): Readonly<Params> => {
+  const returnVal: Readonly<Params>= {} ;
+  return arr.reduce(function (done,curr){
     return  {...done, ...curr.recognizedRoute?.route.params, ...flattenParams(curr.children)};
-    }, {} as Readonly<Params>) ;
-}
-  
+    }, returnVal) ;
+};
+
 export const RouteCallback =  Registration.cachedCallback(IRoute, (container) => {
     const route: IRoute = {
       path: '',
@@ -53,4 +54,4 @@ export const RouteCallback =  Registration.cachedCallback(IRoute, (container) =>
     container.register(AppTask.deactivated(() => navEndSubscription.dispose()));
 
     return route;
-  })
+  });
--- a/packages/router-lite/src/route-callback.ts
+++ b/packages/router-lite/src/route-callback.ts
@@ -1,10 +1,9 @@
 import { DI, Registration } from '@aurelia/kernel';
 import { batch } from '@aurelia/runtime';
 import { IWindow, AppTask } from '@aurelia/runtime-html';
-import { Route, RouteType } from './route';
-import { IRouterEvents } from './router-events';
-import { IRouteViewModel } from './component-agent';
-import { IRouteConfig } from './options';
+import { IRouterEvents, NavigationEndEvent } from './router-events';
+import { IRouter } from './router';
+import { Params, ViewportInstructionTree } from './instructions';
 
 export type IRoute<TParams = Record<string, string | undefined>, TQueryParams = Record<string, string | undefined>> = {
     title: string;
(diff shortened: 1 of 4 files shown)


bigopon:
@jwx @Sayan751 can you help with reviewing the respective part in this PR? If you feel like to, tests are also needed 😉

packages/router/src/route-callback.ts
```diff
@@ -0,0 +1,49 @@
...
+export type IRoute<TParams = Record<string, string | undefined>, TQueryParams = Record<string, string | undefined>> = {
+    title: string;
+    path: string;
+    url: string;
+    params: TParams;
+    queryParams: TQueryParams;
+    snapshot: Partial<Navigation>;
+  };
+
+export const IRoute = DI.createInterface<IRoute>();
```
bigopon, line 15: we need to decorate this with `/*@__PURE__*/` comment for tree shaking
```suggestion
export const IRoute = /*@__PURE__*/ DI.createInterface<IRoute>();
```

packages/router/src/route-callback.ts
```diff
@@ -0,0 +1,49 @@
...
+
+export const RouteCallback =  Registration.cachedCallback(IRoute, (container) => {
```
bigopon, line 17: likely to be the same
```suggestion
export const RouteCallback = /*@__PURE__*/ Registration.cachedCallback(IRoute, (container) => {
```

packages/router-lite/src/route-callback.ts
```diff
@@ -0,0 +1,57 @@
...
+  export const IRoute = DI.createInterface<IRoute>();
+
+const flattenParams = (arr: ViewportInstructionTree['children']): Readonly<Params> => {
+  const returnVal: Readonly<Params>= {} ;
+  return arr.reduce(function (done,curr){
+    return  {...done, ...curr.recognizedRoute?.route.params, ...flattenParams(curr.children)};
+    }, returnVal) ;
+};
+
+export const RouteCallback =  Registration.cachedCallback(IRoute, (container) => {
```
bigopon, line 25: tree shaking related comment
```suggestion
export const RouteCallback = /*@__PURE__*/ Registration.cachedCallback(IRoute, (container) => {
```

packages/router-lite/src/route-callback.ts
```diff
@@ -0,0 +1,57 @@
...
+
+export type IRoute<TParams = Record<string, string | undefined>, TQueryParams = Record<string, string | undefined>> = {
+    title: string;
+    path: string;
+    url: string;
+    params: TParams;
+    queryParams: TQueryParams;
+    snapshot: Partial<NavigationEndEvent>;
+  };
+  export const IRoute = DI.createInterface<IRoute>();
```
bigopon, line 16:
```suggestion
export const IRoute = /*@__PURE__*/ DI.createInterface<IRoute>();
```

bigopon (commented):
Looks good to me, though we need some, at least preliminary, tests

packages/router-lite/src/route-callback.ts
```diff
@@ -0,0 +1,57 @@
...
+    const router = container.get(IRouter);
+
+    const navEndSubscription = routerEvents.subscribe('au:router:navigation-end', (event) => {
+      const { finalInstructions, finalInstructions: { children, queryParams} } = event;
+
+      batch(() => {
+        route.path = finalInstructions.toPath();
+        route.url = finalInstructions.toUrl(true, router.options._urlParser);
+        route.title = localWindow.document.title;
+        route.queryParams = Array.from(queryParams).reduce<Record<string, string>>((acc, [key, value]) => {
```
Sayan751, line 46: This is not needed when `queryParams` is of type [URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams).

packages/router-lite/src/route-callback.ts
```diff
@@ -0,0 +1,57 @@
...
+          acc[key] = value;
+          return acc;
+        }, {});
+        route.params = flattenParams(children);
```
Sayan751, line 50: Why flatten parameter? This results in loss of information.

packages/router-lite/src/route-callback.ts
```diff
@@ -0,0 +1,57 @@
+import { DI, Registration } from '@aurelia/kernel';
+import { batch } from '@aurelia/runtime';
+import { IWindow, AppTask } from '@aurelia/runtime-html';
+import { IRouterEvents, NavigationEndEvent } from './router-events';
+import { IRouter } from './router';
+import { Params, ViewportInstructionTree } from './instructions';
+
+export type IRoute<TParams = Record<string, string | undefined>, TQueryParams = Record<string, string | undefined>> = {
```
Sayan751, line 8: This should be a class.

packages/router-lite/src/route-callback.ts
```diff
@@ -0,0 +1,57 @@
...
+    title: string;
+    path: string;
+    url: string;
+    params: TParams;
+    queryParams: TQueryParams;
```
Sayan751, line 13: Should be of type [URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams).

packages/router-lite/src/route-callback.ts
```diff
@@ -0,0 +1,57 @@
...
+    snapshot: Partial<NavigationEndEvent>;
```
Sayan751, line 14: I find the name `snapshot` to be very ambiguous. I would remove it completely, because semantically, the `event` cannot be a part of the current route. Furthermore, I would like to know the use-cases where this would be beneficial.

If kept, I would suggest removing the `Partial`, as why would the information provided by the framework be partial?

packages/router-lite/src/route-callback.ts
```diff
@@ -0,0 +1,57 @@
...
+import { IWindow, AppTask } from '@aurelia/runtime-html';
+import { IRouterEvents, NavigationEndEvent } from './router-events';
+import { IRouter } from './router';
+import { Params, ViewportInstructionTree } from './instructions';
+
+export type IRoute<TParams = Record<string, string | undefined>, TQueryParams = Record<string, string | undefined>> = {
+    title: string;
+    path: string;
+    url: string;
+    params: TParams;
```
Sayan751, line 12: An extra `TParams` is not needed, as there is already a `Params` type should for this very purpose. 

Moreover, considering multi-level routing, this should be a map (for example, `Map<component,Params>`) of some sort.

packages/router-lite/src/route-callback.ts
```diff
@@ -0,0 +1,57 @@
...
+
+const flattenParams = (arr: ViewportInstructionTree['children']): Readonly<Params> => {
+  const returnVal: Readonly<Params>= {} ;
+  return arr.reduce(function (done,curr){
+    return  {...done, ...curr.recognizedRoute?.route.params, ...flattenParams(curr.children)};
+    }, returnVal) ;
+};
+
+export const RouteCallback =  Registration.cachedCallback(IRoute, (container) => {
+    const route: IRoute = {
```
Sayan751, line 26: Instantiate the class instead.

Sayan751 (changes requested):
@brandonseydel Thank you for this. Reviewed the router-lite part. Seems logical; however, as there are no tests, it is hard to verify that it works for different routing scenarios. Please consider writing tests.

PR #1821, opened 2023-10-13
fix(router-lite): invoke-lifecycles transition plan

Sayan751:
# Pull Request

### DIFF
--- a/packages/__tests__/src/router-lite/smoke-tests.spec.ts
+++ b/packages/__tests__/src/router-lite/smoke-tests.spec.ts
@@ -5410,6 +5410,53 @@ describe('router-lite/smoke-tests.spec.ts', function () {
       await au.stop(true);
     });
 
+    it('invoke-lifecycles - inherited', async function () {
+      @customElement({ name: 'ce-one', template: 'ce1 ${id1} ${id2}' })
+      class CeOne implements IRouteViewModel {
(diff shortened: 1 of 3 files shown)


packages/__tests__/src/router-lite/smoke-tests.spec.ts
```diff
@@ -5407,6 +5412,229 @@ describe('router-lite/smoke-tests.spec.ts', function () {
...
+      await queue.yield();
+      assert.html.textContent(host, 'p1 2 2 ce2 2 3', 'round#3'); // as the parent is replaced, so is the child
+
+      await router.load('p2/1/c1/1');
+      await queue.yield();
+      assert.html.textContent(host, 'p2 1 1 ce1 1 1', 'round#4');
+
+      await router.load('p2/1/c1/2');
+      await queue.yield();
+      assert.html.textContent(host, 'p2 1 1 ce1 2 2', 'round#5');
```
bigopon, line 5628: 👍 though why are we starting with ce2 instead of 1? I got confused reading this test at first since I was assuming we test 1 then 2.

Sayan751: Not any particular reason. I was applying the transitionPlan alternatively, and I ended up with that :P

PR #1826, opened 2023-10-19
Update navigation-model.md

ivanbacher:
Added code for this in example: It additionally shows that from the NavBar#binding, INavigationModel#resolve() is awaited

# Pull Request


Sayan751:
Thank you @ivanbacher for improving of the docs!

PR #1838, opened 2023-11-12
fix(i18n): translating camelCased bindables

Sayan751:
# Pull Request

### 🎫 Issues

This PR fixes the following issue:

```html
<!-- 
  This doesn't work; although it should.
  That is also the default behavior  in au1.
--> 
<my-el my-value="does not work" t="[my-value]displayTextTest"></my-el>

<!--
  This works; although it shouldn't.
  Also this is inconsistent.
  The conversion to the camelCase from the kebab-cased property was missing.
--> 
<my-el my-value="works" t="[myValue]displayTextTest"></my-el>
```

Reproduction: https://stackblitz.com/edit/translating-bindables?file=src%2Fmy-app.ts


bigopon (approved):
Nice fix 👍

PR #1840, opened 2023-11-21
fix(repeater): batched mutation issue

fkleuver:
# Pull Request


packages/__tests__/src/3-runtime-html/repeat.duplicates.spec.ts
```diff
@@ -0,0 +1,94 @@
...
+    it.skip('duplicate primitive number + sort', function () {
+      const { assertText, component } = createFixture(
+        `<div repeat.for="i of items">\${$index}-\${i} </div>`,
+        class { items = [0, 1, 0]; }
+      );
+      assertText('0-0 1-1 2-0 ');
+
+      component.items.sort();
+
+      assertText('0-0 1-0 2-0 3-1 ');
```
bigopon, line 49: this should be `0-0 1-0 2-1`

bigopon (approved):
Looks good for what this PR fixes, we can continue with the other fix afterwards.

PR #1852, opened 2023-12-12
feat(docs): expand and improve introduction

Vheissu:
The introduction didn't espouse some of the benefits of Aurelia properly, it also didn't address community size or lack of virtual DOM.

### DIFF
--- a/docs/user-docs/README.md
+++ b/docs/user-docs/README.md
@@ -6,9 +6,11 @@ description: Get acquainted with Aurelia, the documentation, and how to get star
 
 ## What is Aurelia?
 
-Aurelia is an open-source, JavaScript, front-end platform designed to enable you to easily build even the most demanding web, mobile, or desktop applications. Aurelia stands out for its commitment to open web standards and no-nonsense, get-out-of-your-way conventions that enable vanilla JavaScript development. Of course, Aurelia is also packed full of features, performs at the highest standards, is extensible to its core, and supports a fully testable component-oriented design. But you knew all that already, didn't you? :sunglasses:
+Welcome to Aurelia, the future-oriented, open-source JavaScript platform for crafting extraordinary web, mobile, and desktop applications. At its core, Aurelia is more than just a framework; it believes in the unbridled power of open web standards. With Aurelia, your development experience is streamlined, intuitive, and remarkably flexible.
 
-To get started, jump to our Quick Start Guide, or read on to get more of an overview.
+Aurelia isn't just feature-rich; it's a testament to high performance and extensive extensibility, offering a component-oriented design that's fully testable. It's built for developers who know the value of simplicity and the importance of conventions that don't obstruct but enhance creativity. :sunglasses:
+
+To start, jump to our Quick Start Guide or read on for an overview.
 
 {% content-ref url="getting-started/quick-start-guide/" %}
 [quick-start-guide](getting-started/quick-start-guide/)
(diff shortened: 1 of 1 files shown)


philpax:
Hi there!

I stumbled upon Aurelia through the [NetPad](https://github.com/tareqimbasher/NetPad) project. I don't currently have any need for it, but it looks like an interesting project and I'll keep it in mind in the event that I want to try out something different 🙂 

However, the reason I'm writing this message is precautionary: I *immediately* noticed [the tells](https://www.plusdocs.com/blog/the-most-overused-chatgpt-words) of majority-ChatGPT writing, which many people consider to be [a strongly negative signal](https://x.com/whitequark/status/1800517531607929047) these days. 

I have nothing against the use of AI (I work with it both personally and professionally), but if I can tell, other people will be able to tell, and they are likely to make their own judgements before they have a chance to familiarise themselves with Aurelia proper. These folk will have the impression that documentation was not a priority, which will likely reflect poorly upon the thing being documented.

I would suggest reverting the majority of this commit and organically integrating the bits you liked. Much of the new text is superfluous and/or repetitive and suffers from ChatGPT's perennial sycophancy. I suspect that it could be shrunk to half of its size while retaining its message and feeling more honest.

Good documentation is tight, concise, and conveys exactly what it needs to to the user; it should sell them without overstaying its welcome, and give them direction to where to find more information if required. This version, unfortunately, feels padded in comparison.

I'm very apologetic about expressing this sentiment - I have no bone to pick here, honest - but I believe that a not-insignificant portion of your potential audience will immediately bounce off once they notice the ChatGPT-isms. (Just have a look at Hacker News's response to writing they determine is majority GPT-written!)

Apologies for this, once again, and I understand if this isn't your highest priority. I figured I'd raise the concern and let you decide what you want to do next.

bigopon:
Thanks @philpax for the suggestion.
@Vheissu what do you think? We probably missed the comment from @philpax

Vheissu:
It might surprise you to learn that this wasn't AI written whatsoever, @philpax . My writing style is naturally verbose and not overly concise. For long-text writing, it can be more difficult to be concise and self-edit. The only assistance used here was Grammarly, which helps you improve your grammar and identify any obvious issues.

The intention with the superfluous parts of the docs was to humanise them. And while I agree that docs should be concise, this is an introduction, and we must sell ourselves a little harder. To understand why it's so wordy, you must understand Aurelia's position in the ecosystem. A lot of the same arguments come up when Aurelia is mentioned:

- It's not popular enough
- The community size isn't big enough
- A heap of other things

So, sadly, to get that buy-in and alleviate some of those concerns, we have to be wordy and go beyond, "Aurelia is an open source stands compliant framework" - I think the framework speaks for itself when you use it. But we also need a wordy introduction part of the docs as we do not have the luxury of letting the work speak for itself.

philpax:
Hi,

Fair enough - if you say it wasn't AI-written, that's good enough for me, but I'd still suggest making a few edits to stop people from drawing the same conclusion as me. My suggestions:
- remove all uses of the word "testament", which ChatGPT is deeply associated with
- reduce the number of dot-point lists with bolded keywords: while this is normally a great way of conveying information, and I do it myself, it's also very much associated with ChatGPT's output format
- rephrase some of the sentences to make them sound less generic; "In a landscape where Virtual DOM is often touted as a necessity, Aurelia confidently takes a different path" *feels* generic, even though I know what you're trying to go for
- remove some of the redundancy: you have two dot-points about community in the "Why choose Aurelia" list, and a "Embracing a Boutique Ecosystem" section. I'd consolidate these as much as possible. Ditto the two dot points about web standards adherence and the "The Power of Conventional DOM Manipulation" section.

> The intention with the superfluous parts of the docs was to humanise them. And while I agree that docs should be concise, this is an introduction, and we must sell ourselves a little harder. To understand why it's so wordy, you must understand Aurelia's position in the ecosystem.

I understand your position and sympathise, but I disagree that making it superfluous is the best way to humanise the docs. Your readers are more likely to glaze over the redundant bits and not pick up on the nuggets that are intended to convince them.

I think you'd fare better by tightening this up considerably and injecting personality into what's left - you make great points about the boutique ecosystem and tight access to the core team, for example, but that doesn't come through in the text as-is. You want to hook people with something that speaks to them and makes a compelling case for solving their problems.

An example that comes to mind is [HTMX](https://htmx.org/) ([docs](https://htmx.org/docs/)); while I have my disagreements with how they market themselves externally, they do a good job of conveying what they are, what problem they solve, and how they can solve your problems in a concise way.

Issue #1857, opened 2023-12-15
AUR3352: Unexpected state at swap of VPA


elitastic:
# 🐛 Bug Report

I get the following error when using the `Router` alias from router-lite:

`AUR3352: Unexpected state at swap of VPA`

I reproduced the behaviour in stackblitz:
https://stackblitz.com/edit/au2-conventions-e1suw9?file=src%2Fmy-app.ts,src%2Fmy-app.html,src%2Fmy-component.html,src%2Fmy-component.ts,src%2Fmain.ts

Shouldn't I use this aliases in router-lite? Thank you for any help.

## 🌍 Your Environment

Latest aurelia 2 dev version

PR #1859, opened 2023-12-16
fix(router-lite): Router injection and ignoring null/undefined values for query parameters

### LINKED ISSUE #1855: router-lite: null/undefined in optional parameters
# 🐛 Bug Report

Repro link https://stackblitz.com/edit/au2-conventions-nsmdrz?file=src%2Fmy-app.html

## 🤔 Expected Behavior

When passing null/undefined as optional parameters like:

`<a load="route: test; params.bind: { id: null }">Go to</a>`

I would expected the param to be `null` instead of `'null'`

## 😯 Current Behavior

null/undefined values in optional parameters are converted to "null"/"undefined"

## 🌍 Your Environment

Latest Aurelia 2 dev version

Sayan751:
# Pull Request

### 🎫 Issues

Closes #1855 and #1857

### DIFF
--- a/packages/__e2e__/playwright-util.js
+++ b/packages/__e2e__/playwright-util.js
@@ -12,6 +12,9 @@ module.exports = function getPlaywrightConfig(port, workers) {
     // }
     use: {
       baseURL: 'http://localhost:' + port,
-    }
+    },
+    expect: {
+      timeout: 10_000,
+    },
   }
 };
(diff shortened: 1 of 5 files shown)


bigopon:
Aee the tests covering the null/undefined issue? I dont see any tests that have a similar template to the reported issue.

Sayan751:
> Absence of data should be the less bad option probably?

I agree. If you don't see any issue with the changes, then maybe we can go forward and merge it.

bigopon (approved):

PR #1864, opened 2023-12-29
docs: attribute mapper

### LINKED ISSUE #1297: TODO: document core characteristics of how Aurelia works
From a number of occasions that folks asked similar questions related to attributes, properties & relevant bits in binding, it's necessary to add the following missing parts to the doc:

- Aurelia works with property not attribute
  - there's a mapper to map attribute to corresponding property (e.g label[for] -> label[htmlFor], td/td [colspan] -> colSpan], innerhtml -> innerHTML)
  - attributes are turned into camel-case equivalent properties by default if there's no mapping
  - how to extends the mapping (example https://github.com/aurelia/aurelia/blob/0efe8f914f4f6b426ac6115cb40e290fb8d164db/packages/runtime-html/src/attribute-mapper.ts#L24-L50)
- `.attr` binding command
- `& attr` binding behavior
- etc...

Most recent discourse topic https://discourse.aurelia.io/t/optional-bound-attributes/4386

Sayan751:
# Pull Request

### 🎫 Issues

Closes #1297


docs/user-docs/developer-guides/scenarios/attributemapper.md
```diff
@@ -0,0 +1,161 @@
...
+
+In a broad level, this is what attribute mapping is about. This article provides further information about how it works and how to extend it.
+
+## How it works
+
+To facilitate the attribute mapping, Aurelia uses `IAttrMapper`, which has the information about how to map an attribute to a property. While creating property binding instructions from [binding commands](./bindingcommand.md), it is first checked if the attribute is a bindable or not. If it is a bindable property, then the attribute name (in kebab-case) is converted to the camelCase property name. However, when it is not a bindable, then the the attribute mapper is queried for the target property name. If the attribute mapper returns a property name, then the property binding instruction is created with that property name. Otherwise, the standard camelCase conversion is applied.
+
+This means that if we want to bind a non-standard `<input>` attribute, such as `fizz-buzz`, then we can expect the `input.fizzBuzz` property to be bound. This looks as follows.
```
bigopon, line 33: We may need a different example as by default fizz-buzz is fizzBuzz (turned into camel case version)

bigopon (commented):
Nice work 👍

Vheissu (approved):
Beautiful stuff @Sayan751 I made a few little grammatical changes, nothing major.

Vheissu:
Oops, I didn't see your comment @bigopon sorry. I guess we can change this in a subsequent PR.

bigopon:
Yes we likely need to rework most of the examples as they are all camel case based, which is the automatic conversion.

We can try to use existing mapping as examples, i think thats fine too.

docs/user-docs/developer-guides/scenarios/attributemapper.md
Sayan751: This is an example of that default behaviour. Note that I have not added any custom attribute mapping so far.

Sayan751:
> we likely need to rework most of the examples as they are all camel case based

@bigopon I think we don't need to rework everything. I have found a typo and pushed the fix in #1865. I think that should do.

PR #1865, opened 2023-12-30
docs(attr-mapping): corrected typo

Sayan751:
# Pull Request

## 📖 Description

Corrected a residual change from https://github.com/aurelia/aurelia/pull/1864.


bigopon:
Its a bit subtle (unexpected case) but maybe itll do for now.

PR #1869, opened 2024-01-05
fix(router): store root/default page instruction correctly

### LINKED ISSUE #1812: router: popstate events are broken with empty routes ('')
# 🐛 Bug Report

When navigating back to the default route, the router does not correctly load the component associated with the empty route, but instead either leaves the viewport empty (root viewport), or keeps the component associated with the previous route (nested viewport).

Repro link https://stackblitz.com/edit/au2-conventions-kpnr9p, but the error can also be reproduced in the official routing example at https://simple-routing.vercel.app

## 🤔 Expected Behavior

When I navigate from the default route to another route, then return back with the browser's back button, the home component should be displayed.

When I navigate to the route with nested sub-routes, then to nested page 2, then go back with the browser's back button, the nested page 1 should be displayed.

## 😯 Current Behavior

When I navigate from the default route to another route, and return back with the browser's back button, nothing is displayed.

When I navigate to the route with nested sub-routes, then to nested page 2, then go back with the browser's back button, the nested page 2 is still displayed.

## 💁 Possible Solution

Right now, the only way I know to avoid the error is to not use default routes.

aegenet:
# Pull Request

## 📖 Description

When a user used the browser's "back" button, the root ("/") default page was not correctly reloaded.

### 🎫 Issues

Resolves #1812.

## 👩‍💻 Reviewer Notes

In the current code (without this PR), all routing instructions are always prepended with `-` (`updateNavigation`).

The issue is due to a simple fact: `-` + ` ` *(empty)* equals `-`. And `-` *(minus)* is used to **clear** the current path/page. Thus, when we navigate to `/`, then to another page, the root page is lost and replaced by `-`.

My current fix simply checks whether the instruction is empty, and if so, it prevent the clear instruction to be added.

## 📑 Test Plan

- Added new e2e tests for this specific case.
- Updated the existing tests to reflect a more realistic navigation flow.

### DIFF
--- a/packages/__e2e__/router/src/app.html
+++ b/packages/__e2e__/router/src/app.html
@@ -10,8 +10,8 @@
   <a load="child">Child</a>
   <a load="something/missing">Something missing</a>
 
-  <a load="pages/one-route">Pages one-route</a>
-  <a load="pages/two-route">Pages two-route</a>
+  <a id="page-one-link" load="pages/one-route">Pages one-route</a>
+  <a id="page-two-link" load="pages/two-route">Pages two-route</a>
 </nav>
 
 <au-viewport id="root-vp" name="root-vp" fallback="fallback" fallback-action="abort"></au-viewport>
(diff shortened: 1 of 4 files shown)


jwx (approved):
Thanks for this! Nice job with the test updates.

PR #1870, opened 2024-01-07
BREAKING CHANGE: string literal types in favour of const enums

### LINKED ISSUE #1736: Consumable enums should not be const enum - support TypeScript `verbatimModuleSyntax` (and JS projects)
# Background
When evaluating imports/exports, the TypeScript compiler needs to distinguish _type-only_ from _runtime_ imports/exports.
The compiler's default import elision strategy can not always evaluate this 100% correctly, and the compiled JS may end up with import/export statements of modules - even when only types from that module were actually referenced. 
When explicitly using the `type` keyword with imports/exports in the source code, the TypeScript compiler will evaluate this more correctly.

TypeScript 5.0 introduces a new compiler option `verbatimModuleSyntax` (that merges/deprecates some earlier options).
The idea of this flag is to require/enforce using the `type` keyword for __type-only__ imports and exports.
[Reference](https://devblogs.microsoft.com/typescript/announcing-typescript-5-0/#verbatimmodulesyntax)

I expect `verbatimModuleSyntax` to be the preferred/recommended compiler for TypeScript moving forward, since it more correctly can distinguish types from runtime entities, and avoid unwanted imports/exports in the compiled JS. And with this setting switched on, the TypeScript auto-import refactoring feature automatically includes the `type` keyword correctly when applicable.

# 🐛 Bug Report

When enabling this compiler option in a Aurelia2 project, I get the following compile error whenever a `const enum` from any Aurelia2 package is imported:

_`error TS2748: Cannot access ambient const enums when 'verbatimModuleSyntax' is enabled`_.

The problem is that `const enum` in TypeScript is a type-only enum - the enum itself is not emitted to the JS output and has no runtime representation. A `const enum` can only be used internally within a TypeScript project - since it "exists" at compile-time, but cannot be used by external consumers (f.ex. @aurelia/* package ) as the enums are not defined at runtime.
[Reference](https://www.typescriptlang.org/docs/handbook/enums.html#const-enums)

I therefore suggest that all enums meant to be accessible for @aurelia/* package consumers are turned into regular `enum` instead of `const enum`. (This should not be a too big job, a simple search for "const enum" in the aurelia source gave a manageable number of results.)

I understand that "regular" enums come with a minor size implication. Still; if consumers cannot switch `verbatimModuleSyntax` on, this may have greater bundle size implications for Aurelia2 TypeScript projects. And I guess the enums are not available at all for Aurelia2 JS projects, unless the compiler option `preserveConstEnums` is set to true.

Using "regular" enums is a better option than using const enums with the `preserveConstEnums` compiler option set to true - which I have seen that some packages tsconfig.json files use (i18n/runtime).

FYI: The enums that caused problems for me at this point are:
* LogLevel (packages/kernel/src/logger.ts)
* ColorOptions (packages/kernel/src/logger.ts)
* Signals (packages/i18n/src/utils.ts)

Sayan751:
# Pull Request

## 📖 Description

This PR introduces a BREAKING CHANGE, as it converts the following `export`ed `const enum`s to string literal types, as those are more developer friendly, and does not suffers from the issue outlined in #1736.

- @aurelia/dialog
    - DialogDeactivationStatuses
      ```diff
      - const enum DialogDeactivationStatuses {
      -   Ok = 'ok',
      -   Error = 'error',
      -   Cancel = 'cancel',
      -   Abort = 'abort',
      - }
      +  type DialogDeactivationStatuses = 'ok' | 'error' | 'cancel' | 'abort';
      ```
- @aurelia/kernel
  - ColorOptions
  ```diff
  - const enum ColorOptions {
  -   noColors = 0,
  -   colors = 1,
  - }
  + type ColorOptions = 'no-colors' | 'colors';
  ```
- @aurelia/platform
  - TaskStatus
  ```diff
  - const enum TaskStatus {
  -   pending = 0,
  -   running = 1,
  -   completed = 2,
  -   canceled = 3,
  - }
  + type TaskStatus = 'pending' | 'running' | 'completed' | 'canceled';
  ```
- @aurelia/router-lite
  - ExpressionKind
  ```diff
  - const enum ExpressionKind {
  -   Route,
  -   CompositeSegment,
  -   ScopedSegment,
  -   SegmentGroup,
  -   Segment,
  -   Component,
  -   Action,
  -   Viewport,
  -   ParameterList,
  -   Parameter,
  - }
  + type ExpressionKind = 'Route' | 'CompositeSegment' | 'ScopedSegment' | 'SegmentGroup' | 'Segment' | 'Component' | 'Action' | 'Viewport' | 'ParameterList' | 'Parameter';
  ```
- @aurelia/router
  - ReloadBehavior
  ```diff
  - const enum ReloadBehavior {
  -   default = 'default',
  -   disallow = 'disallow',
  -   reload = 'reload',
  -   refresh = 'refresh',
  - }
  + type ReloadBehavior = 'default' | 'disallow' | 'reload' | 'refresh';
  ```
- @aurelia/runtime-html
  - CommandType
  ```diff
  - const enum CommandType {
  -   None       = 0b0_000,
  -   IgnoreAttr = 0b0_001,
  - }
  + type CommandType = 'None' | 'IgnoreAttr';
  ```
  - DefinitionType
  ```diff
  - const enum DefinitionType {
  -   Element = 1,
  -   Attribute = 2,
  - }
  + type DefinitionType = 'Element' | 'Attribute';
  ```
  - ViewModelKind
  ```diff
  - const enum ViewModelKind {
  -   customElement,
  -   customAttribute,
  -   synthetic
  - }
  + type ViewModelKind = 'customElement' |  'customAttribute' |  'synthetic';
  ```
- @aurelia/runtime
  - ExpressionKind
  ```diff
  - const enum ExpressionKind {
  -   AccessThis,
  -   AccessGlobal,
  -   AccessScope,
  -   ArrayLiteral,
  -   ObjectLiteral,
  -   PrimitiveLiteral,
  -   Template,
  -   Unary,
  -   CallScope,
  -   CallMember,
  -   CallFunction,
  -   CallGlobal,
  -   AccessMember,
  -   AccessKeyed,
  -   TaggedTemplate,
  -   Binary,
  -   Conditional,
  -   Assign,
  -   ArrowFunction,
  -   ValueConverter,
  -   BindingBehavior,
  -   ArrayBindingPattern,
  -   ObjectBindingPattern,
  -   BindingIdentifier,
  -   ForOfStatement,
  -   Interpolation,
  -   ArrayDestructuring,
  -   ObjectDestructuring,
  -   DestructuringAssignmentLeaf,
  -   DestructuringAssignmentRestLeaf,
  -   Custom,
  - }
  + type ExpressionKind = 'AccessThis' | 'AccessGlobal' | 'AccessScope' | 'ArrayLiteral' | 'ObjectLiteral' | 'PrimitiveLiteral' | 'Template' | 'Unary' | 'CallScope' | 'CallMember' | 'CallFunction' | 'CallGlobal' | 'AccessMember' | 'AccessKeyed' | 'TaggedTemplate' | 'Binary' | 'Conditional' | 'Assign' | 'ArrowFunction' | 'ValueConverter' | 'BindingBehavior' | 'ArrayBindingPattern' | 'ObjectBindingPattern' | 'BindingIdentifier' | 'ForOfStatement' | 'Interpolation' | 'ArrayDestructuring' | 'ObjectDestructuring' | 'DestructuringAssignmentLeaf' | 'DestructuringAssignmentRestLeaf' | 'Custom';
  ```
  - ExpressionType
  ```diff
  - const enum ExpressionType {
  -           None = 0,
  -  Interpolation = 0b0_000001,
  -     IsIterator = 0b0_000010,
  -    IsChainable = 0b0_000100,
  -     IsFunction = 0b0_001000,
  -     IsProperty = 0b0_010000,
  -     IsCustom   = 0b0_100000,
  - }
  + type ExpressionType = 'None' | 'Interpolation' | 'IsIterator' | 'IsChainable' | 'IsFunction' | 'IsProperty' | 'IsCustom';
  ```
  - CollectionKind
  ```diff
  - const enum CollectionKind {
  -   indexed = 0b1000,
  -   keyed   = 0b0100,
  -   array   = 0b1001,
  -   map     = 0b0110,
  -   set     = 0b0111,
  - }
  + type CollectionKind = 'indexed' | 'keyed' | 'array' | 'map' | 'set';
  ```
- @aurelia/validation-html
  - ValidateEventKind
  ```diff
  - const enum ValidateEventKind {
  -   validate = 'validate',
  -   reset = 'reset',
  - }
  + type ValidateEventKind = 'validate' | 'reset';
  ```

### 🎫 Issues

Closes #1736.

## 👩‍💻 Reviewer Notes

There are still `const enum`s in code. These fall under the following categories:
- Used internally; not for public consumption.
- Flags enum, and hence a string literal type won't make sense.
- Used widely, and hence a replacement will cause larger bundle size.
- Ordinal enums like `LogLevel` from `@aurelia/kernel` (`LogLevel.warn < LogLevel.error`), where the string literal type would make the comparison needlessly difficult.

Here is a list of publicly consumable `const enums`:
- @aurelia/kernel: LogLevel
- @aurelia/runtime: AccessorType 
- @aurelia/runtime-html: InstructionType, BindingMode, State

The docs have been updated to include information for TypeScript users.


bigopon:
There may be enums that feel uncomfortable to express in string, as they used bits to express multiple characteristics. But probably we can try to ignore that since atm, its only the binding command type.
Ill do a simple check for size based on the examples/helloworld app.

packages/runtime-html/src/resources/binding-command.ts
```diff
@@ -29,18 +28,21 @@ import { AttrSyntax, IAttributeParser } from './attribute-pattern';
 import type { BindableDefinition } from '../bindable';
 import type { CustomAttributeDefinition } from './custom-attribute';
 import type { CustomElementDefinition } from './custom-element';
+import { dtElement } from './resources-shared';
+
+export const ctNone = 'None' as const;
+export const ctIgnoreAttr = 'IgnoreAttr' as const;
```
bigopon, line 34: These 2 can be internal. Also there doesn't seem to be more kinds of command types so this is enough I think 👍

packages/dialog/src/plugins/dialog/dialog-interfaces.ts
```diff
@@ -209,17 +209,15 @@ export interface IDialogCustomElementViewModel<T = unknown> extends ICustomEleme
   readonly $dialog: IDialogController;
 }
 
-export const enum DialogDeactivationStatuses {
-  Ok = 'ok',
-  Error = 'error',
-  Cancel = 'cancel',
+export type DialogDeactivationStatuses =
+    'ok'
```
bigopon, line 213: we can also start with `| ok`

packages/dialog/src/plugins/dialog/dialog-interfaces.ts
```diff
@@ -209,17 +209,15 @@ export interface IDialogCustomElementViewModel<T = unknown> extends ICustomEleme
...
+  | 'error'
+  | 'cancel'
   /**
    * If a view model refused to deactivate in canDeactivate,
    * then this status should be used to reflect that
    */
```
bigopon, line 219: an issue with dropping enum is the loss of this comment from usages. Devs wont be able to see suggestions like this anymore while selecting the values.
But maybe we can move all individual comments to the type instead if it works.

bigopon (approved):
Good work 👍 
Quite a tedious PR thanks @Sayan751 
For bundle size, it's actually few hundred bytes smaller for a helloworld app 😄

bigopon:
I think we can convert binding mode to numbers & type as well instead of enum.
```ts
const toView = 1;
const ...

/**
 * Comment explaining the modes here
 */
export type BindingMode =
  | typeof toView
  | typeof ...
```
It'll be fine I think as the places that use `BindingMode` are limited, and normally devs know what to be the values with literal primitive type easily if we have enough comments.

bigopon:
@Sayan751 I've merged it for now. For the comments, please address them in a follow up PRs, if you prefer to have me handle them, pls let me know. Nice work 👍

packages/runtime-html/src/resources/binding-command.ts
Sayan751: Nope. Tried that already. The '@internal' will remove the const declaration altogether, leaving the following string literal type in a weird situation:

```ts
// due to the removed const declaration, TS at the consumption side has
// no idea what `ctNone` or `ctIgnoreAttr` are.
export type CommandType =
  | typeof ctNone
  | typeof ctIgnoreAttr;
```

Sayan751:
> I think we can convert binding mode to numbers & type as well instead of enum.
> 
> ```ts
> const toView = 1;
> const ...
> 
> /**
>  * Comment explaining the modes here
>  */
> export type BindingMode =
>   | typeof toView
>   | typeof ...
> ```
> 
> It'll be fine I think as the places that use `BindingMode` are limited, and normally devs know what to be the values with literal primitive type easily if we have enough comments like this:
> 
> ```ts
> export interface BindableDefinition {
>   /**
>    * Some more comments here if really necessary <<<<<<<
>    */
>   mode: BindingMode;
> }
> ```
> 
> We can also export a literal `BindingMode` object for convenience:
> 
> ```ts
> export const BindingMode = {
>   toView = ...
> } as const;
> ```
> 
> So it'll be just as good compared to the enum.

This is a good idea. We can do this.

> @Sayan751 I've merged it for now. For the comments, please address them in a follow up PRs, if you prefer to have me handle them, pls let me know. Nice work 👍

I will do a follow-up PR.

Issue #1871, opened 2024-01-09
[🐛] Calling store.dispatch in sequence will only set the new state once


gtroxler:
When store.dispatch is queuing multiple dispatch requests it seems that only the first dispatch requests results in setting the new value. Then afterDispatch is called for the each of the queued dispatch events which seems never to call _setState.

How to reproduce, build a page which dispatches multiple actions which call a rest api, subscribe to state changes and observe that only the first one will published.

Assumption every store.dispatch call that changes the state will be published to subscribers.

Code in question: https://github.com/aurelia/aurelia/blob/master/packages/state/src/store.ts
Function: dispatch

bigopon:
This is actually a bug in `afterDispatch` and the bug is as you where you located in the PR at #1872 .
I'll be having a PR for this. Thanks for reporting.

The current behavior ignores the last `dispatch` call, which means if you have 2 dispatch call, and the handlers only return promise, it will give a feeling like the store only ever call `setState` once. You can validate this by calling dispatch 3 times, `setState` will be called twice instead.

PR #1872, opened 2024-01-11
disptach updated

Siddharth1605:
# Pull Request

## 📖 Description
• Identified the code responsible for queuing and executing dispatch requests
• Reproduced the issue by building a page that dispatches multiple actions
• Implemented a fix to ensure that all dispatch requests set the new value
• Tested the fix by reproducing the issue and verifying that all dispatch requests are published to subscribers
 
This change ensures that the recursive call is properly handled for both synchronous and asynchronous cases, and it also accounts for the case when there are no more queued dispatches.

### 🎫 Issues
#1871

### DIFF
--- a/packages/state/src/store.ts
+++ b/packages/state/src/store.ts
@@ -81,6 +81,9 @@ export class Store<T extends object, TAction = unknown> implements IStore<T> {
         } else {
           return afterDispatch(newState);
         }
+      }else {
+        // Ensure that in case of no queued dispatches, the result is returned
+        return $state instanceof Promise ? $state.then(() => {}) : undefined;
       }
     };
     const newState = reduce(this._state, action);


Siddharth1605:
> Thanks @Siddharth1605 can you add tests for this PR?

Here https://docs.aurelia.io/community-contribution/building-and-testing-aurelia there are 4 kinds of test to be done, as I'm a newbie to this community, can you say what test I need to do?

bigopon:
> > Thanks @Siddharth1605 can you add tests for this PR?
> 
> Here https://docs.aurelia.io/community-contribution/building-and-testing-aurelia there are 4 kinds of test to be done, as I'm a newbie to this community, can you say what test I need to do?

Right thanks for pointing that out and sorry for the confusion. There should be existing tests in a file state.spec.ts you only need to follow it to create new tests.
After that ill have a closer look at the issue and proceed with this PR/discussion.

packages/state/src/store.ts
```diff
@@ -81,6 +81,9 @@ export class Store<T extends object, TAction = unknown> implements IStore<T> {
         } else {
```
gtroxler, line 81: I would also set the state on line 80 (instead of just calling afterDispatch($), i would previously call set_state

packages/state/src/store.ts
```diff
@@ -81,6 +81,9 @@ export class Store<T extends object, TAction = unknown> implements IStore<T> {
...
           return afterDispatch(newState);
         }
+      }else {
+        // Ensure that in case of no queued dispatches, the result is returned
+        return $state instanceof Promise ? $state.then(() => {}) : undefined;
       }
     };
     const newState = reduce(this._state, action);
```
gtroxler, line 89: I also noticed that line 94 is incorrect. this._dispatching--; If I understand the code correctly, `this._dispatching--;` should only be called after the whole afterDispatch returns. `this._dispatching--;`  seems to work like a lock to ensure only one dispatch is happening and all the other requests are chained since the afterDispatch consumes the next item from the queue.

packages/state/src/store.ts
bigopon: the job of `afterDispatch` is to run the state through the whole action queue to get the final state.
Setting state should only happen after the whole action queue has been exhausted.

packages/state/src/store.ts
bigopon: I think you are right. The lines 89 - 97 currently looks like this
```ts
      return newState.then($state => {
        this._setState($state);
        this._dispatching--;

        return afterDispatch(this._state);
      }, ex => {
        this._dispatching--;
        throw ex;
      });
```
It seems to have 2 problems:
- calling `setState` too early, if there's more state changes in `afterDispatch` (this includes the wrong `this._dispatching--`)
- calling `setState` only once, it'll miss all the other changes in `afterDispatch`.

I think it should be changed to something like this:
```ts
      return newState.then($state => {

        return afterDispatch(this._state);
      }, ex => {
        this._dispatching--;
        throw ex;
      }).then(state => {
        this._setState(state);
        this._dispatching--;
      });
```

If you agree with this @gtroxler @Siddharth1605 , can you help with the tests for these changes?

bigopon: On a 2nd thought, maybe we should set state after every action so that it'll be easier & more correct to implement.

gtroxler: Yes that makes sense

PR #1875, opened 2024-01-11
fix(runtime-html): template wrapping 

Sayan751:
# Pull Request

## 📖 Description

This PR fixes the template wrapping for following templates:

```html
before
<template></template>
after
```

For this template, only the `template` element is considered, and the rest is ignored. This PR ensures that the whole template is wrapped with a `template` element.

### DIFF
--- a/packages/runtime-html/src/compiler/template-element-factory.ts
+++ b/packages/runtime-html/src/compiler/template-element-factory.ts
@@ -77,8 +77,7 @@ export class TemplateElementFactory {
       const prevSibling = node.previousSibling;
       if (prevSibling != null) {
         switch (prevSibling.nodeType) {
-          case 1: // Element
-            return true;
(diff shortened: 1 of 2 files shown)


bigopon (approved):
Nice work @Sayan751

packages/runtime-html/src/compiler/template-element-factory.ts
```diff
@@ -61,5 +61,41 @@ export class TemplateElementFactory {
...
+      // low-hanging fruit: check the next element sibling
+      const nextElementSibling = node.nextElementSibling;
+      if (nextElementSibling != null) return true;
+
+      // check the previous sibling
+      const prevSibling = node.previousSibling;
+      if (prevSibling != null) {
+        switch (prevSibling.nodeType) {
+          case 1: // Element
+            return true;
```
Sayan751, line 81: Remove dead code.

PR #1878, opened 2024-01-16
fix(validation): evaluation of tagged rules from bindings

Sayan751:
# Pull Request

## 📖 Description

This PR fixes the issue where a validate instruction with tags is used to validate the bindings. So far, such tags are ignored while collecting the rules from all the bindings. That results in validating all the bindings, irrespective of the tags provided with the instruction. This PR fixes that issue.

### DIFF
--- a/packages/validation/src/validator.ts
+++ b/packages/validation/src/validator.ts
@@ -22,17 +22,6 @@ export class ValidateInstruction<TObject extends IValidateable = IValidateable> 
     public objectTag: string = (void 0)!,
     public propertyTag: string = (void 0)!,
   ) { }
-
-  public static create<TObj extends IValidateable = IValidateable>(input: Partial<ValidateInstruction<TObj>>): ValidateInstruction<TObj> {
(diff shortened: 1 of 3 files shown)


packages/validation/src/validator.ts
```diff
@@ -22,6 +22,17 @@ export class ValidateInstruction<TObject extends IValidateable = IValidateable>
     public objectTag: string = (void 0)!,
     public propertyTag: string = (void 0)!,
   ) { }
+
+  public static create<TObj extends IValidateable = IValidateable>(input: Partial<ValidateInstruction<TObj>>): ValidateInstruction<TObj> {
```
bigopon, line 26: where are we using this? If it's only for testing, this should be a test util. Also, we don't have to follow this `create` pattern if it's not expected to be used directly in applications.

bigopon (approved):

packages/validation/src/validator.ts
Sayan751: That's a missed cleanup. Removed now.

PR #1880, opened 2024-01-21
chore: literal types instead of enums

Sayan751:
# Pull Request

## 📖 Description

This is a follow-up PR of #1870. In this PR, the rest of the publicly consumed enums are converted to literal union types.


bigopon:
Thanks @Sayan751 nice work. This frees up app TS config options.

PR #1882, opened 2024-01-24
Removed Call binding (some-prop.call="...") in documentation

ivanbacher:
Adding a `.call` binding will result is the following error:

> The ".call" binding command has been removed in v2. If you want to pass a callback that preserves the context of the function call, you can use lambda instead. Refer to lambda expression doc for more details.

Hence, I think it makes sense to remove it from the migrating to Aurelia 2 section


bigopon:
Thanks @ivanbacher

PR #1891, opened 2024-01-29
feat(event): ability to add modifier

### LINKED ISSUE #796: How to define custom event modifiers in Aurelia 2?
# 🙋 Feature Request

As recently mentioned in a [blog post](https://aurelia.io/blog/2019/10/31/aurelia-vnext-2019-fall-update), Aurelia 2 has wonderful extensibility capability. We can even simulate the syntax of other frameworks. This is really amazing.

Vue and Angular introduced a feature as `Event binding`. You can find some resources as the following links:

[Vue Event Modifiers](https://vuejs.org/v2/guide/events.html#Event-Modifiers)

```html
<form v-on:submit.prevent="onSubmit"></form>
```

[Vue Key Modifiers](https://vuejs.org/v2/guide/events.html#Key-Modifiers)

```html
<input v-on:keyup.page-down="onPageDown">
```

[Vue System Modifier Keys](https://vuejs.org/v2/guide/events.html#System-Modifier-Keys)

```html
<!-- Alt + C -->
<input @keyup.alt.67="clear">

<!-- Ctrl + Click -->
<div @click.ctrl="doSomething">Do something</div>
```

Even `Vue` lets you declare [custom key modifier](https://vuejs.org/v2/api/#keyCodes)

```ts
Vue.config.keyCodes = {
  v: 86,
  f1: 112,
  // camelCase won`t work
  mediaPlayPause: 179,
  // instead you can use kebab-case with double quotation marks
  "media-play-pause": 179,
  up: [38, 87]
}

<input type="text" @keyup.media-play-pause="method">
```

[Event Modifiers for Angular Applications](https://github.com/NetanelBasal/ngx-event-modifiers)

```ts
import { EventModifiersModule } from 'ngx-event-modifiers';

@NgModule({
  imports: [
    EventModifiersModule
  ]
})
export class AppModule { }

<button (click.stop)="onClick($event, extraData)">Click Me!!</button>
<button (click.prevent)="onClick($event, extraData)" type="submit">Click Me!!</button>
```

Is it possible to have/implement this capability in Aurelia 2? As a plugin or even a native feature?

### LINKED ISSUE #939: Allow AttrSyntax to carry parsed parts information
# 🙋 Feature Request

At the moment, attr syntax does not allow reuse, or support multiple parts. An example of this is the following:
```html
<div click.trigger.stop.prevent="invadeClick($event)"></div>
or
<div click.trigger:stop:prevent="invadeClick($event)"></div>
or
<div @click.stop.prevent="invalidClick($event)"></div>
```
With attr pattern:
```ts
  { pattern: 'PART.trigger.PART.PART', symbols: '.' },
```

The context of this is there needs to be some way to declare many configuration options of for attaching an event listener. An example of this is Chrome now  makes touch event non cancellable by default, unless you explicitly declare `passive: false`.

Maybe we could just extend AttrSyntax, or maybe we could do something else, but it seems having an extra property `parts` on attrsyntax seems to be the simplest option.

### LINKED ISSUE #1255: [RFC] Introduce on binding command
# 💬 RFC

While Aurelia settles on the `attribute.bind|delegate|trigger|call="bindable"` this RFC proposes the introduction of a secondary approach with an `on` binding command. 

## 🔦 Context

The purpose of this approach is to:

* Support of camelCase event names (e.g. custom events fired by the users app or e.g. consumed 3rd party web-components)
```html
<ion-input on="randoEvent: doSomething()">
```
* Multi event bindings
```html
// current
<div this.trigger="doSomething()" that.trigger="doMore()" bail.trigger="quit()">

// alternative with on
<div on="this: doSomething(); that: doMore(); bail: quit()">
```
* Allow custom expressions similar to i18n
```html
<div t="[html]...">
```

## 💻 Examples

The following examples would be open for debate:

* Dynamic composition with bind
```html
<div on.bind="bindable"></div>

// bindable -> randoEvent: doSomething()
```
* Binding to an object of type {[Key: EventName]: Callback}
```html
<div on="{ [myEventExpression1]: doThis(), [myEventNameExpression2]: doThat() }">
```
* Controlling initOptions of addEventListener
```html
<div on="this:{ passive: false, once: true }: doSomething(); that: doMore(); bail: quit()">
// leading to e.g.
<div on="drag:{ passive: false }: onDrag(); dragStart: $event.preventDefault(); dragover: $event.preventDefault()">
```
* Support Vue like key-specific event bindings `@keydown.esc="myfunction"`

bigopon:
## 📖 Description

Sometimes it's more desirable to pack more information in the event binding rather than event binding expression:
```html
<button click.trigger:ctrl="onCtrlClick()">
vs
<button click.trigger="onClick() & click:'ctrl'">
```

This PR adds the ability to declare modifier for event bindings, with 2 default target event groups: mouse and keyboard.

## Basic usage & syntax:
```
[event].trigger:[modifiers]="[expression]"
```
There can be multiple modifiers. When declaring multiple modifiers, use `+` as delimiter, example:
```html
<button keydown.trigger:enter+ctrl+shift="onShiftCtrlEnter()">Go</button>
```

### Prevent default and stop propagation

`preventDefault` and `stopPropagation` are two functions commonly called on any events. Event modifiers can be used to declaratively and easily call those functions, as per following example:

```HTML
<button click.trigger:stop:prevent="validate()">Validate</button>
```

### Mouse button modifiers

When handling mouse event, it sometimes requires a specific mouse button. By default, Aurelia provides 3 modifiers `left`, `middle` and `right` to support mouse button verification. An example is as follow:

```HTML
<button click.trigger:middle="newTab()">Open in new tab</button>
```

### Keyboard mapping

When using keyboard event modifier, sometimes a certain key is used as modifier.
You can use the char code representing the key as the modifier, like the following example, where we want to handle the combo `Ctrl + K` (notice its the upper `K`):

```HTML
<textarea keydown.trigger:ctrl+75="openSearchDialog()">
```
`75` is the charcode of the upper case letter `K`.

Even though direct, it's not always clear `75` means when looking at the template, so it's often desirable to use the real letter `K` instead.
Though Aurelia is not taught to, by default, understand the letter `K` means the code `75`. You can teach Aurelia by adding to the `IKeyMapping`:

```typescript
import { AppTask, IKeyMapping } from 'aurelia';

Aurelia.register(
  AppTask.creating(IKeyMapping, mapping => {
    mapping.keys.upper_k = 75;
  })
)
```
After this enhancement, the `:ctrl+upper_k` modifier will be understood as `ctrl+75`.

Note that we cannot use the upper case letter `K` as modifier in HTML because HTML is case insensitive. We use, in this example, `upper_k` as an alternative for that, and add the mapping accordingly.

> By default, Aurelia provides mapping for all lower case a-z letters in both keycode and leter so both `:ctrl+a` and `:ctrl+97` works.
For upper case letter, only keycode mapping is provided, for example: `:65` for upper letter A works.

Closes #796
Closes #939
Close #1255 - the main features wanted in this issue are covered enough with this PR


packages/__tests__/src/3-runtime-html/custom-elements.spec.ts
```diff
@@ -136,35 +136,147 @@ describe('3-runtime-html/custom-elements.spec.ts', function () {
...
+      const { trigger } = createFixture(
+        '<button @keydown:enter+ctrl+shift="enter()"></button>',
+        { enter: () => entered = 1 },
+        [ShortHandBindingSyntax]
+      );
+      trigger('button', 'keydown', { key: 'enter', ctrlKey: true });
+      assert.strictEqual(entered, 0);
+      trigger('button', 'keydown', { key: 'enter', ctrlKey: true, shiftKey: true });
+      assert.strictEqual(entered, 1);
+    });
```
Sayan751, line 279: Will it be possible to combine `prevent` and/or `stop` with other modifiers? Or even with each other?

Seeing the code, it seems that it is possible. Can you please add a couple of tests to assert that?

Sayan751 (commented):
Awesome work @bigopon! Just have one comment regarding test.

PR #1895, opened 2024-02-01
fix(router): prevent multiple navigation at the same time

### LINKED ISSUE #1860: Aurelia Router - Duplicate content
# 🐛 Bug Report

With `@aurelia/router`, a quick click (in a real application) can result in duplicate content.

Repro link https://gist.dumber.app/?gist=2dd2eeb92319dde10108cc26161b776a

## 🤔 Expected Behavior

No duplications

## 😯 Current Behavior

Duplicate content

## 💁 Possible Solution

It's not directly related to the `portal attribute`, the `@aurelia/router-lite` doesn't have that bug (*but... it does have a potential infinite loop issue* see [Demo 2](https://gist.dumber.app/?gist=c0f7eca5f8f48aa8a5c9cad68bce68f5)).

It looks very much like a timing problem.

## 🔦 Context

In my production application, when a user clicks quickly from one page to another, the content can be duplicated (outch).

Even more so when they use the mouse to go back or forward.

## 💻 Code Sample

https://gist.dumber.app/?gist=2dd2eeb92319dde10108cc26161b776a

## 🌍 Your Environment

| Aurelia          |  2.0.0-beta.8/2.0.0-beta.9
| Language         | TypeScript
| Browser          | Chrome 120.0.6099.72 / Firefox 121.0
| Bundler          | Dumber / Webpack
| Operating System | Windows 10/11
| NPM/Node/Yarn    |

aegenet:
# Pull Request

## 📖 Description

When a user clicks quickly from one page to another, the content can be duplicated.

### 🎫 Issues

Resolves #1860.

## 👩‍💻 Reviewer Notes

I prevent multiple navigations from occurring simultaneously and store the most recent one. This ensures that users don't find it odd to end up on their last desired page.

I was delayed in creating this PR because I noticed another issue with the error handling in @aurelia/router. However, it requires a separate issue/PR, as it's not just linked to this current issue.

## 📑 Test Plan

- Added new integration/e2e tests for this specific case.


bigopon:
@aegenet , thanks for the PR. It seems the fix is not always working, I can see this with the e2e tests:

image

That test result means there's something that pulls back the URL from `two-route` to `one-route`, based on this code
```ts
    await Promise.all([
      page.waitForURL(`${baseURL}/pages/two-route`),
      page.click('#page-two-link'),
    ]);
```

I've checked the fix, it seems ok, so I'm not sure what could be the reason.

cc @jwx

bigopon:
When running locally, without the commits from master which has the changes at #1900, I also experienced the same failure, even though it's really rare. Not even sure whether this is a playwright thing.

aegenet:
@bigopon  I remember there was an issue with the unit test. I think it's the same issue... and yes, it's the same: Playwright doesn't wait for the end of the navigation, so we need to wait for the end of all navigation before navigating back to the last requested page.

bigopon:
@aegenet thanks for checking that.
> Playwright doesn't wait for the end of the navigation

Can this actually affect the test? Once the url has been changed to `two-route`, can it go back to `one-route` via something inside a navigation (or latest navigation), briefly or eventual back to `two-route`?

aegenet:
Yes, because we do a lot of navigation (page-one-link, `page-two-link`, & auth-link etc), so after, when we execute the following statement:

```typescript
await Promise.all([
  page.waitForURL(`${baseURL}/pages/two-route`),
  page.click('#page-two-link'),
]);
```

The:

```typescript
page.waitForURL(`${baseURL}/pages/two-route`)
```
 can be swallowed by the previous loop "click"

bigopon:
That sounds like inside a navigation the URL would be updated somehow, I'm not sure on this, but the fix isn't doing any related to that behavior so I think it's ok. Thanks @aegenet.

packages/__e2e__/router/src/router-configured.spec.ts
```diff
@@ -155,4 +157,90 @@ test.describe('router', () => {
     expect(page.url()).toBe(`${baseURL}/`);
     await expect(page.locator('#root-vp')).toContainText('Home page');
   });
+
+  for (let i = 0; i < 5; i++) {
```
bigopon, line 161: what is this loop for?

aegenet: This loop helps me reproduce the navigation issue _(wait for pending nav_) every time; without it, the issue occurs only once in 100 times (with my computer). I can remove it if you want.

bigopon: Right ok I think that's good to have, it wouldn't affect the test runtime much anyway. I'll add a comment to it to explain what it is. Thanks

bigopon:
Nice work, thanks @aegenet

bigopon:
If you wanna test this out, our dev version should be on npm in about 15 mins @aegenet

aegenet:
It's ok with the dev version: https://gist.dumber.app/?gist=93e7586ed697b010ed48517e7ad3c423

No more 'clone wars', thanks for your review and involvement.

bigopon:
Thank you too @aegenet for the work and the confirmation 🙏

PR #1900, opened 2024-02-05
fix(router): properly handle false in conditional router hooks

jwx:
# Pull Request

## 📖 Description

This PR fixes issues that could occur when several denying async router hooks were applied to the same component.

## 📑 Test Plan

- Add tests for denying async router hooks


packages/router/src/endpoints/viewport-content.ts
```diff
@@ -343,38 +326,35 @@ export class ViewportContent extends EndpointContent {
...
+    if (instance.canUnload != null) {
+      hooks.push((innerStep: Step | null) => {
+        if (innerStep?.previousValue === false) {
           return false;
         }
-        return hooksResult as Promise<boolean>;
+        // TODO: If requested, pass previous value into hook
+        const result = instance.canUnload?.(this.instruction, navigation);
+        if (typeof result !== 'boolean' && !(result instanceof Promise)) {
+          throw new Error(`Method 'canUnload' in component "${this.instruction.component.name}" needs to return true or false or a Promise resolving to true or false.`);
```
bigopon, line 344: if we are throwing when it's not a boolean, don't we need to await the resolved value of the promise to ensure that it's also a boolean?

packages/router/src/endpoints/viewport-content.ts
```diff
@@ -343,38 +326,35 @@ export class ViewportContent extends EndpointContent {
...
-        if (result === false) {
-          innerStep.exit();
-        }
-        return result;
+    const hooks = this.getLifecycleHooks(instance, 'canUnload').map(hook => ((innerStep: Step | null) => {
+      if (innerStep?.previousValue === false) {
+        return false;
       }
-      return result as Promise<boolean>;
+      return hook(instance, this.instruction, navigation);
```
bigopon, line 333: the main fix looks good to me, though why are we moving the handling from here to the block below?
Shouldn't we be able to just say
```ts
if (result === false) {
  innerStep.exit();
}
return result;
```

bigopon (commented):
Looking good @jwx I have a few small comments for now.

packages/__tests__/src/router/router.spec.ts
```diff
@@ -1775,47 +1775,416 @@ describe('router/router.spec.ts', function () {
...
+          return expected;
+        }
+        expected.push('Hooks:my-two:loading', 'VM:my-two:loading', 'VM:my-two:binding');
+
+        expected.push('Hooks:my-two:canUnload');
+        if (!config.Hooks_canUnload) return expected;
+        expected.push('VM:my-two:canUnload');
+        if (!config.Two_canUnload) return expected;
+        expected.push('Hooks:my-one:canUnload');
+        if (!config.Hooks_canUnload) return expected;
```
bigopon, line 1955: @jwx this cannot happen as the condition is already covered by the line 1951 above. There' also other similar conditions, per our deepscan report. Can you have a look https://deepscan.io/dashboard/#view=project&tid=3446&pid=5117&bid=39795&subview=pull-request&prid=1839539

PR #1904, opened 2024-02-14
rename load() hook to loading() in docs

elitastic:
Docs adjusted so that the hook names correspond to those of the router and router-lite


bigopon:
thanks @elitastic

Issue #1806, opened 2023-07-25
templating engine drops attributes from au-compose tag


ekzobrain:
# 🐛 Bug Report

When using au-compose with additional html attributes (like style for example) in template - they are dropped in output HTML.
It worked in earlier versions.

Repro link https://stackblitz.com/edit/au2-conventions-peynua?file=src%2Fmy-app.html,src%2Fmy-app.ts,src%2Fmy-app.css

## 🤔 Expected Behavior

output html is like this:
```html
<au-compose style="width: 20%"></au-compose>
<au-compose style="width: 30%"></au-compose>
```

## 😯 Current Behavior

output html is like this:
```html
<au-compose></au-compose>
<au-compose></au-compose>
```

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Aurelia          | 2.0.0-beta.8
| Language         |
| Browser          |
| Bundler          |
| Operating System |
| NPM/Node/Yarn    |

bigopon:
Thanks @ekzobrain , the style attribute should also be transferred to the host element of the composed custom element. This is unintentional.

ekzobrain:
Yes, in case of rendering a custom element - it works. But does not work with `containerless` feature (in this case i think attributes whould remain on `au-compose`, but they just disappear).
`au-compose` also seems not to work with non-custom element (with convensions) at all. Nothing gets renderer in this case

bigopon:
> non-custom element at all. Nothing gets renderer in this case.

Non custom elements don't get the same treatment with custom elements, so the `static template` part wasn't used. Were you expecting it to work as if it was a custom element?

> attributes should remain on au-compose, as with non-custom element composition

We could, but this will make it incredibly complex I think. Shouldn't it be simpler for apps to have
```html
<au-compose containerless ...>
```
?

ekzobrain:
I think you need an RFC about dynamic composition behavior, which will be transferred to docs afterwards :) The first thing to note is that all dynamic composition parts (au-compose, dialog, au-viewport) should work the same regarding:
1. explicit custom elements / custom elements with convensions / non-custom elements / etc (in fact I do not know exact differences between those)
2. containerless support
3. attribute transferring support

This issue is also relative #1695 

I would try to extract a general dynamic composition function to the core, and then reuse it where applicable. This will result in:
1. more reliable code
2. more convinient overall behavior
3. less code
4. better framework extension capaibilities for plugin authors

[...]

PR #1906, opened 2024-02-18
refactor(au-compose): always create host for non custom element composition

bigopon:
## 📖 Description

Currently the `<au-compose/>` composition tries to be smart when composing non-custom element:
- if the `<au-compose/>` usage is not `containerless` (via `<au-compose containerless/>` usage): it'll use its own host element (`<au-compose/>` element) as the host for the composition
- if the `<au-compose/>` usage is a containerless: it'll use its comment markers as the comment markers for the composition

This has a few issues:
- unclear & hard to understand behavior, as questioned by #1806
- complicated implementation

This PR refactors the `<au-compose/>` composing behavior:
- `<au-compose/>` itself will always be containerless
- non custom element host element is always created, but will be replaced by a marker if there's no host element name specified
    - For example, the following template
    ```html
    <au-compose component.bind="{ a: 1 }" template="<div>hey</div>">
    ```
    will result in rendered composed html (ignore the `<au-compose/>` related own html):
    ```html
    <!--au-start--><div>hey</div><!--au-end-->
    ```
    This is often the desired behavior, but sometimes it's also necessary to have a wrapping element for regardless what the template for the composition is, imagine composing form fields wrapping in a field container:
    Usage:
    ```html
    <au-compose component.bind="{  name: firstName }" template="<input class='field-control' value.bind='name'>">
    ```
    Desirable output:
    ```html
    <div>
      <input class="field-control">
    </div>
    ```
    This can be achieved via `tag` binding on the `<au-compose>` element. `tag` is a new bindable property of the `<au-compose>` element and is used to specify what element to create in the non custom element composition. For the above example, we can do:
    ```html
    <au-compose
        component.bind="{  name: firstName }"
        tag="div"
        template="<input class='field-control' value.bind=name>">
    ```
    All the extra bindings on the `<au-compose/>` usage will also be transferred to this host element, which means we can do
    ```html
    <au-compose tag="div" class="field-container" template="...">
    ```
    to have the following composed html:
    ```html
    <div class="field-container">...
    ```


bigopon:
cc @Sayan751 @fkleuver @ekzobrain

ekzobrain:
@bigopon give me a couple of days, please. Need to dive in again.

bigopon:
doc change is here https://github.com/aurelia/aurelia/pull/1906/files#diff-8cd6af03635c2537aa4b1ff7d67527c37afa3205f4e823d8c79620fe9cfcea54R62

bigopon:
@ekzobrain no worries. I'll go ahead first, will accommodate any issues we discover

PR #1907, opened 2024-02-19
[docs]  Added JS examples using resolve for IHttpClient

ivanbacher:


bigopon (approved):
Nice work, thanks @ivanbacher

PR #1916, opened 2024-03-02
feat(enhance): call app tasks with `.enhance` API

### LINKED ISSUE #1787: [Feature request] Enhance should call app tasks
# 🙋 Feature Request

Currently, the enhance functionality does not fire app tasks, which means you cannot use the router with enhance (possibly other things that rely on app tasks).

## 🤔 Expected Behavior

App tasks should be called regardless of whether you're starting an entire instance or using enhance.

## 😯 Current Behavior

App tasks are not fired.

bigopon:
## 📖 Description

In a PR awhile ago, the return value of the `enhance` call was changed from an `IAppRoot` to a controller for simplicity. This was necessary at the time because the app root created by enhance would override an existing app root on an Aurelia instance. This , however, resulted in the loss of the ability to call app tasks in the `.enhance` call.

This PR makes `.enhance` create an app root again, but this app root would not be used by the owning Aurelia instance.

Changes:
- BREAKING CHANGE: enhance results in an app root instance instead of a controller
- improve typings of `.enhance` & `IAppRoot` to auto infer component instance type
- script dist target changed from es2020 to es2021
- remove `.create` API on container

Resolves #1787 

cc @Sayan751 @Vheissu

### DIFF
--- a/docs/user-docs/getting-to-know-aurelia/enhance.md
+++ b/docs/user-docs/getting-to-know-aurelia/enhance.md
@@ -27,13 +27,13 @@ Key Points to Understand:
 1. **Anonymous Custom Element Hydration:** The enhancement treats the target node as an anonymous custom element, allowing Aurelia to apply its behavior to the existing DOM structure.
 2. **Component Flexibility:** The component parameter in enhance can be a custom element class, a class instance, or an object literal. If a class is provided, it's instantiated by Aurelia's dependency injection container, which can be either provided or automatically created.
 3. **Host Element:** The host is typically an existing DOM node that is not yet under Aurelia's control. It's crucial to note that `enhance ' neither detaches nor attaches the `host` to the DOM. Existing event handlers on the `host` or its descendants remain unaffected.
-4. **Controller Deactivation:** an `enhance` call results in a custom element controller that requires manual deactivation or integration into an existing controller hierarchy for automatic framework management.
+4. **Controller Deactivation:** an `enhance` call results in an application root that requires manual deactivation or integration into an existing controller hierarchy for automatic framework management.
 
 Example of deactivating a controller:
 
 ```typescript
-const controller = au.enhance({ host, component });
-controller.deactivate(controller, null, LifecycleFlags.none);
+const root = au.enhance({ host, component });
+root.deactivate();
 ```
 
 ## Enhancing During Application Startup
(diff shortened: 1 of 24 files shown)


ghiscoding:
@bigopon 
this code change seems to be breaking my lib Aurelia-Slickgrid and I don't know how to fix it. I got a Stackblitz available in my opened PR to update Aurelia to latest in this [PR](https://github.com/ghiscoding/aurelia-slickgrid/pull/1166) and going to "Example 19 - Row Detail" will display the error. 

I'm enhancing dynamic components on the fly and my current code is as shown below (ref of original code [here](https://github.com/ghiscoding/aurelia-slickgrid/blob/master/packages/aurelia-slickgrid/src/services/aureliaUtil.service.ts))
```ts
@singleton()
export class AureliaUtilService {
  constructor(@IAurelia private readonly au: IAurelia) { }

  async createAureliaViewModelAddToSlot(viewModel: Constructable, bindableData?: ViewModelBindableInputData, targetElement?: HTMLElement | Element): Promise<AureliaViewOutput | null> {
    if (!targetElement) {
      return null;
    }

    const def = CustomElement.getDefinition(viewModel);
    const addonBindable = bindableData?.addon ? 'addon.bind="bindableData.addon"' : '';
    const gridBindable = bindableData?.grid ? 'grid.bind="bindableData.grid"' : '';
    const dataViewBindable = bindableData?.dataView ? 'data-view.bind="bindableData.dataView"' : '';
    const parentBindable = bindableData?.parent ? 'parent.bind="bindableData.parent"' : '';

    targetElement.innerHTML = `<${def.name} model.bind="bindableData.model" ${addonBindable} ${gridBindable} ${dataViewBindable} ${parentBindable}></${def.name}>`.trim();

    return { controller: await this.au.enhance({ host: targetElement, component: { bindableData } }) };
  }
}
```
with your PR change, my previously working code is now throwing with this weird error, I have no clue why it fails with Router !??
```sh
index.mjs:4789 Uncaught (in promise) Error: Router has already been started
    at Router.start (index.mjs:4789:1)
    at RouterConfiguration.configurationCall (index.mjs:5755:1)
    at $AppTask.run (index.mjs:525:1)
    at eval (index.mjs:4966:1)
    at Array.reduce (<anonymous>)
    at AppRoot.ae (index.mjs:4964:1)
    at eval (index.mjs:4958:1)
    at onResolve (index.mjs:259:1)
    at AppRoot.activate (index.mjs:4958:1)
    at Aurelia.enhance (index.mjs:5039:1)
```
if I change my code with the following code (taken from your PR)
```ts
constructor(private readonly au = resolve(Aurelia)) {}
```
then it fails with this different error which is not very helpful to me
```sh
index.mjs:19 Uncaught (in promise) Error: AUR0010:function Object() { [native code] }
    at createMappedError (index.mjs:19:1)
    at Container.L (index.mjs:717:1)
    at Container.get (index.mjs:580:1)
    at Container.containerGetKey (index.mjs:779:1)
    at Array.map (<anonymous>)
    at Factory.construct (index.mjs:751:1)
    at Resolver.resolve (index.mjs:1201:1)
    at Container.get (index.mjs:581:1)
    at Container.containerGetKey (index.mjs:779:1)
    at Array.map (<anonymous>)
```
I'm confused into what I'm supposed to change into my lib to get it working again. If you could help by opening my Stackblitz from my open [PR](https://github.com/ghiscoding/aurelia-slickgrid/pull/1166) and opening `aureliaUtil.service.ts` and tell me what I'm supposed to change, it would be really helpful. If not, I might just stop supporting Aurelia-Slickgrid altogether or maybe just that enhancing part

bigopon:
Thanks for the issue @ghiscoding , this is unintentional, I've commented here https://github.com/ghiscoding/aurelia-slickgrid/pull/1166/files#r1517169458

ghiscoding:
Thanks a lot, the last suggestion was the one I went ahead with since it work the same as before. However, I could not find such approach in the Documentation, it might be a good idea to add it to the official Docs. Thanks again for your help, you are always helpful as usual :)

PR #1917, opened 2024-03-02
feat(dev): better DI error messages for instantiation

### LINKED ISSUE #913: DI tracing - tell me what thing is trying to be injected into
errors like this are almost completely useless
```
error:    Error: Attempted to jitRegister an intrinsic type: Object. Did you forget to add @inject(Key)
    at Container.jitRegister (/Users/calebcushing/IdeaProjects/service-graph/node_modules/@aurelia/kernel/src/di.ts:1132:13)
    at Container.get (/Users/calebcushing/IdeaProjects/service-graph/node_modules/@aurelia/kernel/src/di.ts:1066:27)
    at Object.invoke (/Users/calebcushing/IdeaProjects/service-graph/node_modules/@aurelia/kernel/src/di.ts:758:35)
    at Factory.construct (/Users/calebcushing/IdeaProjects/service-graph/node_modules/@aurelia/kernel/src/di.ts:698:22)
    at Resolver.resolve (/Users/calebcushing/IdeaProjects/service-graph/node_modules/@aurelia/kernel/src/di.ts:632:30)
    at Container.get (/Users/calebcushing/IdeaProjects/service-graph/node_modules/@aurelia/kernel/src/di.ts:1067:27)
    at Resolver.resolve (/Users/calebcushing/IdeaProjects/service-graph/node_modules/@aurelia/kernel/src/di.ts:650:24)
    at buildAllResponse (/Users/calebcushing/IdeaProjects/service-graph/node_modules/@aurelia/kernel/src/di.ts:1347:29)
    at Container.getAll (/Users/calebcushing/IdeaProjects/service-graph/node_modules/@aurelia/kernel/src/di.ts:1095:16)
    at Framework.after (/Users/calebcushing/IdeaProjects/service-graph/src/shared/injection/Framework.ts:91:34)
    at Framework.run (/Users/calebcushing/IdeaProjects/service-graph/src/shared/injection/Framework.ts:121:25)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
```

I'd like to be able to provide a flag that could tell me what the parent (bonus points if you can also tell me the position/signature) of what is trying to be injected. I'm fine with this being behind a flag as @fkleuver told me it'd be expensive.

bigopon:
## 📖 Description

Currently, when there is an error during the instantiation of a class, there is not enough information what went wrong. This is quite an issue once there are multiple changes being added in at once, like in a dependency upgrade.

Enhance the DEV mode error message by catching the dependency resolution error while instantiate, so that there will be more information for debugging.

### Examples

1. for the following code with `inject`:
    ```ts
    
    export class App {
      static inject = [class Abc {
        constructor() {
          throw new Error('testing testing');
        }
      }]
    }
    ```
    The error messages logged will look like this:
    
    image

2. For the following code with `resolve`:
    ```ts
    
    export class App {
      abc = resolve(class Abc {
        constructor() {
          throw new Error('testing testing');
        }
      });
    }
    ```
    
    The error messages logged will look like this
    image

### 🎫 Issues

Resolves #913 

cc @xenoterracide @Sayan751 @fkleuver

### DIFF
--- a/packages-tooling/rollup-utils.mjs
+++ b/packages-tooling/rollup-utils.mjs
@@ -123,7 +123,8 @@ export function getRollupConfig(pkg, configure = identity, configureTerser, post
     input: inputFile,
     external: Object.keys(pkg.dependencies ?? {})
       .concat(Object.keys(pkg.devDependencies ?? {}))
-      .concat('os', 'path', 'fs', 'http', 'https', 'http2', 'url', 'stream'),
+      .concat('os', 'path', 'fs', 'http', 'https', 'http2', 'url', 'stream')
+      .concat('node:module', 'node:path'),
     output: [
       {
         file: esmDist,
--- a/packages-tooling/vite-plugin/src/index.ts
+++ b/packages-tooling/vite-plugin/src/index.ts
@@ -2,6 +2,9 @@ import { IOptionalPreprocessOptions, preprocess } from '@aurelia/plugin-conventi
 import { createFilter, FilterPattern } from '@rollup/pluginutils';
 import { resolve, dirname } from 'path';
 import { promises } from 'fs';
+import { createRequire } from 'module';
+
+const require = createRequire(import.meta.url);
 
 export default function au(options: {
   include?: FilterPattern;
@@ -25,7 +28,7 @@ export default function au(options: {
   const devPlugin: import('vite').Plugin = {
     name: 'aurelia:dev-alias',
     config(config) {
-      const isDev = useDev || (!useDev && config.mode !== 'production');
+      const isDev = useDev === true || (useDev == null && config.mode !== 'production');
       if (!isDev) {
         return;
       }
--- a/packages-tooling/vite-plugin/tsconfig.json
+++ b/packages-tooling/vite-plugin/tsconfig.json
@@ -1,6 +1,7 @@
 {
   "extends": "../tsconfig-build.json",
   "compilerOptions": {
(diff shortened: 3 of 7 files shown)


Sayan751 (commented):
Thanks @bigopon for this nice change! I think most of the devs won't use the dev/builds by default. Hence, I think we should have the same information also for non-dev builds.

bigopon:
@Sayan751 if a serious app was to be public, it needs to be first developed in dev mode before a prod build, and thus, error like this will be prominent and get ridden of, so this scenario we don't need these errors in prod mode.
If an app was a non serious one, it can just use dev build for both development and production, and will still be fine.
I think we shouldn't carry this experience enhancement into our prod build.

Sayan751:
In that case, I think we should have "recipe" documentation on how to enable the dev mode. As it might not always be straightforward.

bigopon:
@Sayan751 thanks, and agree, Ive been thinking of an faq doc or something to describe these.

tareqimbasher:
@bigopon @Sayan751 This is a great addition but can't find any documentation on how to use dev mode. Can you provide some info on how I would go about that? Thank you!

Sayan751:
@tareqimbasher Here is an example how to use development builds: https://github.com/aurelia/new/blob/master/webpack/webpack.config.js#L120

tareqimbasher:
@Sayan751 Great, thank you! Plugged in those aliases and its working.

PR #1920, opened 2024-03-07
feat(convention): add import as support

### LINKED ISSUE #622: Local Resource Renaming

bigopon:
## 📖 Description

In v1, we could do this
```html
<require from="..." as="...">
```

But it v2, this was a bit different, we need to declare alias on the resource itself:
```ts
@customElement({
  ...,
  aliases: ['name-1', 'name-2']
})
export class MyElement {
  ...
}
```
While this new way of declaring alias is useful in some cases, it doesn't cover the scenario often seen in v1: aliasing an resource and only register the aliased name. This PR adds back the support for this `<import as>` in our convention, with the following syntax:

```html
<import from="my-resource" as="the-name">
<import from="my-element.html" as="the-element">
<import from="multi-resource-module" resource-1.as="res1" resource-2.as="res2">
```

### 🎫 Issues

Resolves #622 

cc @brandonseydel @3cp


3cp (approved):

bigopon:
Just FYI @brandonseydel @3cp, I didn't go with export keys renaming as it requires uppercase letters, which are not valid  in html.
Most of the time, we expect the main (single resource export) to be aliased anyway. If there' enough request/cases for multiple resource exports renaming via key, we can add something like this later:
```html
<import from="./multi-exports" as="{ Key1: 'Somehing', Key2: 'something-else' }">
```

PR #1924, opened 2024-03-13
feat(template-controller): ability to have a container per factory

### LINKED ISSUE #1923: Replacing view factory containers
# 🙋 Feature Request
Using custom template controllers to inject dependencies into parts of a template would be a useful feature. The example below currently doesn't work because all of the view factories share the same container when used in the same template. This is fine for most template controller use cases, but it prevents this one from working correctly:
```ts
const IExample = DI.createInterface<IExample>("IExample");

class ExampleTemplateController {
   @bindable value;
   viewFactory = resolve(IViewFactory);
   location = resolve(IRenderLocation);

   bound() {
      viewFactory.container.register(Registration.instance(IExample, this.value));
      const view = this.viewFactory.create();
      view.setLocation(this.location);
   }
}
```
```html
<!-- IExample should not exist here. -->
<template example="a">
   <!-- IExample should resolve to "a" -->
   <template example="b">
      <!-- IExample should resolve to "b" -->
   </template>
</template>
```

## 💁 Possible Solution
There should be an option to replace the view factory container before calling `viewFactory.create()` which would allow creating a new child container:
```ts
viewFactory.container = viewFactory.container.createChild();
// or something like this:
viewFactory.useContainer(viewFactory.container.createChild());

viewFactory.container.register(Registration.instance(IExample, this.value));
const view = viewFactory.create();
```

## 🔦 Context
We are currently just overwriting the view factory container with a new child container although the `container` property is marked as readonly. This seems to work fine for now as the container is never used before `viewFactory.create()` is called:
```tsx
(viewFactory as any).container = viewFactory.container.createChild();
```

bigopon:
## 📖 Description

Currently template controllers reuse the container from their owning renderer (custom element), as an optimization, also a requirement since they need to retrieve the resources information from there. This works well for all built in template controllers as they don't need to do anything special with their view factories.
Though, for application custom controller, sometimes it's more desirable to have separate container for dynamic DI work.

This PR adds a new option `containerStrategy` for custom attribute definition to enable this. The acceptable values are `reuse` and `new`, and the default is `reuse`, which is the existing behavior.
image

### 🎫 Issues

Resolves #1923

Thanks @mxjp

### DIFF
--- a/docs/user-docs/advanced-scenarios/css-in-js-with-emotion.md
+++ b/docs/user-docs/advanced-scenarios/css-in-js-with-emotion.md
@@ -79,7 +79,7 @@ export class MyApp {
 
 Go to your view and add emotion custom attribute to an HTML tag.
 
-```markup
+```html
 <div class="message" emotion.bind="cssObject">${message}</div>
 ```
 
--- a/docs/user-docs/advanced-scenarios/strongly-typed-template.md
+++ b/docs/user-docs/advanced-scenarios/strongly-typed-template.md
@@ -60,7 +60,7 @@ The idea behind the code is really simple. First, we separate strings and variab
 
 string\(s\):
 
-```markup
+```html
 <button class="btn btn-primary btn-
 
 " ref="atButtonTemplate">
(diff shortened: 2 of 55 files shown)


Sayan751 (commented):
Thanks @bigopon!  Can you please add a test case of nesting?

```html
<div example.bind="1">
  <my-ce /><!-- value of injected IExample is 1 in my-ce -->
  <div example.bind="2">
   <!-- value of injected IExample is 2 in my-ces -->
    <my-ce />
    <my-ce />
  </div>
  <my-ce /><!-- value of injected IExample is 1 in my-ce -->
</div>
```

PR #1926, opened 2024-03-15
refactor(event): no longer call prevent default by default

bigopon:
## 📖 Description

In v1, we have `event.preventDefault()` always called on event bindings, causing unexpected behavior sometimes (keydown not allowing input to happen etc...). While this can be controlled via `return true` in the event handler, it causes a weird situation where we are doing something that is not commonly expected.
- BREAKING CHANGE: this PR removes the behavior of always calling `preventDefault()` on events, and adds an option for application to switch it back:
    ```ts
    import { AppTask } from 'aurelia';
    import { IListenerBindingOptions } from '@aurelia/runtime-html';
    
    Aurelia.register(AppTask.creating(IListenerBindingOptions, options => options.prevent = true);
    ...
    ```
    
    Alternatively, using compat package will also enable this behavior, via a similar change like the above code example.
- fix an issue with the router-lite trying to register component unnecessarily
- fix an issue with typings for `createFixture.component(...)`
- BREAKING CHANGE: remove `fetch-client` re-export from the `aurelia` package


ghiscoding:
upgrading to Beta 13 caused a regression in Aurelia-Slickgrid that I support, I'm assuming it might be a regression coming from this PR. In SlickGrid, it takes advantage of Custom Event that could potentially call `return false` and expect this return value to be propagated to SlickGrid and then run some conditions in relation to that returned value (`false`), but it doesn't seem to be propagated anymore (it was working fine in Beta 12). 

For example, in SlickGrid, it has a Custom Event named `onBeforeEdit` and if the user returns `false`, then the Edit is blocked (meaning the user cannot edit when that event returns false). So something like this
```ts
handleOnBeforeEdit (e, args) {
  e.preventDefault();
  return false;
}
```
I added some break points and so far, I see this unexpected `returnedValue` within Aurelia, it says that it's `true` but that is incorrect since in my code it returns `false`. So why is that? Please note that returning `false` in Angular-Slickgrid or in plain JS works correctly, it just regressed in Aurelia-Slickgrid with Beta 13 (worked fine in Beta 12). The print screen below is coming from Aurelia Runtime-html (index.mjs)

image

It's a bit late and I need to go to bed but I could try to provide a better repro in the next couple days if required

bigopon:
@ghiscoding where are you getting these returned values? the event handler signature should be void return, and shouldn't return any value to the application, and application probably shouldn't rely on that either.

ghiscoding:
it's the return value from the `dispatchEvent`  and it is default JavaScript behavior as quoted from MDN.

> [Return value](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/dispatchEvent#return_value)
false if event is cancelable, and at least one of the event handlers which received event called [Event.preventDefault()](https://developer.mozilla.org/en-US/docs/Web/API/Event/preventDefault). Otherwise true.

this doesn't work in Beta 13 but was fine in Beta 12.

bigopon:
Ahh right i wasnt expecting dispatcher to communicate with listeners that way, and I think I only ever used it probably less than 5 ti.ea. thanks for the link.

For your main issue, its because Aurelia is no longer calling preventDefault() by default, no matter what the return value of the handler is.

From MDN doc:
```
if (event is cancellable && event.defaultPrevented)
  return true
else
  return false
```
So you will need to call preventDefault yourself if you wish to leverage that returnValue.

bigopon:
I see, i didnt read your issue carefully. I better not use my phone to read anything not short next time I guess. Thanks @ghiscoding , will have a look.

bigopon:
@ghiscoding , btw, is your event cancelable with `new CustomEvent('...', { cancelable: true })`?

ghiscoding:
yes it's cancellable, I think the repro in my case is that the `dispatchEvent` is sent to the Aurelia-Slickgrid custom component and might be altered at that time? I'll see if I can somehow make a repro in the coming days... or actually a repro could be the PR that I upgrade Aurelia-Slickgrid to Au2.Beta.13 in this [PR](https://github.com/ghiscoding/aurelia-slickgrid/pull/1172), there's Stackblitz link if you don't want to install it yourself, then if you go to Example 30 (as shown below), you should see the "Finish Date" as disabled. NOTE: I did merge the PR and got all my Cypress E2E tests to pass because I manually assigned `returnValue = false` in this [commit](https://github.com/ghiscoding/aurelia-slickgrid/pull/1172/files#diff-a1e4bbf70a60584aebfb7c70a668c399c9d8d38ff39e436379c130ff5da73302), if you remove that then you'll see the regression

msedge_Jod3YIoQ9D

bigopon:
I got this error with the stackblitz 
image

Is it expected?

For the actual issue, I tried to debug it a bit by changing this
```html
<div
  ...
  on-validation-error.trigger="handleValidationError($event.detail.eventData, $event.detail.args)">
```
to
```html
<div
  ...
  on-validation-error.trigger="handleValidationError($event.detail.eventData, $event.detail.args, $event)">
```
and inside the ts file:
```diff
  handleValidationError(e: Event, args: any, event: any) {
    console.log({event});
```
I got this log:
image

You can see that the first parameter `e` is not an event at all (it's from `$event.detail.eventData`), are you working with this object and expecting some value from it? Shouldn't you call `preventDefault()` on the `$event` instead?

ghiscoding:
oops yeah not an expected error for Stackblitz, I run the project in a pnpm workspace and I need to run an initial build, I will fix that later by simply adding an extra npm script to prebuild the project so that the demo always reference the built dist folder. I don't have that problem locally because I already ran the project multiple times so I already have an initial build, but on Stackblitz it always start fresh (so no build to start with). 

> not an event... 

right in this case it's a SlickEventData, it's a wrapper of an Event. I need to investigate further, I'm sorry to have make you waist time on this. I really have to provide you a better repro, so you can put that aside and I will investigate. SlickGrid has its own PubSub Service, so it's a bit confusing when using that in combo with another framework... so let me investigate further and come up with a better repro. Thanks a lot for investigating. Once I get a better repro I will open a new issue to track it as well.

bigopon:
No worries, thanks for all the issues. They are a source of correctness for Aurelia.

ghiscoding:
@bigopon 
BTW, I was going to push a PR to possibly add Stackblitz to Aurelia 2 project but if you try it yourself, you'll see that the `npm install` fails for some reason. I think Stackblitz is pretty nice to use once installed because it creates a small sandbox for each PRs and any user can start/debug Aurelia project by themselve without requiring a full install (as you saw when you tried my project Stackblitz). So I'm not sure if you know what the error is and if you're interested in having Stackblitz in Aurelia 2 project?

here's the simply link to use
https://stackblitz.com/~/github.com/aurelia/aurelia

and here's the npm install error (you can test it yourself via the link just above)

image

bigopon:
Hmm stackblitz installation error may have something to do with the web container tech not supporting wasm related deps?

> ... interested in having Stackblitz in Aurelia 2 project?

I'm interested, but I've never managed to get to set it up. Also, another work I think is of higher value is to integrate community projects/libs into the pipeline as some kind of e2e tests that run against latest commits to guard against regression.

bigopon:
> Hmm stackblitz installation error may have something to do with the web container tech not supporting wasm related deps?

Actually deleting package lock and reinstalling deps worked, I'll need to investigate that.

ghiscoding:
just remember that Stackblitz always starts fresh, it never has any cache, which was the problem I had with my Stackblitz that you saw in Aurelia-Slickgrid

PR #1928, opened 2024-03-18
feat(custom-attribute): ability to find closest attr by name or ctor

bigopon:
## 📖 Description

Part of the improvement in v2 compared to v1, where attribute and elements are both participating in the container hierarchy and causing resolver shadowing/issues with container resolution, is that only element participates in the container hierarchy.
This, however, makes finding/injecting custom attribute in a group of attributes a challenge. This is a valid need, since often times that custom attributes are developed in grouped, as discussed in #660 or https://discourse.aurelia.io/t/au2-access-parent-component-viewmodel-in-a-clean-way/5378/5

This PR adds a new utility `closest` in `CustomAttribute`, to enable walking the DOM and finding a custom attribute controller via name or constructor, like the following example:

```html
<div foo="1">
  <div bar="2">
  </div>
</div>
```

```ts
import { CustomAttribute, resolve, INode } from 'aurelia';

@customAttribute('bar')
export class Bar {
  host = resolve(INode);
  parent = CustomAttribute.closest(this.host, 'foo');
}
```

Or the following also works if you want to search using the constructor:

```ts
import { CustomAttribute, resolve, INode } from 'aurelia';
import { Foo } from './foo';

@customAttribute('bar')
export class Bar {
  host = resolve(INode);
  parent = CustomAttribute.closest(this.host, Foo); // <--- use constructor instead of string 'foo'
}
```

In the discussion in #660 , `closest` was chosen instead of a  `@parent` injection as it's closer to DOM API, and the way we walk & find, also it's less ambiguous since `@parent` doesn't indicate whether it's immediate or any of the ancestors.

Thanks @migajek @rluba

### DIFF
--- a/docs/user-docs/templates/custom-attributes.md
+++ b/docs/user-docs/templates/custom-attributes.md
@@ -8,7 +8,7 @@ A custom attribute allows you to create special properties to enhance and decora
 
 ## Creating custom attributes
 
-On a simplistic level, custom attributes resemble [components](broken-reference) quite a lot. They can have [bindable properties](../components/bindable-properties.md), and they use classes for their definitions.
+On a simplistic level, custom attributes resemble [components](../components/components.md) quite a lot. They can have [bindable properties](../components/bindable-properties.md), and they use classes for their definitions.
 
 A basic custom attribute looks something like this:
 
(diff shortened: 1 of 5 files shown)


Sayan751:
Thanks @bigopon! Looks good!

PR #1931, opened 2024-03-27
fix(form): prevent actionless submission

bigopon:
## 📖 Description

Currently, form actionless submission is a part of the compat package, though it's not necessarily most useful there. Move it to be part of the main application via config `allowActionlessForm`, so that the submit event on form without an `action` attribute will be called `preventDefault()`. This is to avoid reloading the whole page, which is 99.99% of the times not wanted in an SPA.

This behavior can be configured like the following examples:

```ts
Aurelia.app({
  host: ...,
  component: ...,
  allowActionlessForm: true, // <-- this will reload page by default
})

// or

Aurelia.register(AppTask.creating(IAppRoot, root => root.config.allowActionlessForm = true));
```

BREAKING CHANGE:
- Move actionless form submit prevention from compat to application config, though this behavior is the same with v1, so it's mostly a change compared to previous version of v2 only.

cc @Sayan751 @fkleuver @CollinHerber

### DIFF
--- a/packages/runtime-html/src/aurelia.ts
+++ b/packages/runtime-html/src/aurelia.ts
@@ -161,6 +161,13 @@ export interface ISinglePageAppConfig<T = unknown> {
    * The root component of the app
    */
   component: T | Constructable<T>;
+  /**
+   * When a HTML form is submitted, the default behavior is to "redirect" the page to the action of the form
+   * This is not desirable for SPA applications, so by default, this behavior is prevented.
+   *
+   * This option re-enables the default behavior of HTML forms.
+   */
+  allowActionlessForm?: boolean;
 }
 
 export interface IEnhancementConfig<T> {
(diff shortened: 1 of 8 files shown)


packages/runtime-html/src/aurelia.ts
```diff
@@ -173,4 +180,11 @@ export interface IEnhancementConfig<T> {
    * A predefined container for the enhanced view.
    */
   container?: IContainer;
+  /**
+   * When a HTML form is submitted, the default behavior is to "redirect" the page to the action of the form
+   * This is not desirable for SPA applications, so by default, this behavior is prevented.
+   *
+   * This option re-enables the default behavior of HTML forms.
+   */
+  allowActionlessForm?: boolean;
```
CollinHerber, line 189: Do you think it's worth it to have this be on by default?
If 99.99% of usecases don't want this behavior don't you think it should be opt-out, not opt-in?

CollinHerber (approved):

packages/runtime-html/src/aurelia.ts
bigopon: Yes the name indicates opt out, or can you elaborqte a bit more?

CollinHerber: Ahhh yeah I was thinking opposite, apologies!

bigopon: no worries, thanks for the review.

PR #1932, opened 2024-03-30
chore: migration to TC39 decorators + metadata simplification

Sayan751:
# Pull Request


packages/runtime-html/src/resources/binding-behavior.ts
```diff
@@ -57,8 +58,8 @@ export class BindingBehaviorDefinition<T extends Constructable = Constructable>
 
     return new BindingBehaviorDefinition(
       Type,
-      firstDefined(getBehaviorAnnotation(Type, 'name'), name),
-      mergeArrays(getBehaviorAnnotation(Type, 'aliases'), def.aliases, Type.aliases),
+      firstDefined(getBehaviorAnnotation(Type, 'name', context), name),
```
bigopon, line 61: is there a way to let the decorator set the medata using `Symbol.metadat` so that we don't need to pass decorator context into our resource APIs, similar to my comment in the other place, our APIs shouldn't be polluted by decorator related stuff, we just need to work with metadata.

packages/runtime-html/src/resources/binding-command.ts
```diff
@@ -69,21 +69,21 @@ export type BindingCommandInstance<T extends {} = {}> = {
 export type BindingCommandType<T extends Constructable = Constructable> = ResourceType<T, BindingCommandInstance, PartialBindingCommandDefinition>;
 export type BindingCommandKind = IResourceKind & {
   // isType<T>(value: T): value is (T extends Constructable ? BindingCommandType<T> : never);
-  define<T extends Constructable>(name: string, Type: T): BindingCommandType<T>;
-  define<T extends Constructable>(def: PartialBindingCommandDefinition, Type: T): BindingCommandType<T>;
-  define<T extends Constructable>(nameOrDef: string | PartialBindingCommandDefinition, Type: T): BindingCommandType<T>;
-  getAnnotation<K extends keyof PartialBindingCommandDefinition>(Type: Constructable, prop: K): PartialBindingCommandDefinition[K];
+  define<T extends Constructable>(name: string, Type: T, decoratorContext?: DecoratorContext): BindingCommandType<T>;
+  define<T extends Constructable>(def: PartialBindingCommandDefinition, Type: T, decoratorContext?: DecoratorContext): BindingCommandType<T>;
+  define<T extends Constructable>(nameOrDef: string | PartialBindingCommandDefinition, Type: T, decoratorContext?: DecoratorContext): BindingCommandType<T>;
```
bigopon, line 74: similar comment on keeping decorator out of resource APIs.

packages/runtime-html/src/resources/custom-attribute.ts
```diff
@@ -58,19 +58,19 @@ export type CustomAttributeKind = IResourceKind & {
...
-  isType<T>(value: T): value is (T extends Constructable ? CustomAttributeType<T> : never);
-  define<T extends Constructable>(name: string, Type: T): CustomAttributeType<T>;
-  define<T extends Constructable>(def: PartialCustomAttributeDefinition, Type: T): CustomAttributeType<T>;
-  define<T extends Constructable>(nameOrDef: string | PartialCustomAttributeDefinition, Type: T): CustomAttributeType<T>;
-  getDefinition<T extends Constructable>(Type: T): CustomAttributeDefinition<T>;
+  isType<T>(value: T, context?: DecoratorContext | null): value is (T extends Constructable ? CustomAttributeType<T> : never);
+  define<T extends Constructable>(name: string, Type: T, decoratorContext?: DecoratorContext): CustomAttributeType<T>;
+  define<T extends Constructable>(def: PartialCustomAttributeDefinition, Type: T, decoratorContext?: DecoratorContext): CustomAttributeType<T>;
+  define<T extends Constructable>(nameOrDef: string | PartialCustomAttributeDefinition, Type: T, decoratorContext?: DecoratorContext): CustomAttributeType<T>;
```
bigopon, line 64: Similar comment on keeping decorator related stuff out of resources APIs.

packages/runtime-html/src/resources/custom-element.ts
```diff
@@ -102,26 +102,26 @@ export type CustomElementKind = IResourceKind & {
...
   define<C extends Constructable>(def: PartialCustomElementDefinition, Type: C): CustomElementType<C>;
   define<C extends Constructable>(def: PartialCustomElementDefinition, Type?: null): CustomElementType<C>;
   define<C extends Constructable>(nameOrDef: string | PartialCustomElementDefinition, Type: C): CustomElementType<C>;
-  getDefinition<C extends Constructable>(Type: C): CustomElementDefinition<C>;
+  getDefinition<C extends Constructable>(Type: C, context?: DecoratorContext | null): CustomElementDefinition<C>;
   // eslint-disable-next-line
-  getDefinition<C extends Constructable>(Type: Function): CustomElementDefinition<C>;
-  annotate<K extends keyof PartialCustomElementDefinition>(Type: Constructable, prop: K, value: PartialCustomElementDefinition[K]): void;
-  getAnnotation<K extends keyof PartialCustomElementDefinition>(Type: Constructable, prop: K): PartialCustomElementDefinition[K];
+  getDefinition<C extends Constructable>(Type: Function, context?: DecoratorContext | null): CustomElementDefinition<C>;
```
bigopon, line 112: similar comment on decorator and resource APIs.

packages/runtime-html/src/resources/custom-element.ts
```diff
@@ -102,26 +102,26 @@ export type CustomElementKind = IResourceKind & {
    * @returns The controller associated with the provided node, if it is a custom element, otherwise `null`
    */
   for<C extends ICustomElementViewModel = ICustomElementViewModel>(node: Node, opts: { optional: true }): ICustomElementController<C> | null;
-  isType<C>(value: C): value is (C extends Constructable ? CustomElementType<C> : never);
+  isType<C>(value: C, context?: DecoratorContext): value is (C extends Constructable ? CustomElementType<C> : never);
```
bigopon, line 105: here you'll see the effect of the decorator context on the API, there shouldn't be any other factor than the class itself to determine whether it's a custom element or not.

packages/runtime-html/src/resources/custom-elements/au-compose.ts
```diff
@@ -105,7 +110,7 @@ export class AuCompose {
   /** @internal */ private readonly _platform = resolve(IPlatform);
   /** @internal */ private readonly _rendering = resolve(IRendering);
   /** @internal */ private readonly _instruction = resolve(IInstruction) as HydrateElementInstruction;
-  /** @internal */ private readonly _contextFactory = resolve(transient(CompositionContextFactory));
+  /** @internal */ private readonly _contextFactory = resolve(transient(CompositionContextFactory, null!));
```
bigopon, line 113: this change reflect how cumbersome it becomes, if there's a way to avoid it, we should probably try to have it not cumbersome and noisy like this.

packages/runtime-html/src/resources/value-converter.ts
```diff
@@ -26,24 +26,24 @@ export type PartialValueConverterDefinition = PartialResourceDefinition;
 export type ValueConverterType<T extends Constructable = Constructable> = ResourceType<T, ValueConverterInstance>;
 export type ValueConverterKind = IResourceKind & {
   isType<T>(value: T): value is (T extends Constructable ? ValueConverterType<T> : never);
-  define<T extends Constructable>(name: string, Type: T): ValueConverterType<T>;
-  define<T extends Constructable>(def: PartialValueConverterDefinition, Type: T): ValueConverterType<T>;
-  define<T extends Constructable>(nameOrDef: string | PartialValueConverterDefinition, Type: T): ValueConverterType<T>;
+  define<T extends Constructable>(name: string, Type: T, decoratorContext?: DecoratorContext): ValueConverterType<T>;
+  define<T extends Constructable>(def: PartialValueConverterDefinition, Type: T, decoratorContext?: DecoratorContext): ValueConverterType<T>;
+  define<T extends Constructable>(nameOrDef: string | PartialValueConverterDefinition, Type: T, decoratorContext?: DecoratorContext): ValueConverterType<T>;
```
bigopon, line 31: similar comment on keeping decorator stuff out of the resource APIs

packages/runtime/src/binding/connectable.ts
```diff
@@ -119,7 +119,7 @@ type Connectable = { oL: IObserverLocator } & IConnectable & Partial<ISubscriber
 type DecoratableConnectable<TProto, TClass> = Class<TProto & Connectable, TClass>;
 type DecoratedConnectable<TProto, TClass> = Class<TProto & Connectable, TClass>;
 
-function connectableDecorator<TProto, TClass>(target: DecoratableConnectable<TProto, TClass>): DecoratedConnectable<TProto, TClass> {
+function connectableDecorator<TProto, TClass>(target: DecoratableConnectable<TProto, TClass>, _context: ClassDecoratorContext<DecoratableConnectable<TProto, TClass>>): DecoratedConnectable<TProto, TClass> {
```
bigopon, line 122: Maybe we can have a core connectable mixing and a connectable decorator instead. So in places where we are using it as mixin (which is what it is), we don't have to have weird `..., null`

packages/runtime/src/observation/observable.ts
```diff
@@ -27,135 +27,125 @@ const noValue: unknown = {};
...
+    };
+  }
 
-    // todo(bigopon/fred): discuss string api for converter
+  // case: @observable({...}) class | @observable({...}) prop
+  return function (target: Constructable | undefined, context: ClassFieldDecoratorContext | ClassDecoratorContext) {
+    switch (context.kind) {
+      case 'field': return createFieldInitializer(context);
+      case 'class': return createDescriptor(target, config.name!, () => noValue, true);
+      default: throw createMappedError(ErrorNames.invalid_observable_decorator_usage);
```
bigopon, line 98: observable can also be used on getter/setter. we may need to bundle it together with `field`.

packages/runtime/src/observation/observer-locator.ts
```diff
@@ -57,14 +57,14 @@ class DefaultNodeObserverLocator implements INodeObserverLocator {
...
   get?: ObservableGetter;
-  set?: ObservableSetter;
+  // set?: ObservableSetter;
 };
 export type ObservableGetter = PropertyDescriptor['get'] & {
-  getObserver?(obj: unknown, requestor: IObserverLocator): IObserver;
-};
-export type ObservableSetter = PropertyDescriptor['set'] & {
-  getObserver?(obj: unknown, requestor: IObserverLocator): IObserver;
+  getObserver?(obj: unknown/* , requestor: IObserverLocator */): IObserver;
```
bigopon, line 63: Is this removal of requestor intentional?

packages/runtime/src/observation/proxy-decorators.ts
```diff
@@ -8,66 +8,30 @@ import { nowrapClassKey, nowrapPropKey } from './proxy-observation';
...
+  context: ClassDecoratorContext<Class<TThis>> | ClassFieldDecoratorContext<TThis>
+): void {
+  switch (context.kind) {
+    case 'class':
+      defineHiddenProp(target!, nowrapClassKey, true);
+      break;
+    case 'field':
+      context.addInitializer(function (this: object) {
+        defineHiddenProp(this.constructor, `${nowrapPropKey}_${safeString(context.name)}__`, true);
+      });
```
bigopon, line 33: the working of field makes it quite an annoyance. If we have multiple instance, this will be call again, probably we will need to guard against multiple define calls.

packages/state/src/state-decorator.ts
```diff
@@ -14,35 +14,30 @@ import { StateGetterBinding } from './state-getter-binding';
...
+    let dependencies = context.metadata[dependenciesKey] as unknown[];
     if (dependencies == null) {
-      CustomElement.annotate(target, dependenciesKey, dependencies = []);
+      dependencies = context.metadata[dependenciesKey] = [];
     }
-    dependencies.push(new CreatedLifecycleHooks(getValue, key));
+    dependencies.push(
+      new HydratingLifecycleHooks(getValue, key),
+        new CreatedLifecycleHooks(getValue, key)
+    );
```
bigopon, line 36: indent 😉 
If we are adding both hydrating and created lifecycles to a custom element, we will add two bindings that target the same property. Will probably need to tweak this. Custom element can use `hydrating` so the binding can be added early, while custom attribute can use `created` as it's the earliest lifecycle.

packages/ui-virtualization/src/virtual-repeat.ts
```diff
@@ -52,6 +52,14 @@ const noScrollInfo: IScrollerInfo = {
...
 
+@customAttribute({
+  isTemplateController: true,
+  name: 'virtual-repeat',
+  bindables: {
+    local: { name: 'local' },
+    items: { name: 'items', primary: true }
+  }
+})
 export class VirtualRepeat implements IScrollerSubscriber, IVirtualRepeater {
```
bigopon, line 63: Code generation is a reason, another is the display of the instance in the console, it's empty and sometimes it's confusing.
Just giving the reason why it's done like this in some places 😃

packages/validation-i18n/src/localization.ts
```diff
@@ -18,13 +18,10 @@ export const I18nKeyConfiguration = /*@__PURE__*/DI.createInterface<I18nKeyConfi
 export class LocalizedValidationController extends ValidationController {
   private readonly localeChangeSubscription: IDisposable;
   public constructor(
-    @IServiceLocator locator: IServiceLocator,
-    @IEventAggregator ea: EventAggregator,
-    @IValidator validator: IValidator,
-    @IExpressionParser parser: IExpressionParser,
-    @IPlatform platform: IPlatform,
+    ea: EventAggregator = resolve(IEventAggregator),
+    platform: IPlatform = resolve(IPlatform),
```
bigopon, line 22: maybe they don't need to be in the constructor?

packages/validation/src/rule-provider.ts
```diff
@@ -473,10 +474,10 @@ export class ValidationRules<TObject extends IValidateable = IValidateable> impl
...
 
   public constructor(
-    @IServiceLocator private readonly locator: IServiceLocator,
-    @IExpressionParser private readonly parser: IExpressionParser,
-    @IValidationMessageProvider private readonly messageProvider: IValidationMessageProvider,
-    @IValidationExpressionHydrator private readonly deserializer: IValidationExpressionHydrator,
+    private readonly locator: IServiceLocator = resolve(IServiceLocator),
+    private readonly parser: IExpressionParser = resolve(IExpressionParser),
+    private readonly messageProvider: IValidationMessageProvider = resolve(IValidationMessageProvider),
+    private readonly deserializer: IValidationExpressionHydrator = resolve(IValidationExpressionHydrator),
```
bigopon, line 480: I think `resolve` + `constructor param as property` is quite a bad combo, we normally can just declare the property  and resolve the values without the constructor. Having it there will just give confusion. If we want to override the default resolved value, we can do it with just normal constructor parameters?

packages/validation/src/rules.ts
```diff
@@ -47,14 +47,14 @@ export interface ValidationRuleDefinition {
 }
 export type RuleType<TRule extends IValidationRule> = Class<TRule, { $TYPE: string }>;
 export const ValidationRuleAliasMessage = Object.freeze({
-  aliasKey: Protocol.annotation.keyFor('validation-rule-alias-message'),
-  define<TRule extends IValidationRule>(target: RuleType<TRule>, definition: ValidationRuleDefinition): RuleType<TRule> {
-    ValidationRuleAliasMessage.setDefaultMessage(target, definition);
+  aliasKey: getAnnotationKeyFor('validation-rule-alias-message'),
+  define<TRule extends IValidationRule>(target: RuleType<TRule>, definition: ValidationRuleDefinition, context: DecoratorContext): RuleType<TRule> {
```
bigopon, line 51: similar comment on keeping resource APIs separate from decorator related stuff.

packages/validation/src/serialization.ts
```diff
@@ -94,9 +94,9 @@ export class ValidationDeserializer implements IValidationExpressionHydrator {
   }
   public readonly astDeserializer: Deserializer = new Deserializer();
   public constructor(
-    @IServiceLocator private readonly locator: IServiceLocator,
-    @IValidationMessageProvider public readonly messageProvider: IValidationMessageProvider,
-    @IExpressionParser public readonly parser: IExpressionParser
+    private readonly locator: IServiceLocator = resolve(IServiceLocator),
+    public readonly messageProvider: IValidationMessageProvider = resolve(IValidationMessageProvider),
+    public readonly parser: IExpressionParser = resolve(IExpressionParser),
   ) { }
```
bigopon, line 100: similar comment on no need for constructor + resolve as default value.

packages/validation/src/serialization.ts
```diff
@@ -196,9 +196,9 @@ export class ModelValidationExpressionHydrator implements IValidationExpressionH
   public readonly astDeserializer: Deserializer = new Deserializer();
   public constructor(
     /** @internal */
-    @IServiceLocator public readonly l: IServiceLocator,
-    @IValidationMessageProvider public readonly messageProvider: IValidationMessageProvider,
-    @IExpressionParser public readonly parser: IExpressionParser
+    public readonly l: IServiceLocator = resolve(IServiceLocator),
+    public readonly messageProvider: IValidationMessageProvider = resolve(IValidationMessageProvider),
+    public readonly parser: IExpressionParser = resolve(IExpressionParser),
   ) { }
```
bigopon, line 202: similar comment on constructor  + resolve as fallback value combo

bigopon (commented):
Nice work @Sayan751, we are on final steps to RC with this. I've had a few comments mostly around the API changes.

packages/__tests__/src/3-runtime-html/process-content.spec.ts
```diff
@@ -128,8 +128,53 @@ describe('3-runtime-html/process-content.spec.ts', function () {
         }
       );
     }
+    // The following tests don't work. Refer: https://github.com/microsoft/TypeScript/issues/57366
```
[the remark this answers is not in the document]
Sayan751: I wouldn't say that it was illogical, as it was better than using the name of the method, which is a string. Thus, it was easier to rename the method etc. However, there are already various other options to mark/define a processContent hook. Thus, loosing this one, won't hurt us much.

packages/kernel/src/di.container.ts
```diff
@@ -637,22 +648,22 @@ function containerGetKey(this: IContainer, d: Key) {
...
+  ? Resolved<R> | undefined
+  : K extends IFactoryResolver<infer R>
+  ? IResolvedFactory<R>
+  : K extends IResolver<infer R>
+  ? Resolved<R>
+  : K extends [infer R1 extends Key, ...infer R2]
+  ? [IResolvedInjection<R1>, ...IResolvedInjection<R2>]
+  : K extends InterfaceSymbol<infer T>
+  ? T
+  : Resolved<K>;
```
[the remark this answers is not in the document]
Sayan751: I needed to resolve the `__resolved__` to counter the issues with the typings. Unfortunately, I cannot exactly remember what the issues were. Hence, the change. As the both the unit and API tests worked, I didn't bother about it. Would it be possible for you to give an example, where we needed it? If the same example still works with the new typedef, t should be okay.

packages/kernel/src/di.resolvers.ts
```diff
@@ -31,20 +31,20 @@ export function createResolver<T extends Key>(getter: (key: T, handler: IContain
...
   resolver.$isResolver = true;
   resolver.resolve = (handler: IContainer, requestor: IContainer) => requestor.getAll(key, searchAncestors);
 
   return resolver as IAllResolver<T>;
 };
-export type IAllResolver<T> = IResolver<Resolved<T>[]> & {
+export type IAllResolver<T> = IResolver<readonly Resolved<T>[]> & {
   // type only hack
-  __isAll: undefined;
+  __isAll?: undefined;
```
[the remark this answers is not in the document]
Sayan751: It looks like that we don't need that at all. Removed it now.

If you see any use-case, not satisfied by that, let me know.

packages/kernel/src/di.resolvers.ts
```diff
@@ -123,9 +123,11 @@ export type IOptionalResolver<K extends Key = Key> = IResolver<K | undefined> &
...
  * ignore tells the container not to try to inject a dependency
  */
-export const ignore: IResolver<undefined> = /*@__PURE__*/objectAssign((target: Injectable, property?: string | number, descriptor?: PropertyDescriptor | number): void => {
-  inject(ignore)(target, property, descriptor);
-}, { $isResolver: true, resolve: () => void 0 } as const);
+export const ignore = (decorated: unknown, context: DecoratorContext): void => {
+  inject(ignore)(decorated, context);
+};
+ignore.$isResolver = true;
+ignore.resolve = () => undefined;
```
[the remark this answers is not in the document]
Sayan751: You are right. This is an unindented change on my part. Or rather, a result of impatient git conflict resolution. :P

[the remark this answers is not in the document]
Sayan751: Cleaned up now

packages/kernel/src/di.resolvers.ts
```diff
@@ -270,6 +272,6 @@ const createNewInstance = (key: any, handler: IContainer, requestor: IContainer)
 };
 
 // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
-const isInterface = <K>(key: any): key is InterfaceSymbol<K> => isFunction(key) && key.$isInterface === true;
+const isInterface = <K>(key: any): key is InterfaceSymbol<K> => key.$isInterface === true;
```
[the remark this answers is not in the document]
Sayan751: Good point. However, the `isFunction` check is now invalid, as a `key` can no longer be a parameter decorator. However, I can add a `isObject` check.

packages/kernel/src/logger.ts
```diff
@@ -176,18 +175,17 @@ interface SinkDefinition {
...
+  getHandles<TSink extends ISink>(target: TSink) {
+    return target.constructor[Symbol.metadata]?.[this.key] as LogLevel[] | undefined;
   },
 });
 
 export const sink = (definition: SinkDefinition) => {
-  return <TSink extends ISink>(target: Constructable<TSink>): Constructable<TSink> =>
-    LoggerSink.define(target, definition);
+  return <TSink extends ISink>(_target: Constructable<TSink>, context: ClassDecoratorContext): void =>
+    LoggerSink.define(context, definition);
```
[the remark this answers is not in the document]
Sayan751: It depends on the timing when the decorator is invoked. As far as I know, the `System.metadata` is assigned to the class, when it is fully defined. That happens when all the decorators are invoked. Till that time, there is no `System.metadata` in the target class. Plus, going forward, it also depends on how the transpilers are generating code for that. As far as I have seen so far, TS simply assigns the `System.metadata` value to the class, without merging the existing `System.metadata`. It annoys me as well, that we need to pull the decorator context everywhere, but I would wait a bit before prematurely optimizing it.

packages/runtime-html/src/bindable.ts
```diff
@@ -156,14 +184,14 @@ export class BindableDefinition {
     public readonly set: InterceptorFunc,
   ) { }
 
-  public static create(prop: string, target: Constructable<unknown>, def: PartialBindableDefinition = {}): BindableDefinition {
```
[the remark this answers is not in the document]
Sayan751: I am a bit critical about this change to be honest. As the design-type metadata are not emitted by TS when using the TC39 decorators, the TS users will lose the ease of usage wrt. coercion. Maybe, we can later activate it, if the TS decides to emit those metadata again.

packages/runtime-html/src/renderer.ts
```diff
@@ -387,7 +387,17 @@ export interface IRenderer<TType extends string = string> {
...
+// type DecoratableInstructionRenderer<TType extends string, TProto, TClass> = Class<TProto & Partial<IInstructionTypeClassifier<TType> & Pick<IRenderer, 'render'>>, TClass> & Partial<IRegistry>;
+// type DecoratedInstructionRenderer<TType extends string, TProto, TClass> =  Class<TProto & IInstructionTypeClassifier<TType> & Pick<IRenderer, 'render'>, TClass> & IRegistry;
+
+// type InstructionRendererDecorator<TType extends string> = <TProto, TClass>(target: DecoratableInstructionRenderer<TType, TProto, TClass>, context: ClassDecoratorContext) => DecoratedInstructionRenderer<TType, TProto, TClass>;
+
+// export function renderer<TType extends string>(targetType: TType): InstructionRendererDecorator<TType> {
+//   return function decorator<TProto, TClass>(target: DecoratableInstructionRenderer<TType, TProto, TClass>): DecoratedInstructionRenderer<TType, TProto, TClass> {
+//     target.register = function (container: IContainer): void {
+//       singletonRegistration(IRenderer, this).register(container);
+//     };
```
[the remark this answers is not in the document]
Sayan751: Done

packages/runtime-html/src/resources/binding-behavior.ts
Sayan751: Let us discuss this issue under this thread: https://github.com/aurelia/aurelia/pull/1932#discussion_r1546247909. Closing this one.

packages/runtime-html/src/resources/binding-command.ts
Sayan751: Let us discuss this issue under this thread: https://github.com/aurelia/aurelia/pull/1932#discussion_r1546247909. Closing this one.

packages/runtime-html/src/resources/custom-attribute.ts
Sayan751: Let us discuss this issue under this thread: https://github.com/aurelia/aurelia/pull/1932#discussion_r1546247909. Closing this one.

packages/runtime-html/src/resources/custom-element.ts
Sayan751: Let us discuss this issue under this thread: https://github.com/aurelia/aurelia/pull/1932#discussion_r1546247909. Closing this one.

packages/runtime-html/src/resources/custom-element.ts
Sayan751: Let us discuss this issue under this thread: https://github.com/aurelia/aurelia/pull/1932#discussion_r1546247909. Closing this one.

packages/kernel/src/logger.ts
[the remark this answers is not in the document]
Sayan751: Maybe we can keep the public API without the decorator context. And use the decorator context only for internal API.

packages/runtime-html/src/resources/custom-elements/au-compose.ts
Sayan751: Let us discuss this issue under this thread: https://github.com/aurelia/aurelia/pull/1932#discussion_r1546247909. Closing this one.

packages/runtime-html/src/resources/value-converter.ts
Sayan751: Let us discuss this issue under this thread: https://github.com/aurelia/aurelia/pull/1932#discussion_r1546247909. Closing this one.

packages/runtime/src/binding/connectable.ts
Sayan751: Without the 2nd parameter, the type definition won't pass, and the compiler will be angry.

packages/runtime/src/observation/observer-locator.ts
Sayan751: As per my understanding, it was completely unused. Hence, I have removed it. Let me know if I have missed something.

packages/runtime/src/observation/proxy-decorators.ts
Sayan751: That's a good point. I have somehow overlooked it. Thanks for pointing it out.

packages/validation-i18n/src/localization.ts
Sayan751: I would say yes. The alternative would be to keep those as class property, although we don't need to keep those around outside the ctor scope.

packages/validation/src/rule-provider.ts
Sayan751: > I think resolve + constructor param as property is quite a bad combo... Having it there will just give confusion.

I think you need to clarify more.

I really like this pattern. Here is my justification: keeping the ctor arguments makes the class easier to unit test. At the same time, having default values for those properties makes the life easier for the devs. You are right that 

> If we want to override the default resolved value, we can do it with just normal constructor parameters?

We can do that, but IMO, that's exactly the use-case for having optional parameters. By avoiding optional parameters, we will do the same thing, but with more code.

packages/validation/src/rules.ts
Sayan751: Let us discuss this issue under this thread: https://github.com/aurelia/aurelia/pull/1932#discussion_r1546247909. Closing this one.

packages/__tests__/src/3-runtime-html/resources.spec.ts
```diff
@@ -158,7 +136,9 @@ describe('3-runtime-html/resources.spec.ts', function () {
     assert.strictEqual(created, true, 'created');
   });
 
-  it('works with aliases static property', function () {
+  // The following won't work with TC39 decorators due to the timing when the decorators are executed.
+  // When the decorator customAttribute is executed, the static property aliases is not yet defined.
+  it.skip('works with aliases static property', function () {
```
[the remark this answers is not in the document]
Sayan751: I think that we can probably leverage the class level initializers to read the values from the static properties. Thus, the idea is to define the resource using normal decorators work, and also add a class initializer to read the values from the static properties and overwrite the values. I will give that a try.

packages/kernel/src/di.resolvers.ts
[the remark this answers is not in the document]
bigopon: If it's working, it's probably because of a change in TS where the type will be preserved as much as possible. In the pass, it's resolved to as atomic as possible. Maybe it's fine. We may not even need to have these typing hack anymore I think.

packages/validation/src/rule-provider.ts
bigopon: > Here is my justification: keeping the ctor arguments makes the class easier to unit test

We should not test our code without container in any way in our repo, it's not how the code is used, it shouldn't be how the code is tested, unless no other way.

> > I think resolve + constructor param as property is quite a bad combo... Having it there will just give confusion.
>
> I think you need to clarify more.

The presence of the combo gives me a question whether it's valid or invalid to use without a container. Even for just a moment, I think it's an unnecessary oddness in the pattern. Additionally, I think creating things consistently helps.

bigopon (commented):
Thanks for the replies, I think I agree on the initializer part, we can just let all our decorator wait until the initializer phase and call the `define`, so we can just deal with `Symbol.metadata` instead of having to drag context everywhere.

packages/__tests__/src/3-runtime-html/resources.spec.ts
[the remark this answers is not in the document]
Sayan751: The strategy worked.

packages/validation/src/rule-provider.ts
Sayan751: > The presence of the combo gives me a question whether it's valid or invalid to use without a container.

I don't want to sound argumentative, but why is that important? While defining a component, you declare the dependencies of the component. Whether those dependencies are fulfilled by a container or not should be secondary. 

Additionally, I honestly think that having the ctor parameters better outlines the dependencies needed by a component. It brings transparency to the API. Converting those to pure class properties with the `resolve(...)` as initializer hides the information, and that is problematic even when you are using the container to register the dependencies, because as a dev you now have no information what all you needs to register to the container.

> We should not test our code without container in any way in our repo, it's not how the code is used, it shouldn't be how the code is tested, unless no other way.

Following the above argumentation, I would say that even if we don't test our code without container, the devs might want to mock it easily without doing much of the container gymnastic. We should keep that in mind.

> Additionally, I think creating things consistently helps.

I agree that being consistent helps greatly. However, this topic of injecting dependencies is highly context specific. Sometimes, you truly don't use the injected stuffs outside the ctor, and thus it makes very little sense to have those as class properties and giving those a lifetime same as the class instance. In such cases, you directly use the `resolve(...)` in the ctor. One might find that to be inconsistent, even though it fits the context perfectly.

If you are still not convinced, I will gladly make the change. :)

packages/runtime/src/observation/observable.ts
Sayan751: Both getter and setter? If you implement the setter, you can invoke the changeHandler by yourself. Do we really need observable support there? Even better is if you can give me an example, then I can test it directly.

packages/runtime-html/src/renderer.ts
```diff
@@ -387,7 +387,7 @@ export interface IRenderer<TType extends string = string> {
 
 export const IRenderer = /*@__PURE__*/createInterface<IRenderer>('IRenderer');
 
-export function renderer<TType extends string, T extends Constructable<IRenderer<TType>>>(targetType: TType): (target: T) => T {
+export function renderer<TType extends string, T extends Constructable<IRenderer<TType>>>(targetType: TType): (target: T, context: ClassDecoratorContext) => T {
   return function decorator(target) {
```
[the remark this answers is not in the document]
Sayan751: It is not noise as it is rather forced by TS. I guess that's how it is determined if a function can be used as a decorator or decorator factory, as well as what can be valid targets for that decorator.

packages/runtime/src/observation/observer-locator.ts
```diff
@@ -165,7 +161,7 @@ export class ObserverLocator {
     if (pd !== void 0 && !hasOwnProp.call(pd, 'value')) {
       let obs: IObserver | undefined | null = this._getAdapterObserver(obj, key, pd);
       if (obs == null) {
-        obs = (pd.get?.getObserver ?? pd.set?.getObserver)?.(obj, this);
+        obs = (pd.get?.getObserver /* ?? pd.set?.getObserver */)?.(obj/* , this */);
```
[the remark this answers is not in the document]
Sayan751: Yup. Will clean it up. Missed it. Thanks for noticing it.

packages/runtime-html/src/renderer.ts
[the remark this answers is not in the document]
bigopon: Right thanks for that info. Maybe it's enforced during the transition and will be removed at a later time.

packages/kernel/src/di.container.ts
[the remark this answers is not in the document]
bigopon: It's for injectable token, but maybe if it's already working fine then it's fine. If anything, I'll fix it later, no worries.

packages/validation/src/rule-provider.ts
bigopon: > ... While defining a component, you declare the dependencies of the component. Whether those dependencies are fulfilled by a container or not should be secondary.

We support dynamic dependencies during instantiation with container, the presence of any parameters raises the confusion, hence I think it's better to avoid it if possible.

> ... Additionally, I honestly think that having the ctor parameters better outlines the dependencies needed by a component. It brings transparency to the API
> ... the devs might want to mock it easily without doing much of the container gymnastic

If you expect any one other than ourselves constructing that instance directly, then yeah, maybe it's a reasonable thing to provide. What I'm seeing is it's not beneficial or realistic that anyone would do that. We probably shouldn't encourage people or make it easy to test their code in a different way with how it is used.

packages/kernel/src/logger.ts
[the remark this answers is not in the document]
Sayan751: Done.

packages/runtime/src/observation/proxy-decorators.ts
Sayan751: Fixed.

packages/runtime/src/observation/observer-locator.ts
Sayan751: As per my understanding, it was an internal API. Henece, probably we won't need any compat code for this.

packages/state/src/state-decorator.ts
Sayan751: Thanks for pointing this out. I think as of now, as we don't have a way to grab the constructor in the field/getter decorator, we need to tweak the hook implementation itself, so that only one binding is attached.

Sayan751: Changed now.

packages/validation/src/rule-provider.ts
Sayan751: Changes done.

packages/validation/src/serialization.ts
Sayan751: Changes done.

Sayan751: We need here optional parameters as the ctor is used by the static deserialize method.

packages/runtime-html/src/bindable.ts
```diff
@@ -31,82 +31,117 @@ type PartialBindableDefinitionPropertyOmitted = Omit<PartialBindableDefinition,
...
+  // eslint-disable-next-line @typescript-eslint/ban-types
+  | ((target: Function, context: ClassGetterDecoratorContext) => void) {
+
+  let configOrProp: PartialBindableDefinition | string | undefined = void 0;
+  function decorator(_target: unknown, context: ClassDecoratorContext | ClassFieldDecoratorContext | ClassGetterDecoratorContext): void {
+    let $prop: string;
+
+    switch (context.kind) {
+      case 'getter':
+      case 'field': {
```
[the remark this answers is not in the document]
Sayan751: Not sure what you meant by that, as there are already tests for that. Closing this for now, without any further changes. Feel free to unresolve.

packages/ui-virtualization/src/virtual-repeat.ts
Sayan751: Done.

packages/runtime/src/observation/observable.ts
bigopon: what about I think we also need to consider accessor, which can e added later, maybe this is fine for now. For the independent getter/setter I cant seem to recall when I added/removed it.

Sayan751:
@bigopon it is ready for review.

packages/runtime/src/observation/proxy-decorators.ts
```diff
@@ -77,15 +44,8 @@ function testNowrap() {
   @nowrap
   class MyModel {}
 
-  @nowrap()
```
bigopon, line 80: Is `nowrap()` an invalid expression to be used by `@` in the new decorator proposal?

packages/metadata/src/index.ts
```diff
@@ -64,1022 +76,35 @@ export function isNullOrUndefined(value: unknown): value is null | undefined {
...
+  defineMetadata(value: any, type: any,...keys: string[]): void {
+    const metadata = type[Symbol.metadata] ??= Object.create(null);
+    const length = keys.length;
+    switch (length) {
+      case 0: throw new Error('At least one key must be provided');
+      case 1: metadata[keys[0]] = value; return;
+      case 2: metadata[keys[0]] = metadata[keys[1]] = value; return;
+      default: {
+        for (let i = 0; i < length; ++i) {
+          metadata[keys[i]] = value;
```
bigopon, line 92: this line is not covered, though we can add the `/* istanbul ignore next*/` above `default:` case so that it won't cause noise in our coverage.

bigopon (commented):
Nice work @Sayan751 , only a last few comments.

bigopon:
@Sayan751 I'll go ahead with the merge, the rest of the comments are minor non blocking, we can continue the discussion here and get them in later. Fantastic work 👍

packages-tooling/plugin-conventions/src/hmr.ts
```diff
@@ -55,8 +55,8 @@ export const getHmrCode = (className: string, moduleText: string = 'module', mod
 
     if (hot.data?.aurelia) {
       const newDefinition = $$CE.getDefinition(currentClassType);
-      $$M.define(newDefinition.name, newDefinition, currentClassType);
-      $$M.define(newDefinition.name, newDefinition, newDefinition);
+      $$M.defineMetadata(newDefinition, currentClassType, null, newDefinition.name);
```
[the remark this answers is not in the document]
Sayan751: You are right. Also, there is a minor error here; gone unnoticed, as this part of the code is not compiled. I will open up a new branch for the residual changes.

packages-tooling/plugin-conventions/src/preprocess-resource.ts
```diff
@@ -301,36 +434,65 @@ function findResource(node: Node, expectedResourceName: string, filePair: string
...
-
   } else {
     if (type === 'customElement') {
       // Custom element can only be implicit resource
       if (isImplicitResource && filePair) {
         return {
           className,
           implicitStatement: { pos: pos, end: node.end },
-          runtimeImportName: isViewPair ? 'view' : 'customElement'
+          runtimeImportName: 'defineElement'
```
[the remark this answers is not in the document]
Sayan751: We can do that. But the function `CustomElement()` in PascalCase looks odd :P; and we already  `@customElement`. 

I have not put much thought into this. Just that the `defineElement` seemed to be in my head low ceremony. I think that we don't need to convert the `CustomElement` to a function. If you prefer the `CustomElement.define`, I can make the changes.

packages/kernel/src/di.resolvers.ts
```diff
@@ -269,7 +267,6 @@ const createNewInstance = (key: any, handler: IContainer, requestor: IContainer)
   return handler.getFactory(key).construct(requestor);
 };
 
-// eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
-const isInterface = <K>(key: any): key is InterfaceSymbol<K> => isFunction(key) && key.$isInterface === true;
+const isInterface = <K>(key: any): key is InterfaceSymbol<K> => isObject(key) && (key as {$isInterface?: boolean}).$isInterface === true;
```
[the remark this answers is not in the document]
Sayan751: Unsure about that. Seems super lax.

packages-tooling/plugin-conventions/src/hmr.ts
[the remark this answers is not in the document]
Sayan751: Changed in the follow-up PR.

packages/runtime-html/src/resources/binding-command.ts
```diff
@@ -65,21 +65,24 @@ export type BindingCommandInstance<T extends {} = {}> = {
 export type BindingCommandType<T extends Constructable = Constructable> = ResourceType<T, BindingCommandInstance, PartialBindingCommandDefinition>;
 export type BindingCommandKind = IResourceKind & {
   // isType<T>(value: T): value is (T extends Constructable ? BindingCommandType<T> : never);
-  define<T extends Constructable>(name: string, Type: T): BindingCommandType<T>;
-  define<T extends Constructable>(def: PartialBindingCommandDefinition, Type: T): BindingCommandType<T>;
-  define<T extends Constructable>(nameOrDef: string | PartialBindingCommandDefinition, Type: T): BindingCommandType<T>;
-  getAnnotation<K extends keyof PartialBindingCommandDefinition>(Type: Constructable, prop: K): PartialBindingCommandDefinition[K];
+  define<T extends Constructable>(name: string, Type: T, decoratorContext?: DecoratorContext): BindingCommandType<T>;
```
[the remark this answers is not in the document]
Sayan751: Made the changes in the follow-up PR.

packages/runtime-html/src/resources/template-controllers/switch.ts
```diff
@@ -355,16 +342,35 @@ export class Case implements ICustomAttributeViewModel {
...
+        case 'false': return false;
+        // eslint-disable-next-line @typescript-eslint/strict-boolean-expressions
+        default: return !!v;
+      }
+    }
+  }
+];
+
+defineAttribute({ name: 'default-case', bindables, isTemplateController: true }, DefaultCase);
+defineAttribute({ name: 'case', bindables, isTemplateController: true }, Case);
```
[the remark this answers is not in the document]
Sayan751: Nope. The issue is slightly different. Once we define the metadata in `Class[Symbol.metadata]` (it is irrelevant how it is defined) on a parent class, it is inherited by the child class. And that causes the problem. From a different angle, it is a reasonable design that the child inherits the parents' metadata. If you want to override that, you can do so by using decorators, and the generated code for that will take care of it. But in the absence of a decorator, the metadata gets inherited, and that seems reasonable. If we had used decorators directly in our core code, ignoring the bundle size aspect, such gymnastic, like the code in question, would not be necessary.

packages/runtime-html/src/resources/custom-attribute.ts
```diff
@@ -63,19 +63,19 @@ export type CustomAttributeKind = IResourceKind & {
   for<C extends ICustomAttributeViewModel = ICustomAttributeViewModel>(node: Node, name: string): ICustomAttributeController<C> | undefined;
   closest<A extends object | Constructable, TType extends A extends Constructable<infer T extends object> ? Constructable<T> : Constructable<A> = A extends Constructable<infer T extends object> ? Constructable<T> : Constructable<A>>(node: Node, Type: CustomAttributeType<TType>): ICustomAttributeController<InstanceType<TType>> | null;
   closest<A extends object | Constructable, TType extends A extends Constructable<infer T extends object> ? Constructable<T> : Constructable<A> = A extends Constructable<infer T extends object> ? Constructable<T> : Constructable<A>>(node: Node, name: string): ICustomAttributeController<InstanceType<TType>> | null;
-  isType<T>(value: T): value is (T extends Constructable ? CustomAttributeType<T> : never);
+  isType<T>(value: T, context?: DecoratorContext | null): value is (T extends Constructable ? CustomAttributeType<T> : never);
```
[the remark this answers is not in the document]
Sayan751: Done in the follow-up PR.

packages/runtime-html/src/watch.ts
```diff
@@ -50,58 +50,65 @@ export function watch<T extends object, D = unknown>(
...
+      core(target as Constructable);
+    } else {
+      context.addInitializer(function (this: T) {
+        core(this.constructor as Constructable);
+      });
     }
-    if (isElementType(Type)) {
-      getElementDefinition(Type).watches.push(watchDef as IWatchDefinition);
+
+    function core(type: Constructable) {
```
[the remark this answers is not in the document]
Sayan751: Done in the follow-up PR.

packages/metadata/src/index.ts
Sayan751: That's a surprise, TBH. That should have been covered by the usage of `@customElement` alone. Hence, I think we don't need to ignore that line.

packages/runtime-html/src/bindable.ts
```diff
@@ -126,23 +159,16 @@ export const Bindable = objectFreeze({
...
     let Class: Constructable;
-    let i: number;
     while (--iProto >= 0) {
       Class = prototypeChain[iProto];
-      keys = getAllAnnotations(Class).filter(isBindableAnnotation);
-      keysLen = keys.length;
-      for (i = 0; i < keysLen; ++i) {
-        defs[iDefs++] = getOwnMetadata(baseName, Class, keys[i].slice(propStart)) as BindableDefinition;
-      }
+      const bindableMetadata = getMetadata<Record<PropertyKey, BindableDefinition>>(baseName, Class);
```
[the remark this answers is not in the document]
Sayan751: It depends on whether the metadata gets overridden by the child class, via the decorator infra or not. However, it might be of no consequence, as the `getAll` method:
- traverses the prototypal chain, making sure all the bindables are collected from the chain, and
- prioritize the bindables from the child class.

packages-tooling/plugin-conventions/src/preprocess-resource.ts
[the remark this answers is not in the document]
bigopon: We probably shouldn't provide two ways to do thing, `defineElement` competing against `CustomElement.define` is a weird situation that we can get ourselves in. Let's tweak it to `CustomElement.define` and mark `defineElement` internal?

packages/kernel/src/di.resolvers.ts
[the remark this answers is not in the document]
bigopon: Well that's the key though, anything that is intentionally decorated with `$isInterface` is enough, they are not any laxer than `object or function` condition

packages/runtime-html/src/resources/template-controllers/switch.ts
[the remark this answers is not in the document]
bigopon: We can use static blocks, so that the code will stay with the definition and we won't have to deal with loose calls like this, it'll help with tree shaking I believe.

packages/runtime/src/observation/proxy-decorators.ts
bigopon: so all decorators either have to be plain (no empty call `()` allowed)` or have context as their 2nd param?

packages-tooling/plugin-conventions/src/preprocess-resource.ts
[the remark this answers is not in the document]
Sayan751: Sure. I can refactor the convention plugin such that it uses the `CustomElement.define`, and `CustomAttribute.define`. That should do the job.

packages/kernel/src/di.resolvers.ts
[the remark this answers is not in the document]
Sayan751: True. I think it can simply be `key?.$isInterface === true`.

packages/runtime/src/observation/proxy-decorators.ts
Sayan751: I did not find a way to make TS happy. If it is a major annoyance, I can take a look again.

Sayan751: Done

ghiscoding:
@Sayan751 @bigopon hey guys, it would be helpful to get more details of the migration in either the PR or the release (preferable in the release) when an impactful breaking change like this one comes into play. I'm trying hard to keep my open source project Aurelia-Slickgrid up to date with Aurelia 2 but there's breaking change at every beta without much migration info (none in this case). After looking at the PR and searching for something similar to my code, like `@I18N` decorator, and then I found out that I had to do the following changes

```diff
- constructor(@I18N private readonly i18n: I18N) {
+ constructor(private readonly i18n: I18N = resolve(I18N)) {
```
or this other line

```diff
- constructor(@newInstanceOf(IHttpClient) readonly http: IHttpClient) {
+ constructor(readonly http: IHttpClient = resolve(newInstanceOf(IHttpClient))) {
```

Showing these kind of changes would be **really** helpful to be displayed directly in the release (or at the minimum in the PR). Angular, esbuild, they all do that in their release and it's very helpful and quicker to migrate.

Sayan751:
@ghiscoding You are absolutely right. It is my bad. I have plan to update the docs, but I am yet to do so. For now, I have updated the release notes with relevant details. Thank you for drawing our attention to this.

ghiscoding:
Thanks for that and all the work you do on Aurelia 2 :)

Sayan751:
Added another point related to coercing in the release notes.

tareqimbasher:
@Sayan751 @bigopon Hi guys. I am also a maintainer of an open-source project, NetPad, that is built using Aurelia 2 and appreciate you adding some documentation on this. I rely pretty heavily on decorators for all my constructor injection and after reviewing the changes I had a few questions before I embark on the non-trivial changes to the codebase needed for this update:

**1.** You mention that

> The decorator proposal does not support parameter decorators (yet)

Will parameter decorators eventually be supported, and thus be brought back to Au2 before release? Or is that up in the air. If the former, I'd rather hold off on upgrading to beta15 and wait for when they come back(?)

**2.** Migration instructions for parameter decorators is to use the `resolve` function. In your example the `@IBar private readonly bar: IBar` was moved outside of the `constructor` declaration. Is that a requirement because its a property while the `IFoo` remained as a constructor parameter since its not a class-scoped property and you used the `resolve` function in place?

I'd rather keep the class property in the constructor and use the `resolve` function there instead of moving it out, if possible.

**3.** You mention that

> If you are using only Aurelia decorators in your code, it should be fine

Just want to confirm that built-in decorators like `@bindable` and `@singleton` need no changes at all besides the TypeScript specific changes mentioned.

**4.** Lastly, what about decorators applied to classes or methods. Are those affected by this as well? ie. For built-in decorators they are fine, but custom one need to be migrated as shown in linked articles?

Thank you guys for all the amazing work you do on Aurelia 2. Really love the framework and try to push it in business settings.

ghiscoding:
@tareqimbasher on my side, all the `@bindable` and `@singleton` weren't touched at all by this new beta.15, so no worries for them. The real change is really all related to the dependency injection side of things. For reference you can take a look at my migration in this Aurelia-Slickgrid [PR](https://github.com/ghiscoding/aurelia-slickgrid/pull/1184) for Beta 15. I try to keep up to date with Au2 because it seems that every Beta version has breaking changes, so the more you wait, the more changes it will require you to do. I'd be curious to know when the Beta cycle ends? I mean RC was advertised not long ago I think!?

tareqimbasher:
Thank you @ghiscoding for the input and the PR, will take a look. I'm upgrading from beta.7 and currently got it working up to beta.14 besides a bug I'm trying to figure out related to the changes with `au-compose` introduced with beta.8, but ya I'm going to take your advice and try to keep more up to date with every release. Recent updates look like we're gearing up for an RC soon, and the release cycle is pretty regular at every 2 weeks which is great.

Also great lib, Aurelia-Slickgrid, looking forward to using it soon!

Sayan751:
@tareqimbasher 

> Will parameter decorators eventually be supported, and thus be brought back to Au2 before release?

That purely depends on whether and when this [proposal](https://github.com/tc39/proposal-class-method-parameter-decorators) reaches stage 3. It is out of our control to predict anything on that front. But yes, once the proposal is standardized, we will be glad to bring back parameter decorators. However, if I remember correctly, only the DI-related decorators were parameter decorators. You can easily replace those using `resolve`.

> 2. Migration instructions for parameter decorators is to use the resolve function. In your example the @IBar private readonly bar: IBar was moved outside of the constructor declaration. Is that a requirement because its a property while the IFoo remained as a constructor parameter since its not a class-scoped property and you used the resolve function in place? 
>I'd rather keep the class property in the constructor and use the resolve function there instead of moving it out, if possible.

Feel free to do so. The release note merely shows both possibilities. Plus, if those are classes are meant for inheritance, avoiding ctor parameters in the base class, makes it easier (less code) for the child class. Having said that, you can very well do something like below.
```ts
import { resolve } from 'aurelia';
class MyClass {
  public constructor(
    private readonly bar: IBar = resolve(IBar),
  ) {  }
}
```

> Just want to confirm that built-in decorators like @bindable and @singleton need no changes at all besides the TypeScript specific changes mentioned.

That is correct. Apart from the parameter decorators, every other decorator should be as it was before this change.

> 4. Lastly, what about decorators applied to classes or methods. Are those affected by this as well? ie. For built-in decorators they are fine, but custom one need to be migrated as shown in linked articles?

That is correct. If you have written any (experimental/TS) decorators, those needs to be migrated to the new proposal. Here is an [example](https://github.com/Netatwork-de/odata-edm-generator/pull/95/files#diff-aef260957872ec31f716dbdf309715e9fcd40d4da17e5c664850a01a1cc08e84), how we migrated our decorators at work.

Happy Migration!

bigopon:
@ghiscoding  @tareqimbasher 

> I'd be curious to know when the Beta cycle ends? I mean RC was advertised not long ago I think!?

with the completion of the decorator migration, RC is considered around the corner. We will have a few more beta versions as "buffer" to be safe, and to resolve potential issues with the new decorator implementation. The decorator proposal also [getting new capabilities considered](https://github.com/tc39/proposal-decorators/issues/521) so it's not even bad for us to delay a bit.

>  every Beta version has breaking changes, so the more you wait, the more changes it will require you to do

Thanks for the reminder @ghiscoding , I should say that while there' breaking changes, it's not the kind of changes that rock your apps, so the benefit of having checkpoints for applications to migrate will be appreciated later (at least from the maintainers perspective, maybe they'll be cursed less), especially there' probably many apps in that category.

PR #1947, opened 2024-04-17
refactor: cleanup deco code

bigopon:
## 📖 Description

Refactor some deco code so that it'll be consistent with the rest, simplify subscriber collection typings, employ static blocks to prepare for es2022 compilation target

### DIFF
--- a/packages/runtime-html/src/resources/template-controllers/switch.ts
+++ b/packages/runtime-html/src/resources/template-controllers/switch.ts
@@ -342,6 +342,34 @@ export class Case implements ICustomAttributeViewModel {
 }
 
 export class DefaultCase extends Case {
+  static {
+
+    // Notes:
+    // - The usage of $au is intentionally avoided here.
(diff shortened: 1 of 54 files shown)


packages/runtime-html/src/resources/template-controllers/switch.ts
```diff
@@ -342,35 +342,38 @@ export class Case implements ICustomAttributeViewModel {
 }
 
 export class DefaultCase extends Case {
+  static {
```
Sayan751, line 345: @bigopon Are you sure about it? What happens if an app does not use the DefaultCase at all? Will the Case still work?

bigopon: static blocks are considered side-effectful atm, so these will always run. Undesirable, but also desirable. We will adjust accordingly later, I'm just grouping the code so it's not too loose, and not too tight.

PR #1954, opened 2024-04-27
refactor: extract template compiler into own package

### LINKED ISSUE #1746: refactor: extract template compiler into a separate package
## 🔦 Context

As part of the pre compilation work, the template compiler, and likely the expression parsing will need to be extracted into a separate package so that it can be reused by both applications and build tools.

Maybe we can have like this
```diff
+ @aurelia/expression-parser
  @aurelia/runtime
+ @aurelia/template-compiler
  @aurelia/runtime-html
```

One challenge could be around the typings & instructions exports, though they shouldn't be difficult to work out.

bigopon:
## 📖 Description

Currently the template compiler and compilation related infra are part of the runtime html, but it doesn't need to be, splitting this out into its own separate package to promote a reusable bit for build time pre compilation.

The following exports will be moved to the new package `@aurelia/template-compiler` from `@aurelia/runtime-html`
```
- ITemplateCompiler
- ITemplateCompilerHooks,
- TemplateCompilerHooks,
- templateCompilerHooks,
- IAttrMapper
- ITemplateElementFactory
- IInstruction
- InstructionType
- ... instruction classes
- IAttrParser
- AttrSyntax
- IAttributePattern
- AttributePattern
- BindingCommand
- ... default binding command classes
```
Applications importing the above from the `aurelia` package won't need to adjust their code, since it'll be the same from within the `aurelia` package.

Side note: this also allows me to have a clearer look of what a hello world Aurelia app weights like, it's about ~100KB (minified) so this'll be our achievable target with build time pre-compilation, after probably quite some amount of work. With moderate amount of work, its about ~140KB (minified), which is quite acceptable for most apps.

### 🎫 Issues

Resolve #1746 

cc @Sayan751 @fkleuver


Sayan751:
Nice work @bigopon! Does this trigger a change in the convention package? For the `XXX.define` calls? Such as `BindingCommand.define` etc.?

bigopon:
> ... Does this trigger a change in the convention package? For the `XXX.define` calls? Such as `BindingCommand.define` etc.?

It does, I forgot about this. Maybe we can keep things as is by reexporting `BindingCommand` from runtime-html.

PR #1960, opened 2024-05-04
feat(convention): rewrite runtime-html decorators

Sayan751:
# Pull Request

## 📖 Description

As the wide support for TC39 decorators are still lacking, this PR refactors the convention-plugin to rewrite the following decorators:

- bindable
- containerless
- useShadowDOM
- capture
- alias 
- inject

Discord thread: https://discord.com/channels/448698263508615178/1230047298084409464

### DIFF
--- a/packages-tooling/plugin-conventions/src/preprocess-resource.ts
+++ b/packages-tooling/plugin-conventions/src/preprocess-resource.ts
@@ -70,9 +70,13 @@ interface CustomElementDecorator {
   namePosition: IPos;
 }
 
-type ReplaceableDecorators = {
+type ReplaceableDecorator = {
   position: IPos;
   modifiedContent: string;
+};
+
+type BindableDecorator = ReplaceableDecorator & {
+  isClassDecorator: boolean;
 };
 
 interface DefineElementInformation {
@@ -81,6 +85,7 @@ interface DefineElementInformation {
 }
 
 interface IFoundResource {
+  type?: ResourceType;
   className?: string;
   localDep?: string;
   needDefinition?: [number, string];
@@ -88,7 +93,8 @@ interface IFoundResource {
   runtimeImportName?: string;
   customElementDecorator?: CustomElementDecorator;
   defineElementInformation?: DefineElementInformation;
-  ceDecorators?: ReplaceableDecorators[];
+  ceDecorators?: ReplaceableDecorator[];
+  bindables?: BindableDecorator[];
 }
 
 interface IResourceDecorator {
@@ -104,7 +110,13 @@ interface IModifyResourceOptions {
   customElementDecorator?: CustomElementDecorator;
   transformHtmlImportSpecifier?: (path: string) => string;
   defineElementInformation?: DefineElementInformation;
-  ceDecorators?: ReplaceableDecorators[];
+  ceDecorators?: ReplaceableDecorator[];
+  ceBindables?: BindableDecorator[];
+}
+
+interface DecoratorInformation {
+  resourceType: IResourceDecorator | undefined;
+  ceDecorators: ReplaceableDecorator[];
 }
 
 export function preprocessResource(unit: IFileUnit, options: IPreprocessOptions): ModifyCodeResult {
@@ -117,7 +129,9 @@ export function preprocessResource(unit: IFileUnit, options: IPreprocessOptions)
   let implicitElement: IPos | undefined;
   let customElementDecorator: CustomElementDecorator | undefined; // for @customName('custom-name')
   let defineElementInformation: DefineElementInformation | undefined;
-  let ceDecorators: ReplaceableDecorators[] | undefined;
+  let ceDecorators: ReplaceableDecorator[] | undefined;
+  let ceBindables: BindableDecorator[] | undefined;
+  const bindables: BindableDecorator[] = [];
 
   // When there are multiple exported classes (e.g. local value converters),
   // they might be deps for composing the main implicit custom element.
@@ -155,7 +169,8 @@ export function preprocessResource(unit: IFileUnit, options: IPreprocessOptions)
       runtimeImportName,
       customElementDecorator: customName,
       defineElementInformation: $defineElementInformation,
-      ceDecorators: $ceDecorators
+      ceDecorators: $ceDecorators,
+      bindables: $bindables
     } = resource;
 
     if (localDep) localDeps.push(localDep);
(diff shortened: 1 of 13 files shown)


Sayan751:
@CollinHerber Do these changes covers most of your use-cases?

CollinHerber:
> @CollinHerber Do these changes cover most of your use-cases?

Yeah only one I think might be missing is @inject

3cp:
Question, are we going to switch to #1939 in future for conventions code rewrite? Not sure whether it covers all the use cases.

CollinHerber:
> @CollinHerber Do these changes cover most of your use-cases?

And @customElement

Sayan751:
> Question, are we going to switch to #1939 in future for conventions code rewrite? Not sure whether it covers all the use cases.

That's a good idea @3cp. However, it should be trivial to remove code later.

Sayan751:
@CollinHerber The resource decorators such as custom element, attribute etc. are already taken care of. I'll try to do something for the `@inject`.

packages-tooling/plugin-conventions/src/preprocess-resource.ts
```diff
@@ -596,6 +619,16 @@ function findResource(node: Node, expectedResourceName: string, filePair: string
       case 'tc': return `\nCustomAttribute.define({ name: '${name}', isTemplateController: true${bindableOption} }, ${className});\n`;
     }
   }
+  function rewriteNonDefinitionDecorators() {
+    const remove: IPos[] = [];
+    let content = '';
+    for (const d of decorators) {
```
CollinHerber, line 625: Should this have ?? []   
Incase no decorators

CollinHerber (approved):

packages-tooling/plugin-conventions/src/preprocess-resource.ts
Sayan751: The array is guaranteed to be truthy here.

packages-tooling/plugin-conventions/src/preprocess-resource.ts
```diff
@@ -434,72 +521,294 @@ function findResource(node: Node, expectedResourceName: string, filePair: string
...
 }
+
+function collectBindables(node: ClassDeclaration, code: string): BindableDecorator[] {
+  const bindables: BindableDecorator[] = [];
+
+  // class-level decorators
+  if (canHaveDecorators(node)) {
+    for (const decorator of (getDecorators(node) ?? [])) {
+      const de = decorator.expression;
+      if (!isCallExpression(de)) continue;
```
bigopon, line 743: I think we should do it the other way:
if it's bindable, then throw when not a call expression. This will cover both misuse in JS and TS while this current way will leave JS users confused when they forgot to specify the property.

packages-tooling/plugin-conventions/src/preprocess-resource.ts
```diff
@@ -216,42 +254,110 @@ function modifyResource(unit: IFileUnit, m: ReturnType<typeof modifyCode>, optio
...
   return m;
 }
 
+function createDecoratorStatement(decorators: ReplaceableDecorator[] | undefined, classPos: number, m: ReturnType<typeof modifyCode>): { statement: string; effectivePos: number } {
+  let statement = '';
+  if (decorators == null) return { statement, effectivePos: classPos };
+  for (const d of decorators) {
+    if (!d.isDefinitionPart) continue;
+    const end = d.position.end;
+    m.replace(d.position.pos, end, '');
```
bigopon, line 338: is it appropriate to modify the passed parameter when the function name starts with `create`? To me, this seems like something that may be overlooked later.

packages-tooling/plugin-conventions/src/preprocess-resource.ts
```diff
@@ -434,72 +521,294 @@ function findResource(node: Node, expectedResourceName: string, filePair: string
...
+        modifiedContent = `aliases: ${ceDefinitionOptions}`;
+        break;
       }
 
-      return result;
+      case 'inject': {
+        if (!isCallExpression(d.expression)) continue;
+
+        isDefinitionPart = false;
+        modifiedContent = `Reflect.defineProperty(${getText(node.name!, code)}, 'inject', { value: [${d.expression.arguments.map(a => getText(a, code)).join(', ')}], writable: true, configurable: true, enumerable: true });`;
```
bigopon, line 718: if we use `...define(...)` then this won't matter, as this `Reflect.defineProperty` will be following a side-effectful statement. If we ever switch to static `$au` then this will likely need some tweak as it'll be consider side-effectful. Probably not that important for resources since they are gonna be used but just mentioning this in case.

packages-tooling/plugin-conventions/src/preprocess-resource.ts
```diff
@@ -434,72 +521,294 @@ function findResource(node: Node, expectedResourceName: string, filePair: string
...
+        } else if (args.length === 1) {
+          // case 3: @bindable({...}) x
+          const definition = getText(args[0], code);
+          const name = getText(member.name, code);
+          bindables.push({
+            isDefinitionPart: true,
+            isClassDecorator: false,
+            position: getPosition(decorator, code),
+            modifiedContent: `{ name: '${name}', ...${definition} }`
+          });
```
bigopon, line 800: we likely need to handle here too, throwing is also ok, so JS app can have some signals. Writing wrong code + having no clue + not seeing anything maybe a bit tough to debug.
Or are we trying to leave custom `bindable` untouched, just in case?

packages/runtime-html/src/resources/custom-element.ts
```diff
@@ -36,7 +36,6 @@ import { ErrorNames, createMappedError } from '../errors';
 import { dtElement, getDefinitionFromStaticAu, type IResourceKind } from './resources-shared';
 
 export type PartialCustomElementDefinition<TBindables extends string = string> = PartialResourceDefinition<Omit<IElementComponentDefinition<TBindables>, 'type'> & {
-  readonly cache?: number | '*';
```
bigopon, line 39: 👍

bigopon (commented):
Nice work @Sayan751 , looks good to me though I'd prefer to have @3cp approve this 😺

packages-tooling/plugin-conventions/src/preprocess-resource.ts
Sayan751: Got it. I think that all these codes are temporary and can be removed as soon as the decorators are supported widely enough.

packages-tooling/plugin-conventions/src/preprocess-resource.ts
Sayan751: I originally gave benefit of doubt; if some other decorator is imported from other package that is also named bindable etc. However, not throwing error for potentially obvious wrong code is also bad strategy. Made the changes.

packages-tooling/plugin-conventions/src/preprocess-resource.ts
Sayan751: Renamed the methods.

packages-tooling/plugin-conventions/src/preprocess-resource.ts
Sayan751: Made the change.

3cp (approved):

bigopon:
@CollinHerber you can test it out with the latest dev version, thanks everyone, nice work 👍

PR #1962, opened 2024-05-06
feat(repeat): allow custom repeatable value

### LINKED ISSUE #851: repeat.for doesn't work on all Iterable's
maybe it can't but I think it should... the other day I was trying to iterate over collections from 
https://www.npmjs.com/package/jstreemap I thought it should work, but it didn't
```
Uncaught (in promise) Error: Code 0: Cannot add observers to object.
    at Object.error (reporter.js:39)
    at ForOfStatement.iterate (ast.js:939)
    at Repeat.normalizeToArray (repeat.js:175)
    at Repeat.beforeBind (repeat.js:37)
    at Controller.bindCustomAttribute (controller.js:399)
    at Controller.bind (controller.js:257)
    at Controller.bindControllers (controller.js:453)
    at Controller.bindBindings (controller.js:443)
    at Controller.bindCustomElement (controller.js:379)
    at Controller.bind (controller.js:255)
```

simple repro

```html
  <ul>
    <li repeat.for="e of set">${e}</li>
  </ul>
```

```ts
  set = new TreeSet();
  constructor() {
    this.set.add('c');
    this.set.add('a');
    this.set.add('b');
  }
```

### LINKED ISSUE #1703: More general repeat.for for FileList and other specialized collections
# 🙋 Feature Request

I have noticed `repeat.for` does not work for several specialized collections, like for example [`FileList`](/https://developer.mozilla.org/en-US/docs/Web/API/FileList).

This applies to both Aurelia 1

https://github.com/aurelia/templating-resources/blob/master/src/repeat-strategy-locator.ts#L31-L40

and Aurelia 2:

https://github.com/aurelia/aurelia/blob/0937b6308a7fdf77d27ff75c1e82bfd78913f9bd/packages/runtime-html/src/resources/template-controllers/repeat.ts#L749-L760

## 🤔 Expected Behavior

Since `for...of` works for `FileList` I expected the equivalent and equally structured Aurelia construct to work.

## 😯 Current Behavior

Repeating does not work since a strategy is not explicitly defined for `FileList`.

## 💁 Possible Solution

In Aurelia 1 I guess it would be possible for consumers to configure extra repeater strategies.
This seems harder to impossible to do in Aurelia 2 due to how the code is structured (at least in the snippet I found).

Explicitly defining a strategy for every possible typed collection does not seem viable. Moreover, according to [`MDN`](/https://developer.mozilla.org/en-US/docs/Web/API/FileList) these kinds of specialized collections are getting rarer.

Maybe some kind of duck-typed check (e.g. whether the object has a `length` or `size` property, whether it is [iterable](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols#the_iterable_protocol), ...) could be more appropriate.

Ultimately, note that this missing functionality can be substituted by simply iterating over indexes:

```html
<ul>
    <li repeat.for="i of files.length">
        ${files[i].name}
    </li>
</ul>
```

bigopon:
## 📖 Description

Currently the repeater only knows a fixed set of repeatable value, as we wanted to start simple. Though this is a missing capability from v1, where applications can register custom collection strategies to iterate and create views. In v2, we want to centralize view related work in the repeater, so exposing the ability to have a collection strategy to mostly iterate.

### 🎫 Issues

close #1703
close #851

cc @fkleuver @Sayan751


docs/user-docs/templates/repeats-and-list-rendering.md
```diff
@@ -306,3 +306,78 @@ Our value converter uses the Javascript `Reflect` API to get the keys of our obj
...
 {% endhint %}
+
+### Custom collection handling strategy
+
+In the previous section, we see that any object can be iterated after using a value converter to convert it into an array, or any other form that Aurelia understands how to iterate. There is another way to teach Aurelia how to iterate an object using the `IRepeatableHandler` interface.
+
+An example of teaching Aurelia to iterate over any array like objects (HTMLCollection, NodeList, FileList etc...), is as follow:
+
+```typescript
+import Aurelia, { Registration, IRepeatableHandler } from 'aurelia';
```
elitastic, line 317: Shouldn't `IRepeatableHandler` be used anywhere in the code?

bigopon: Thanks, fixed 👍

bigopon:
Doc looks like this https://docs.aurelia.io/~/revisions/bZNKO72tg2q8awKheGLm/templates/repeats-and-list-rendering#custom-collection-handling-strategy

PR #1963, opened 2024-05-07
feat(template): auto infer binding expression when empty

### LINKED ISSUE #1259: [RFC] Ability to not double specify on bindable match on name
# 💬 RFC

Allow user to ignore specifying the same property name if matched on the attribute binding to.

## 🔦 Context

This would shrink html and make more readable

## 💻 Examples

```ts
class TestCE {
@bindable value:string;
}

class App {
value = 'test'
}
```

Current State
```html
<test-ce value.bind="value" />
```

Proposed
```html
<test-ce value.bind  />
// or using shorthand
<test-ce :value />
```

bigopon:
## 📖 Description

Our binding syntax comprises of 3 parts: target, command and expression as per the following formular:
```
target.command="expression"
```
Currently, when `expression` is absent, or empty, like the following examples:
```html
<input value.bind="">
or
<input value.bind>
```
it's treated as an empty string, which means the template above works like this:
```html
<input value.bind="``">
```

This is a missed-opportunity in v1, where we could have better dev experience, like suggested in #1259 . When the `expression` is either empty or absent, we can infer the expression based on the `target`. so for the above example, it should be understood as:
```html
<input value.bind="value">
```

All binding commands that will share this behavior are:
- `.bind`
- `.one-time`
- `.to-view`
- `.from-view`
- `.two-way`
- `.attr`

This is technically a breaking change, but no-one would write anything meaningful with empty expression so the effect of this should be small, or none. Though it's still safer to have this done in beta so we don't cause unnecessary discomfort.

### 🎫 Issues

Close #1259 

cc @fkleuver @Sayan751 @brandonseydel

### DIFF
--- a/docs/user-docs/developer-guides/migrating-to-aurelia-2/README.md
+++ b/docs/user-docs/developer-guides/migrating-to-aurelia-2/README.md
@@ -149,10 +149,19 @@ Read more about dynamic composition in v2 in this [dynamic composition doc](../.
 * Templates no longer need to have `<template>` tags as the start and ending tags. Templates can be pure HTML with enhanced Aurelia markup but `<template>` doesn't need to be explicitly defined.
 * `PLATFORM.moduleName` is gone. This was to address a limitation in Aurelia 1. Aurelia 2 now works well with all bundlers and does not require the addition of this code to use code splitting or tell the bundler where template code is.
 * Better intellisense support for TypeScript applications. Using the new injection interfaces, you can now inject strongly typed Aurelia packages such as Fetch Client, Router or Internationalization. These packages are prefixed with an "I" such as `IHttpClient`, `IRouter` and so on.
+* empty binding expressions are automatically inferred based on the target of the binding, like the following example:
+    ```html
+    <input value.bind>
+    <input value.bind="">
+    ```
+    means:
+    ```html
+    <input value.bind="value">
+    ```
 
 ## Plugins:
 
 ### Web-Components plugin
 
-* Remove automatic au- prefix
+* Remove automatic `au-` prefix
 * Remove auto-conversion of Aurelia element -> WC element. Applications need to explicitly define this. This should make mix-matching & controlling things easier.
--- a/docs/user-docs/templates/template-syntax/attribute-binding.md
+++ b/docs/user-docs/templates/template-syntax/attribute-binding.md
@@ -18,7 +18,12 @@ The basic syntax for binding to attributes in Aurelia is straightforward:
 You can bind to almost any attribute listed in the comprehensive HTML attributes list, which can be found [here](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes).
 
 {% hint style="info" %}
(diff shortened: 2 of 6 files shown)


ghiscoding:
@bigopon I'm trying to upgrade to Beta 17 and I'm getting a few errors as shown below in my Aurelia-Slickgrid demos, all of the code that fails have the `@bindable` decorator in this [CI Failing Workflow](https://github.com/ghiscoding/aurelia-slickgrid/actions/runs/9048658136/job/24861923782)

>  TS2488: Type 'Record<string, PartialBindableDefinition>' must have a '[Symbol.iterator]()' method that returns an iterator.

I don't exactly understand what I'm supposed to do in order to migrate? If I remove all the `@bindable` then it compiles without error but all my E2E tests then fail. 

Here's code for 1 example that I have (code ref [link](https://github.com/ghiscoding/aurelia-slickgrid/blob/70a88a3f814ea4f9e2626e1b1fee9033ffad7dc0/packages/demo/src/examples/slickgrid/example21.ts#L12C1-L15C32))

```ts
export class Example21 {
  @bindable() selectedColumn!: Column;
  @bindable() selectedOperator!: string;
  @bindable() searchValue = '';
```

Do you have a clue what the errors are all about? Thanks

bigopon:
@ghiscoding this likely is a typing issur. Can you try `bindable() as any` to see if theres any runtime error?

bigopon:
Nvm, sorry I forgot we rewrite decorator code with convention plugin. This likely is a types issues from rewritten code.

bigopon:
@ghiscoding we have a test for this scenario too in our e2e, though a difference is in our setup we use `transpileOnly` here https://github.com/aurelia/aurelia/blob/master/packages/__e2e__/3-hmr-webpack/webpack.config.js#L36-L42
This is the reason why we didn't catch the issue. For now, you can apply the same change to temporarily work around it, will fix soon. Thanks!

@obedm503 you don't like this change?

obedm503:
I don't like implicit behavior. An empty binding should result in an error. Even `something.bind=""` should be an error instead of an empty string.

ghiscoding:
> @ghiscoding we have a test for this scenario too in our e2e, though a difference is in our setup we use `transpileOnly` here https://github.com/aurelia/aurelia/blob/master/packages/__e2e__/3-hmr-webpack/webpack.config.js#L36-L42 This is the reason why we didn't catch the issue. For now, you can apply the same change to temporarily work around it, will fix soon. Thanks!
> 
> @obedm503 you don't like this change?

Thanks that worked, even if it wasn't an obvious fix. 👍🏻

bigopon:
> ... implicit behavior. An empty binding should result in an error ...

I think when faced with an empty string, we have 2 choices: strict (throw) and helpful (infer).

We can choose to be strict, but I think being helpful would be nicer, especially if it's not obscure, which is the case for our change here. Whatever the expression is can be seen from the attribute, there's nothing hidden at all.

ghiscoding:
> We can choose to be strict, but I think being helpful would be nicer, especially if it's not obscure, which is the case for our change here

I agree which is why I wrote here because I wasn't sure what to do since the message was confusing

PR #1965, opened 2024-05-08
feat(template): support spread syntax ...

### LINKED ISSUE #1258: [RFC] Bindables built in attribute (Spread like)
# 💬 RFC
Aurelia could benefit in a syntax to leverage bindable multiple properties of an object to individual bindables rather than a state object.  This can be useful when dealing with UI libs and such.

# Examples
The purpose is to unlock the capability to send an object to a CE and let properties bind to bindables of that CE without having to use one state object.

```ts
export class PersonViewer {
  @bindable firstName:string;
  @bindable lastName:string;
}
```

Originally had to always go this route

```html
<person-view first-name.bind="person.firstName" last-name.bind="person.lastName" />
```
Proposed
```html
<person-view bindables.bind="person" />
```

Bindables will assign properties of the object to bindables of the CE matched by name.  This would include functions and such.

### Some talking points:
If there is a bindable say middle name and the object passed is not there should be create the property on the object?  If we do not do this then if another CE adds the property the current CE wouldnt be notified of the change.  So is auto creating properties on an object OK, optional, or no way?  

The most obvious use case of this is to be able to use UI libs without tons of bindables.
```html
<md-button bindables.bind="{ disabled:true, outline:true, small:true }" />
// compared to 
<md-button disabled.bind="true" outline.bind="true" small.bind="true" />
```

If we include both does order trump?

```html
<mb-button small.bind="false" bindables.bind="{small:true}" />
```
Is small bound to false or the objects true?

### LINKED ISSUE #1477: Ability to spread an object to bindables
# 🙋 Feature Request

When looping through sometimes I would like to spread the object I am looping through on the component rather than oen by one

## 🤔 Expected Behavior

```ts
        <k-button  repeat.for="action of actions" ...action>${action.content}</k-button>
```

## 😯 Current Behavior

```ts
        <k-button color.bind="action.color"
                          size.bind="action.size"
                          type.bind="action.type" 
                          epeat.for="action of actions">${action.content}</k-button>

```

## 💁 Possible Solution

Allow some form of spread to bind al properties from object that match

bigopon:
## 📖 Description

When the number of bindable properties grows bigger, and the binding are repetitive, it's quite a boilerplate having to specify everything, as per the examples in #1477 #1258 :

Desirable:

```html
  <k-button  repeat.for="action of actions" ...action>${action.content}</k-button>
  <k-button  repeat.for="action of actions" ...$bindables="action">${action.content}</k-button>
  <k-button  repeat.for="action of actions" $bindables.spread="action">${action.content}</k-button>
```

Not desirable
```html
  <k-button color.bind="action.color"
            size.bind="action.size"
            type.bind="action.type" 
            repeat.for="action of actions">${action.content}</k-button>  
```

This PR adds support for spread syntaxes:
- full: `$bindables.spread="expression"`
- short: `...$bindables="expression"` 
- shorter: `...expresion`

Note:
- binding created via `$bindables` will have mode `to-view`, ignoring any default mode of the target bindable properties
- If the same object is returned from evaluating the expression, the spread binding won't try to rebind its inner bindings. This means mutating and then reassigning won't result in new binding, instead, give the spread binding a new object.
- fix an issue with spreading `...$attrs` not consider element bindable properties before custom attribute

### 🎫 Issues

Close #1477
Close #1258

cc @fkleuver @Sayan751 @brandonseydel @Vheissu

### DIFF
--- a/docs/user-docs/components/bindable-properties.md
+++ b/docs/user-docs/components/bindable-properties.md
@@ -597,8 +597,24 @@ Sometimes when the expression of the spread binding is simple, we can simplify t
 
 {% hint style="warning" %}
 - Remember that HTML is case insensitive, so `...firstName` actually will be seen as `...firstname`, for example
-- If the expression contains space, it will result into multiple attributes and thus won't work as intended with spread syntax `...`
+- If the expression contains space, it will result into multiple attributes and thus won't work as intended with spread syntax `...`.
+For example `...a + b` will be actually turned into 3 attributes: `...a`, `+` and `b`
 {% endhint %}
+
+### Binding orders
+
+The order of the bindings created will be the same with the order declared in the template. For example, for the `NameTag` component above, if we have a usage
+
+```html
+<name-tag id="1" first="John" ...$bindables="{ first: 'Jane' }">
+<name-tag id="2" ...$bindables="{ first: 'Jane' }" first="John">
+```
+Then the value of the `first` property in `NameTag` with `id=1` will be `Jane`, and the value of `first` property in `NameTag` with `id=2` will be `John`.
+
+{% hint style="warning" %}
+- An exception of this order is when bindables spreading is used together with [`...$attrs`](#attributes-transferring), `...$attrs` will always result in bindings after `...$bindables`/`$bindables.spread`/`...expression`.
+{% endhint %}
+
 
 ## Attributes Transferring
 
--- a/docs/user-docs/components/bindable-properties.md
+++ b/docs/user-docs/components/bindable-properties.md
@@ -615,6 +615,37 @@ Then the value of the `first` property in `NameTag` with `id=1` will be `Jane`, 
 - An exception of this order is when bindables spreading is used together with [`...$attrs`](#attributes-transferring), `...$attrs` will always result in bindings after `...$bindables`/`$bindables.spread`/`...expression`.
 {% endhint %}
 
+### Observation behavior
+
+Bindings will be created based on the keys available in the object evaluated from the `expression` of a spread binding. The following example illustrate the behavior:
+
+For the `NameTag` component above:
+```html
+<let item.bind="{ first: 'John' }">
+<name-tag ...item></name-tag>
+<button click.trigger="item.last = 'Doe'">Change last name</button>
+```
+
+The rendered HTML of `<name-tag>` will be
+```html
+<b>JOHN</b>
+```
+
+When clicking on the button with text `Change last name`, the rendered html of `<name-tag>` won't be changed,
+as the original object given to `<name-tag>` doesn't contain `last`, hence it wasn't observed, which ignore our new value set from the button click.
+If it's desirable to reset the observation, give a new object to the spread binding, like the following example:
+
+```html
+<let item.bind="{ first: 'John' }">
+<name-tag ...item></name-tag>
+<button click.trigger="item = { first: item.name, last: 'Doe' }">Change last name</button>
+```
+
+{% hint style="success" %}
+With the above behavior of non-eager binding, applications can have the opportunity to leave some bindable properties alone,
+while with the opposite of always observing all properties on the given object based on the number of bindable properties,
+missing value (`null`/`undefined`) will start flowing in in an unwanted way.
+{% endhint %}
 
 ## Attributes Transferring
 
--- a/docs/user-docs/components/bindable-properties.md
+++ b/docs/user-docs/components/bindable-properties.md
@@ -583,6 +583,9 @@ The `...$bindables="..."` syntax will only connect properties that are matching 
 <name-tag $bindables.spread="customer1">
 <name-tag $bindables.spread="customer.details">
 <name-tag $bindables.spread="customer[this_that]">
+<name-tag $bindables="customer1 | mapDetails">
+<name-tag $bindables="customer.details | simplify">
+<name-tag $bindables="customer[this_that] | addDetails">
 ```
 
 ### Shorthand syntax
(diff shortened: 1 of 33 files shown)


brandonseydel:
How does order work or is there an error?  e.g.

```html
<mb-button small.bind="false" bindables.bind="{small:true}" />
```

bigopon:
> How does order work or is there an error? e.g.
> 
> ```
> <mb-button small.bind="false" bindables.bind="{small:true}" />
> ```

It will create 2 binding in the order seen in the template.

Vheissu (approved):
Amazing work @bigopon it's actually awesome how very little code this is to implement.

docs/user-docs/components/bindable-properties.md
```diff
@@ -547,6 +547,60 @@ export class MyEl {
...
+```html
+<name-tag ...bindables="{ first: 'John', last: 'Doe' }"></name-tag>
+```
+
+The rendered html will be:
+```html
+<b>JOHN</b> Doe
+```
+
+Here we are using `...bindables` to express we want to bind all properties in the object `{ first: 'John', last: 'Doe' }` to bindable properties on `<name-tag>` component.
```
Sayan751, line 577:
```suggestion
Here we are using `...bindables` to express that we want to bind all properties in the object `{ first: 'John', last: 'Doe' }` to bindable properties on `<name-tag>` component.
```

packages-tooling/rollup-utils.mjs
```diff
@@ -109,8 +105,8 @@ export function getRollupConfig(pkg, configure = identity, configureTerser, post
   const inputFile = 'src/index.ts';
   // const esmDevDist = 'dist/esm/index.dev.mjs';
   // const cjsDevDist = 'dist/cjs/index.dev.cjs';
-  const esmDist = 'dist/esm/index.mjs';
-  const cjsDist = 'dist/cjs/index.cjs';
+  // const esmDist = 'dist/esm/index.mjs';
+  // const cjsDist = 'dist/cjs/index.cjs';
```
Sayan751, line 109: Missing cleanup? :) Also the same in the other rollup-utils file.

Sayan751:
Thanks @bigopon for this nice addition!

bigopon:
@fkleuver  @Sayan751 I ended up giving a special treatment to `...` syntax, embed the knowledge of it into the compiler. I think it's more appropriate this way, as there are syntax, parser and order matters to juggle, separating the work out into different concepts likely gonna results in leaky abstraction.

bigopon:
@Sayan751 I'll go ahead with a new beta for the weekend, but will address any comments you have.

packages/runtime-html/src/binding/spread-binding.ts
```diff
@@ -148,3 +154,183 @@ export class SpreadBinding implements IBinding, IHasController {
...
+    for (key in this._bindingCache) {
+      this._bindingCache[key].unbind();
+    }
+  }
+
+  /**
+   * @internal
+   */
+  private _createBindings(value: Record<string, unknown> | null, unbind: boolean) {
+    if (value == null) {
```
Sayan751, line 290: Maybe I have missed something, however, aren't these two cases are candidates for early terminations?

packages/template-compiler/src/template-compiler.ts
```diff
@@ -707,15 +710,21 @@ export class TemplateCompiler implements ITemplateCompiler {
...
         }
 
-        canCapture = realAttrTarget !== auslotAttr && realAttrTarget !== 'slot';
+        canCapture = realAttrTarget !== auslotAttr
+          && realAttrTarget !== 'slot'
+          && ((spreadIndex = realAttrTarget.indexOf('...')) === -1
+            // the following condition will allow syntaxes:
+            // ...$bindables
+            // ...some.expression
+            || (spreadIndex === 0 && (realAttrTarget === '...$attrs' /* || realAttrTarget === '...$element' */))
```
Sayan751, line 719: `/* || realAttrTarget === '...$element' */` can be removed, right?

packages/__tests__/src/3-runtime-html/spread.spec.ts
```diff
@@ -1,316 +1,633 @@
...
-      }
+        assert.html.innerEqual(appHost, '');
+      },
     });
 
-    $it('spreads custom attribute (with interpolation)', {
-      template: `<my-input my-attr="\${size}">`,
-      component: { size: 20 },
-      registrations: [MyAttr],
+    $it('does not apture slot', {
```
Sayan751, line 79: Typo; I am assuming you meant 'capture'.

Sayan751 (commented):
Great work @bigopon! Just some minor comments and questions.

packages/runtime-html/src/binding/spread-binding.ts
bigopon: I thought i was efficient ... this actually is the same with the next condition. Thanks

packages/template-compiler/src/template-compiler.ts
bigopon: resolved in #1968

packages/__tests__/src/3-runtime-html/spread.spec.ts
bigopon: resolved in #1968

packages/runtime-html/src/binding/spread-binding.ts
```diff
@@ -148,3 +154,183 @@ export class SpreadBinding implements IBinding, IHasController {
...
+    if (!this.isBound) {
+      /* istanbul ignore next */
+      return;
+    }
+    this.updateTarget();
+  }
+
+  public bind(scope: Scope) {
+      /* istanbul ignore if */
+    if (this.isBound) {
```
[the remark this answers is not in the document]
bigopon: resolved in #1968

PR #1966, opened 2024-05-12
feat(router-lite): current route

Sayan751:
# Pull Request

## 📖 Description

Related to #1819

### DIFF
--- a/packages/router-lite/src/current-route.ts
+++ b/packages/router-lite/src/current-route.ts
@@ -1,0 +1,57 @@
+import { DI, IDisposable, Writable, emptyArray, resolve } from '@aurelia/kernel';
+import { batch } from '@aurelia/runtime';
+import { IRouter } from './router';
+import { IRouterEvents } from './router-events';
+
+import type { Params, ViewportInstruction } from './instructions';
+import type { RouteConfig } from './route';
+
+let _currentRouteSubscription: IDisposable | null = null;
+/** @internal */
+export function _disposeCurrentRouteSubscription(): void {
+  _currentRouteSubscription?.dispose();
+}
+
+export const ICurrentRoute = /*@__PURE__*/ DI.createInterface<ICurrentRoute>('ICurrentRoute', x => x.singleton(CurrentRoute));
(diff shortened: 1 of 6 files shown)


Sayan751:
Doc: https://docs.aurelia.io/~/revisions/MY9zJSoqAPCuKM0A8jcf/getting-to-know-aurelia/routing/aurelia-router-lite/router-events#current-route

packages/router-lite/src/current-route.ts
```diff
@@ -0,0 +1,65 @@
...
+  public readonly path: string = '';
+  public readonly url: string = '';
+  public readonly title: string = '';
+  public readonly query: URLSearchParams = new URLSearchParams();
+  public readonly parameterInformation: ParameterInformation[] = emptyArray;
+
+  public constructor() {
+    const router = resolve(IRouter);
+    const options = router.options;
+    _currentRouteSubscription = resolve(IRouterEvents)
```
bigopon, line 29: We are not even disposing the subscription, and I guess the intention is to fire & forget, maybe we don't even need a variable to hold the sub

bigopon (approved):
Nice work @Sayan751 👍 only a cleanup comment.

packages/router-lite/src/current-route.ts
Sayan751: It is done in the [app task](https://github.com/aurelia/aurelia/pull/1966/files#diff-f1c72185a40f97fb259580ebad4cc0da926cfacfb404b7874543ca1c0cb20572R75).

Sayan751:
@bigopon I have merged this PR; in case you have further comments, I will gladly open a follow-up PR for those.

packages/router-lite/src/current-route.ts
bigopon: Is this a bit dangerous? any new instance of this would create a forgotten sub, do we not want to deal with it right now?

Sayan751: That's a very valid point. I will make the change.

PR #1967, opened 2024-05-13
fix(convention): use array for bindables isntead of object

bigopon:
## 📖 Description

In the previous PR at #1960 , because of the presence of `transpileOnly: true` config in our webpack e2e, we accidentally introduced broken bindables code when the html template of a component has `<bindable>` tag. The following scenario will break:
```html
<bindable name="prop1">
```
```ts
class MyEl {
  @bindable value
}
```
as the generated code will be something like this:
```ts
...
CustomElement.define({ ...au2ViewDef, bindables: [...{prop1:{}}, 'value'] }, ...)
```
 This PR fixes this.

Thanks @ghiscoding for reporting this at https://github.com/aurelia/aurelia/pull/1963#issuecomment-2106112004

cc @Sayan751 @3cp 
Should we change the bindable generation code to use object syntax instead of array? like the following code for the above example:
```ts
...
CustomElement.define({ ...au2ViewDef, bindables: {...{prop1:{}}, value: {} }, ...)
```

### DIFF
--- a/packages-tooling/__tests__/src/plugin-conventions/preprocess-html-template.spec.ts
+++ b/packages-tooling/__tests__/src/plugin-conventions/preprocess-html-template.spec.ts
@@ -454,7 +454,7 @@ export const template = "<template ></template>";
 export default template;
 export const dependencies = [  ];
 export const containerless = true;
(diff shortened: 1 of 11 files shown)


Sayan751:
Thanks @bigopon! Somehow it went unnoticed.

PR #1969, opened 2024-05-13
chore(router-lite): current route subscription disposal

Sayan751:
# Pull Request

### 🎫 Issues

Related to https://github.com/aurelia/aurelia/pull/1966#discussion_r1598854920


packages/router-lite/src/current-route.ts
```diff
@@ -37,6 +33,7 @@ export class CurrentRoute {
           (this as Writable<CurrentRoute>).parameterInformation = vit.children.map((instruction) => ParameterInformation.create(instruction));
         });
       });
+    container.register(AppTask.deactivated(() => subscription.dispose()));
```
bigopon, line 36: Will doing this make the updates stop coming after stopping and starting an application again?

Sayan751: Alright! Might be something for future.

PR #1972, opened 2024-05-15
refactor(i18n-validation): replace errors with error codes

bigopon:
## 📖 Description

Cleaning up error code of i18n, validation & validation html packages.

@Sayan751 with this I've removed all the logger todos, so pinging you for awareness.


bigopon:
Coverage is so off, probably will need to merge all coverage before sending it to codecov.

PR #1973, opened 2024-05-16
Updated ISignaler documentation example

ivanbacher:
Based on this: https://discourse.aurelia.io/t/signaler-signal-is-not-a-funcion-aurelia-2-esnext/5104

updated `signaler.signal` to `signaler.dispatchSignal`


bigopon:
Ahh, thanks @ivanbacher

PR #1976, opened 2024-05-18
feat(kernel): add last resolver

Vheissu:
# Pull Request

## 📖 Description

This pull request introduces support for retrieving the last registered value from the Aurelia DI (Dependency Injection) container. In scenarios where multiple values are registered under the same key, the new last resolver ensures that the most recently registered value is always returned.

In the Aurelia DI system, it's possible to register multiple values under the same key. This feature can be useful for various purposes, such as overriding previously registered values without losing them. However, when retrieving these values, the container typically returns all instances associated with the key. There are cases where it's desirable to get only the most recent value, such as when the latest configuration or instance should take precedence.

The implementation of the last resolver involves creating a new resolver that:

- Checks for the presence of multiple values under a specified key.
- Retrieves and returns the last (most recently registered) value among them.

This approach leverages the existing functionality of the getAll method to fetch all instances associated with a key and then accesses the last element in the array.

- **Single Call to getAll**: The last resolver optimises performance by making a single call to the getAll method and storing the result.

- **Direct Element Access**: It accesses the last element directly using array indexing, ensuring no modifications are made to the array, which preserves immutability.

- **Graceful Handling of Empty Arrays**: The resolver includes checks to handle cases where no values are registered under the key, returning undefined in such situations.

### Usage

The following example demonstrates what happens when you register three different values under the same key. 

```typescript
container.register(Registration.instance('key', 'value1'));
container.register(Registration.instance('key', 'value2'));
container.register(Registration.instance('key', 'value3'));

const myValue = container.get(last('key'));
console.log(myValue); // Outputs: value3
```

### DIFF
--- a/packages/__tests__/src/1-kernel/di.last.spec.ts
+++ b/packages/__tests__/src/1-kernel/di.last.spec.ts
@@ -44,6 +44,6 @@ describe('1-kernel/di.last.spec.ts', function () {
 
   it('should return undefined if the key is not registered', function () {
     const result = container.get(last('key'));
-    assert.equal(result);
+    assert.strictEqual(result, undefined);
   });
 });
--- a/packages/__tests__/src/1-kernel/di.last.spec.ts
+++ b/packages/__tests__/src/1-kernel/di.last.spec.ts
@@ -16,12 +16,12 @@ describe('1-kernel/di.last.spec.ts', function () {
     assert.equal(result, 'value3');
   });
 
-  it('should return the last registered instance in a parent container', function () {
-    const parent = container.createChild();
-    parent.register(Registration.instance('key', 'parentValue1'));
-    parent.register(Registration.instance('key', 'parentValue2'));
-    const result = parent.get(last('key'));
-    assert.equal(result, 'parentValue2');
+  it('should return the last registered instance in a child container', function () {
+    const child = container.createChild();
+    child.register(Registration.instance('key', 'childValue1'));
+    child.register(Registration.instance('key', 'childValue2'));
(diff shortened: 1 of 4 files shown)


packages/__tests__/src/1-kernel/di.last.spec.ts
```diff
@@ -0,0 +1,49 @@
...
+
+  it('should return the last registered instance of a key', function () {
+    container.register(Registration.instance('key', 'value1'));
+    container.register(Registration.instance('key', 'value2'));
+    container.register(Registration.instance('key', 'value3'));
+    const result = container.get(last('key'));
+    assert.equal(result, 'value3');
+  });
+
+  it('should return the last registered instance in a parent container', function () {
```
bigopon, line 19: this `parent` should be called `child`

packages/__tests__/src/1-kernel/di.last.spec.ts
```diff
@@ -0,0 +1,49 @@
...
+    container.register(Registration.instance('key', instance1));
+    container.register(Registration.singleton('key', class { name = 'instance3'; }));
+    container.register(Registration.instance('key', instance2));
+    const result = container.get(last('key'));
+    assert.deepEqual(result, instance2);
+  });
+
+  it('should return undefined if the key is not registered', function () {
+    const result = container.get(last('key'));
+    assert.equal(result);
```
bigopon, line 47:
```suggestion
    assert.strictEqual(result, undefined);
```

bigopon (approved):
Nice work, thanks @Vheissu

PR #1977, opened 2024-05-18
fix: container specific registrables

Sayan751:
# Pull Request

## 📖 Description

This PR fixes the problem demonstrated in this [SB project](https://stackblitz.com/edit/au2-registrable-issue?file=src%2Fmain.ts). In short, some resources were registered globally, which can cause problem when dealing with multi-app setup. For example, app1 uses the `t` as translation attribute, and app2 uses `i18n` for the same purpose. As the patterns were registered globally, the configuration from one app overwrites the configuration from the another app. This can cause unwanted surprises.

This PR changes that behaviour by lazily registering the formerly globally registered resources (Registrables) to the container.

### DIFF
--- a/packages/__tests__/src/i18n/t/translation-integration.spec.ts
+++ b/packages/__tests__/src/i18n/t/translation-integration.spec.ts
@@ -1637,4 +1637,102 @@ describe('i18n/t/translation-integration.spec.ts', function () {
     }
 
   });
+
+  // This test doubles down as the containerization of the attribute patterns; that is not global registry of patterns other than the container.
+  it('different aliases can be used for different apps', async function () {
+    @customElement({ name: 'app-one', template: `<span t="\${key11}"></span><span t.bind="key12"></span>` })
+    class AppOne {
+      public key11: string = 'key11';
+      public key12: string = 'key12';
+    }
+
+    @customElement({ name: 'app-two', template: `<span i18n="\${key21}"></span><span i18n.bind="key22"></span>` })
+    class AppTwo {
+      public key21: string = 'key21';
+      public key22: string = 'key22';
+    }
+
+    const translation = {
+      key11: 'a',
+      key12: 'b',
+      key21: 'c',
+      key22: 'd',
+
+      key13: 'e',
+      key14: 'f',
+      key23: 'g',
+      key24: 'h',
+    };
+
+    const i18nInitOptions = {
+      fallbackLng: 'en',
+      fallbackNS: 'translation',
+      resources: { en: { translation } },
+    };
+    let checkPoint1 = false;
+    let checkPoint2 = false;
+    async function createAppOne() {
+      const ctx = TestContext.create();
+      const host = PLATFORM.document.createElement('app-one');
+      const au = new Aurelia(ctx.container);
+      au.register(
+        I18nConfiguration.customize((opt) => {
+          opt.translationAttributeAliases = ['t'];
+          opt.initOptions = i18nInitOptions;
+        })
+      );
+      checkPoint1 = true;
+
+      while (!checkPoint2) {
+        await new Promise((res) => setTimeout(res, 1));
+      }
+
+      au.app({ host, component: AppOne });
+      await au.start();
+
+      return { au, host, vm: au.root.controller.viewModel as AppOne, queue: ctx.platform.domQueue };
+    }
+
+    async function createAppTwo() {
+      while (!checkPoint1) {
+        await new Promise((res) => setTimeout(res, 1));
+      }
+
+      const ctx = TestContext.create();
+      const host = PLATFORM.document.createElement('app-two');
+      const au = new Aurelia(ctx.container);
+      au.register(
+        I18nConfiguration.customize((opt) => {
+          opt.translationAttributeAliases = ['i18n'];
+          opt.initOptions = i18nInitOptions;
+        })
+      );
+      checkPoint2 = true;
+
+      au.app({ host, component: AppTwo });
+      await au.start();
+      return { au, host, vm: au.root.controller.viewModel as AppTwo, queue: ctx.platform.domQueue };
+    }
+
+    const [
+      { au: au1, host: host1, vm: appOne, queue: q1 },
+      { au: au2, host: host2, vm: appTwo, queue: q2 },
+    ] = await Promise.all([createAppOne(), createAppTwo()]);
+
+    assert.html.textContent(host1, 'ab');
+    assert.html.textContent(host2, 'cd');
+
+    appOne.key11 = 'key13';
+    appOne.key12 = 'key14';
+    q1.flush();
+
+    appTwo.key21 = 'key23';
+    appTwo.key22 = 'key24';
+    q2.flush();
+
+    assert.html.textContent(host1, 'ef');
+    assert.html.textContent(host2, 'gh');
+
+    await Promise.all([au1.stop(), au2.stop()]);
+  });
 });
--- a/packages/kernel/src/di.container.ts
+++ b/packages/kernel/src/di.container.ts
@@ -171,6 +171,10 @@ export class Container implements IContainer {
       } else if ((def = getMetadata(resourceBaseName, current)!) != null) {
         def.register(this);
       } else if (isClass<StaticResourceType>(current)) {
+        if (hasRegistrable(current)) {
+          current.getRegistrable().register(this);
+          continue;
+        }
         const registrable = current[Symbol.metadata]?.[registrableMetadataKey] as IRegistry;
         if (isRegistry(registrable)) {
           registrable.register(this);
@@ -749,3 +753,7 @@ const isClass = <T>(obj: unknown): obj is Class<any, T> =>
 
 const isResourceKey = (key: Key): key is string =>
   isString(key) && key.indexOf(':') > 0;
+
+// Limit this method for now only to classes
+const hasRegistrable = (obj: Class<any, unknown>): obj is Class<any, unknown> & { getRegistrable(): IRegistry } =>
+  isFunction((obj as Class<any, unknown> & { getRegistrable(): IRegistry }).getRegistrable);
--- a/packages/kernel/src/di.container.ts
+++ b/packages/kernel/src/di.container.ts
@@ -33,6 +33,7 @@ import { emptyArray } from './platform';
 import { ResourceDefinition, StaticResourceType, resourceBaseName, type ResourceType } from './resource';
 import { getMetadata, isFunction, isString, objectFreeze } from './utilities';
 
+export const registrableMetadataKey = Symbol.for('au:registrable');
 export const Registrable = /*@__PURE__*/(() => {
   const map = new WeakMap<WeakKey, (container: IContainer) => IContainer | void>();
   return objectFreeze({
@@ -193,7 +194,14 @@ export class Container implements IContainer {
       } else if ((def = getMetadata(resourceBaseName, current)!) != null) {
         def.register(this);
       } else if (isClass<StaticResourceType>(current)) {
-        if (isString((current).$au?.type)) {
+        if (hasRegistrable(current)) {
+          current.getRegistrable().register(this);
+          continue;
+        }
+        const registrable = current[Symbol.metadata]?.[registrableMetadataKey] as IRegistry;
+        if (registrable != null && isRegistry(registrable)) {
+          registrable.register(this);
+        } else if (isString((current).$au?.type)) {
           const $au = current.$au;
           const aliases = (current.aliases ?? emptyArray).concat($au.aliases ?? emptyArray);
           let key = `${resourceBaseName}:${$au.type}:${$au.name}`;
@@ -770,3 +778,7 @@ const isClass = <T>(obj: unknown): obj is Class<any, T> =>
 
 const isResourceKey = (key: Key): key is string =>
   isString(key) && key.indexOf(':') > 0;
+
+// Limit this method for now only to classes
+const hasRegistrable = (obj: Class<any, unknown>): obj is Class<any, unknown> & { getRegistrable(): IRegistry } =>
+  isFunction((obj as Class<any, unknown> & { getRegistrable(): IRegistry }).getRegistrable);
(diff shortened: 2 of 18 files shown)


packages/__tests__/src/i18n/t/translation-integration.spec.ts
```diff
@@ -1637,4 +1637,102 @@ describe('i18n/t/translation-integration.spec.ts', function () {
...
+      while (!checkPoint1) {
+        await new Promise((res) => setTimeout(res, 1));
+      }
+
+      const ctx = TestContext.create();
+      const host = PLATFORM.document.createElement('app-two');
+      const au = new Aurelia(ctx.container);
+      au.register(
+        I18nConfiguration.customize((opt) => {
+          opt.translationAttributeAliases = ['i18n'];
```
bigopon, line 1706: This gives the feeling i18n package has so much control what to register. It's quite unclear if we need to rock the whole attr pattern + DI register infra.
Shouldn't i18n be able to create different registrations for different apps based on this line?

The following is what I meant
```ts
  class TranslationAttributePattern {}
  class TranslationBindAttributePattern {};

  for (const alias of aliases) {
    const bindAlias = `${alias}.bind`;

    patterns.push({ pattern: alias, symbols: '' });
    TranslationAttributePattern.prototype[alias] = (rawName, rawValue, parts) => new AttrSyntax(rawName, rawValue, '', alias);

    bindPatterns.push({ pattern: bindAlias, symbols: '.' });
    TranslationBindAttributePattern.registerAlias(alias);

    if (alias !== 't') {
      commandAliases.push(alias);
      bindCommandAliases.push(bindAlias);
    }
  }

  const renderers = [
    AttributePattern.define(patterns, TranslationAttributePattern),
    BindingCommand.define({name:'t', aliases: commandAliases}, TranslationBindingCommand),
    ...
```
We declare `TranslationAttributePattern` and `TranslationBindingCommand` on the spot, rather than using a global static class, and thus will be able to avoid conflicts between apps.

packages/kernel/src/di.container.ts
```diff
@@ -188,12 +168,17 @@ export class Container implements IContainer {
...
       } else if ((def = getMetadata(resourceBaseName, current)!) != null) {
         def.register(this);
       } else if (isClass<StaticResourceType>(current)) {
-        if (isString((current).$au?.type)) {
+        if (hasRegistrable(current)) {
+          current.getRegistrable().register(this);
+          continue;
+        }
+        const registrable = current[Symbol.metadata]?.[registrableMetadataKey] as IRegistry;
+        if (registrable != null && isRegistry(registrable)) {
```
bigopon, line 179: feels like `isRegistry` should do the optional chaining instead.

packages/kernel/src/di.container.ts
```diff
@@ -188,12 +168,17 @@ export class Container implements IContainer {
...
-        Registrable.get(current)!.call(current, this);
       } else if ((def = getMetadata(resourceBaseName, current)!) != null) {
         def.register(this);
       } else if (isClass<StaticResourceType>(current)) {
-        if (isString((current).$au?.type)) {
+        if (hasRegistrable(current)) {
+          current.getRegistrable().register(this);
+          continue;
+        }
+        const registrable = current[Symbol.metadata]?.[registrableMetadataKey] as IRegistry;
```
bigopon, line 178: This seems like much better than `Registrable` 👍 💯

packages/template-compiler/src/attribute-pattern.ts
```diff
@@ -491,125 +514,139 @@ export class AttributeParser implements IAttributeParser {
...
 export const AttributePattern = /*@__PURE__*/ objectFreeze<AttributePatternKind>({
   name: getResourceKeyFor('attribute-pattern'),
   define(patternDefs, Type) {
-    patterns.set(Type, patternDefs);
-    return Registrable.define(Type, (container: IContainer) => {
-      singletonRegistration(IAttributePattern, Type).register(container);
-    });
+    return {
+      register(container: IContainer) {
+        container.get(IAttributeParser).registerPattern(patternDefs, Type);
```
bigopon, line 544: this will make registration quite order sensitive, I think we should be careful introducing this.
At the moment, everything kind of will be resolved the moment they are needed, with this, its right at the registration time. This maybe fine but I think it's something we should avoid as it's not how registration is perceived in the rest of the core modules.

bigopon (commented):
Nice changes @Sayan751 the metadata is a good improvement. I've added some comments for other parts.

packages/template-compiler/src/attribute-pattern.ts
Sayan751: I understand this is a different pattern.

However, it won't make the attribute patterns order sensitive. Because, the `registerPattern` only collects the patterns. What is important is the `ISyntaxInterpreter#add` is called only once lazily, and the `ISyntaxInterpreter#add` sorts the patterns before creating the parser states. Hence, the order of the registration of those patterns would not matter.

If you meant the order of registration of the resources to the container generally, you might be right. A more preferable solution to me would be to use the AppTasks. However, here, the AppTasks aren't available for use.

I think here it will be alright, as the `IAttributeParser`, and `ISyntaxInterpreter` both have default singleton resolvers.

packages/__tests__/src/i18n/t/translation-integration.spec.ts
Sayan751: I understand what you are saying. I would say that irrespective of what i18n package does, it might still be worth it to limit the former `Registrable`s to a container rather than to global static state.

Coming back to the i18n, I know that your solution avoids the problem of having duplicate registrations. Assuming that we agree on the removal of the global `Registrable`s, doing so will result in loss of standard API like `registerAlias`. I think that's not a nice change.

bigopon: From the look of `registerAlias`, it doesn't seem encouraging any mid application usage. If this is correct, will it make a better API to have the configuration take care of the aliasing?
If it encouraged mid application usage, how would the code look like? It feels like it won't be pleasant to look at. To me, sounds like we can have a special higher level thing in i18n to take care of aliasing,

packages/kernel/src/di.container.ts
```diff
@@ -188,12 +168,17 @@ export class Container implements IContainer {
...
       if (isRegistry(current)) {
         current.register(this);
-      } else if (Registrable.has(current)) {
-        Registrable.get(current)!.call(current, this);
       } else if ((def = getMetadata(resourceBaseName, current)!) != null) {
         def.register(this);
       } else if (isClass<StaticResourceType>(current)) {
-        if (isString((current).$au?.type)) {
+        if (hasRegistrable(current)) {
+          current.getRegistrable().register(this);
```
bigopon, line 175: Isn't this just another version of `isRegistry`? Is there anything `getRegistrable` can do but `register` cannot?

packages/__tests__/src/i18n/t/translation-integration.spec.ts
Sayan751: Hmm... on a second thought, I can make the change.

packages/template-compiler/src/attribute-pattern.ts
```diff
@@ -462,27 +464,48 @@ export class AttributeParser implements IAttributeParser {
...
-      },
-      emptyArray as AttributePatternDefinition[]
-    );
-    interpreter.add(allDefs);
+    this._interpreter = resolve(ISyntaxInterpreter);
+    this._container = resolve(IContainer);
+  }
+
+  public registerPattern(patterns: AttributePatternDefinition[], Type: Constructable<IAttributePattern>): void {
+    if (this._initialized) throw createMappedError(ErrorNames.attribute_pattern_already_initialized);
```
bigopon, line 484: Here we kind of create a conflict with `registerAlias` API in i18n, if it was to be used for mid application aliasing, then this will prevent that.
Our previous API also doesn't allow it since it eagerly resolves, I guess we never intended `registerAlias` to be a non startup thing.

Sayan751: Nope. All these are supposed to be bootstrap-phase code, including `registerAlias`.

packages/kernel/src/di.container.ts
Sayan751: Yes, you are right. This can be cleaned up. I initially thought to keep the `register` and this method separate. But that probably does not have much utility.

Sayan751: We cannot really reuse register for this purpose, as the `AttributePattern.define` registers a singleton registration with the container and that once again triggers the `register` method. Hence, it is best to avoid that. I have reused the metadata, and it seems to be working.

packages/__tests__/src/i18n/t/translation-integration.spec.ts
Sayan751: Done.

packages/kernel/src/di.container.ts
bigopon: Reusing metadata sounds good, for the comment it cannot be used with attr pattern define, I think we should do it the same way we were handling with getRegistrable:
```ts
class Pattern {
  static register(c) {
    return c.get(AttrParser).register(patterns)
  }
}
```

packages/template-compiler/src/attribute-pattern.ts
```diff
@@ -462,27 +464,48 @@ export class AttributeParser implements IAttributeParser {
...
+
+    const $patterns = this._patterns;
+    for (const { pattern } of patterns) {
+      if ($patterns[pattern] != null) throw createMappedError(ErrorNames.attribute_pattern_duplicate, pattern);
+      $patterns[pattern] = { factory: container => container.get(Type) };
```
bigopon, line 489: it shouldn't have to be a callback, I think it can just be the type itself. This is all internal code so we don't have to create a factory to be passed around.

packages/template-compiler/src/attribute-pattern.ts
```diff
@@ -462,27 +464,48 @@ export class AttributeParser implements IAttributeParser {
...
+    }
+    this._allDefinitions.push(...patterns);
+  }
+
+  /** @internal */
+  private _initialize(): void {
+    this._interpreter.add(this._allDefinitions);
+    const _container = this._container;
+    for (const [, value] of Object.entries(this._patterns)) {
+      value.pattern = value.factory(_container);
```
bigopon, line 499: with the above change, it can be just
```ts
      value.pattern = container.get(value.Type);
```

packages/template-compiler/src/attribute-pattern.ts
```diff
@@ -491,125 +514,138 @@ export class AttributeParser implements IAttributeParser {
...
 }
 
-const getAllPatternDefinitions = <P extends Constructable>(Type: P): AttributePatternDefinition[] =>
-  patterns.get(Type as Constructable<IAttributePattern>) ?? emptyArray;
-
-const patterns = new WeakMap<Constructable<IAttributePattern>, AttributePatternDefinition[]>();
-
 export const AttributePattern = /*@__PURE__*/ objectFreeze<AttributePatternKind>({
   name: getResourceKeyFor('attribute-pattern'),
   define(patternDefs, Type) {
```
bigopon, line 541: maybe let's rename this as `create` or something similar? we are no longer "defining" like other places.

packages/template-compiler/src/attribute-pattern.ts
```diff
@@ -462,27 +464,48 @@ export class AttributeParser implements IAttributeParser {
...
-        return allDefs.concat(patternDefs);
-      },
-      emptyArray as AttributePatternDefinition[]
-    );
-    interpreter.add(allDefs);
+    this._interpreter = resolve(ISyntaxInterpreter);
+    this._container = resolve(IContainer);
+  }
+
+  public registerPattern(patterns: AttributePatternDefinition[], Type: Constructable<IAttributePattern>): void {
```
bigopon, line 483: I just realized this pattern of API for registration will disable child/per element attr parser, as child parser won't have access to patterns anymore. Do we want to preserve it?

packages/kernel/src/di.container.ts
Sayan751: I was referring to this code in my last comment:

```ts
export const AttributePattern = /*@__PURE__*/ objectFreeze<AttributePatternKind>({
  name: getResourceKeyFor('attribute-pattern'),
  create(patternDefs, Type) {
    return {
      register(container: IContainer) {
        container.get(IAttributeParser).registerPattern(patternDefs, Type);
        singletonRegistration(IAttributePattern, Type).register(container); //L1
      }
    };
  },
});
```

The `L1` essentially cause, later in the processing pipeline, one more execution of the `register` method. As the `register` method is a semi-reserved method from DI, having a `register` method on the pattern classes will always create unwanted calls to the method. This is the reason, it worked for the `getRegistrable` method, as it was not called `register` :D

And as we already the metadata infra in place, we can leverage that without inventing any new methods. And this way we can circumvent the issues with the `register` quite easily.

packages/template-compiler/src/attribute-pattern.ts
Sayan751: As far as I have understood, the attribute parser was never per element basis. Both the attribute parser and syntax interpreter are singletons. As the child containers are mostly inheriting the parents' registrations, registering a new instance of parser or interpreter will essentially resolve multiple instances of those! Plus, I have not seen any tests for this. 

I think that it is a rare edge-case where anyone would want to use two (or more) exclusive sets of attribute patterns; one for parent and another for a set of child elements. Doing that would essentially mean that 2 different binding languages are in use in different parts of the same app. 

Having said that, I think with this change the patterns are more local to the containers rather than the previous model, where everything is global, and one pattern definition can potentially replace a former definition. In that sense, this change should be less disruptive. I am assuming that if any such child element specific attribute parser did work before, might work here as well.

If you find it otherwise, please suggest a test case.

bigopon: I see, I think that's a pretty good argument for not caring about it. I was thinking of local binding syntax but maybe we don't need to encourage it right now.

bigopon (approved):
nice work @Sayan751

Issue #1979, opened 2024-05-21
RFC: Enhancements for Overriding Registrations in Aurelia DI Container


Vheissu:
# 💬 RFC

In the current implementation of Aurelia's dependency injection (DI) container, registering a new value under an existing key does not replace the original value but adds to it. This behaviour can lead to unintended side effects, such as users being surprised to find multiple instances registered under the same key (especially during testing). This RFC proposes multiple solutions to this issue by allowing more explicit control over registration overrides.

The need to override dependencies arises primarily in tests, where it might be desirable to stub/mock dependencies.

## Current behaviour

In the current implementation of Aurelia 2's dependency injection (DI) container, registering a value under a key does not replace any existing value associated with that key. Instead, it adds to the list of values registered under that key. This behaviour allows for multiple instances to be associated with the same key, which can be useful in some scenarios but may lead to unintended side effects in others.

When you register multiple instances under the same key, the DI container maintains all these instances. You can then retrieve these instances using methods like `getAll` or `all`. Here’s a brief overview of how this works:

```typescript
const container = DI.createContainer();
const instance1 = new MyService();
const instance2 = new MyService();

container.register(Registration.instance('myKey', instance1));
container.register(Registration.instance('myKey', instance2));

// Retrieve all instances registered under 'myKey'
const instances = container.getAll('myKey');
console.log(instances); // Will contain two instances of MyService[]
```

In this example, both `instance1` and `instance2` are registered under the key 'myKey'. When you call `getAll('myKey')`, you get an array containing both instances.

```typescript
// In a class constructor
@inject(all('myKey'))
class MyClass {
  constructor(public instances: MyService[]) {}
}

// Later in the code
const container = DI.createContainer();
const instance1 = new MyService();
const instance2 = new MyService();

container.register(Registration.instance('myKey', instance1));
container.register(Registration.instance('myKey', instance2));

const myClassInstance = container.get(MyService);
console.log(myClassInstance.instances); // Will contain two instances of MyService[]
```

Here, the `all('myKey')` usage in the `@inject` decorator ensures that instances will be an array containing all instances registered under 'myKey'.

### Enhancement 1: Override Option in Registration Methods

Extend the current registration methods to accept an `options` parameter with an override flag. When this flag is true, the registration will replace any value associated with the key.

```typescript
const container = DI.createContainer();
const instance1 = new MyService();
const instance2 = new MyService();

container.register(Registration.instance('myKey', instance1));
[...]

PR #1981, opened 2024-05-22
feat(kernel): add ability to deregister di dependencies

Vheissu:
# Pull Request

## 📖 Description

This relates to the RFC I created around the ability to unregister the DI container's dependencies to work around how DI works with multiple registered values under a key. Through this functionality, users will be able to completely remove all instances registered under a specific key and register a new instance.

### 🎫 Issues

References RFC: https://github.com/aurelia/aurelia/issues/1979

## 📑 Test Plan

New test cases have been added to the existing Kernel DI test suite. Specifcally, tests for determining different types of registrations and child container usage.


bigopon (approved):
Nice work @Vheissu

PR #1982, opened 2024-05-23
fix(router-lite): replace transition plan for same component with different paths

Sayan751:
# Pull Request

## 📖 Description

This fixes the issue reported in this [discourse post](https://discourse.aurelia.io/t/transition-plan-for-routes-when-parameters-do-not-change/5431) by @ivanbacher.

Additionally, it add some assertions to some existing tests.


bigopon (approved):
Well I'd say this is a very economic changes for app, maybe a bit too much 😃

PR #1985, opened 2024-05-26
feat(validation): state rule

Sayan751:
# Pull Request

## 📖 Description

This PR adds a new feature to the validation packages. So far, it was assumed that a validation rule can have 2 states, valid or invalid. Even though it is generally true that a rule has a single valid state, there are instances where a rule can have several invalid states. A classic example might be [certificate validation errors](https://x509errors.org/). 

The Aurelia validation has not yet supported a validation rule with multiple invalid states and state-specific error messages. This PR introduces the support for that. An example looks like as follows.

```ts
type Error = 'none' | 'fooError' | 'barError';

let state: Error = 'none';
validationRules
  .on(obj)
  .ensure('name')
  .satisfiesState<Error, string>(
    /* valid state                 */ 'none',
    /* state (validation) function */ (_value, _object) => state, 
    /* state -> message mapper     */ ($state) => $state === 'fooError' ? 'foo' : 'bar');
```

### DIFF
--- a/packages/validation/src/rule-interfaces.ts
+++ b/packages/validation/src/rule-interfaces.ts
@@ -11,8 +11,10 @@ export type ValidationRuleExecutionPredicate<TObject extends IValidateable = IVa
 export interface IValidationRule<TValue = any, TObject extends IValidateable = IValidateable, TState = unknown> {
   tag?: string;
   messageKey: string;
-  state?: TState;
-  isStateful: boolean;
+  /** @internal */
+  _state?: TState;
+  /** @internal */
+  _isStateful: boolean;
   canExecute(object?: IValidateable): boolean;
 
   /**
--- a/packages/validation/src/rule-interfaces.ts
+++ b/packages/validation/src/rule-interfaces.ts
@@ -8,9 +8,11 @@ import { IValidationMessageProvider } from './rules';
 export type IValidateable<T = any> = (Class<T> | object) & { [key in PropertyKey]: any };
 export type ValidationRuleExecutionPredicate<TObject extends IValidateable = IValidateable> = (object?: TObject) => boolean;
 
-export interface IValidationRule<TValue = any, TObject extends IValidateable = IValidateable> {
+export interface IValidationRule<TValue = any, TObject extends IValidateable = IValidateable, TState = unknown> {
   tag?: string;
   messageKey: string;
+  state?: TState;
+  isStateful: boolean;
   canExecute(object?: IValidateable): boolean;
 
   /**
--- a/packages/validation/src/rule-interfaces.ts
+++ b/packages/validation/src/rule-interfaces.ts
@@ -22,7 +22,7 @@ export interface IValidationRule<TValue = any, TObject extends IValidateable = I
    */
   execute(value: TValue, object?: TObject): boolean | Promise<boolean>;
   accept(visitor: IValidationVisitor): any;
-  getStateMessage?(): string;
+  getMessage?(): string;
 }
 
 export interface IRequiredRule extends IValidationRule { }
--- a/packages/validation/src/rule-interfaces.ts
+++ b/packages/validation/src/rule-interfaces.ts
@@ -8,13 +8,9 @@ import { IValidationMessageProvider } from './rules';
 export type IValidateable<T = any> = (Class<T> | object) & { [key in PropertyKey]: any };
 export type ValidationRuleExecutionPredicate<TObject extends IValidateable = IValidateable> = (object?: TObject) => boolean;
 
-export interface IValidationRule<TValue = any, TObject extends IValidateable = IValidateable, TState = unknown> {
+export interface IValidationRule<TValue = any, TObject extends IValidateable = IValidateable> {
   tag?: string;
   messageKey: string;
-  /** @internal */
-  _state?: TState;
-  /** @internal */
-  _isStateful: boolean;
   canExecute(object?: IValidateable): boolean;
 
   /**
@@ -26,6 +22,7 @@ export interface IValidationRule<TValue = any, TObject extends IValidateable = I
    */
   execute(value: TValue, object?: TObject): boolean | Promise<boolean>;
   accept(visitor: IValidationVisitor): any;
+  getStateMessage?(): string;
 }
 
 export interface IRequiredRule extends IValidationRule { }
--- a/packages/validation/src/rule-provider.ts
+++ b/packages/validation/src/rule-provider.ts
@@ -25,6 +25,7 @@ import {
   BaseValidationRule,
   StateRule,
   explicitMessageKey,
+  isStatefulRule,
 } from './rules';
 import {
   IValidateable,
@@ -237,7 +238,7 @@ export class PropertyRule<TObject extends IValidateable = IValidateable, TValue 
     const rule = this.latestRule;
     this.assertLatestRule(rule);
     // eslint-disable-next-line no-console
-    if (__DEV__ && rule._isStateful) { console.debug('Setting message to stateful rule will override the message mapper set to the rule.'); }
+    if (__DEV__ && isStatefulRule(rule)) { console.debug('Setting message to stateful rule will override the message mapper set to the rule.'); }
     this.messageProvider.setMessage(rule, message, explicitMessageKey);
     return this;
   }
@@ -643,8 +644,7 @@ export class ValidationMessageProvider implements IValidationMessageProvider {
       if (parsedMessage !== void 0) { return parsedMessage; }
     }
 
-    // In case of stateful rule, the message key is the message itself.
-    if (rule._isStateful) return this.setMessage(rule, messageKey);
+    if (isStatefulRule(rule)) return this.setMessage(rule, rule.getStateMessage());
 
     const validationMessages = ValidationRuleAliasMessage.getDefaultMessages(rule);
     let message: string | undefined;
--- a/packages/validation/src/rule-provider.ts
+++ b/packages/validation/src/rule-provider.ts
@@ -1,4 +1,4 @@
-import { Class, DI, ILogger, IServiceLocator, resolve } from '@aurelia/kernel';
+import { Class, DI, ILogger, IServiceLocator, isFunction, resolve } from '@aurelia/kernel';
 import {
   IExpressionParser,
   Interpolation,
@@ -25,7 +25,7 @@ import {
   BaseValidationRule,
   StateRule,
   explicitMessageKey,
-  isStatefulRule,
+  providesMessage,
 } from './rules';
 import {
   IValidateable,
@@ -237,8 +237,6 @@ export class PropertyRule<TObject extends IValidateable = IValidateable, TValue 
   public withMessage(message: string) {
     const rule = this.latestRule;
     this.assertLatestRule(rule);
-    // eslint-disable-next-line no-console
-    if (__DEV__ && isStatefulRule(rule)) { console.debug('Setting message to stateful rule will override the message mapper set to the rule.'); }
     this.messageProvider.setMessage(rule, message, explicitMessageKey);
     return this;
   }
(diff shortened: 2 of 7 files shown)


Sayan751:
[Docs](https://docs.aurelia.io/~/revisions/TZsvvlqCRO0QmAZsgzGz/aurelia-packages/validation/defining-rules)

image

packages/validation/src/rule-interfaces.ts
```diff
@@ -8,9 +8,13 @@ import { IValidationMessageProvider } from './rules';
...
 export type ValidationRuleExecutionPredicate<TObject extends IValidateable = IValidateable> = (object?: TObject) => boolean;
 
-export interface IValidationRule<TValue = any, TObject extends IValidateable = IValidateable> {
+export interface IValidationRule<TValue = any, TObject extends IValidateable = IValidateable, TState = unknown> {
   tag?: string;
   messageKey: string;
+  /** @internal */
+  _state?: TState;
+  /** @internal */
+  _isStateful: boolean;
```
bigopon, line 17: can this be combined into 1 single official API somehow? This feels like an invitation design oversight. Maybe `undefined`/`null` should just be enough to consider not stateful?

packages/validation/src/rule-provider.ts
```diff
@@ -628,11 +636,17 @@ export class ValidationMessageProvider implements IValidationMessageProvider {
...
-    if (parsedMessage !== void 0) { return parsedMessage; }
+    const messageKey = rule.messageKey;
+    const lookup = this.registeredMessages.get(rule);
+    if (lookup != null) {
+      const parsedMessage = lookup.get(explicitMessageKey) ?? lookup.get(messageKey);
+      if (parsedMessage !== void 0) { return parsedMessage; }
+    }
+
+    // In case of stateful rule, the message key is the message itself.
+    if (rule._isStateful) return this.setMessage(rule, messageKey);
```
bigopon, line 647: the rule provider knowing too much about the internal of the rule is quite undesirable. do we need to resort to this? It seems we already take dependency on `StateRule` inside this RuleProvider, maybe just be with it here?
```suggestion
    if (rule instanceof StateRule) return this.setMessage(rule, rule.message);
```

packages/validation/src/rule-provider.ts
```diff
@@ -234,7 +236,9 @@ export class PropertyRule<TObject extends IValidateable = IValidateable, TValue
   public withMessage(message: string) {
     const rule = this.latestRule;
     this.assertLatestRule(rule);
-    this.messageProvider.setMessage(rule, message);
+    // eslint-disable-next-line no-console
+    if (__DEV__ && rule._isStateful) { console.debug('Setting message to stateful rule will override the message mapper set to the rule.'); }
```
bigopon, line 240: can we do a direct check instead? Trying to think about this in an abstract way with a private property will cause confusion.
```suggestion
    if (__DEV__ && rule instanceof StateRule) { console.debug('Setting message to stateful rule will override the message mapper set to the rule.'); }
```

bigopon (commented):
Nice new feature @Sayan751 , I have a few comments wrt to the code only.

Sayan751:
Thanks @bigopon for the review. Addressed your comments differently.

packages/validation/src/rule-provider.ts
```diff
@@ -628,11 +637,16 @@ export class ValidationMessageProvider implements IValidationMessageProvider {
...
-    const parsedMessage = this.registeredMessages.get(rule);
-    if (parsedMessage !== void 0) { return parsedMessage; }
+    const messageKey = rule.messageKey;
+    const lookup = this.registeredMessages.get(rule);
+    if (lookup != null) {
+      const parsedMessage = lookup.get(explicitMessageKey) ?? lookup.get(messageKey);
+      if (parsedMessage !== void 0) { return parsedMessage; }
+    }
+
+    if (isStatefulRule(rule)) return this.setMessage(rule, rule.getStateMessage());
```
bigopon, line 647: is still reads very weird when we have private API (`isStatefulRule`) combined with a public one `getStateMessage()`. Also, I don't think stateful express the meaning of this multi state validate rule. What about we explore it from a different name, based on the perspective that, let's say, currently we have 2 states: false/true

```ts
export interface IValidationRule<TValue = any, TObject extends IValidateable = IValidateable> {
  ...
  isBinary?: boolean;
  getMessage?(): string;

  isNonBinary?: boolean;
  getMessage?(): string;

  isMultiResults?: boolean;
  
  isMulti?: boolean;
}
```

and we can have the following change:
```diff
- export class StateRule {
+ export class MultiStateRule {
```

packages/validation/src/rules.ts
```diff
@@ -247,6 +246,45 @@ export class EqualsRule extends BaseValidationRule implements IEqualsRule {
...
+    super(void 0);
+    this._state = validState;
+  }
+
+  public execute(value: unknown, object?: TObject): boolean | Promise<boolean> {
+    return onResolve(
+      this.stateFunction(value as TValue, object),
+      state => {
+        this._state = state as TState;
+        this.messageKey = this.messageMapper(state as TState);
```
bigopon, line 273: is the intention of this line about providing messages for all possible states, including valid one?

packages/validation/src/rules.ts
```diff
@@ -247,6 +246,45 @@ export class EqualsRule extends BaseValidationRule implements IEqualsRule {
...
+const statefulFlag: unique symbol = Symbol.for('au:validation:rule:isStateful');
+export function isStatefulRule(rule: IValidationRule): rule is IValidationRule & { getStateMessage(): string } {
+  return (rule as IValidateable & {[statefulFlag]?: boolean})[statefulFlag]  // this for short-circuiting the most common case
+    ?? 'getStateMessage' in rule;                                            // this is just for better DX
+}
+export class StateRule<TValue = any, TObject extends IValidateable = IValidateable, TState = unknown> extends BaseValidationRule<TValue, TObject> {
+  public static readonly $TYPE: string = 'StateRule';
+
+  private readonly [statefulFlag]: boolean = true;
+  public _state: TState;
```
bigopon, line 258: private 🐱

packages/validation/src/rules.ts
```diff
@@ -247,6 +246,45 @@ export class EqualsRule extends BaseValidationRule implements IEqualsRule {
...
+        this._state = state as TState;
+        this.messageKey = this.messageMapper(state as TState);
+        return state === this.validState;
+      });
+  }
+
+  public accept(_visitor: IValidationVisitor) {
+    if (__DEV__) {
+      // eslint-disable-next-line no-console
+      console.warn('Serialization of a StateRule is not supported.');
```
bigopon, line 281: if `stateFunction` is a string, then the serialization can be supported. Which means if we can allow it to be
```diff
-    private readonly stateFunction: ...,
+    private readonly stateFunction: string | ...,
```
then this can be serialized. The only problem is the mapper for invalid states to their corresponding messages. It can be a record too, since state "reducer", i assume, often results in primitive values.

packages/validation/src/rule-provider.ts
```diff
@@ -646,9 +660,14 @@ export class ValidationMessageProvider implements IValidationMessageProvider {
...
-  public setMessage(rule: IValidationRule, message: string): Interpolation | PrimitiveLiteralExpression {
+  public setMessage(rule: IValidationRule, message: string, messageKey?: symbol): Interpolation | PrimitiveLiteralExpression {
     const parsedMessage = this.parseMessage(message);
-    this.registeredMessages.set(rule, parsedMessage);
+    const rm = this.registeredMessages;
+    let messageLookup = rm.get(rule);
+    if (messageLookup === void 0) {
+      rm.set(rule, messageLookup = new Map());
+    }
+    messageLookup.set(messageKey ?? rule.messageKey, parsedMessage);
```
bigopon, line 670: Looks ok, but a record would make the hoop so much simpler then a map on this 2nd layer, since it's only string | symbol.

packages/validation/src/rule-provider.ts
```diff
@@ -646,9 +660,14 @@ export class ValidationMessageProvider implements IValidationMessageProvider {
     return this.setMessage(rule, message);
   }
 
-  public setMessage(rule: IValidationRule, message: string): Interpolation | PrimitiveLiteralExpression {
+  public setMessage(rule: IValidationRule, message: string, messageKey?: symbol): Interpolation | PrimitiveLiteralExpression {
```
bigopon, line 663: Suggestion for discussion:
Instead of having to extend this, we can make `messageKey` special on `StatefulRule`, since it's already a special rule, (changes are marked with number in comments)
```ts
export class BaseValidationRule {
  // 1.
  // we just need this to force the presence of messageKey on all rules
  abstract messageKey: string;
}

export class StatefulRule extends BaseValidationRule {
  _messageKey: string = '';
  // 2.
  // provide a signal for overriding key
  _overrideMessageKey?: string;
  
  set messageKey() {
    return this._overrideMessageKey ?? this._messageKey;
  }

  get messageKey(v: string) {
    this._overrideMessageKey = v;
  }

  public execute(value: unknown, object?: TObject): boolean | Promise<boolean> {
    return onResolve(
      this.stateFunction(value as TValue, object),
      state => {
        this._state = state as TState;
        // 3.
        // set the internal key instead
        this._messageKey = this.messageMapper(state as TState);
        return state === this.validState;
      }
    );
  }
}
```

With the above changes, we don't need to mess with layered of rules, messages, and keys

bigopon (commented):
Nice changes @Sayan751 , though the combination of APIs is still quite rubbing, gave it some more thoughts, let me know how you see the suggestions 😃

packages/validation/src/rules.ts
Sayan751: Thanks for noticing this. Chnaged now.

packages/validation/src/rules.ts
Sayan751: Good catch. The message is always for invalid states. Refactored it to lazily resolve the message.

packages/validation/src/rule-provider.ts
Sayan751: I agree with you. Refactored such that the message provider is agnostic of stateful rules. It just focuses on the fact whether the rule provides a message or not.

packages/validation/src/rules.ts
Sayan751: I think this rule is comparable to the `.satisfies(LAMBDA)`, which is also non-serializable. I think for now, it is not that critical that this rule is not serializable. We can always come back to this if we receive enough feedback to make this serializable.

Also, in the worst case one can actually have their own implementation of stateful rule, because the message provider will always respect the `getMessage` implemented by a rule. In that case, one can also make it serializable if needed.

For now, I am marking this discussion as resolved.

packages/validation/src/rule-provider.ts
Sayan751: Not sure what you meant. The 2nd level map is necessary as now, unlike before the messages are not solely tied to the rule instance, rather also to the (stateful) message/messageKey in the rule.

packages/validation/src/rule-provider.ts
Sayan751: As already pointed [here](https://github.com/aurelia/aurelia/pull/1985#discussion_r1617707743), the hierarchical message lookup is necessary. 

I think the usage of an explicit message key symbol is troubling you. Well, as I see it, with the absence of that will make the message provider dependent on the "stateful"ness of a rule. After the last changes, it makes no/less special provisions depending on the type of the rule.

bigopon: > with the absence of that will make the message provider dependent on the "stateful"ness of a rule

This is not necessarily true, but the opposite, I think is true: the presence of it makes the message provider aware of the statefulness of a rule. Why is there an explicit message key when there's already a public `messageKey` property.

packages/validation/src/rule-provider.ts
bigopon: I meant
```ts
rm.set(rule, messageLookup = {})
```

bigopon (approved):
It feels like we can do better 😃 but I think the current one is also good, nice work 👍

Issue #1948, opened 2024-04-18
CSS modules support for Vite


3cp:
# 💬 RFC

Currently Aurelia 2 setup with CSS modules does not work with Vite bundler.

## 🔦 Context

Aurelia 2 has conventions that when using CSS modules, the css file name should still match js/ts/html file names. For example `my-comp.css(scss/less)` to match `my-comp.html` and `my-comp.ts(js)`.

However Vite has [different conventions](https://vitejs.dev/guide/features.html#css-modules), it only works with file name `my-comp.module.css(scss/less)`, and it decided not to make it flexible with alternative setup. vitejs/vite#2981 

For us to support CSS module in Vite,
1. one option is that we can utilize our vite-plugin's "load" and "transform" hooks, similar to what we did with our html file.
2. another option is to update our conventions to support `.module.css` naming conventions.

Both options seem approachable. Option 2 might be slightly easier because it's in our control.

3cp:
Existing app should still work, we would not remove the existing global config for css modules.

Options 2 adds file based css-module support without explicit configuration. The question is do we want to support that vite conventions in other bundlers' setup. If it's a vite only conventions, it's bit confusing for other setups. It might be tricky to support it in other bundlers.

bigopon:
> The question is do we want to support that vite conventions in other bundlers' setup. If it's a vite only conventions, it's bit confusing for other setups. It might be tricky to support it in other bundlers.

What if css module can be configured from our html?
```html
<import from="my.css" type="css-module">
```
With this as the base, we can configure the webpack loader/vite plugin to do necessary work to turn the css into a css module file/virtual file.

With this, we will have the ability to have standard css together with css module
```html
<import from="my.css">
<import from="my.modules.css" type="css-module">
```

3cp:
`<import from="my.css" type="css-module">` does not work for vite. Actually vite only supports its own naming convention, nothing else.

We might just leave vite as special.

I think your proposal needs option 1 instead, the vite plugin.

bigopon:
> We might just leave vite as special.

You meant if an apps use Vite and wants to have css module it needs to have `.module.css` as the extension for its css file?

3cp:
> > We might just leave vite as special.
> 
> You meant if an apps use Vite and wants to have css module it needs to have `.module.css` as the extension for its css file?

Yes.

bigopon:
[...]

PR #1988, opened 2024-05-28
feat(plugin-conventions): support foo.module.css CSS module convention

3cp:
Related to #1948

This enables foo.module.css file to be processed with CSS module.


bigopon (approved):
nice work @3cp

bigopon:
A new dev version has been released for this change.

PR #1990, opened 2024-05-29
fix(router): fix redirect issue

jwx:
# Pull Request

## 📖 Description

This PR fixes an issue with redirect from `canLoad` when there are multiple `canLoad` hooks. It also adds `guardedLoad` to the default sync states to behave more like v1 router out of the box.


bigopon:
We cannot proceed this PR as it is right now, if we still want the changes here it's better to create a new PR.

PR #1992, opened 2024-05-30
fix(metadata): own metadata

Sayan751:
# Pull Request

## 📖 Description

This PR brings two changes:
- It ensures that when defining metadata for a type, it is defined on the type itself; that is, the type owns the metadata instead of inheriting metadata.
- It refactors such that the bindables of the parent resources can be inherited by the child resources.
- Removes the decorator rewriting infra from tooling.

 
## ⏭ Next Steps

- ~~Duplicate the vite e2e tests without convention.~~
- ~~If that works, remove the decorator rewrite infra from tooling.~~


packages/__tests__/src/1-kernel/metadata.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+      static {
+        Metadata.define('child', this, 'foo');
+      }
+    }
+
+    assert.strictEqual(Metadata.get('foo', Parent), 'parent', 'parent');
+    assert.strictEqual(Metadata.get('foo', Child), 'child', 'child');
+  });
+
+  it('metadata is defined for every type - without static block - parent then child', function () {
```
bigopon, line 37:
```suggestion
  it('defines metadata for every type - without static block - parent then child', function () {
```
similar to other tests

packages-tooling/plugin-conventions/src/preprocess-resource.ts
```diff
@@ -342,9 +344,9 @@ function processDecorators(decorators: ReplaceableDecorator[] | undefined, class
   return { statement, effectivePos: classPos };
 }
 
-function processBindables(bindables: BindableDecorator[] | undefined, classPos: number, m: ReturnType<typeof modifyCode>, viewDef: string): { statement: string; effectivePos: number } {
-  let statement = '';
-  if (!bindables) return { statement, effectivePos: classPos };
+function processBindables(bindables: BindableDecorator[] | undefined, classPos: number, m: ReturnType<typeof modifyCode>, viewDef: string, superDefinitionIdentifier: string): { statement: string; effectivePos: number } {
+  let statement = `, bindables: { ...(${superDefinitionIdentifier}.bindables ?? {}), ...${viewDef}.bindables`;
```
bigopon, line 348: shouldn't this work be moved to `Bindables.getAll()`? Complicating generated code I think is unnecessary for this, it also requires us to change the types of `bindables` in resources unnecessarily (I like object more than array form though)

bigopon (commented):
I think bindable inheritance can be left to run time, it won't be a costly thing to do, I believe.

packages-tooling/plugin-conventions/src/preprocess-resource.ts
Sayan751: I think the `Bindables.getAll()` should not be changed. We needed to do this in the convention-plugin because we previously rewrote bindables. Which meant that the subclass got no information regarding the bindables from the super class. In other words, those `@bindable`s aren't inherited, although those properties are. 

Hence, I see the current plugin code only as a temporary solution, and should be removed as soon as the support around decorators are stabilized in vite and esbuild. 

This is what I noted under the next steps in the PR description.

bigopon: We can do this with convention, but we cant expect normal code to be written like this. Even when temporary I think we should avoid that. Or we can just leave bindable alone while waiting for decorator instead.

Sayan751: Done.

packages-tooling/plugin-conventions/src/preprocess-resource.ts
bigopon, line 1: it probably should have been done this record way, at least now we are back with simple generated code

bigopon (approved):
Nice work @Sayan751 , I only have a few non blocking comments 👍

packages/runtime-html/src/bindable.ts
```diff
@@ -170,6 +170,14 @@ export const Bindable = objectFreeze({
     }
     return defs;
   },
+  /** @internal */
+  _addToMetadata(bindable: BindableDefinition, Type: Constructable) {
```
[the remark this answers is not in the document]
Sayan751: I will keep this internal for now, as I don't see it to be used in directly in dev land. I have renamed the method though.

packages/__tests__/src/3-runtime-html/custom-attributes.spec.ts
```diff
@@ -1043,4 +1045,75 @@ describe('3-runtime-html/custom-attributes.spec.ts', function () {
       assert.throws(() => CustomAttribute.closest(appHost, class NotAttr {}));
     });
   });
+
+  describe('bindable inheritance', function () {
```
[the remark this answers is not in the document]
Sayan751: Done.

bigopon (approved):

PR #1993, opened 2024-06-04
test(e2e): vite without convention

Sayan751:
# Pull Request

## 📖 Description

This PR adds an e2e test project for vite-based apps without using convention. This ensures that vite-based apps works without convention, which in turns means that the rewrite of decorators via the convention plugin is no-longer required.


packages-tooling/vite-plugin/package.json
```diff
@@ -51,7 +51,7 @@
     "@aurelia/plugin-conventions": "2.0.0-beta.18",
     "@aurelia/runtime": "2.0.0-beta.18",
     "@rollup/pluginutils": "5.0.2",
-    "vite": "^5.0.12",
+    "vite": "5.3.0-beta.0",
```
bigopon, line 54: we probably should keep the `^`

bigopon (approved):
Nice work @Sayan751 👍

packages-tooling/vite-plugin/package.json
Sayan751: Done

PR #1997, opened 2024-06-26
refactor(router): navigation coordinator refactor

jwx:
# Pull Request

## 📖 Description

This PR moves responsibilities and recursive code from routing scope to iterative code in navigation coordinator. It also fixes some race condition issues, primarily with multiple async routing hooks.

BREAKING CHANGE: Refresh and browser back and forward now behave the same way deep links does; if a viewport with a default component/route has been emptied/cleared the mentioned browser actions will now load the default (instead of keeping the viewport empty).

## 📑 Test Plan


packages/__tests__/src/router/router.lifecycle-hooks.spec.ts
```diff
@@ -225,6 +225,9 @@ describe('router/router.lifecycle-hooks.spec.ts', function () {
 
         assert.strictEqual(calledHooks.join('|'), expected.join('|'), `calledHooks`);
 
+        // Make sure there's time to store the result before teardown. Should only be needed in testing.
+        await new Promise((resolve) => setTimeout(resolve, 0));
```
bigopon, line 229: what result are we storing?

packages/router/src/endpoints/viewport-content.ts
```diff
@@ -347,7 +353,16 @@ export class ViewportContent extends EndpointContent {
         }
         // TODO: If requested, pass previous value into hook
         const result = instance.canUnload?.(this.instruction, navigation);
-        return result instanceof Promise ? result.then(Boolean) : Boolean(result);
+        if (result instanceof Promise) {
+          void result.then(canUnload => {
+            if (typeof canUnload !== 'boolean') {
+              throw new Error(`Method 'canUnload' in component "${this.instruction.component.name}" needs to return true or false or a Promise resolving to true or false.`);
+            }
```
bigopon, line 360: enforcing boolean type is a weirdly strict API in JS. Is there a reason why we cannot just do coerce the values to boolean ourselves?

packages/router/src/instructions/instruction-endpoint.ts
```diff
@@ -84,8 +84,8 @@ export class InstructionEndpoint {
     if (this.instance !== null && other.instance !== null) {
       return this.instance === other.instance;
     }
-    return (this.endpointType === null ||
-      other.endpointType === null ||
+    return (this.endpointType !== null &&
+      other.endpointType !== null &&
```
bigopon, line 88: if `null` and `undefined` are of the same behavior, we can just do `!= null`, it'll be less work having to watchout whether null and undefined will be treated differently.

packages/router/src/instructions/routing-instruction.ts
```diff
@@ -372,6 +395,18 @@ export class RoutingInstruction {
     return (this.nextScopeInstructions?.length ?? 0) > 0;
   }
 
+  /**
+   * Get the dynasty of the routing instruction. The dynasty is the instruction
+   * itself and all its descendants (next scope instructions iteratively).
+   */
+  public get dynasty(): RoutingInstruction[] {
```
bigopon, line 402: maybe hierarchy instead? Generally any terms that is plain and obvious is better.

packages/router/src/navigation-coordinator.ts
```diff
@@ -174,6 +185,348 @@ export class NavigationCoordinator {
...
+      if (scope == null) {
+        throw new Error('No scope for instruction');
+      }
+      // eslint-disable-next-line no-await-in-loop
+      changedEndpoints.push(...await this.processInstructionsForScope(scope));
+    }
+    return changedEndpoints;
+  }
+
+  public async processInstructionsForScope(scope: RoutingScope): Promise<IEndpoint[]> {
```
bigopon, line 277: I'll have to read this quite a lot to get the change, though I think if the router can be simplified and code is extracted from there, it'll be better.

packages/router/src/router.ts
```diff
@@ -443,28 +454,40 @@ export class Router implements IRouter {
     }
 
     if (typeof transformedInstruction === 'string') {
-      transformedInstruction = transformedInstruction === '' // || transformedInstruction === '-'
-        ? [new RoutingInstruction('')] // Make sure empty route is also processed
-        : RoutingInstruction.parse(this, transformedInstruction);
+      if (transformedInstruction === '') { // || transformedInstruction === '-'
```
bigopon, line 457: what's the comment for? it seems similar to the line 460 below.

packages/router/src/router.ts
```diff
@@ -443,28 +454,40 @@ export class Router implements IRouter {
...
-    // for (const instr of instructions) {
-    //   instr.topInstruction = true;
-    // }
+
+    coordinator.appendInstructions(transformedInstruction);
+
+    // If router options defaults to navigations being full state navigation (containing the
+    // complete set of routing instructions rather than just the ones that change), ensure
+    // that there's an instruction to clear all non-specified viewports in all the scopes of
+    // the top instructions
```
bigopon, line 478: I think the comment is good but can be made more obvious with maybe some inlined examples of the full state navigation and partial navigation. Since there' are a lot of scenarios to support in the router, any amount of clarity is good.

packages/router/src/router.ts
```diff
@@ -443,28 +454,40 @@ export class Router implements IRouter {
...
+    if (!options.additiveInstructionDefault) {
```
bigopon, line 479: is there a better name than this? It reads quite weird and confusing to me. What about `partialChange`?

packages/router/src/utilities/utils.ts
```diff
@@ -19,6 +19,21 @@ export function arrayRemove<T>(arr: T[], func: (value: T, index?: number, obj?:
...
+ */
+export function arrayAddUnique<T>(arr: T[], values: T | T[]): T[] {
+  if (!Array.isArray(values)) {
+    values = [values];
+  }
+  for (const value of values) {
+    if (!arr.includes(value)) {
+      arr.push(value);
+    }
+  }
```
bigopon, line 33: We can have a shorter version for this:
```suggestion
  arr.push(...values.filter(v => !arr.includes(v));
```

bigopon (commented):
Looking good @jwx , I have only added a few preliminary comments so far, will continue to review.

packages/__tests__/src/router/router.lifecycle-hooks.spec.ts
jwx: Lifecycle calls to the test components are stored in the `calledHooks` to be asserted after the tests are done.

packages/router/src/endpoints/viewport-content.ts
jwx: We can. This is just a dx thing so forgetting a return value won't be that bad.

packages/router/src/instructions/instruction-endpoint.ts
jwx: The return type for this getter doesn't allow undefined, but I've got no problem changing the check. Should I?

packages/router/src/instructions/routing-instruction.ts
jwx: I'm not opposed to changing it, I do however think hierarchy is more about showing the position within the tree and could also include the parents.

packages/router/src/router.ts
jwx: Missed it in clean up. Will fix.

packages/router/src/router.ts
jwx: Will look into it.

packages/router/src/router.ts
jwx: Yeah, it's a bit convoluted. Will look into a new name.

packages/router/src/utilities/utils.ts
jwx: Haha, yes, indeed. Think it was late one night and I asked Copilot to create this func and then just didn't review it enough. Will fix.

packages/router/src/router.ts
jwx: I'm leaning towards changing it to `completeStateNavigations` (with a switched value). Any thoughts?

packages/router/src/utilities/utils.ts
bigopon: actually there's a difference in behavior between my suggestion and the original code where the original also does a unique check for incoming values.

packages/router/src/endpoints/viewport-content.ts
bigopon: we are throwing here, so it cant be argued that this is a DX thing I think. If we are not throwing then there's no reason for us to discriminate between boolean vs non boolean.

packages/router/src/router.ts
```diff
@@ -582,11 +607,23 @@ export class Router implements IRouter {
...
-      // Can't return current navigation promise since it can lead to deadlock in load
-      return Promise.resolve();
+    const append = options.append ?? false;
+    if (append !== false) {
+      if (append instanceof NavigationCoordinator) {
+        if (!append.closed) {
+          instructions = RoutingInstruction.from(this, instructions);
+          this.appendInstructions(instructions as RoutingInstruction[], scope, append);
+          // Can't return current navigation promise since it can lead to deadlock in load
+          return Promise.resolve();
```
bigopon, line 617: is it ok to return a loose promise like this? When it's not connected with any of the router infra, do we have race condition?

bigopon (approved):
changes look ok to me, I'm not able to go through all so I'll do more review later

bigopon:
Also, deepscan picked up a few issues with the code, can you fix them before merge @jwx?

packages/router/src/router.ts
bigopon: maybe `fullNavigation`?

packages/router/src/instructions/routing-instruction.ts
bigopon: at least `hierarchy` describes a tree while `dynasty` doesn't do anything. I've never seen an API using it, or maybe I've not seen enough, feels quite weird.

packages/router/src/instructions/instruction-endpoint.ts
bigopon: if it's null only then it's all good, thanks.

packages/router/src/endpoints/viewport-content.ts
jwx: I didn't have much time when I answered these questions the first time so my answers are a bit too short. Sorry about that. 

The only return values that actually make sense from `canUnload` are true or false (or a promise to it). During a navigation, if _any_ `canUnload` returns `false` the navigation is silently aborted. If a dev forgets to add a return value in some code path or there's just an empty `canUnload` method somewhere the `undefined` will be coerced into `false`. And then finding the source of that unintended silently aborted navigation can be a bit tricky (speaking from experience). When trying to improve the situation a bit, the two easy alternatives I came up with was to either consider `undefined` as `true` or throw if not a boolean. The later felt less weird to me at the time. Any other suggestions are welcome, of course.

packages/router/src/instructions/routing-instruction.ts
jwx: The result isn't a tree but a list (of the dynasty). I've seen `dynasty` used, but it might've been some time ago. As far as I understand the word, it describes the result quite well. But again, other suggestions welcome. (Also, this doesn't have to be a part of public api.)

jwx: (Out of curiosity, I asked my new colleague, ChatGPT, which had this to say:
In English, the term "dynasty" typically refers to a sequence of rulers from the same family or lineage, extending over a period of time. While it generally implies a succession of individuals, it often includes the founder or current head of the family as part of the broader context of the dynasty.

For example, when discussing the "Ming Dynasty" in China, it includes the initial emperor who founded the dynasty as well as all subsequent rulers from the same family line.

Therefore, in the context of describing a tree structure in code, using "dynasty" would be understood to include the current node (or "ruler") and all its descendants. This makes "dynasty" an appropriate and evocative term for a method that returns the current node and its descendants, aligning with both the technical and typical English usage of the term.)

packages/router/src/router.ts
jwx: I could go with `fullNavigations` since it's a setting for navigations.

packages/router/src/utilities/utils.ts
jwx: Yeah, I realized this as well out of the blue yesterday. Will be restored in the next push.

packages/router/src/router.ts
jwx: Yeah, should be okay. An `append` navigation is an internal thing that appends the navigation's instructions onto an already existing navigation (as in a previous call to `load`) and the appended instructions become a part of the existing navigation's promise resolve.

jwx:
> Also, deepscan picked up a few issues with the code, can you fix them before merge @jwx?

Fixed and pushed, @bigopon. Thanks for reviewing!

packages/router/src/endpoints/viewport-content.ts
bigopon: > the undefined will be coerced into false

I think this is the source of our issue, coercing `undefined` into `false` is not the "right" mental model I believe, and it's also not consistent with how we are treating these `can....` APIs in the rest of the framework. The general design is we assume it can do unless `false` was explicitly returned. Only `false` is false, and the rest is true should be the coercion.

packages/router/src/instructions/routing-instruction.ts
bigopon: right, ok. I guess it's my issue. But this is just my general feeling towards to vocab of the router code, why use so many unfamiliar terms, are they necessary, not only for situation like this but also other internal & public APIs, the names are either long or ambiguous or lack clarity.

packages/router/src/endpoints/viewport.ts
```diff
@@ -494,13 +494,13 @@ export class Viewport extends Endpoint {
...
         if (this.isActiveNavigation(coordinator)) {
-          return this.canLoad(coordinator, step) as boolean | LoadInstruction | LoadInstruction[];
+          return this.canLoad(coordinator, step);
         }
       },
 
       (step: Step) => {
         if (this.isActiveNavigation(coordinator)) {
-          let canLoadResult = step.previousValue as boolean | LoadInstruction | LoadInstruction[];
+          let canLoadResult = (step.previousValue ?? true) as boolean | LoadInstruction | LoadInstruction[];
```
bigopon, line 503: I think we can skip the cast but it can be for the next PR

packages/router/src/endpoints/viewport.ts
```diff
@@ -637,7 +637,7 @@ export class Viewport extends Endpoint {
         return this.getContent().connectedScope.canUnload(coordinator, innerStep);
       },
       (innerStep: Step<boolean>) => {
-        if (!(innerStep.previousValue as boolean)) { // canUnloadChildren
+        if ((innerStep.previousValue ?? true) === false) { // canUnloadChildren
```
bigopon, line 640: do we need this comment?

bigopon (approved):
Nice work @jwx 👍

packages/router/src/router.ts
bigopon: I think we should clarify better around the return code then. When something loose like this, we should just return void to signal it's connected somewhere else.

bigopon:
@jwx does this PR close any "issues"? either in GH or Discord or Discourse, if yes, can you please help update the PR description?

PR #2002, opened 2024-07-04
refactor(router-lite): can* hook semantics

Sayan751:
# Pull Request

## 📖 Description

This PR changes the semantics of `canLoad` and `canUnload` hooks.

If `canLoad` returns no value, `null`, or en explicit `undfined`, the component will be loaded. Returning a routing instruction to perform redirect instead is still supported. To prevent loading altogether, the return of an explicit `boolean` `false` is required.

Similarly, if the `canUnload` returns no value, `null`, or en explicit `undfined`, the component will be unloaded. To prevent unload, the return of an explicit `boolean` `false` is required.


bigopon (approved):
Nice work @Sayan751 👍

Issue #732, opened 2019-10-31
Support increment operators in view


davismj:
# 🙋 Feature Request

Support the following operators in the view:

- `+=`
- `-=`
- `++`
- `--`
- `*=` 
- `/=`

## 🤔 Expected Behavior

Particularly, this is for event handlers.

```html
<button click.delegate="value++">Increment</button>
```

## 😯 Current Behavior

Throws an error.

3cp:
add `*=` and `/=`.

Vheissu:
@bigopon @3cp Do we still want to add support for these?

bigopon:
We may want to support it. I'll work out the AST for this before we initiate the parser work for @fkleuver

bigopon:
resolved in #2004

PR #2004, opened 2024-07-10
feat(ast): Increment/decrement assign

fkleuver:
# Pull Request

## 📖 Description

Implementation of unary increment/decrement and assignment operators:
`++`
`--`
`+=`
`-=`
`/=`
`*=`

Example usage:
```html
<button @click="value++">Increment</button>

<button @click="value += 2">Increment by 2</button>
```

Using these operators in other bindings will result in an error to prevent infinite update-eval loops:
```html
<!-- throws AUR0113 -->

<div>${value++}</div>

<my-component :prop="value++"> 
```

### 🎫 Issues

https://github.com/aurelia/aurelia/issues/732

## 👩‍💻 Reviewer Notes

I put the increment and decrement operators in `UnaryExpression` even though the spec describes them as `UpdateExpression`:
https://262.ecma-international.org/#sec-update-expressions

This doesn't cause any problems for precedence or parsing order since they're consecutive anyway.

## 📑 Test Plan

Came across a small unrelated bug in the parser that would cause parenthesized expressions to become assignable even if the body isn't.

The test case is addressed by the parenthesized `SimplyUnaryList` 
image

In this loop:
image

Other than that, added the new variations to existing AST lists to utilise the existing test loops


bigopon:
Nice work @fkleuver 👍

PR #2006, opened 2024-07-11
fix(vite-plugin): fix vitest Windows issue with html file

3cp:
Solves some vitest+win32 issue identified in aurelia/new#108


bigopon (approved):
ahh, nice fix 👍

Issue #1542, opened 2022-09-05
Platform taskQueue multiple instances all use the same delay value


alexgustafson:
# 🐛 Bug Report

When using multiple instances of PLATFORM.taskQueue.queueTask as a replacement for setTimeout. It appears that the longest delay value is used for all instances.

Repro link https://github.com/alexgustafson/multiple-queuetask-timing

## 🤔 Expected Behavior

I would expect that each instance of queueTask would repeat at its own set delay time.

## 😯 Current Behavior

All instances use the same delay time. Its seems that the longest ( largest ) value wins.
There is no error.

## 💁 Possible Solution

I am using native setTimeout instead.

## 💻 Code Sample
Here is a link to an example Aurelia v2 project. There are three components, typing text into the view. Each should be typing at a different speed. 

https://github.com/alexgustafson/multiple-queuetask-timing

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Aurelia          |   v2 2.alpha 39 |
| Language         | typescript |
| Browser          |
| Bundler          |
| Operating System |
| NPM/Node/Yarn    |

superspartak:
I can confirm that as of March 2024, it still persists

bigopon:
The `PLATFORM.taskQueue` is uses `setTimeout` under the hood so it sounds weird for what we are seeing. Are you able to reproduce the issue with just setTimeout instead of the taskQueue @alexgustafson @superspartak?

superspartak:
Switching to setTimeout seems to fix the issue. as @alexgustafson mentioned, initially, all events run on their expected time if the delay for the earlier tasks are shorter than the later once. In the next iterations, whatever delay is the maximum, it seems to be the same for all tasks.

PR #2007, opened 2024-07-12
fix(task-queue): fix timing issue

fkleuver:
# Pull Request

## 📖 Description

Fixes an issue in the task queue when using multiple persistent delayed tasks with different timings.

### 🎫 Issues

- close https://github.com/aurelia/aurelia/issues/1542
- resolve https://discourse.aurelia.io/t/platform-taskqueue-does-not-run-concurrent-tasks/5379
- resolve https://discord.com/channels/448698263508615178/448699089513611266

## 👩‍💻 Reviewer Notes

There was a small oversight in the queue processing loop where it assumed the smallest delay task to be in the front of the queue. It now checks all tasks in the delayed task list.

## 📑 Test Plan

A single test covers this bug properly. Without the fix, it fails with `counter10` and `counter20` being equal. Now it's doubled (though I added some leeway in there to account for flakiness)

### DIFF
--- a/packages/__tests__/src/2-runtime/scheduler.spec.ts
+++ b/packages/__tests__/src/2-runtime/scheduler.spec.ts
@@ -1526,7 +1526,7 @@ describe('2-runtime/scheduler.spec.ts', function () {
     });
   });
 
-  it('multiple persistent delayed tasks with different delays retain correct timing', async function () {
+  it('2 persistent delayed tasks with different delays retain correct timing', async function () {
     let counter10 = 0;
     let counter20 = 0;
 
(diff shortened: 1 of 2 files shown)


packages/__tests__/src/2-runtime/scheduler.spec.ts
```diff
@@ -1524,4 +1525,60 @@ describe('2-runtime/scheduler.spec.ts', function () {
...
+      assert.ok(true, `counter10 - 1`);
+    } else if (counter10 - 2 === counter20 * 2) {
+      // this might happen rarely, we don't want to fail the whole test suite for this
+      assert.ok(true, `counter10 - 2`);
+    } else if (counter10 + 1 === counter20 * 2) {
+      // a little bit of drift the other direction can also happen
+      assert.ok(true, `counter10 + 1`);
+    } else {
+      assert.fail(`counter10: ${counter10}, counter20: ${counter20}`);
+    }
```
bigopon, line 1576: Should/shouldn't we make it more "tolerant"?

```suggestion
    if (Math.abs(counter10  - counter20 * 2) > 2) {
      assert.fail('too far apart')
    }
```

packages/__tests__/src/2-runtime/scheduler.spec.ts
```diff
@@ -1524,4 +1525,60 @@ describe('2-runtime/scheduler.spec.ts', function () {
...
+
+    task10.cancel();
+    task20.cancel();
+    task.cancel();
+
+    assert.areTaskQueuesEmpty();
+  });
```
bigopon, line 1583: can we have another test with 3 tasks together? 10 + 20 + 30

bigopon (commented):
Nice work @fkleuver , only a few minor comments

packages/__tests__/src/2-runtime/scheduler.spec.ts
```diff
@@ -1524,4 +1525,106 @@ describe('2-runtime/scheduler.spec.ts', function () {
...
+
+    const task = platform.taskQueue.queueTask(
+      noop,
+      {
+        persistent: true,
+        delay: 200,
+      }
+    );
+
+    await task.result;
```
bigopon, line 1617: if there's a way to implement fake timer like jest so we can do `global.advanceTimer(200)` instead of having to wait for 200 ms, it'd be much nicer for these tests. Maybe an idea for improvements related to our tests.

bigopon (approved):
Nice work @fkleuver

packages/__tests__/src/2-runtime/scheduler.spec.ts
fkleuver: Yeah that's be great

PR #2008, opened 2024-07-12
refactor(task-queue): remove 'reusable' param

fkleuver:
# Pull Request

## 📖 Description

Removes the `reusable` task queue option. This was originally intended as a performance optimization for animations but the mechanism for animations is going to be optimized separately.


bigopon (approved):
👍 nice cleanup

PR #2011, opened 2024-07-14
chore: updated deps to remove vulnerable deps

Sayan751:
# Pull Request


Sayan751:
@bigopon I am merging this, since there is no change in the logic, just deps update and associated refactoring. If you see something funny, we can revert or push changes in follow-up PRs.

PR #2012, opened 2024-07-20
fix(router-lite): element injection for routed view-model

### LINKED ISSUE #2009: Host element Injection bug?
I noticed something, not sure if it's a bug or not.

When injecting a components/routes host into the view model (not sure if this is correct wording)

```
@inject(Element)

class MyComponent {
  constructor(host) {
    this.host = host
  }
}
```

if the host element differs depending on whether it's a component or a route.

For a route the host element is `au-view`

For a component the host element is the component.

e.g

my-route -> host is `<au-view>
my-component -> host is `<my-component>

Should it not also be `<my-route>` for a route.

Here is an example: https://stackblitz.com/edit/aurelia-2-host-injection-bug

Sayan751:
# Pull Request

## 📖 Description

This PR fixes the issue where the host for routed view-model in router-lite was not registered for injection.

### 🎫 Issues

Closes #2009.

### DIFF
--- a/packages/router-lite/src/route-context.ts
+++ b/packages/router-lite/src/route-context.ts
@@ -23,6 +23,7 @@ import {
   isCustomElementController,
   isCustomElementViewModel,
   PartialCustomElementDefinition,
+  registerHostNode,
 } from '@aurelia/runtime-html';
 
 import { ComponentAgent, IRouteViewModel } from './component-agent';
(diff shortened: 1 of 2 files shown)


packages/router-lite/src/route-context.ts
```diff
@@ -186,6 +190,7 @@ export class RouteContext {
     this._moduleLoader = parentContainer.get(IModuleLoader);
 
     const container = this.container = parentContainer.createChild();
+    this._platform = container.get(IPlatform);
```
bigopon, line 193:
```suggestion
    const platform = this._platform = container.get(IPlatform);
```
though it seems we don't use `_platform`?

bigopon (approved):
Nice work @Sayan751

packages/router-lite/src/route-context.ts
Sayan751: We have not used the platform in the ctor but in `_createComponentAgent`.
image

Sayan751:
@bigopon I have merged this one. In case you have any further comments, let me know.

PR #2013, opened 2024-07-24
fix(router): fix missing filter part for available endpoints

jwx:
# Pull Request

## 📖 Description

This PR adds a filter that was overlooked in the navigation coordinator refactor, fixing an issue mentioned in Discord that could occur when using default child routes.

## 📑 Test Plan

- Add test for the specific occasion


packages/__tests__/src/router/router.link-click-defaults.spec.ts
```diff
@@ -0,0 +1,170 @@
...
+    const test = tests[i];
+    it(`for "${test.load}"`, async function () {
+      const { platform, host, router, $teardown } = await $setup({}, [Nav, Parent, Child, GrandChild]);
+
+      await $load('/nav', router, platform);
+      await platform.domQueue.yield();
+
+      const links = host.getElementsByTagName('A') as unknown as HTMLElement[];
+      const link = links[i];
+      link.click();
```
bigopon, line 152: these dom targeting will be confusing, and prone to breaking during refactoring. Is there any better way to express this?

packages/__tests__/src/router/router.link-click-defaults.spec.ts
```diff
@@ -0,0 +1,170 @@
...
+    ];
+  });
+
+  const tests = [
+    { load: '/parent/child', result: '!parent!!child!!grandchild!', },
+  ];
+
+  for (let i = 0; i < tests.length; i++) {
+    const test = tests[i];
+    it(`for "${test.load}"`, async function () {
```
bigopon, line 144: is there a better name for this test?

bigopon (approved):
Looks good @jwx , just some small comments for clarity

packages/__tests__/src/router/router.link-click-defaults.spec.ts
jwx: Yes. Updated.

PR #2015, opened 2024-07-28
feat(validation): link two or more properties' validation

XmasCarola:
# Pull Request

## 📖 Description

Possibility to connect properties if they are related to each other, so that the validation of one property acts as a trigger for the other(s). This is useful when two variables are interconnected, for example start and end date that are supposed to function as comparison values. Providing multiple checks with subsequent highlights can point out all the values that need to be modified to properly fill out a form.

### DIFF
--- a/docs/user-docs/aurelia-packages/validation/defining-rules.md
+++ b/docs/user-docs/aurelia-packages/validation/defining-rules.md
@@ -91,6 +91,19 @@ validationRules
 
 With TypeScript support, intellisense is available for both the variants.
 
+## Linking a property's validation with others
+
+This functionality can be useful when a change in one property has to trigger additional validations because the values are intrinsically related.
+Linked properties must be provided as an array of object keys.
+
+```typescript
+validationRules
+  .on(person)
+  .ensure('name')
+    .linkProperties(['age', 'address']);
+```
+
+
 ## Associating validation rules with property
 
 After selecting a property with `.ensure` the next step is to associate rules. The rules can be built-in or custom. Irrespective of what kind of rule it is, at the low-level it is nothing but an instance of the rule class. For example, the "required" validation is implemented by the `RequiredRule` class. This will be more clear when you will define custom validation rules. However, let us take a look at the built-in rules first.
--- a/packages/validation/src/rule-provider.ts
+++ b/packages/validation/src/rule-provider.ts
@@ -123,6 +123,7 @@ export class PropertyRule<TObject extends IValidateable = IValidateable, TValue 
   private latestRule?: IValidationRule;
   /** @internal */
   public readonly l: IServiceLocator;
+  public linkedProperties: PropertyKey[] = [];
 
   public constructor(
     locator: IServiceLocator,
@@ -265,6 +266,15 @@ export class PropertyRule<TObject extends IValidateable = IValidateable, TValue 
     if (latestRule === void 0) {
       throw createMappedError(ErrorNames.rule_provider_no_rule_found);
     }
+  }
+
+  /**
+   * Links the PropertyRule to other key values.
+   * @param properties - Array of property keys to link.
+   */
+  public linkProperties(properties: PropertyKey[]) {
+    this.linkedProperties.push(...properties);
+    return this;
   }
   // #endregion
 
--- a/packages/validation/src/validator.ts
+++ b/packages/validation/src/validator.ts
@@ -57,6 +57,11 @@ export class StandardValidator implements IValidator {
     const scope = Scope.create({ [rootObjectSymbol]: object });
 
     if (propertyName !== void 0) {
+      const propertyRule: PropertyRule | undefined = rules.find((r) => r.property.name === propertyName);
+      if (propertyRule !== void 0 && propertyRule.linkedProperties.length > 0) {
+        const additionalRules = propertyRule.linkedProperties.map(lp => rules.find(r => r.property.name === lp)).filter(Boolean) as PropertyRule[];
+        return (await Promise.all([propertyRule?.validate(object, propertyTag, scope), ...additionalRules.map(async (rule) => rule.validate(object, propertyTag, scope))])).flat();
+      }
       return (await rules.find((r) => r.property.name === propertyName)?.validate(object, propertyTag, scope)) ?? [];
     }
 
--- a/packages/__tests__/src/validation/rule-provider.spec.ts
+++ b/packages/__tests__/src/validation/rule-provider.spec.ts
@@ -1179,6 +1179,27 @@ describe('validation/rule-provider.spec.ts', function () {
       validationRules.off();
     });
 
+    it('can be linked to other properties', async function () {
+      const { validationRules } = setup();
+      const obj: Person = new Person((void 0)!, (void 0)!, (void 0)!);
+      const properties: PropertyKey[] = ['age', 'address'];
+      const rule = validationRules
+        .on(obj)
+        .ensure('name')
+        .required()
+        .satisfies((value) => value === 'foobar')
+        .linkProperties(properties)
+        .ensure('age').min(18)
+        .ensure('address').required()
+        .rules[0];
+
+      assert.equal(rule.linkedProperties.length, 2);
+      assert.deepEqual(rule.linkedProperties[0], 'age');
+      assert.deepEqual(rule.linkedProperties[1], 'address');
+
+      validationRules.off();
+    });
+
     // The state rule is tested here as the individual unit tests are somewhat pointless.
     describe('StateRule', function () {
       it('stateful message - sync state function', async function () {
--- a/packages/__tests__/src/validation/validator.spec.ts
+++ b/packages/__tests__/src/validation/validator.spec.ts
@@ -235,6 +235,37 @@ describe('validation/validator.spec.ts', function () {
 
           validationRules.off();
         });
+
+        it('if given, validates linked properties', async function () {
+          const { sut, validationRules } = setup();
+          const obj: Person = new Person((void 0)!, (void 0)!, (void 0)!);
+          const linkedProperties = ['age', 'address.line1'];
+
+          const rules = validationRules
+            .on(defineRuleOnClass ? Person : obj)
+
+            .ensure(getProperty1() as any)
+            .required()
+            .linkProperties(linkedProperties)
+
+            .ensure(getProperty2() as any)
+            .required()
+
+            .ensure(getProperty3() as any)
+            .required()
+            .withMessage('Address is required.')
+
+            .rules;
+
+          const result = await sut.validate(new ValidateInstruction(obj, 'name'));
+          assert.equal(result.length, 3);
+
+          assertValidationResult(result[0], false, 'name', obj, RequiredRule, 'Name is required.');
+          assertValidationResult(result[1], false, 'age', obj, RequiredRule, 'Age is required.');
+          assertValidationResult(result[2], false, 'address.line1', obj, RequiredRule, 'Address is required.');
+
+          validationRules.off();
+        });
       }
 
       const properties3 = [


packages/validation/src/validator.ts
```diff
@@ -57,6 +57,11 @@ export class StandardValidator implements IValidator {
     const scope = Scope.create({ [rootObjectSymbol]: object });
 
     if (propertyName !== void 0) {
+      const propertyRule: PropertyRule | undefined = rules.find((r) => r.property.name === propertyName);
+      if (propertyRule !== void 0 && propertyRule.linkedProperties.length > 0) {
+        const additionalRules = propertyRule.linkedProperties.map(lp => rules.find(r => r.property.name === lp)).filter(Boolean) as PropertyRule[];
+        return (await Promise.all([propertyRule?.validate(object, propertyTag, scope), ...additionalRules.map(async (rule) => rule.validate(object, propertyTag, scope))])).flat();
+      }
       return (await rules.find((r) => r.property.name === propertyName)?.validate(object, propertyTag, scope)) ?? [];
```
Sayan751, line 65: `propertyRule` should be reused here.

packages/validation/src/validator.ts
```diff
@@ -57,6 +57,11 @@ export class StandardValidator implements IValidator {
     const scope = Scope.create({ [rootObjectSymbol]: object });
 
     if (propertyName !== void 0) {
+      const propertyRule: PropertyRule | undefined = rules.find((r) => r.property.name === propertyName);
+      if (propertyRule !== void 0 && propertyRule.linkedProperties.length > 0) {
+        const additionalRules = propertyRule.linkedProperties.map(lp => rules.find(r => r.property.name === lp)).filter(Boolean) as PropertyRule[];
+        return (await Promise.all([propertyRule?.validate(object, propertyTag, scope), ...additionalRules.map(async (rule) => rule.validate(object, propertyTag, scope))])).flat();
```
Sayan751, line 63: Why the `.filter` is needed? Are we expecting the rules to be `falsy`?

docs/user-docs/aurelia-packages/validation/defining-rules.md
```diff
@@ -91,6 +91,19 @@ validationRules
...
+This functionality can be useful when a change in one property has to trigger additional validations because the values are intrinsically related.
+Linked properties must be provided as an array of object keys.
+
+```typescript
+validationRules
+  .on(person)
+  .ensure('name')
+    .linkProperties(['age', 'address']);
+```
+
```
Sayan751, line 105: Maybe add a bit more details. To me at least, after reading the docs, the purpose was not clear.

packages/validation/src/rule-provider.ts
```diff
@@ -123,6 +123,7 @@ export class PropertyRule<TObject extends IValidateable = IValidateable, TValue
   private latestRule?: IValidationRule;
   /** @internal */
   public readonly l: IServiceLocator;
+  public linkedProperties: PropertyKey[] = [];
```
Sayan751, line 126: I think that it would be great if these linked properties can be serialized; refer [model based validation](https://docs.aurelia.io/aurelia-packages/validation/model-based-validation). 

To that end, the `accept` method(s) and the rule hydrators needs to be refactored.

packages/validation/src/rule-provider.ts
```diff
@@ -266,6 +267,15 @@ export class PropertyRule<TObject extends IValidateable = IValidateable, TValue
       throw createMappedError(ErrorNames.rule_provider_no_rule_found);
     }
   }
+
+  /**
+   * Links the PropertyRule to other key values.
+   * @param properties - Array of property keys to link.
+   */
+  public linkProperties(properties: PropertyKey[]) {
```
Sayan751, line 275: I think that we should also support lambda expressions as input along with `PropertyKey`, like we also do for the `.ensure`.

Sayan751 (commented):
Hi @XmasCarola, thanks for the contribution. This is an interesting change. Left a couple of suggestions. 

Coming to bike-shading now :D 
The method name `linkProperties` seems a bit ambiguous to me. The purpose is not immediately clear whether the validation of the current property is dependent on the other properties, or is it the other way around. As your test shows, it is the other way around. From that perspective, a `.dependsOn` method will be more intuitive.

As an example from your tests

```
 const rules = validationRules
  .on(defineRuleOnClass ? Person : obj)

    .ensure('name')
    .required()

    .ensure('age')
    .dependsOn('name')
    .required()

    .ensure('address.line1')
    .dependsOn('name')
    .required()
    .withMessage('Address is required.')
```

It don't have to be `dependsOn`, any other intuitive name should also work.

packages/validation/src/validator.ts
```diff
@@ -57,6 +57,11 @@ export class StandardValidator implements IValidator {
     const scope = Scope.create({ [rootObjectSymbol]: object });
 
     if (propertyName !== void 0) {
+      const propertyRule: PropertyRule | undefined = rules.find((r) => r.property.name === propertyName);
```
Sayan751, line 60: When validating multiple properties, we might want to do that when the validation instruction involves validating a single property or a single binding. We might want to avoid repetitive executions of the same rules when the validation is triggered for the entire object or for all bindings.

Sayan751:
Maybe slightly differently, we can mark a set of properties as one group. When one of those properties is validated, all the properties in the set are validated.

For example:
```ts
validationRules
  .on(defineRuleOnClass ? Person : obj)
  .group('name', 'age', 'address.line1') // <-- validate all 3 properties when one of those are validated.

    .ensure('name')
    .required()

    .ensure('age')
    .required()

    .ensure('address.line1')
    .required()
    .withMessage('Address is required.')
```
@bigopon What do you think about that?

bigopon:
> ... When one of those properties is validated, all the properties in the set are validated.

If we had a group of
- [radio input]: 1 way / 2 way flight
- [date input]: departure date
- [date input]: return date

I think I want to express something like this
```ts
  .ensureGroup('direction', 'departureDate', 'returnDate', (direction, departureDate, returnDate) => {
    if (direction === '1 way') {
      if (returnDate != null) { throw new Error('1 way flight has no return date'); }
      if (startDate < Date.now()) { throw new Error('No time travel possible'); }
    } else {
      if (startDate > returnDate) { throw new Error('not possible to go back in time'); }
      if (!hasFlight(returnDate)) { throw new Error('no flights on the date back, boat?'); }
    }
  });
```
and the trigger point will probably be after the radio and a date (either departure or return) has been selected. I wouldn't want to trigger any date independently I think.

Sayan751:
@bigopon I like your example a lot. Only instead of throwing error, it needs to return `false`. Or we can utilize the newly introduced state rule in background for multiple error messages.

XmasCarola:
Hi guys, thank you for your inputs! :)

bigopon:
Just an idea, maybe a serializable version of the `ensureGroup` above would be something like
```ts
.ensureGroup(
  ['direction', 'departureDate', 'returnDate'],
  `$direction === '1-way' && $returnDate == null && $departureDate > Date.now()
    || $returnDate > $departureDate
  `
)
```

XmasCarola:
Hi guys, I've made some additional integrations to the code. After analyzing the structure of ensureGroup(), I realized that what @bigopon is aiming for can be achieved with the current setup. As a result, I designed a test to emulate his request (you can refer to "can validate linked properties w/ custom logic" inside validator.spec.ts). Let me know your thoughts. Thanks!

packages/validation/src/validator.ts
```diff
@@ -57,7 +57,12 @@ export class StandardValidator implements IValidator {
     const scope = Scope.create({ [rootObjectSymbol]: object });
 
     if (propertyName !== void 0) {
-      return (await rules.find((r) => r.property.name === propertyName)?.validate(object, propertyTag, scope)) ?? [];
+      const propertyRule: PropertyRule | undefined = rules.find((r) => r.property.name === propertyName);
+      if (propertyRule !== void 0 && propertyRule.linkedProperties.length > 0) {
+        const additionalRules = propertyRule.linkedProperties.map(lp => rules.find(r => r.property.name === lp)).filter(Boolean) as PropertyRule[];
+        return (await Promise.all([propertyRule.validate(object, propertyTag, scope), ...additionalRules.map(async (rule) => rule.validate(object, propertyTag, scope))])).flat();
```
bigopon, line 63: is the `.flat()` necessary?

packages/validation/src/validator.ts
```diff
@@ -57,7 +57,12 @@ export class StandardValidator implements IValidator {
     const scope = Scope.create({ [rootObjectSymbol]: object });
 
     if (propertyName !== void 0) {
-      return (await rules.find((r) => r.property.name === propertyName)?.validate(object, propertyTag, scope)) ?? [];
+      const propertyRule: PropertyRule | undefined = rules.find((r) => r.property.name === propertyName);
+      if (propertyRule !== void 0 && propertyRule.linkedProperties.length > 0) {
```
bigopon, line 61: I think with this addition, what we have is the ability to trigger a set of rules together with another rule.  I also can't answer whether the depended rules will run first or in parallel with the dependent rule. For those reasons, I think the `.dependsOn()` is inappropriate. Maybe let's try to come up with a more accurate name?
What about:
- `afterRules(...)`
- `withRules(...)`
- `whenRules(...)`
- `with(...)`
- `useRules(...)`
- `withGroup(...)`

bigopon:
btw, the benchmark jobs can be fixed by simply updating this branch with latest master @XmasCarola

packages/validation/src/validator.ts
Sayan751: After revisiting the code, I also agree that the `.dependsOn` might not be a correct name for this. If we have the following:

```ts
rules
  .on(object)
  .ensure('prop1')
  .rule1()
  .dependsOn(['prop2', 'prop3'])
```

then the rules associated with `prop2` and `prop3` are executed when the `prop1` is validated. So, in other words, the validation of `prop2` and `prop3` are triggered by the validation of `prop1`, making `prop2` and `prop3` dependent on `prop1` and not the other way around. I am unsure if that is the goal? If yes, then `.dependsOn` for that might be a confusing name.

I was under the impression that the goal is to validate a set of properties as a group. Therefore, I am inclined towards the `.ensureGroup` idea from @bigopon. @XmasCarola Do you think, implementing that satisfies your use-case as well?

Sayan751: This is not a blocking comment, but on this note, we should also evaluate the UX aspect for this. My preference is that when one property is changed and therefore validated, only the rules associated with that property are executed, and the errors are shown accordingly.
However, with the introduction of this change, many properties will potentially be validated when the value of one property is changed, showing validation errors for potentially many properties. IMO, that increases the cognitive load for the end user.

Sayan751: @XmasCarola I too have experience such need for validating multiple properties as a group. Hence, there is no question about the legitimate need of this feature. The point on the UX is more or less an emphasis on seeing it in action and evaluating how it is affecting the UX for a legitimate use-case. I appreciate you working on it!

Sayan751:
@XmasCarola @bigopon I think that if we agree on the `.ensureGroup` thing, then I propose to keep it simple for the first version. Then, depending on how the devs are using it, we can always make it more complicated (read fine-tuning :)). My proposal is as follows:

```js
.ensureGroup(
  ['property', or => accessor],
  (value1, value2, object) => boolean | Promise<boolean>
)
```

The rule is executed when any one of the property value is changed. The order of the rule execution is determined by how the rule is registered. I mean that that is the default order of executing the rules. We should maintain that here as well, without treating this rule too specially.

My mental model is as follows: when we encounter an `.ensureGroup`, we attach the same rule for all the given properties. That way when one of those property changes, the rule will automatically be executed, with a small tweak that using the property accessors or expression, the current value of all properties will be fetched and supplied to the lambda for the evaluation of the rule. When generating an error from this rule, the error needs to be generated for all the properties (needs to be replicated for alll properties), so that the information can be shown in the UI via the subscribers (this makes me a bit uncomfortable; maybe we need smart out-the-box subscribers that can show only one error for a single UI component; however, this might be a secondary issue, as showing errors is very much context-specific, and one might need custom subscriber anyway).

bigopon:
> ... the error needs to be generated for all the properties ...

I don't feel like this is an acceptable default at all. Adding a group validation and suddenly I won't be able to figure out where it went wrong in my tests is an undesirable scenario to be in.

In a group validation, issues may not always be with a single property. Like in my example, sometimes it's the direction, sometimes it's the departure, sometimes it's the return date. Do we want to always notify all those direction departure and return dates when only one of those fails?
If we don't want that, and want to be precise where to notify, where should be have that information? To me, maybe the lambda that is responsible for validation should also be responsible for signaling: maybe it should return more than just a boolean:

```diff
.ensureGroup(
  ['property', or => accessor],
  (value1, value2, object) =>
    | { valid: true }
    | Promise<{ valid: true }>
+   | { valid: false; property: string }
+   | Promise<{ valid: false; property: string }>
)
```

Sayan751:
@bigopon Like the idea. I would suggest an extension of that to the following:

```diff
.ensureGroup(
  ['property', or => accessor],
  (value1, value2, object) =>
+    | { valid: boolean, { property: string, valid: boolean, messageOrMessageKey?: string }[] }
+    | Promise<{ valid: boolean, { property: string, valid: boolean, messageOrMessageKey?: string }[] }>
)
```

If no message or message key is provided in the result, then the usual workflow of resolving the message should take over the task. We need to remember that it should work with i18n as well.

XmasCarola:
@Sayan751 @bigopon Sorry guys, I committed by mistake on master.

Sayan751:
Hey @XmasCarola! Just checking in. Are you still interested in concluding the changes? In case, you need a hand or want us to take over, just let us know ✌️

XmasCarola:
> Hey @XmasCarola! Just checking in. Are you still interested in concluding the changes? In case, you need a hand or want us to take over, just let us know ✌️

Sayan751:
@XmasCarola Thanks for getting back. I will have a look at the changes.

Sayan751:
@XmasCarola I had a look at the changes. It seems to me that your implementation differs from the goal of `ensureGroup` as expressed by @bigopon's [comment](https://github.com/aurelia/aurelia/pull/2015#issuecomment-2326477591) or my [comment](https://github.com/aurelia/aurelia/pull/2015#issuecomment-2327150402).

If my understanding of your implementation is correct, then you have attempted to trigger the validations of the grouped properties when one of those is validated. However, the idea, according to our previous discussion, was that using the `ensureGroup` a new group level rule is defined, that is executed whenever one of the grouped properties is validated.

XmasCarola:
@bigopon Thanks for the feedback — you're absolutely right that the implementation doesn't define a new group-level rule per se, but instead groups the rules of the specified properties. This was the outcome of evaluating how best to introduce multi-property validation within the boundaries of the current rule engine.

Rather than requiring users to define an entirely new rule just to group existing validations, I opted to leverage the already defined rules and their execution priorities. This approach avoids redundancy, keeps the API surface minimal, and integrates cleanly with the current structure — while still achieving the intended effect: triggering all grouped validations when one property is validated.

I couldn't identify a clear use case where defining a separate group-level rule would bring additional value — unless there’s a need to decouple or redefine validation logic already tied to individual properties. But in most cases, when a property is validated, it makes sense to apply all of its conditions. That said, I’m open to discussing if we foresee future scenarios where a true group-level rule might be necessary.

XmasCarola:
As further notice, if there's a future need to distinguish between grouped and ungrouped rules for the same property, I could consider introducing that logic. In such a case, I believe the most straightforward and scalable approach would be to add a flag at the rule level, rather than introducing a new layer of group-level rules. Please, let me know you considerations on this!

Sayan751:
@bigopon those are interesting points. 

I thought that API similar to `depends` was discussed previously and decided against. Similar arguments go for `with`; although this was not discussed before, in this context this sounds similar to `depends`.

My advice is to not use `observes` as that will potentially raise the question whether the other properties are not observed.

What I particularly like about having an API that defines a _rule_ with `ensureGroup` is that it gives the opportunity to define the validation logic for the whole group that is associated with all the properties in the group in a single place. Any API that only expresses the dependency between properties cannot inherently do that, resulting in distributing the validation logic which might be difficult to follow IMO. The question is which kind of API is generally preferred in this context?

XmasCarola:
I agree with what @bigopon is saying. My concern wasn’t about the feasibility of the change or its potential to improve the UX — I believe it could. That said, I think it would involve a more structural shift in how the validation flow is currently designed. My goal with this PR wasn’t to overhaul the validation API, but to introduce grouped validations in a way that aligns with the existing architecture. That said, it’s great that @Sayan751 pointed out that the API could be improved — it’s definitely a valuable insight for future iterations. Happy to help if needed — just give me a shout.

bigopon:
Thanks @XmasCarola, its nice you want to have some extra fun. I wasnt aware that we had this API in v1. Probably been awhile. Im not sure if the 2 APIs can give the same sort of code flow in an application. We can still test and see what we come up with before the final merge. Just my thought. @Sayan751 makes the call though

Sayan751:
@XmasCarola I don't want to change the existing API as well. My implementation approach would be as follows:
- Create a "GroupRule" class that has the list of properties and the validation function. Both of those come from the `ensureGroup`.
- A call to the `ensureGroup` creates an instance of the rule and adds that to all the properties in the group.
- That way, whenever a single property in the group is validated, the rule is guaranteed to be executed.
- The new group rule takes care of procuring the value of all the properties in the group by evaluating the property expressions. I am on the fence on this particular step, but it should be doable. Then the rule invokes the validation function with the values.
- The validator class might need to be changed to work with the complex object from the group rule execution; as of now it works with Boolean. Additionally it takes care of the properties indicated from the group rule; for example `direction` is changed, but as per the group rule validation function, the invalid property is now `departureDate`, for example. The validator should construct the validation result in such a way that the error is for the `departureDate` property instead of the `direction` property.
- Lastly a couple of tests, especially for the validator and for the validation-controller + validate binding behavior should be able to assert the validity of this approach.
- Note that there are no breaking changes to the existing public API introduced in this way.

Do you think: 
- this can pose a conflict in terms of breaking the existing API? 
- doesn't satisfy your use-case?
- is in conflict with the concerns, you have expressed?

I can create a branch and push implementation following this approach, so that the idea is a bit easier to showcase. Let me know if you would prefer that.

XmasCarola:
@Sayan751 The proposed implementation looks good to me. You can go ahead with a new branch.

PR #2016, opened 2024-07-29
docs(router-lite): route-context injection in app-root

Sayan751:
# Pull Request

## 📖 Description

Added docs in response to the issue reported here: https://discourse.aurelia.io/t/cannot-resolve-iroutercontext-for-router-lite/5503/2

### DIFF
--- a/docs/user-docs/router-lite/navigation-model.md
+++ b/docs/user-docs/router-lite/navigation-model.md
@@ -69,6 +69,10 @@ Note that in the example above we aren't dealing with async routing.
 Therefore, *for that example* waiting the `INavigationModel#resolve()` can be avoided.
 {% endhint %}
 
+{% hint style="info" %}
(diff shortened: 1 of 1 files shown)


bigopon:
@Sayan751 shouldn't we be able to provide a special app root injection object with everything in it empty, and then assign the data later?

Sayan751:
@bigopon In order to instantiate a root route-context, we need to know the root view-model (from the app-root); this is unavoidable, as the router-lite has a first-class support for non-static routing configuration. Thus, an attempt to resolve a route-context at the ctor-level of the app-root view-model creates a conflict.

I am not sure if there is a way around this, but I am assuming that such attempt has very low ROI. As the lazy injection of ÌRouteContext`at the app-root works sufficiently well, maybe we can leave this infra to its current state and wait for user reports those demonstrate the shortcoming of the current infra.

PR #2024, opened 2024-08-09
refactor(parser): no longer provide default to exp parser interface

bigopon:
## 📖 Description

One possible optimization of build time template compilation is that template expression can be precompiled into AST. The fully benefit from the potential tree shaking that comes with this optimisation, the parser interface shouldn't be tied to the parser implementation. This PR removes that connection.

This shouldn't affect any application that uses standard `StandardConfiguration`, but if for some reason a minimal setup was used, then the container of that setup needs to register `ExpressionParser` from `@aurelia/expression-parser` to make `container.get(IExpressionParser)` work.

cc @fkleuver @Sayan751


bigopon:
@fkleuver with this we kind of complete the objective of separating default implementation from interface, at least for the major interfaces ... after sometime 😸

PR #2029, opened 2024-08-12
fix(validation-i18n): prioritization of translated messages

Sayan751:
# Pull Request

### 🎫 Issues

Reported in https://discourse.aurelia.io/t/au2-validation-in-conditionally-rendered-component/5507/8.


packages/validation-i18n/src/localization.ts
```diff
@@ -82,7 +84,7 @@ export class LocalizedValidationMessageProvider extends ValidationMessageProvide
       key ??= messageKey;
     }
 
-    return this.setMessage(rule, this.i18n.tr(this.getKey(key)));
+    return this.setMessage(rule, this.i18n.tr(hasTranslation ? i18nKey : this.getKey(key)));
```
bigopon, line 87: shouldn't we return early with `hasTranslation` instead? If the style of the package is to have a single return then all good.

.circleci/config.yml
```diff
@@ -864,7 +864,7 @@ workflows:
       - e2e_test:
           <<: *filter_ignore_develop_release
           name: e2e_validation_issue_2025
-          path: "packages/__e2e__/11-validation-issue-2025"
+          path: "packages/__e2e__/11-validation"
```
bigopon, line 867: 👍

bigopon (approved):

packages/validation-i18n/src/localization.ts
Sayan751: Done.

bigopon:
nice work @Sayan751

Issue #2019, opened 2024-08-06
[Router lite] attached Lifecycle hooks execution hierarchy


ivanbacher:
This might be a bug or expected behaviour. Not sure, so feel free to close if it's expected.

Example app:

```
MyApp
    RouteOne
        ComponentOne
            ComponentTwo
```

From reading the docs the order of the `attached` hook should be bottom up. So

```
ComponentTwo - attached()
ComponentOne  - attached()
RouteOne - attached()
MyApp  - attached()
```

However it is currently:

```
MyApp  - attached()
ComponentTwo - attached()
ComponentOne  - attached()
RouteOne - attached()
```

`MyApp  - attached()` gets called first. Maybe because it's the root component?

[StackBlitz example](https://stackblitz.com/edit/au2-attached-hook-hierarchy)

Sayan751:
> MyApp  - attached() gets called first. Maybe because it's the root component?

@ivanbacher You are right. This is not a bug. A routed view always gets activated when the hosting `au-viewport` is attached. This explains the order of invocations, as you have observed.

I think the docs for router-lite can be adjusted to reflect this. In case you feel like contributing to the docs, feel free to submit a PR. Otherwise, I will try to push the changes in docs.

bigopon:
Is this intuitive? I thought the viewport would only call attached after it has finished loading. Just asking based on the perspective of this issue.

Sayan751:
@bigopon @ivanbacher I was not very clear with my explanation before. Let me add additional information.

The initial load of routed route happens post-activation of the app.

https://github.com/aurelia/aurelia/blob/0f3dbee6b350ef6b7288f901ac9d3df525125c0e/packages/router-lite/src/configuration.ts#L72

This means that even if the `au-viewport` in the root component is attached and the app gets completely activated first. At this point, the viewport is still without any routed view-model, as the initial load by the router is yet to happen. Once the app is activated, as a post-activation task, the router loads a route for the first time. And this is the reason behind the observed order of lifecycle hook invocations.

I am inclined towards keeping the order of invocations in its current state, unless it is causing a problem that cannot be solved without changing the order of invocations.

bigopon:
An issue with that timing is we cannot tell when the initial routing has finished in app root `attached`, and likely will have to opt to listen to some events with a more cumbersome setup. It'd be better if we can have the initial timing more "smooth".
Or did I mis-analyze it?

Sayan751:
[...]

PR #2032, opened 2024-08-13
docs(router-lite): order of hooks invocation

Sayan751:
# Pull Request

## 📖 Description

Addresses the documentation need from #2019.

### DIFF
--- a/docs/user-docs/router-lite/routing-lifecycle.md
+++ b/docs/user-docs/router-lite/routing-lifecycle.md
@@ -315,3 +315,44 @@ In the context of swapping two views/components it is as follows.
 - `loading` hook (when present) of the current component is invoked.
 
 Note that the last 2 steps may run in parallel, if the hooks are asynchronous.
+
+## Order of invocations of component lifecycle hooks
+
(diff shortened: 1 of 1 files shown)


docs/user-docs/router-lite/routing-lifecycle.md
```diff
@@ -315,3 +315,44 @@ In the context of swapping two views/components it is as follows.
...
+```
+
+In this case, the component lifecycle hooks are invoked in the following order.
+
+1. app-root `attached`.
+2. component-two `attached`.
+3. component-one `attached`.
+4. routed-view `attached`.
+
+Note that the application root is attached before any other components are attached. This happens because the router-lite starts loading the first route only after the app-root, and thereby the viewport(s) it is hosting, are fully activated/attached. In order to load a route, the router needs registered viewports. The registration process of a viewport only happens during the `attaching` phase of a viewport. More details on this topic, can be found in this [GitHub issue](https://github.com/aurelia/aurelia/issues/2019).
```
bigopon, line 358: unrelated to this PR: `attaching` is kind of unnecessarily late to register a viewport.

bigopon (approved):

Issue #2025, opened 2024-08-10
Validation in conditionally rendered component


Sayan751:
# 🐛 Bug Report

This issue was originally reported in Discourse: https://discourse.aurelia.io/t/au2-validation-in-conditionally-rendered-component/5507

Repro link: https://stackblitz.com/edit/au2-validation-issue-reproduction-discourse-5507?description=Starter%20project%20for%20Node.js,%20a%20JavaScript%20runtime%20built%20on%20Chrome%27s%20V8%20JavaScript%20engine&file=src%2Fedit.ts,src%2Fapp-root.ts&title=node.new%20Starter

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Aurelia          |
| Language         |
| Browser          |
| Bundler          |
| Operating System |
| NPM/Node/Yarn    |

PR #2033, opened 2024-08-14
feat(bindable): support aggregate callback

### LINKED ISSUE #2021: [Feature request] ability to aggregate multiple bindable change callbacks
_Originally posted by @ekzobrain in https://github.com/aurelia/aurelia/issues/1957#issuecomment-2132673808_

---

I know about propertyChanged, but that's not what I meant. I meat exactly propertiesChanged(), which might have the following definition:

```typescript
function propertiesChanged(changes: Record<string, { name: string, newValue: unknown, oldValue: unknown }>): void
```

and may be used like this:
```typescript
propertiesChanged({ prop1, prop2 }) {
  if (prop1 && prop2) {
    // both bindables were changed at the same time (in a batch for example). apply first update strategy
  } else if (prop1) {
    // only prop1 was changed - apply second update strategy
  } else {
    // only prop2 was changed - apply third update strategy
  }
}
```

So it will not fire for each changed property, but only once for all batch changes.

---

Some extra comment from @ekzobrain

> There is no way to set a single change callback for multiple properties. In Angular there's an ngOnChanges() callback (https://angular.io/api/core/OnChanges) which fires once for all changed bindables. I would like to have something similar in Aurelia, for example "propertiesChanged()". There are cases when component update strategy heavily depends on wheather one or several related properties changed at once.

bigopon:
## 📖 Description

Currently, as per noted in #2021 by @ekzobrain:

> There is no way to set a single change callback for multiple properties. In Angular there's an ngOnChanges() callback (https://angular.io/api/core/OnChanges) which fires once for all changed bindables. I would like to have something similar in Aurelia, for example "propertiesChanged()". There are cases when component update strategy heavily depends on wheather one or several related properties changed at once.

This is a missing feature of v1 where it was quite difficult to aggregate the changes during 1 flush cycle of a component. Thanks @ekzobrain for bringing this up.

This PR adds support for the ability to declare a aggregated callback on the component instances (attributes/elements) to receive all the changes at once, with the following signature:

```ts
export interface ICustomElementViewModel {
  ...
  propertiesChanged?(changes: Record<string, { newValue: unknown; oldValue: unknown }>): void;
}
```

- Note that if there' multiple writes to the same property, the changes record given to `propertiesChanged` will only have the latest values, i.e:
    ```ts
    class Component {
      prop = 0
    }
    ...
    component.prop = 1;
    ...
    component.prop = 2;
    ```
    Changes record will be:
    ```ts
    {
      prop: { newValue: 2, oldValue: 1 }
    }
    ```
    not:
    ```ts
    {
      prop: { newValue: 2, oldValue: 0 }
    }
    ```
- The timing of the aggregated changes callback will be the immediate next tick, queued via `Promise.resolve().then(...)` Though this may change if it's deemed unsuitable. We will test for a few versions before RC.

### 🎫 Issues

Close #2021

cc @Sayan751 @fkleuver @ekzobrain 

Note: about the timing, we can try to make it synchronous after a `batch`, but I'm not sure if that's desirable.

cc @3cp with this next tick group update, we kind of have a way to deal with state tearing.

### DIFF
--- a/docs/user-docs/components/bindable-properties.md
+++ b/docs/user-docs/components/bindable-properties.md
@@ -99,6 +99,43 @@ export class NameComponent {
 }
 ```
 
+If you have multiple bindable properties like `firstName`/`lastName` in the above example, and want to use a single callback to react to those changes,
+you can use `propertyChanged` callback. `propertyChanged` callback will be called immediately after the targeted change callback.
+The parameters of this callback will be `key`/`newValue`/`oldValue`, similar like the following example:
+
+```ts
+export class NameComponent {
+    @bindable firstName = '';
+    @bindable lastName  = '';
+
+    propertyChanged(key, newVal, oldVal) {
+      if (key === 'firstName') {
+
+      } else if (key === 'lastName') {
+
+      }
+    }
+```
+
+In the above example, even though `propertyChanged` can be used for multiple properties (like `firstName` and `lastName`), it's only called individually for each of those properties.
+If you wish to act on a group of changes, like both `firstName` and `lastName` at once in the above example, `propertiesChanged` callback can used instead, like the following example:
+```ts
+propertiesChanged({ firstName, lastName }) {
+  if (firstName && lastName) {
+    // both firstName and lastName were changed at the same time
+    // apply first update strategy
+    const { newValue: newFirstName, oldValue: oldFirstName } = firstName;
+    const { newValue: newLastName, oldValue: oldLastName } = lastName;
+  } else if (firstName) {
+    // only firstName was changed - apply second update strategy
+    // ...
+  } else {
+    // only lastName was changed - apply third update strategy
+    // ...
+  }
+}
+```
+
 ## Configuring bindable properties
 
 Like almost everything in Aurelia, you can configure how bindable properties work.&#x20;
--- a/docs/user-docs/components/bindable-properties.md
+++ b/docs/user-docs/components/bindable-properties.md
@@ -136,6 +136,35 @@ propertiesChanged({ firstName, lastName }) {
 }
 ```
 
+For the order of callback when there' multiple callbacks involved, refer the following example:
+If we have a component class that looks like this:
+```ts
+class MyComponent {
+  @bindable prop = 0
+  
+  propChanged() { console.log('prop changed'); }
+
+  propertyChanged(name) { console.log(`property "${name}" changed`) }
+
+  propertiesChanged(changes) {
+    console.log('changes are:', changes)
+  }
+}
+```
+When we do
+```ts
+myComponent.prop = 1;
+console.log('after assign');
+```
+the console logs will look like the following:
+
+```
+propChanged
+property "prop" changed
+after assign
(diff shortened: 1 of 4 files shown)


ekzobrain:
Does current timing conform to synchronous binding system? Or it is considered async as an exception?

docs/user-docs/components/bindable-properties.md
```diff
@@ -99,6 +99,43 @@ export class NameComponent {
...
+  } else if (firstName) {
+    // only firstName was changed - apply second update strategy
+    // ...
+  } else {
+    // only lastName was changed - apply third update strategy
+    // ...
+  }
+}
+```
+
```
Sayan751, line 138: I think that it needs to be documented, how the callbacks will be invoked when all of them `firstNameChanged`, `propertyChanged`, and `propertiesChanged`. This is already tested, but missing in docs.

3cp:
From reading I don't think this handles deep change? That was the source of my state-tearing of aurelia dnd service.

```js
class VM {
  someService // Injected service, might be 3rd party
}

class SomeService {
  mutate() {
    this.a = ...;
    this.b = ...;
  }
}
```

bigopon:
> Does current timing conform to synchronous binding system? Or it is considered async as an exception?

It is an exception, everything is synchronous except this aggregated one. Maybe it should be made clearer in the doc.

> Also why not use PropertyKey instead of string in propertiesChanged() signature just as in propertyChanged?

I originally had more than `string`, but then realised we don't support anything other than `string` for `@bindable`, since you wouldn't be able to use it in the template anyway, so changed it to `string`. `propertyChanged` should be updated to reflect this too.

bigopon:
> From reading I don't think this handles deep change? That was the source of my state-tearing of aurelia dnd service.

Indeed, this is only for `@bindable` properties. It won't handle deep mutation. Can you please help note how the observation & mutation are done when state tearing happens, at least for the awareness?

3cp:
Like this:
```js
class VM
  service: SomeService;
  
  // Au1 will get true all the time
  // Au2 will have a false spike between a and b assignments.
  get computedState() {
    return this.service.a === this.service.b;
  }
}

class SomeService {
  a = "1";
  b = "1";

  someMethod() {
    this.a = "2";
    // Au2 introduced side effects between this two lines.
    // This is way against intuition.
    //
    // Note "SomeService" can be 3rd party lib that
    // User has no control of.
    //
    // I don't think this is a good practice.
    this.b = "2";
  }
}
```

3cp:
As I repeatedly said, au1 has no issue with 3rd party lib.

No matter what you fix in au2 sync mode, it only applies to the code that user can modify, not 3rd party lib.

It's not humble to introduce side effects to 3rd party lib.

docs/user-docs/components/bindable-properties.md
bigopon: thanks, added the doc for this 👍

bigopon:
Thanks @3cp , I've created a separate RFC at #2025 to deal with the timing, this PR is good to go without having to wait for the finalisation of that RFC.

PR #2036, opened 2024-08-15
chore(router-lite): removed IRouteNode implementation in RouteNode

Sayan751:
# Pull Request

## 📖 Description

The inheritance was relatively superfluous. The `IRouteNode` is also not a significant part of the public API. This commit corrects that by renaming the said interface to `IRouteNodeInitializationOptions` as the interface is merely used for creating a `RouteNode`. However, the interface is exported from the package as `RouteNode#create` is a part of the public API.

Thanks @ekzobrain for [reporting this](https://discord.com/channels/448698263508615178/1243519563283435520/1273318182798622936).

### DIFF
--- a/packages/router-lite/src/index.ts
+++ b/packages/router-lite/src/index.ts
@@ -67,6 +67,7 @@ export {
 } from './route-expression';
 
 export {
+  IRouteNodeInitializationOptions,
   RouteNode,
   RouteTree,
 } from './route-tree';
--- a/packages/router-lite/src/route-tree.ts
+++ b/packages/router-lite/src/route-tree.ts
@@ -49,7 +49,7 @@ import { resolveCustomElementDefinition, resolveRouteConfiguration, RouteConfig,
 import { Events, getMessage } from './events';
 import { pathUrlParser } from './url-parser';
 
-export interface IRouteNode {
+export interface IRouteNodeInitializationOptions {
   path: string;
   finalPath: string;
   context: IRouteContext;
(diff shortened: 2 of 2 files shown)


ekzobrain:
What would then be the signature of `loading()` view model hook? Previously I did it like this:

```ts
loading(params: Params, next: IRouteNode)
```

bigopon (approved):
Do we need to remove the interface? Designing/learning the design with the concrete implementation is a bit more stressful than interfaces/headers

Sayan751:
> What would then be the signature of `loading()` view model hook? Why doesn't RouteNode implement any interface? Previously I did it like this:
> 
> ```ts
> loading(params: Params, next: IRouteNode) {
>   // use next.queryParams
> }
> ```

@ekzobrain You can simply use `RouteNode` instead of `IRouteNode`. The signatures of all the lifecycle hooks are documented here: https://docs.aurelia.io/getting-to-know-aurelia/routing/aurelia-router-lite/routing-lifecycle.



> Do we need to remove the interface? Designing/learning the design with the concrete implementation is a bit more stressful than interfaces/headers

@bigopon I agree with you that it is not the best thing to do. In this case, the interface was useless for implementation. It was merely used only for the initialization purpose of `RouteNode`. Also, as the `RouteNode` is no candidate for DI, it also takes away the need for an additional interface. We can still have an interface that is implemented by `RouteNode`, but that will be just for the sake of satisfying the design pattern, instead of serving any real purpose. In case you are still keen on having that interface, let me know, I can make the change.

Sayan751:
@bigopon I am going ahead and merging this.

Issue #2042, opened 2024-08-22
[router-lite] Issue with host injection in route (v2.0.0-beta.21)


ivanbacher:
When injecting host element into a route and then using `querySelector` on host element to select a html element within the route template, no html is found and null is returned. New behaviour since `v2.0.0-beta.21`

E.g

route-one.js
```
import { inject } from 'aurelia';

@inject(Element)

export class RouteOne {
  private host

  constructor(host) {
    this.host = host
  }

  attached() {
    this.host.querySelector('.my-class') // this is null 
  }
}
```

route-one.html
```
<div class="my-class"></div>
```

`this.host.querySelector('.my-class')` should return the correct element.

StackBlitz example: https://stackblitz.com/edit/aurelia-2-route-host-injected-bug-2

`2.0.0-beta.20` working example: https://stackblitz.com/edit/aurelia-2-route-host-injected-bug-1

ivanbacher:
workaround atm is: 
```
attached() {
  this.host = document.body.querySelector(<name of route>)
}
```

PR #2043, opened 2024-08-27
feat(runtime-html): reuse host and hostController

Sayan751:
# Pull Request

## 📖 Description

This enables reusing the given host and hostController without trying to resolve the hostController again in the Controller. This fix is likely a predecessor for the fix for #2042.


bigopon:
This looks good to me, though I think we should eventually remove the host controller part.

bigopon (approved):

PR #2045, opened 2024-08-28
fix(router-lite): element injection issue

### LINKED ISSUE #2042: [router-lite] Issue with host injection in route (v2.0.0-beta.21)

Sayan751:
# Pull Request

## 📖 Description

This uses the changes from the #2043 to ensure that the injected element in the routed viewmodel is indeed hydrated. 

### 🎫 Issues

Fixes #2042.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -165,13 +165,16 @@
         "--recursive",
         "--timeout",
         "5000",
+        "-r",
+        "source-map-support/register",
         "--watch-extensions",
         "js",
         "${workspaceRoot}/packages/__tests__/dist/setup-node.js",
         "${workspaceRoot}/packages/__tests__/dist/router-lite/**/*.spec.js"
       ],
       "cwd": "${workspaceRoot}",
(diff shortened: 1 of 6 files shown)


Sayan751:
@bigopon I am merging this one. In case you notice something funny, let me know. I will take care of that in a follow-up PR.

PR #2046, opened 2024-08-31
fix(router): fix redirect to async raise cond issue

jwx:
# Pull Request

## 📖 Description

This PR fixes a race condition issue with async redirects to async loaded components.

## 📑 Test Plan

- Add tests for the specific occasions


packages/router/src/instructions/instruction-component.ts
```diff
@@ -215,17 +216,36 @@ export class InstructionComponent {
    *
    * Throws instantiation error if there was an error during instantiation.
    */
-  public toInstance(parentContainer: IContainer, parentController: IHydratedController, parentElement: HTMLElement, instruction: RoutingInstruction): IRouteableComponent | null {
-    // TODO: Allow instantiation from a promise here, but awaiting resolve?
-    void this.resolve(instruction);
+  public toInstance(parentContainer: IContainer, parentController: IHydratedController, parentElement: HTMLElement, instruction: RoutingInstruction): IRouteableComponent | null | Promise<IRouteableComponent> {
```
bigopon, line 219: shouldn't the type be `... | Promise<IRouteableComponent | null>`?

packages/router/src/utilities/open-promise.ts
```diff
@@ -24,10 +24,13 @@ export class OpenPromise<T = void> {
    */
   private _reject!: (reason?: unknown) => void;
 
-  public constructor() {
+  public static promises: OpenPromise<any>[] = [];
```
bigopon, line 27: Why do we need this?

packages/router/src/utilities/runner.ts
```diff
@@ -137,6 +137,21 @@ export class Runner {
...
+   * @param valueOrPromise - The value or promise to run the function with
+   * @param func - The function to run
+   * @returns The result of the function or a promise that resolves to the result
+   */
+  public static runWith<T = unknown, R = unknown>(valueOrPromise: T | Promise<T>, func: (value: T) => R | Promise<R>): R | Promise<R> {
+    if (valueOrPromise instanceof Promise) {
+      return valueOrPromise.then(func);
+    }
+    return func(valueOrPromise);
+  }
```
bigopon, line 153: let's use `onResolve` from `kernel` instead of this

bigopon (commented):
nice work @jwx I 've a few preliminary comments

packages/router/src/endpoints/viewport-content.ts
```diff
@@ -195,26 +195,22 @@ export class ViewportContent extends EndpointContent {
...
-        if (!(e as Error).message.startsWith('AUR0009:')) {
-          throw e;
-        }
-
-        if (__DEV__) {
-          const componentName = this.instruction.component.name as string;
-          // eslint-disable-next-line no-console
-          console.warn(createMappedError(ErrorNames.endpoint_instantiation_error, componentName, e));
-        }
+        return Runner.runWith<IRouteableComponent | null>(
```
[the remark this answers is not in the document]
jwx: I'll replace runWith with onResolve throughout.

packages/router/src/endpoints/viewport.ts
```diff
@@ -529,9 +542,12 @@ export class Viewport extends Endpoint {
               canLoadResult = instructions;
             }
             return Runner.run(step,
+              () => {
+                void this.router.load(canLoadResult, { append: coordinator });
```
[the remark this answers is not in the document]
jwx: Not when appending to a coordinator since the instruction will be added to the the coordinator's set of instructions.

packages/router/src/utilities/open-promise.ts
jwx: We don't, it's purely a debug thing (to make it possible to track unresolved promises).

packages/__tests__/src/router/router.redirect.spec.ts
```diff
@@ -79,20 +79,32 @@ describe('router/router.redirect.spec.ts', function () {
...
-  ];
-
   const DefaultPage = CustomElement.define({ name: 'defaultpage', template: '!root!' });
-  const Zero = CustomElement.define({ name: 'zero', template: '!zero!' });
+  const Zero = CustomElement.define({ name: 'zero', template: '!zero!' }, class Zero { });
   const One = CustomElement.define({ name: 'one', template: '!one!<au-viewport name="one-vp"></au-viewport>' });
   const Two = CustomElement.define({ name: 'two', template: '!two!', }, class { public canLoad() { return 'route-zero'; } });
   const Three = CustomElement.define({ name: 'three', template: '!three!', }, class { public canLoad() { return 'zero'; } });
   const Four = CustomElement.define({ name: 'four', template: '!four!', }, class { public canLoad() { return 'zero-id'; } });
+  const Five = CustomElement.define({ name: 'five', template: '!five!', }, class { public async canLoad() { return 'route-zero'; } });
```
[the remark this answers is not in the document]
jwx: It's a set of failures that occurs after a couple of steps we're trying to fix. I don't think new tests would be much cleaner, tbh.

packages/router/src/navigation-coordinator.ts
```diff
@@ -354,11 +354,6 @@ export class NavigationCoordinator {
           // of a configured route.
           // Inform endpoint of new content and retrieve the action it'll take
           const action = endpoint.setNextContent(matchedInstruction, this.navigation);
-          if (action !== 'skip') {
-            // Add endpoint to changed endpoints this iteration and to the coordinator's purview
-            changedEndpoints.push(endpoint);
-            this.addEndpoint(endpoint);
-          }
```
bigopon, line 361: why do we move this to after swap check?

packages/router/src/instructions/instruction-component.ts
```diff
@@ -215,17 +215,36 @@ export class InstructionComponent {
...
     }
+    return this.name;
+  }
 
+  /**
+   * Creates the component instance for this instruction.
+   *
+   * Throws instantiation error if there was an error during instantiation.
+   */
+  private _createInstance(parentContainer: IContainer, parentController: IHydratedController, parentElement: HTMLElement, instruction: RoutingInstruction): IRouteableComponent | null {
```
bigopon, line 247: should this be marked with `@internal`?

packages/router/src/instructions/instruction-component.ts
```diff
@@ -215,17 +215,36 @@ export class InstructionComponent {
    *
    * Throws instantiation error if there was an error during instantiation.
    */
-  public toInstance(parentContainer: IContainer, parentController: IHydratedController, parentElement: HTMLElement, instruction: RoutingInstruction): IRouteableComponent | null {
-    // TODO: Allow instantiation from a promise here, but awaiting resolve?
-    void this.resolve(instruction);
+  public toInstance(parentContainer: IContainer, parentController: IHydratedController, parentElement: HTMLElement, instruction: RoutingInstruction): IRouteableComponent | null | Promise<IRouteableComponent | null> {
+    return onResolve(this.resolve(instruction) as IRouteableComponent, (_notUsed) => {
```
bigopon, line 219: we don't need `_notUsed`

packages/router/src/endpoints/viewport-content.ts
```diff
@@ -235,21 +231,21 @@ export class ViewportContent extends EndpointContent {
...
+                this.contentStates.set('created', void 0);
+              }) as void | Promise<void>;
           } catch (ee) {
-            if (!(ee as Error).message.startsWith('AUR0009:')) {
-              throw ee;
-            }
-            const componentName = this.instruction.component.name as string;
-            throw createMappedError(ErrorNames.endpoint_instantiation_error, componentName, ee);
-            // throw new Error(`'${this.instruction.component.name as string}' did not match any configured route or registered component name - did you forget to add the component '${this.instruction.component.name}' to the dependencies or to register it as a global dependency?`);
+            this._assertInstantiationError(ee);
```
bigopon, line 243: it seems we are missing tests to assert instantiation error, which could be a really good exercise for asserting the router fallback behaviors. Probably not for this PR but we should have it.

packages/__tests__/src/router/router.redirect.spec.ts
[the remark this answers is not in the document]
bigopon: But this is only one more failure we are trying to have a test for, shouldn't we be able to know exactly what we need to demonstrate the case? It doesn't need to be clean tbh, it just need to be plain. I can't make head or tail if this is added in a loop of 80 other scenarios.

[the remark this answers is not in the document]
jwx: It's just not one failure, it's several. Manifesting in different stages in the tests.

packages/router/src/utilities/open-promise.ts
jwx: I don't really understand what you mean. I don't think the promise will be resolved outside the wrapper. But in any case, this is just for debugging, I can just remove it entirely if you want.

packages/router/src/navigation-coordinator.ts
jwx: We don't. It was moved as part of a solution I decided against. Would've been moved back when cleaning up.

packages/router/src/utilities/open-promise.ts
bigopon: I actually looked at it wrong, it's just a promise being exposed, not the resovle/reject.

bigopon (approved):
The fix looks good, thanks @jwx , I'll get a dev version out.

Issue #2037, opened 2024-08-15
Repeat for with primitive types - invalid contextual properties


aegenet:
# 🐛 Bug Report

I've just caught a bug with the repeat.for template controller: most contextual properties _($index, $first...)_ are invalid with primitive arrays _(number[], boolean[], string[]...)_.

```html
<ul>
  <!-- <li repeat.for="val of ['a','a','b','c','d','d','d','d','h']"> -->
  <!-- <li repeat.for="val of [true,true,false,false,false,true,false,true,true]"> -->
  <!-- <li repeat.for="val of [undefined,true,'a',false,false,'a',false,true,undefined]"> -->
  <li repeat.for="val of [1,1,2,3,4,4,4,4,5]">
    index: ${$index}
    first: ${$first}
    even: ${$even}
    value: ${val}
  </li>
</ul>
```

Repro link: https://gist.dumber.app/?gist=6a430716e4df710a5de3816124fa8018

## 🤔 Expected Behavior

```bash
index: 0; first: true; even: true; value: 1;
index: 1; first: false; even: false; value: 1;
index: 2; first: false; even: true; value: 2;
index: 3; first: false; even: false; value: 3;
index: 4; first: false; even: true; value: 4;
index: 5; first: false; even: false; value: 4;
index: 6; first: false; even: true; value: 4;
index: 7; first: false; even: false; value: 4;
index: 8; first: false; even: true; value: 5;
```

## 😯 Current Behavior

```bash
index: 1; first: false; even: false; value: 1;
index: 1; first: false; even: false; value: 1;
index: 2; first: false; even: true; value: 2;
index: 3; first: false; even: false; value: 3;
index: 7; first: false; even: false; value: 4;
index: 7; first: false; even: false; value: 4;
index: 7; first: false; even: false; value: 4;
index: 7; first: false; even: false; value: 4;
index: 8; first: false; even: true; value: 5;
```

## 💁 Possible Solution

_Workaround: object array_

## 💻 Code Sample

Repro link: https://gist.dumber.app/?gist=6a430716e4df710a5de3816124fa8018

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Aurelia          |  2.0.0-beta.1 to 2.0.0-beta.21
| Language         | 
| Browser          |
| Bundler          |
| Operating System |
| NPM/Node/Yarn    |

ekzobrain:
[...]

PR #2050, opened 2024-09-19
rix(repeat): fix keyed issues & contextual property optimization

fkleuver:
# Pull Request

## 📖 Description

- Adds an optimization for contextual properties
- Fixes contextual properties for duplicate items (including duplicate primitives)
- Significant simplification of update logic by using scope objects as the keys in non-keyed mode
- Memory usage improvements

### 🎫 Issues
https://github.com/aurelia/aurelia/issues/2037


packages/runtime-html/src/binding/scope.ts
```diff
@@ -102,11 +102,11 @@ export class Scope {
     return new Scope(null, bc as IBindingContext, oc ?? new OverrideContext(), isBoundary ?? false);
   }
 
-  public static fromParent(ps: Scope | null, bc: object): Scope {
+  public static fromParent(ps: Scope | null, bc: object, oc: IOverrideContext = new OverrideContext()): Scope {
```
bigopon, line 105: 👍 there' been a few times I added this but removed in the end

packages/runtime-html/src/resources/template-controllers/repeat.ts
```diff
@@ -70,6 +69,8 @@ export class Repeat<C extends Collection = unknown[]> implements ICustomAttribut
 
   public views: ISyntheticView[] = [];
   private _oldViews: ISyntheticView[] = [];
+  private _scopes: Scope[] = [];
+  private _oldScopes: Scope[] = [];
```
bigopon, line 73: should have internal for these

packages/runtime-html/src/resources/template-controllers/repeat.ts
```diff
@@ -723,16 +630,31 @@ interface IRepeatOverrideContext extends IOverrideContext {
...
+  public get $middle(): boolean {
+    return !this.$first && !this.$last;
+  }
+  public get $last(): boolean {
+    return this.$index === this.$length - 1;
+  }
+
+  public constructor(
+    public readonly $index: number = 0,
+    public readonly $length: number = 1,
```
bigopon, line 652: likely a weird default but these won't be use so maybe it's fine

packages/runtime-html/src/resources/template-controllers/repeat.ts
```diff
@@ -588,31 +509,17 @@ export class Repeat<C extends Collection = unknown[]> implements ICustomAttribut
...
-        }
-        setContextualProperties(view.scope.overrideContext as IRepeatOverrideContext, i, newLen);
+        setContextualProperties(view.scope.overrideContext as RepeatOverrideContext, i, newLen);
         view.nodes.insertBefore(view.location!);
       } else {
-        if (_hasDestructuredLocal) {
-          astAssign(dec, view.scope, _forOfBinding, _normalizedItems![i]);
-        } else {
-          view.scope.bindingContext[local] = _normalizedItems![i];
-        }
```
bigopon, line 612: this seems like a good cleanup, items won't be retrieved/checked in so many places

packages/runtime-html/src/resources/template-controllers/repeat.ts
```diff
@@ -588,31 +509,17 @@ export class Repeat<C extends Collection = unknown[]> implements ICustomAttribut
       view.nodes.link(next?.nodes ?? _location);
 
       if (indexMap[i] === -2) {
-        viewScope = getScope(_scopeMap, _normalizedItems![i] as IIndexable, forOf, parentScope, _forOfBinding, local, _hasDestructuredLocal);
```
bigopon, line 591: 👍

bigopon (approved):
Nice work @fkleuver , a nice fix + a good cleanup 👍

ekzobrain:
Does this change fix a sorting issue when the same objects are reordered in an array? In v21 contextual properties brake after that.
Example:
```js
const a = {};
const b = {};
const c = {};
const d = {};

let arr = [a, b, c, d];

// when rendering this list objects get $index'es a=0, b=1, c=2, d=3
// then update update list using the same object in another order:

arr = [b, c, d, a]

// when rendering updated list $index'es are broken: b=0, c=1, d=2, a=0 (removed and reinserted at other position items keep their previous index)
```

ekzobrain:
And also this race condition issue? https://discord.com/channels/448698263508615178/448699089513611266/1274338349297041489

bigopon:
> And also this race condition issue? https://discord.com/channels/448698263508615178/448699089513611266/1274338349297041489

@ekzobrain list mutation while rendering will be a separate issue. This issue is about correctly tracking items movement when there's duplicate primitives

ekzobrain:
> > And also this race condition issue? https://discord.com/channels/448698263508615178/448699089513611266/1274338349297041489
> 
> 
> 
> @ekzobrain list mutation while rendering will be a separate issue. This issue is about correctly tracking items movement when there's duplicate primitives

I asket just in case it was touched alongside )

fkleuver:
> Does this change fix a sorting issue when the same objects are reordered in an array? In v21 contextual properties break after that. Example:
> 
> ```js
> const a = {};
> const b = {};
> const c = {};
> const d = {};
> 
> let arr = [a, b, c, d];
> 
> // when rendering this list objects get $index'es a=0, b=1, c=2, d=3
> // then update update list using the same object in another order:
> 
> arr = [b, c, d, a]
> 
> // when rendering updated list $index'es are broken: b=0, c=1, d=2, a=0 (removed and reinserted at other position items keep their previous index)
> ```

Yes, there is logic in place that ensures the index is updated in all circumstances, including the scenario you mentioned

PR #2060, opened 2024-10-01
Update dialog.md code sample

Lakerfield:
# Pull Request

## 📖 Description

Add missing DefaultDialogEventManager in code sample when using a customized dialog

### DIFF
--- a/docs/user-docs/aurelia-packages/dialog.md
+++ b/docs/user-docs/aurelia-packages/dialog.md
@@ -70,7 +70,7 @@ Aurelia.register(DialogConfiguration.customize(settings => {
 If there's a need to only swap some implementation, say `IDialogDomRenderer` for example, then the default implementation can be imported and mixed like the following example:
 
 ```typescript
-import { DialogConfiguration, DialogService, DefaultDialogGlobalSettings } from '@aurelia/dialog';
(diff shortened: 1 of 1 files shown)


docs/user-docs/aurelia-packages/dialog.md
```diff
@@ -81,6 +81,8 @@ Aurelia.register(DialogConfiguration.customize(settings => {
   MyDialogRenderer,
   // use default dialog global settings
   DefaultDialogGlobalSettings,
+  // use default dialog event manager
+  DefaultDialogEventManager,
```
bigopon, line 85: Thanks @Lakerfield , it's intentionally omitted since the BYO dialog dom renderer may not use the default event manager at all. Though we can doc it different, something along the line why we are not including it in the example. Maybe let's go with that?

Lakerfield: Thanks @bigopon, as far as i can find the DefaultDialogEventManager is used in the DialogController instead of the BYO dialog dom renderer. (The DefaultDialogDomRenderer uses the also left out IDialogDomAnimator) Maybe I overlooked something ;-)

bigopon: Indeed, I somehow remembered it the way in my comment, but it's as you mentioned.

Issue #1971, opened 2024-05-15
[feature request] ability to typecheck template


bigopon:
## 🔦 Context

Currently, the template expressions are compiled at runtime, and evaluated then. Because of this, it doesn't have the information whether there's a typo or typing issue. This is normally fine, until it's kind of less fine: when project size grows bigger, there's a need to  have some assistance in discovering the typo/types validity of expressions in the template. Currently the Webstorm and Vscode plugins are doing a decent job with refactoring, but this should be part of the build process.

Originally this was considered part of AOT, but since we want more realistic and short term quantifiable goals, this should stay in a separate issue as it's quite the first part to tackle.

## 💻 Examples

For the following component:
1. `my-input.ts`
    ```ts
    export class MyInput {
        @bindable value = '';
        items: { id: number; name: string}[] = [];
    
        public constructor(private readonly b: ServiceB) {}
    }
    ```
3. `my-input-html`
    ```html
    <import from="./my-text.html" as="the-text"></import>
    <bindable name="notUsed"></bindable>
    <textarea value.bind="value"></textarea>
    <p>${value}</p>
    <the-text text.bind="value"></the-text>
    <div>${b.doComplex()}</div>
    <div repeat.for="item of items">${item.id} -- ${item.name}</div>
    ```

Generate code for `my-input.ts` should have this **unused** function for type checking:
```ts
// ... import statements

// eslint-disable-next-line
function __typecheck_template_MyInput__() {
    const access = <T extends MyInput>(typecheck: (o: T) => any, expr: string) => expr;

    return `<import from="./my-text.html" as="the-text"></import>
    <bindable name="notUsed"></bindable>
    <textarea value="${access<MyInput>(o => o.value, 'value')}"></textarea>
    <p>${access<MyInput>(o => o.value, 'value')}</p>
    <the-text value="${access<MyInput>(o => o.value, 'value')}"></the-text>
    <div>${access<MyInput>(o => o.b.doComplex(), 'b.doComplex()')}</div>
    <div repeat.for="${access<MyInput & { item: MyInput['items'] }>(o => o.item, 'item')} of ${access<MyInput['items']>(o => o.items, 'items')}">${access<MyInput>(o => o.item.id, 'item.id')} -- ${access<MyInput>(o => o.item.name, 'item.name')}`
}

// ... the rest of my-input.ts code
```

[...]

PR #2066, opened 2024-11-07
feat(tooling): type-checking for templates - Phase1

Sayan751:
# Pull Request

## 📖 Description

This is the first instalment of the type-checking for templates. 

The convention plugin package exposes the `experimentalTemplateTypeCheck` flag which can be set to `true` to enable the type-checking for the templates. Example:

```js
{
  test: /\.ts$/i,
  use: [
    'ts-loader',
    {
      loader: '@aurelia/webpack-loader',
      options: {
        experimentalTemplateTypeCheck: true, //enables type-checking for templates
        // enableConventions: false,         // <-- use this when you are not using conventions
      }
    }
  ], 
  exclude: /node_modules/
}
```

The type checking can be used with or without convention. Use the `enableConventions` to de-/activate conventions.

In this instalment, the following areas are covered.

- [x] Language support: JS, TS, HTML
- [x] Aurelia templating syntax:
  - [x] Property binding for custom elements (partial, only type-checks the property on usage side)
  - [x] Property binding for custom attributes (partial, only type-checks the property on usage side)
  - [x] Binding for template controllers:
    - [x] if/else
    - [x] switch
    - [x] promise
    - [x] repeat
      - [x] array
      - [x] set
      - [x] map
      - [x] range (`i of 10`)

### 🎫 Issues

Phase1 of #1971.

###  🙏 Request to the community

Feel free to try out the type-checking. I would appreciate any feedback. Also keep in mind that the DX around type-checking is not 100% there yet; the `experimental` prefix for the flag should have made that clear, I guess :) I will try to update the linked issue with the next steps and phases of implementation.


packages-tooling/plugin-conventions/src/template-typechecking.ts
```diff
@@ -0,0 +1,511 @@
...
+
+const identifierPrefix = '__Template_TypeCheck_Synthetic_';
+class TypeCheckingContext {
+  private static readonly _o = 'o';
+  private static readonly rootAccessScope = new AccessScopeExpression(this._o, 0);
+  private scope: IdentifierScope;
+  public readonly toReplace: { loc: Location; modifiedContent: () => string }[] = [];
+  public readonly classUnion: string;
+  public readonly accessTypeParts: ((acc: string) => string)[];
+  public readonly accessTypeIdentifier: string;
```
bigopon, line 142: well, does `accessTypeIdent` read weird?

packages-tooling/plugin-conventions/src/template-typechecking.ts
```diff
@@ -0,0 +1,511 @@
...
+  ) {
+    const classNames = classes.map(c => c.name);
+    this.classUnion = `(${classNames.join('|')})`;
+    this.accessTypeParts = [() => `${this.classUnion} & { $parent: any }`];
+    this.accessTypeIdentifier = `__Template_Type_${classNames.join('_')}__`;
+    this.accessIdent = `access${isJs ? '' : `<${this.accessTypeIdentifier}>`}`;
+    this.scope = new IdentifierScope('none', classes);
+  }
+
+  public produceTypeCheckedTemplate(rawHtml: string): string {
```
bigopon, line 170: From the look of the type checking tests, we don't have any thing asserting the emitted code.  I think we can at least have some preliminary tests for that, since it gives good clue what's happening when something fails, also to understand what we are doing top down better, rather than only bottom up at the moment.

packages-tooling/plugin-conventions/src/template-typechecking.ts
```diff
@@ -0,0 +1,511 @@
...
+
+const contextualRepeatProperties = ['$index', '$first', '$last', '$even', '$odd', '$length'] as readonly string[];
+function mutateAccessScope(expr: IsBindingBehavior, ctx: TypeCheckingContext, memberNameResolver: (member: string) => string, needsPath: true): [expr: IsAssign, path: string[]];
+function mutateAccessScope(expr: IsBindingBehavior, ctx: TypeCheckingContext, memberNameResolver: (member: string) => string): [expr: IsAssign, null];
+function mutateAccessScope(expr: IsBindingBehavior, ctx: TypeCheckingContext, memberNameResolver: (member: string) => string, needsPath: boolean = false): [expr: IsAssign, path: string[] | null] {
+  while (expr.$kind === 'ValueConverter' || expr.$kind === 'BindingBehavior') {
+    expr = expr.expression;
+  }
+
+  if ((expr.$kind === 'AccessScope' || expr.$kind === 'AccessMember') && contextualRepeatProperties.includes(expr.name)) {
```
bigopon, line 450: is the `expr.$kind` part for `$this.$index`? It seems to miss the condition that `expr.obj.$kind === 'AccessScope'`

packages-tooling/plugin-conventions/src/template-typechecking.ts
```diff
@@ -0,0 +1,511 @@
...
+  } else {
+    while (!isAccessGlobal(object) && (object.$kind === 'CallMember' || object.$kind === 'AccessMember' || object.$kind === 'AccessKeyed')) {
+      switch (object.$kind) {
+        case 'AccessMember': path?.push(object.name); break;
+        case 'AccessKeyed': {
+          if (object.key.$kind === 'AccessScope' && object.key.ancestor === 0) {
+            const member = object.key.name;
+            const overriddenMember = memberNameResolver(member);
+            (object as Writable<AccessKeyedExpression>).key = ctx.createMemberAccessExpression(overriddenMember);
+          }
```
bigopon, line 472: nice work on this part. `member` naming probably is fine, just not what I'm used to 😃

packages-tooling/plugin-conventions/src/template-typechecking.ts
```diff
@@ -0,0 +1,511 @@
...
+          break;
+        }
+        default: break;
+      }
+      object = object.object;
+      if (object.$kind === 'AccessScope' && object.ancestor === 0) {
```
bigopon, line 478: we should have a comment to handle `$parent....` cases, along with `$parent.$parent...` etc..

packages-tooling/plugin-conventions/src/template-typechecking.ts
```diff
@@ -0,0 +1,511 @@
...
+        const prop = (object as Writable<AccessScopeExpression>).name = memberNameResolver(member);
+        path?.push(prop);
+        break;
+      }
+    }
+  }
+  return [expr, path?.reverse() ?? null];
+}
+
+function isAccessGlobal(ast: IsLeftHandSide): boolean {
```
bigopon, line 489: we also have `CallGlobal`, but it's fine to handle just `AccessGlobal` for now.

bigopon (approved):
Nice work @Sayan751 🙏 , even though I have comments, I think the PR can go in as it is, in case you want to get it out for testing purposes.

For the tests, I think we should have some code emitting check to get some sort of "visualisation".

Also, we likely need an issue to track what's supported, and what's not, like the description of this PR, but with more scenarios and checkboxes, either checked or non.

bigopon:
This will be out as a dev version, if you are interested, please help test it out 😺

packages-tooling/plugin-conventions/src/preprocess-resource.ts
```diff
@@ -344,18 +526,43 @@ function createAuResourceTransformer(): TransformerFactory<SourceFile> {
...
+    const dArgs = decorator.arguments;
+    const numArguments = dArgs.length;
+
+    // map the classes to the template imports to classes - @customElement decorator
+    if (resourceType.type === 'customElement') {
+      if (numArguments === 1) tryCollectTemplateMetadataFromDefinition(dArgs[0], $class, templateMetadata);
+      for (const m of node.members) {
+        // static template property
+        tryCollectTemplateMetadataFromStaticTemplate(m, $class, templateMetadata);
+      }
```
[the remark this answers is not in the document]
Sayan751: As far as I understand, both decorator and static properties are in nature "static". Hence, it might be alright. Note that the code does not terminate after discovering the static properties; it is just collecting information.

packages-tooling/plugin-conventions/src/preprocess.ts
```diff
@@ -3,67 +3,59 @@ import { ModifyCodeResult } from 'modify-code';
...
+  // for testing
+  _fileExists = fileExists,
+  _readFile = readFile,
 ): ModifyCodeResult | undefined {
   const ext = path.extname(unit.path);
   const basename = path.basename(unit.path, ext);
   const allOptions = preprocessOptions(options);
+  const templateExtensions = allOptions.templateExtensions;
+  const useProcessedFilePairFilename = allOptions.useProcessedFilePairFilename;
+  unit.readFile = (path) => _readFile(unit, path);
```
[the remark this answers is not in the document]
Sayan751: I think that we don't need that now. If needed, we can always extend it later.

packages-tooling/plugin-conventions/src/template-typechecking.ts
```diff
@@ -0,0 +1,511 @@
...
+
+  traverse(tree, (node) => processNode(node, ctx));
+  return ctx.produceTypeCheckedTemplate(rawHtml);
+}
+
+function traverseDepth($node: DefaultTreeElement, ctx: TypeCheckingContext): void {
+  // drill down
+  if ($node.childNodes) traverse($node, n => processNode(n, ctx));
+  // For <template> tag
+  if (($node as Template).content?.childNodes) traverse(($node as Template).content, n => processNode(n, ctx));
```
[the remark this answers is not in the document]
Sayan751: Agreed. The problem is that currently the tooling infra is not able to detect resource types 100% correctly, especially if those are coming from the node_modules or being imported from other files. That needs to be extended first.

packages-tooling/plugin-conventions/src/template-typechecking.ts
Sayan751: I prefer `Identifier` over `Ident`. Probably rename the existing ones. `Ident` sometimes look very close to `Indent` 😄

packages-tooling/plugin-conventions/src/template-typechecking.ts
Sayan751: I also thought about that. Decided against that at the end. IMO such tests are difficult to maintain. Once something in the code generation process changes, we need to update all the tests. There is certainly value addition in having such tests. For the phase1 I went for directly asserting the compilation success and failure. Once we are close to marking the feature non-experimental, we can add such tests. I have added a TODO item in the linked issue.

packages-tooling/plugin-conventions/src/template-typechecking.ts
Sayan751: Added tests. Seems to be working.

PR #2068, opened 2024-11-21
feat(expression-parser): NewExpression

fkleuver:
# Pull Request

## 📖 Description

This PR implements the `NewExpression` syntax. This allows you to instantiate global types:

```html
<template>
  <input value.bind="email" type="email" placeholder="Enter email">
  <p if.bind="new RegExp('^\\S+@\\S+\\.\\S+$').test(email)">Valid Email Address</p>
</template>
```

And types that are defined in-scope

```ts
class Foo {}

export class MyElement {
  Foo = Foo;
}
```

```html
<child-element prop.bind="new Foo()"></child-element>
```


bigopon (approved):
nice work @fkleuver

PR #2069, opened 2024-11-29
ADD usage section of web components

TomasKulhanek:
# Pull Request

## 📖 Description

This PR just adds some documentation howto use web components. As it took me some time to investigate current gitbook and source codes. This may be updated using insights from core aurelia developers. However, I've tested the documented configuration on real project.


bigopon (approved):
👍 nice, thanks @TomasKulhanek

PR #2077, opened 2024-12-29
chore(tooling/type-checking): handling TS private members and method calls

Sayan751:
# Pull Request

## 📖 Description

In this iteration, the following supports are added to the type-checking:
- Handling TS private properties #2115.
- Handling method calls.
- Added tests for method overloads, and union typed properties and method arguments.

### DIFF
--- a/packages-tooling/__tests__/src/plugin-conventions/type-checking/without-convention.basic.spec.ts
+++ b/packages-tooling/__tests__/src/plugin-conventions/type-checking/without-convention.basic.spec.ts
@@ -2116,8 +2116,8 @@ ${prop('prop', 'string', isTs)}
       assertSuccess(entry, result.code);
     });
 
-    describe.only('union typed', function () {
-      it(`property - pass - language: ${lang}`, function () {
+    describe(`union typed - language: ${lang}`, function () {
+      it('property - pass', function () {
         const entry = `entry.${extn}`;
         const markupFile = 'entry.html';
         const markup = '${prop.toString()}';
@@ -2139,7 +2139,7 @@ ${prop('prop', 'string', isTs)}
         assertSuccess(entry, result.code);
       });
 
-      it(`property - fail - language: ${lang}`, function () {
+      it('property - fail', function () {
         const entry = `entry.${extn}`;
         const markupFile = 'entry.html';
         const markup = '${prop.toExponential(2)}';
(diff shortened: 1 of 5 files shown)


packages-tooling/__tests__/src/plugin-conventions/type-checking/without-convention.basic.spec.ts
```diff
@@ -2115,5 +2115,473 @@ ${prop('prop', 'string', isTs)}
...
+                      `,
+            readFile: createMarkupReader(markupFile, markup),
+          }, nonConventionalOptions);
+
+        assertFailure(entry, result.code, [/Property 'toExponential' does not exist on type 'string'/, /Property 'toUpperCase' does not exist on type 'number'/]);
+      });
+    });
+
+    describe('access modifier', function () {
+      for (const am of ['public', 'protected', 'private'] as const) {
```
bigopon, line 2444: nit
```suggestion
      for (const modifier of ['public', 'protected', 'private'] as const) {
```
😃 there's a lot of code in these files, not even having to jump once is a lot better.

packages-tooling/plugin-conventions/src/preprocess-resource.ts
```diff
@@ -363,13 +384,48 @@ function captureTemplateImport(s: Statement, code: string): ITemplateMetadata |
...
+        ? 'private'
+        : flags & ModifierFlags.Protected
+          ? 'protected'
+          : 'public'
+        ;
+      acc.push({
+        name: name.escapedText.toString(),
+        accessModifier: accessModifier!,
+        memberType,
+        dataType: ((m as PropertyDeclaration | MethodDeclaration | GetAccessorDeclaration).type ?? getJSDocType(m))?.getText(sf) ?? 'any',
```
bigopon, line 416: 👍

packages-tooling/plugin-conventions/src/template-typechecking.ts
```diff
@@ -160,11 +160,38 @@ class TypeCheckingContext {
...
     const classNames = classes.map(c => c.name);
-    this.classUnion = `(${classNames.join('|')})`;
+
+    // create a class union
+    // - Omit the private and protected members
+    // - Add them back to the union type
+    // - Then create union
+    const classUnion = [];
+    for (const $class of classes) {
+      // Omit<CLASS, 'private' | 'protected'> & { PRIVATE: TYPE, PROTECTED: TYPE }
```
bigopon, line 170: interesting solution, if someone wanted to ensure that they don't accidentally use private member, this kind of will null that, maybe it can be configurable later though.

bigopon (approved):
nice work @Sayan751

packages-tooling/__tests__/src/plugin-conventions/type-checking/without-convention.basic.spec.ts
Sayan751: Changes done.

PR #2083, opened 2025-01-05
chore: used LTS node 22 in CI

Sayan751:
# Pull Request

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -63,7 +63,7 @@ executors:
     parameters:
       node:
         type: string
-        default: "18.19.0"
+        default: "22.12.0"
     working_directory: *working_dir
     resource_class: large
     docker:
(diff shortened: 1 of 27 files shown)


.circleci/config.yml
```diff
@@ -63,7 +63,7 @@ executors:
     parameters:
       node:
         type: string
-        default: "18.19.0"
+        default: "22.12.0"
```
bigopon, line 66: Sometimes it feels like we can just go with the most minimal possible, but maybe not with node version 😃

bigopon (approved):

.circleci/config.yml
Sayan751: As it is just for CI, and not for the packages (engines), it is probably ok. I am going ahead and merging this.

Issue #2078, opened 2024-12-30
repeat.for uses the same custom element instance when the same object is added to array twice


euglv:
# 🐛 Bug Report

Looks like, when you add the same object to the array twice, `repeat.for` does not creates new custom element instance, but takes existing instance from cache. This leads to the wrong `$index` value and other issues.

Consider the following `my-app.html`:
```html
<template>
  <template as-custom-element="array-item">
    <au-slot></au-slot>
  </template>

  <array-item repeat.for="item of items"><div>Item index: ${$index}</div></array-item>
  <br><br>
  <button click.trigger="test()">Test</button>
</template>
```
`my-app.ts`:
```ts
export class MyApp {
  items = [{}, {}]
  test(){
    this.items.unshift(this.items[1]);
  }
}

```

Repro link:
https://stackblitz.com/edit/au2-hello-world-example-5hsujd5a?file=src%2Fmy-app.ts
## 🤔 Expected Behavior

`repeat.for` should not take from cache custom element instances that are attached to DOM.

## 😯 Current Behavior

If you press `Test` button in the above example, the following results are produced: 
![image](https://github.com/user-attachments/assets/6f62f396-0d02-4906-8269-73162808534e)

bigopon:
Interesting, thanks for the issue @euglv , cc @fkleuver

fkleuver:
It's fixed now. Thanks for the repro!

PR #2084, opened 2025-01-06
Fix repeater scope reuse issue

fkleuver:
# Pull Request

## 📖 Description

Fixes an issue in the repeater where the scope cache can get misaligned in indexed mode with duplicate objects

### 🎫 Issues

https://github.com/aurelia/aurelia/issues/2078

## 👩‍💻 Reviewer Notes

The `getScope` method and the way it was storing scopes associated with duplicate items in the scopeMap is not reliable when existing items are moved around. In that case the indexMap is now used to restore existing scopes from the correct source location instead of a naive FIFO item-based identity lookup.

### DIFF
--- a/packages/__tests__/src/3-runtime-html/repeat.duplicates.spec.ts
+++ b/packages/__tests__/src/3-runtime-html/repeat.duplicates.spec.ts
@@ -102,19 +102,93 @@ describe("3-runtime-html/repeat.duplicates.spec.ts", function () {
     });
 
     it('object duplication unshift (issue 2078)', function () {
-      const objA = { toString() { return 'a'; } };
-      const objB = { toString() { return 'b'; } };
+      const objA = { toString() { return 'A'; } };
+      const objB = { toString() { return 'B'; } };
 
       const { assertText, component, flush } = createFixture(
         `<div repeat.for="i of items">\${$index}-\${i} </div>`,
         class { items = [objA, objB]; }
       );
-      assertText('0-a 1-b ');
+      assertText('0-A 1-B ');
 
       component.items.unshift(objB);
       flush();
 
-      assertText('0-b 1-a 2-b ');
+      assertText('0-B 1-A 2-B ');
+    });
+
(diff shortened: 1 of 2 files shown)


packages/runtime-html/src/resources/template-controllers/repeat.ts
```diff
@@ -340,8 +340,22 @@ export class Repeat<C extends Collection = unknown[]> implements ICustomAttribut
     const local = this.local;
     const hasDestructuredLocal = this._hasDestructuredLocal;
 
-    for (let i = 0; i < len; ++i) {
-      scopes[i] = getScope(oldScopeMap, newScopeMap, items[i], forOf, parentScope, binding, local, hasDestructuredLocal);
+    if (indexMap === void 0) {
```
bigopon, line 343: we dont need to make a distinction between null/undefined in our private code. It should only be considered in a public API. From what I see when changing null/undefined specific handling to nullish, it's simpler and there's no downside.

packages/__tests__/src/3-runtime-html/repeat.duplicates.spec.ts
```diff
@@ -100,5 +100,95 @@ describe("3-runtime-html/repeat.duplicates.spec.ts", function () {
 
       assertText('0-a 1-b 2-c ');
     });
+
+    it('object duplication unshift (issue 2078)', function () {
```
bigopon, line 104: nit: normally gh describes its links as `#___`, maybe we should have something similar, so searching will be easier
```suggestion
    it('object duplication unshift (issue #2078)', function () {
```

bigopon (approved):
nice work @fkleuver

packages/runtime-html/src/resources/template-controllers/repeat.ts
fkleuver: Mostly agree. There were specific reasons for it in various places in the private code but this is not one of those places at least

Issue #2087, opened 2025-01-07
Au2 repeat.for keyed iteration is completely broken


haltesh:
# 🐛 Bug Report

Sorry for the potentially inflammatory title, but based on my understanding of the use case for keyed iteration (or `matcher` in Au1), I believe the way it currently works in Aurelia2 is essentially pointless.

My understanding is that  if a key (or `matcher`) is unspecified, `repeat.for`, uses strict equality to compare the items. This means that if your items are mutable and will never be re-assigned in the app's lifespan, **there is no reason to use keys**. The reason you'd want to use keys is if your items are immutable, or may be re-assigned from an external source.

## 😯 Current Behavior

Right now, if you re-assign the array backing a keyed repeat.for, any item that was previously bound will reuse it's previous backing model. This means the only safe way to use a keyed repeat.for is when the items are mutable and will never be re-assigned, which is the exact scenario in which you do not need to use keys in the first place.

Repro link: https://stackblitz.com/edit/au2-repeat-for-keyed-mnw5drd9?file=src%2Fmy-app.html

## 🔦 Context

I'm trying to port our front-end code from Aurelia1 to Aurelia2, this along with [#2035](https://github.com/aurelia/aurelia/issues/2035) are the biggest paint points right now.

fkleuver:
To the best of my understanding, this is behaving as it should, and in your use case, you should **not** be using keyed mode. 

## What "keyed" really means

**Keyed** means “treat items with the same key as the same identity” and preserve their existing DOM/component instance.

**Non-keyed** (in general, more on this below) means “treat items with the same index as the same identity,” which usually translates to a naive create/destroy approach when the array changes.

Put differently: if the key is the same, the item is the same. When you provide `key.bind: item.key`, Aurelia2 will interpret: "Whenever I see an item whose `.key` equals X, I’ll continue to use the same `<tr>` (and the same binding context) that belonged to `X` previously."

Hence if you reassign `items` with new objects **but using the same `item.key` string (e.g. `"item1"`, `"item2"`)**, Aurelia2 sees that the “logical identity” is still `"item1"` and `"item2"`. Since the identity has not changed, Aurelia2 holds onto the old row references (and old data binding contexts), instead of replacing them with the newly created objects.

From `Aurelia2’s` perspective, that’s the entire *purpose* of keyed iteration: stable identities must stay stable. This is precisely what React, Vue, and others do as well.

## Common misunderstanding of keyed mode

[...]

PR #2088, opened 2025-01-07
fix(repeat): change to key-based scope caching in keyed mode

fkleuver:
# Pull Request

## 📖 Description

The scope caching only relied on item identity. This was an oversight. Now it uses the keys in keyed mode and properly retains the scope / binding context when you re-assign a new array or new items with the same keys.

### 🎫 Issues

Fixes https://github.com/aurelia/aurelia/issues/2087

## 👩‍💻 Reviewer Notes

I experimented with a few different methods before settling on full key evaluation. It seems there is no other way to guarantee correct up-to-date keys in all scenarios (including the combination of expression-based keys with destructured locals) other than pre-evaluating them all. This may have performance impact. Therefore it may be a good idea to explicitly point out in the docs that using string-based keys is preferable if possible

## 📑 Test Plan

I added the test case from https://github.com/aurelia/aurelia/issues/2087 as well as a few extra. Also added a test to confirm input focus is retained (this was not verified before)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -684,6 +684,8 @@ jobs:
             - benchmarks/local/dist/app.local.js
             - benchmarks/local/dist/app-repeat-ce.local.js
             - benchmarks/local/dist/app-big-template.local.js
+            - benchmarks/app-repeat-keyed-string/dist
+            - benchmarks/app-repeat-keyed-expr/dist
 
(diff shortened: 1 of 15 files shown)


bigopon:
This looks good enough for now, we'll improve more after we have dealt with the async binding. Thanks @fkleuver , nice work.

PR #2090, opened 2025-01-11
fix(hmr): disposed controllers to avoid memory leak issues

Sayan751:
# Pull Request

### 🎫 Issues

The issue is reported in [discourse](https://discourse.aurelia.io/t/nested-dynamic-custom-element-composition/4231/19?u=sayan751).

The issue was caused by the HMR code. The controller instances were kept around, but were not cleaned up when the components were disposed (in a routing situation, this might mean navigating to a different route). This PR fixes that cleanup.


packages-tooling/vite-plugin/src/index.ts
```diff
@@ -164,15 +164,23 @@ if (${moduleText}.hot) {
...
       // @ts-ignore
-      ogCreated && ogCreated.call(this, controller);
+      $created?.call(this, controller);
       controllers.push(controller);
     }
+    // @ts-ignore
+    proto.dispose = function() {
+      // @ts-ignore
+      $dispose?.call(this);
+      controllers.length = 0;
```
bigopon, line 182: shouldn't we only remove the this controller?

packages/__e2e__/13-hmr-convention-routing-memory-leak/playwright/app.spec.ts
```diff
@@ -0,0 +1,45 @@
+import { test, expect } from '@playwright/test';
+import * as fs from 'fs';
+import * as path from 'path';
+import { parseSnapshot } from 'v8-heapsnapshot';
+
+test.describe.serial('examples/hmr-webpack-e2e/app.spec.ts', function () {
+
+  test.only('hmr does not cause memory leak', async function ({ page, baseURL }) {
```
bigopon, line 8: `.only` ❌

packages/__e2e__/13-hmr-convention-routing-memory-leak/playwright/app.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+    expect(v8Snapshot.nodes.filter(n => n.toString().includes('<p-1>')).length).toBeGreaterThan(0);
+    expect(v8Snapshot.nodes.filter(n => n.toString().includes('<p-2>')).length).toBe(0);
+
+    // navigate to p-2
+    await page.click('text=Page 2');
+    await page.locator('text=p-2').waitFor({ state: 'visible' });
+    snapshot = await takeSnapshot('post-navigation.heapsnapshot');
+    v8Snapshot = await parseSnapshot(snapshot);
+    expect(v8Snapshot.nodes.filter(n => n.toString().includes('<p-2>')).length).toBeGreaterThan(0);
+    expect(v8Snapshot.nodes.filter(n => n.toString().includes('<p-1>')).length).toBe(0);
```
bigopon, line 26: maybe this is enough. At least it'll serve as a base assertion for future improvements.

bigopon (approved):

packages-tooling/vite-plugin/src/index.ts
Sayan751: As per my understanding, the `controllers` made the problem and hold a reference to a controller, that is no longer needed. What you said is also intuitive, and also was my first intuition. However, the `dispose` gets no `controller`, and the `this.$controller` should have been disposed anyway already by Au2 pipeline. All we need to do here is to kill the dangling reference to the controller. 

On a side note, after reviewing the code, it seems to me that there is only one instance of controller in the `controllers` array. I might be wrong, but I could not find a way to simulate the scenario, where multiple controllers get pushed to that array.

Hope this argumentation makes sense.

packages/__e2e__/13-hmr-convention-routing-memory-leak/playwright/app.spec.ts
Sayan751: I think so as well. However, the project is not yet added to the pipeline. If we can finally find a way to deterministically identify the detached node, we can revisit this code and add it to the pipeline. For now, the fix can be verified by manually starting the app.

packages/__e2e__/13-hmr-convention-routing-memory-leak/playwright/app.spec.ts
Sayan751: Thanks! Removed.

PR #2092, opened 2025-01-13
docs(building-plugins): rewrite building plugin docs

Vheissu:
The existing docs were focusing too heavily on Monorepo and not on how to create plugins are their basic core. This makes the docs minimal first and then adds advanced later on.

### DIFF
--- a/docs/user-docs/developer-guides/building-plugins.md
+++ b/docs/user-docs/developer-guides/building-plugins.md
@@ -1,421 +1,269 @@
 ---
 description: >-
-  Aurelia makes it easy to create your own plugins. Learn how you can create
-  individual plugins, register them and work with tasks to run code at certain
+  Aurelia makes it easy to create your plugins. Learn how to create
+  individual plugins, register them, and work with tasks to run code during certain
   parts of the lifecycle process.
 ---
 
-# Building plugins
+# Building Plugins
 
-One of the most important needs of users is to design custom plugins. In the following, we want to get acquainted with how to design a plugin in the form of a mono-repository structure with configuration.
+Aurelia plugins allow you to encapsulate functionality that can be reused across multiple applications. They can include custom elements, value converters, and other resources. The goal is to create packaged, easily shared, ready-to-use functionalities that integrate seamlessly with Aurelia applications.
 
-### What is a mono-repository?
+## Minimal Plugin Example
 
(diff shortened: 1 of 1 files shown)


docs/user-docs/developer-guides/building-plugins.md
```diff
@@ -1,421 +1,269 @@
...
-// BootstrapV5Options.ts
+## Adding Components
 
-import { Size } from "./Size";
+You often want to add custom components to make your plugin more useful.
 
-export interface IBootstrapV5Options {
-    defaultSize?: Size;
+```typescript
+// hello-world.ts
```
bigopon, line 49: maybe we can use the multitab UI component from gitbook?

docs/user-docs/developer-guides/building-plugins.md
```diff
@@ -1,421 +1,269 @@
...
-export * from './src';
+```typescript
+// my-configurable-plugin.ts
+import { DI, IContainer, Registration } from '@aurelia/kernel';
+
+export const IMyPluginOptions = DI.createInterface<MyPluginOptions>('IMyPluginOptions');
+
+export const MyConfigurablePlugin = {
+  configure(options: Partial<MyPluginOptions> = {}) {
+    const finalOptions = { ...defaultOptions, ...options };
```
bigopon, line 98: maybe at least have a simple `defaultOptions` object somewhere so that it's more complete?
```ts
const defaultOptions = {
  // your default options here
}
```

docs/user-docs/developer-guides/building-plugins.md
```diff
@@ -1,421 +1,269 @@
...
+import { IMyPluginOptions } from './my-configurable-plugin';
+
+@customElement({
+  name: 'greeting',
+  template: '<div>${options.greeting}, ${name}!</div>'
+})
+export class Greeting {
+  name = 'World';
+  
+  constructor(@inject(IMyPluginOptions) private options: MyPluginOptions) {}
```
bigopon, line 143: we are no longer supporting parameter injection, let's use `resolve` instead
```ts
import { resolve } from 'aurelia'

...
private options = resolve(IMyPluginOptions);
```

bigopon (commented):
nice work @Vheissu , a few points we should correct, but besides looks a lot better 👍

PR #2094, opened 2025-01-14
feat: ported noView and inlineView decorators

### LINKED ISSUE #2085: Implement @noView and @inlineView like au1
# 🙋 Feature Request

The title says it all. Just for clarity, it would be nice to implement `@noView` and `@inlineView` decorators like au1, for easier migration.

```ts
// component-without-view.js
@noView()
export class ComponentWithoutView {}

// small-component.js
@inlineView(`foo`)
export class SmallComponent {}
```

## 😯 Current Behavior

Repro link:

Sayan751:
# Pull Request

### 🎫 Issues

Closes #2085.

### DIFF
--- a/packages/__e2e__/3-hmr-webpack/src/components/no-view.ts
+++ b/packages/__e2e__/3-hmr-webpack/src/components/no-view.ts
@@ -1,0 +1,10 @@
+import { noView } from '@aurelia/compat-v1';
+import { INode, resolve } from 'aurelia';
+
+@noView
+export class NoView {
+  private readonly node: HTMLElement = resolve(INode) as HTMLElement;
+  public attached() {
+    this.node.textContent = 'Nothing to see here';
+  }
+}
--- a/packages/__e2e__/3-hmr-webpack/src/components/no-view.ts
+++ b/packages/__e2e__/3-hmr-webpack/src/components/no-view.ts
@@ -1,10 +1,4 @@
 import { noView } from '@aurelia/compat-v1';
-import { INode, resolve } from 'aurelia';
 
 @noView
-export class NoView {
-  private readonly node: HTMLElement = resolve(INode) as HTMLElement;
-  public attached() {
-    this.node.textContent = 'Nothing to see here';
-  }
-}
+export class NoView { }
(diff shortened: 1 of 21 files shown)


bigopon (commented):
Thanks @Sayan751 , nice work. A concern is we no longer employ view as a terminology, so this is more about migration. Also the 2 decorators share the same underlying targeted property, so I think we can have something like this
- `@template` decorator in runtime
- `@noView` and `@inlineView` that use `@template`, in the `compat-v1` package

Sayan751:
That sounds good @bigopon! Will make the change. I am wondering if adding the `@template` deco in core is an overkill. Probably, we should add the other 2 decorators only to the compat package.

Sayan751:
@bigopon it is done now.

packages/compat-v1/src/compat-custom-element.ts
```diff
@@ -0,0 +1,52 @@
...
+  };
+}
+
+const elementTypeName = 'custom-element';
+const elementBaseName = /*@__PURE__*/getResourceKeyFor(elementTypeName);
+const annotateElementMetadata = <K extends keyof PartialCustomElementDefinition>(Type: Constructable, prop: K, value: PartialCustomElementDefinition[K]): void => {
+  defineMetadata(value, Type, getAnnotationKeyFor(prop));
+};
+
+/** Manipulates the `template` property of the custom element definition for the type, when present else it annotates the type. */
```
bigopon, line 44: Just repeating some potential improvement: this is kind of a reason why we should try to make the definition recognition lazier. But this should also be good enough, since its user code with clear enough intention.

bigopon (approved):

Sayan751:
Missed the conventional support.

packages/__e2e__/3-hmr-webpack/src/components/no-view.ts
```diff
@@ -0,0 +1,10 @@
+import { noView } from '@aurelia/compat-v1';
+import { INode, resolve } from 'aurelia';
+
+@noView
+export class NoView {
+  private readonly node: HTMLElement = resolve(INode) as HTMLElement;
+  public attached() {
+    this.node.textContent = 'Nothing to see here';
```
bigopon, line 8: if we do this, it's more likely that we run into false positive, probably can just leave alone and assert `no-view` html is empty instead?

bigopon (approved):

packages/__e2e__/3-hmr-webpack/src/components/no-view.ts
Sayan751: Done.

PR #2099, opened 2025-01-20
chore(event-aggregator): order of subscriber invocation

Sayan751:
# Pull Request

## 📖 Description

This PR changes the order of subscriber invocations. 

As of now, the subscribers were invoked in the reverse order of subscription/registration, when `EventAggregator.publish(EVENT, [EVENT_DATA])` is called. Generally, the order of the invocations for the subscribers is not mandatorily needs to be the same as the order of subscription/registration. However, it is more or less a fair expectation. 

The order of invocation can play a crucial role, when some subscription is made using `AppTask.xxx`, and then later a CE also subscribes to the same channel. It might be a fair assumption, the subscription handler made via the `AppTask` is executed before the subscription handler of the CE.

This PR corrects that.

### DIFF
--- a/packages/kernel/src/eventaggregator.ts
+++ b/packages/kernel/src/eventaggregator.ts
@@ -57,17 +57,17 @@ export class EventAggregator {
       let subscribers = this.eventLookup[channelOrInstance];
       if (subscribers !== void 0) {
         subscribers = subscribers.slice();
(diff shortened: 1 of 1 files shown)


Sayan751:
@bigopon I am going ahead and merging this. In case you note something funny, let me know.

PR #2113, opened 2025-02-05
fix(validation): for dynamically accessed fields

### LINKED ISSUE #1623: Validation triggering not working on dynamically accessed props
# 🐛 Bug Report

image

When using dynamic property accessor with the validate attribute it does not validate onchange and only errors on submit when the rules are invoked manually.   This must be due to the fact the validate tries to use the property name when looking for rules to run.

Repro link https://stackblitz.com/edit/au2-conventions?file=src%2Fmy-app.html

## 🌍 Your Environment

| Aurelia          |
| Language         |
| Browser          |
| Bundler          |
| Operating System |
| NPM/Node/Yarn    |

Sayan751:
# Pull Request

### 🎫 Issues

Closes #1623

### DIFF
--- a/packages/__tests__/src/validation-html/validate-binding-behavior.spec.ts
+++ b/packages/__tests__/src/validation-html/validate-binding-behavior.spec.ts
@@ -1716,9 +1716,9 @@ describe('validation-html/validate-binding-behavior.spec.ts', function () {
       await stop(true);
     });
 
-    it.only('works for conditionally accessed fields', async function () {
+    it('works for conditionally accessed fields', async function () {
       class MyApp {
-        private person: Person = new Person((void 0)!, (void 0)!);
+        public readonly person: Person = new Person((void 0)!, (void 0)!);
         public readonly validationController: IValidationController = resolve(newInstanceForScope(IValidationController));
         private readonly fields: (keyof Person)[] = ['name', 'age'];
         public constructor()  {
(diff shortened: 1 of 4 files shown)


packages/validation/src/validator.ts
```diff
@@ -50,14 +50,19 @@ export interface IValidator {
...
+    let propertyName = instruction.propertyName;
     const propertyTag = instruction.propertyTag;
 
     const rules = instruction.rules ?? validationRulesRegistrar.get(object, instruction.objectTag) ?? [];
     const scope = Scope.create({ [rootObjectSymbol]: object });
 
     if (propertyName !== void 0) {
-      return (await rules.find((r) => r.property.name === propertyName)?.validate(object, propertyTag, scope)) ?? [];
+      let rule = rules.find((r) => r.property.name === propertyName);
+      if (rule == null && typeof propertyName === 'string' && propertyName.startsWith('[') && propertyName.endsWith(']')) {
```
bigopon, line 61: does this work with 2 level keyed access?

bigopon (approved):
nice work @Sayan751

packages/validation/src/validator.ts
Sayan751: @bigopon Great catch! I have fixed that. Thanks!

packages/__tests__/src/validation-html/validate-binding-behavior.spec.ts
```diff
@@ -1758,20 +1763,21 @@ describe('validation-html/validate-binding-behavior.spec.ts', function () {
...
       // round #3
       component.person.name = 'foo';
       component.person.age = 18;
+      component.person.address = { pin: 123, city: 'foo', line1: 'foo' };
       await domQueue.yield();
 
       validationResult = await validationController.validate();
       assert.strictEqual(validationResult.valid, true, 'validationResult.valid 3');
       assert.deepStrictEqual(
         validationResult.results.map(x => [x.propertyName, x.valid, x.message]),
```
bigopon, line 1779: it's a bit confusing when we don't have the original form of the expression, since keyed is quite different with member access. Just a note in case we want to add a display name in the future.

bigopon (approved):
👍 nice fix.

Issue #2063, opened 2024-10-07
Vite build watch cannot handle html imports


MaximBalaganskiy:
# 🐛 Bug Report

`npx makes aurelia`
`vite build --watch`

## 🤔 Expected Behavior

The app should be built

## 😯 Current Behavior

Vite errors out
```log
[vite:build-html] Unable to parse HTML; parse5 error code unexpected-character-in-unquoted-attribute-value
```

## 💁 Possible Solution

Modify Aurelia plugin to import html as `.html?raw`, which will load templates as raw text

## 🔦 Context

Consider an asp.net api which serves gui as well to avoid dealing with JWT auth and rely on cookies only.
Vite watch build replaces vite dev run in this scenario.

Vheissu:
Gotta fix for this in a PR here which in my testing seems to resolve the issue: https://github.com/aurelia/aurelia/pull/2124

MaximBalaganskiy:
@Sayan751 this is still relevant in rc1

bigopon:
I'm not having this issue
```
npx vite build --watch
You are using Node.js 20.14.0. Vite requires Node.js version 20.19+ or 22.12+. Please upgrade your Node.js version.
vite v7.3.3 building client environment for production...

watching for file changes...

build started...
✓ 14 modules transformed.
dist/index.html                  0.30 kB │ gzip:  0.23 kB
dist/assets/index-DC-JZFZd.js  252.83 kB │ gzip: 77.95 kB
built in 7701ms.
```
my local output for default esnext.

bigopon:
@MaximBalaganskiy  pls help with a more reproducible steps, including the os/environment

MaximBalaganskiy:
Not relevant anymore, I've switched to a vite+aurelia-plugin based builds and proxying when `vite start`

Issue #1998, opened 2024-06-26
Shadow DOM documentation out of date


ivanbacher:
At the moment the `Webpack Configuration for Shadow DOM` only has this:

```
{
  test: /[/\\]src[/\\].+\.html$/i,
  use: {
    loader: '@aurelia/webpack-loader',
    options: {
      defaultShadowOptions: { mode: 'open' }
    }
  },
  exclude: /node_modules/
}
```

However, when bootstrapping with `npx makes aurelia` and choosing the Shadow DOM option. All of the below is added to the webpack config. Should this also be added to the documentation?

```
rules: [
        { test: /\.(png|svg|jpg|jpeg|gif)$/i, type: 'asset' },
        { test: /\.(woff|woff2|ttf|eot|svg|otf)(\?v=[0-9]\.[0-9]\.[0-9])?$/i,  type: 'asset' },
        {
          test: /\.css$/i,
          // For style loaded in src/main.js, it's not loaded by style-loader.
          // It's for shared styles for shadow-dom only.
          issuer: /[/\\]src[/\\]main\.(js|ts)$/,
          use: [ cssLoader, postcssLoader ]
        },
        {
          test: /\.scss$/i,
          // For style loaded in src/main.js, it's not loaded by style-loader.
          // It's for shared styles for shadow-dom only.
          issuer: /[/\\]src[/\\]main\.(js|ts)$/,
          use: [ cssLoader, postcssLoader, sassLoader ]
        },
        {
          test: /\.css$/i,
          // For style loaded in other js/ts files, it's loaded by style-loader.
          // They are directly injected to HTML head.
          issuer: /(?<![/\\]src[/\\]main)\.(js|ts)$/,
          use: [ 'style-loader', cssLoader, postcssLoader ]
        },
        {
          test: /\.scss$/i,
          // For style loaded in other js/ts files, it's loaded by style-loader.
          // They are directly injected to HTML head.
          issuer: /(?<![/\\]src[/\\]main)\.(js|ts)$/,
          use: [ 'style-loader', cssLoader, postcssLoader, sassLoader ]
        },
        {
          test: /\.css$/i,
          // For style loaded in html files, Aurelia will handle it.
          issuer: /\.html$/,
          use: [ cssLoader, postcssLoader ]
        },
        {
          test: /\.scss$/i,
          // For style loaded in html files, Aurelia will handle it.
          issuer: /\.html$/,
          use: [ cssLoader, postcssLoader, sassLoader ]
        },
[...]

PR #2124, opened 2025-02-21
feat(bundlers): fix issue with html template imports in vite apps

Vheissu:
Fixes issue #2063 around `.html` imports in Vite apps erroring when being attempted to be used for non-conventions component definitions where the HTML file is imported and provided to the `template` property.

Oh, yeah, removed some unused docs files too and fixes a longstanding docs issue #1998 around shadow dom.


packages-tooling/vite-plugin/src/index.ts
```diff
@@ -72,21 +72,33 @@ export default function au(options: {
...
-        getHmrCode,
-        transformHtmlImportSpecifier: (s) => {
-          return this.meta.watchMode
-            ? s
-            : s.replace(/\.html$/, '.$au.ts');
+      // If we are in a .ts or .js file, let's rewrite .html imports to .html?raw
+      if (/\.[cm]?[tj]s$/.test(id)) {
+        // Simple check for lines with .html and no existing query (e.g. ?someFlag).
+        code = code.replace(
+          /(\.html)(?!\?)(['"])/g,
```
mroeling, line 79: You could consider using non-capturing group(s), especially for the quote style. Faster, and prevents the need to use $2 in the replace string.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Regular_expressions/Non-capturing_group

But nice to see the solution is rather simple :)

3cp: As I expected, the `import template from "./foo.html"` actually works with current version of vite plugin, but only in dev mode. the "./foo.html?raw" works in both.

3cp: In vite prod mode, it tries to parse html module as html, but our html module is code.

```
[vite:build-html] Unable to parse HTML; parse5 error code unexpected-character-in-unquoted-attribute-value
 at /Users/huocp/playground/my-app/src/my-app2.html:3:38
1  |  import { CustomElement } from '@aurelia/runtime-html';
2  |  export const name = "my-app2";
3  |  export const template = "<div class=\"message\">TWO ${message}</div>\n";
   |                                       ^
4  |  export default template;
5  |  export const dependencies = [  ];
file: /Users/huocp/playground/my-app/src/my-app2.html
```

3cp: Oh, the manual import of `.$au.ts` actually works in both dev and prod mode. Only editor complains, vite just runs.

```ts
import { customElement } from "aurelia";
import template from "./my-app2.$au.ts";

@customElement({
  name: 'my-app',
  template
})
export class MyApp {
  public message = 'Hello World2!';
}
```

bigopon: I introduced `.$au.ts` as a way to reuse the same code for both dev and production build. Without that it was not possible, at least at the time. Maybe we can try to split the handling of the code, just need to detect what mode vite is running in. It may or may not be complicated, if at all.

bigopon: > Oh, the manual import of .$au.ts actually works in both dev and prod mode. Only editor complains, vite just runs.

If this is the case, maybe in this PR we can try with rewriting `.html` to `.$au.ts` and see if it works for both dev mode and production build mode?

bigopon (approved):
I've tested this locally, it seems to work fine, I think we can go ahead with this. Thanks @Vheissu , nice work.

@3cp did you run into any issues with this?

3cp:
I am fine with it :) Let's merge and try.

3cp:
Wait, there is an use case that this new rewrite blocks!

3cp:
We actually support user to manually import a html module like
```js
import viewDef from "./foo.html"; // import au2 compiled html module

@customElement({...viewDef, name: 'override-name'})
export class Foo
```

This above worked in other modules right now. This new import rewrite in vite will block that usage.

bigopon:
Hmm I don't think we can dismiss that usage, which means we cannot merge this PR as it is. I remember I used to try to make our plugin to process html first, but vite always insert its own html processor before any others. Maybe there's a way to hook into that, or config something so that we can support many scenarios.

MaximBalaganskiy:
no luck with this?

bigopon:
Currently, whenever a `.html` import is encountered, it's automatically converted to `.$au.ts` import. With this behavior, convention gets turned on by default, and requires application to opt-out with `?raw` to get the old text import behavior.

This is probably not great for the reason that importing `.html` is quite common and normally expected to be a text-stuff operation. Maybe we can flip this around, and make convention be an opt-in instead, like the following example:
```ts
import template from './my-template.html'
import template2 from './my-template.html?convention'
```

It will be modified to:
```ts
import template from './my-template.html'
import template2 from './my-template.$au.ts'
```

---

> We actually support user to manually import a html module like
> 
> ```js
> import viewDef from "./foo.html"; // import au2 compiled html module
> 
> @customElement({...viewDef, name: 'override-name'})
> export class Foo
> ```
> 
> This above worked in other bundlers right now. This new import rewrite in vite will block that usage.

@3cp I just checked both webpack & vite, and didn't observe the behavior you mentioned above.

When having
```html
<div>hi</div>
```
```ts
import { customElement } from 'aurelia';
import template from './component.html'

console.log({template});

@customElement({
  name: 'component',
  template: null,
  // ...template
})
export class Component {

}
```
console logs `<div>hi</div>` for me. So I think it's safe that we flip the behavior to requiring `?convention` to opt-in convention. This will also allow us to play better with other plugin as well.

cc @Sayan751

3cp:
@bigopon just noticed my example was wrong, it meant to be:
```js
import * as viewDef from "./foo.html"; // import au2 compiled html module

@customElement({...viewDef, name: 'override-name'})
export class Foo
```

3cp:
With your newly suggested way, `import * as viewDef from "./foo.html?convention";` should work nicely.

3cp:
Let me throw another suggestion, a breaking change, what about change our convention?

Instead of touching foo.html, let's require user to write file `foo.au.html` in order to be automatically picked up by au2 conventions.

So conventions would process foo.js/ts, foo.au.html, and foo.css/scss automatically, but leave foo.html untouched.

In this way, users use file naming convention to control whether a html file (1) is a pain html file, or (2) an au template to be processed. And we can relief the users from using weird bundler specifier.

3cp:
Note `.au.html` was for the editor to show correct syntax. To use `.au` only, we need to ship a vscode plugin.

3cp:
@bigopon we might consider adopting `.au.something` for the whole conventions.
Conventions only touch `foo.au.js/ts`, `foo.au.html`, and `foo.au.css/scss`, so it's very clear what files that au2 conventions process. This also makes `.au` SFC component a natural way going forward. If we are interested, we can make another RFC to start the discussion.

MaximBalaganskiy:
This all seems like more magic - no au.html file but you can import it. Isn't it possible to utilise html?aurelia everyone knows what that means and how it works

bigopon:
> I think we should also add some e2e tests. There are a ton already based on vite, we can take any of those apps, and add a HTML template import and assert if it works correctly. Probably, on the same line, the HMR can also be tested.

Thanks @Sayan751 we do have e2e tests atm, including hmr, similar like how webpack hmr is tested.

---

I've pushed a different attempt to fix this issue in #2194 . I've tested the changes locally and it worked for me.

If the plugin is failing in production mode because it fails to detect the mode, and doesn't convert `.html` to `.$au.ts`, then the first fix I think could simply be trying to detect the mode better. If this works we may be able to put down the matter related to existing convention here.

For the new convention, let's try with an RFC, though I need help with championing that, if we really want to.

---

For the changes proposed in this PR, I don't think we can go with it, it simply will change the code in an unpredictable way, since any random string with `.html"` will be replaced by `.html?raw"`, which could be affecting inline templates as well. Not to mention broken sourcemap, I think we should avoid it.

bigopon:
thanks @MaximBalaganskiy , if you are not using the plugin, how did you get the error mentioned in this issue?

ekzobrain:
> Tbh, I'm not using the plugin in any of our projects. Yes, this means little bit more work importing scss and converters but there's no magic at all - anyone looking at the project can understand what happens. It works on dev, build and watch modes as well

You mean plugin-conventions? I am also not using it and prefer direct imports for the same better readiness.

ekzobrain:
> thanks @MaximBalaganskiy , if you are not using the plugin, how did you get the error mentioned in this issue?

I think he means convertions plugin, not Vite plugin.

ekzobrain:
> @bigopon i caught that when experimenting with new projects. Initially I wanted to use conventions but it did not work well with the plugin I'm authoring - it ships compiled Ts and raw html files, rather than a bundle
> 
> @ekzobrain i mean aurelia vite plugin

Ok. So you are using Vite, but not using Vite plugin? Could you, please, show some of your working Vite configs?

@bigopon you closed this PR, but it also contained some docs fixes as well. May be they need to be ported too?

bigopon:
> ... it also contained some docs fixes as well. May be they need to be ported too?

MaximBalaganskiy:
@ekzobrain 
 Commented out lines are needed if you want to run vite dev. I'm using build watch and don't need that
```Ts
import { defineConfig } from 'vite';
// import { nodePolyfills } from 'vite-plugin-node-polyfills'
// import aurelia from '@aurelia/vite-plugin';
import path from 'path';

export default defineConfig({
  server: {
    open: !process.env.CI,
    port: 9000,
  },
  esbuild: {
    target: 'es2022',
  },
  // optimizeDeps: {
  //   esbuildOptions: {
  //     loader: {
  //       '.html': 'text',
  //     },
  //   },
  // },
  build: {
    emptyOutDir: true,
    outDir: path.resolve(__dirname, '../Apsm.Api/wwwroot'),
    sourcemap: true,
    minify: 'terser',
    terserOptions: {
      keep_classnames: true
    }
  },
  css: {
    preprocessorOptions: {
      scss: {
        api: 'modern-compiler',
        quietDeps: true,
        silenceDeprecations: ['mixed-decls', 'import', 'global-builtin', 'color-functions']
      },
    },
  },
  plugins: [
   
  ]
});
```

3cp:
Well, it's just another suggestion :)
Either way, the conventions plugin still needs to handle the special file name (.au.html) or special import (.html?whatever). Both way is not free from existing bundler behaviour.

ghiscoding:
> Hi everyone, if you are using the vite plugin with convention, the latest dev version should have the fix for the issue mentioned here. Pls try it with yarn resolutions or npm overrides for the following:
> 
> ```ts
> {
>   "resolution": {
>     "vite": "2.1.0-dev.202507081127"
>   },
>   "overrides": {
>     "vite": "2.1.0-dev.202507081127"
>   }
> }
> ```

@bigopon I think you made a major typo on the package name 😆 I believe it should be an override on `@aurelia/vite-plugin` not `vite` 😉

bigopon:
Oops yes, it shhould be `@aurelia/runtime`, `@aurelia...` etc. Thanks @ghiscoding

PR #2125, opened 2025-02-28
docs(components) my-component.css -> my-component.module.css

HIRANO-Satoshi:
# Pull Request

## 📖 Description

In "Example: Using CSS Modules in Components", my-component.css should be my-component.module.css.


bigopon (approved):
thanks @HIRANO-Satoshi

PR #2126, opened 2025-03-01
fix(router-lite): support for conventional HTML-only component

Sayan751:
# Pull Request

## 📖 Description

This PR adds support for routing to conventional HTML-only components.

### 🎫 Issues

The issue is reported in [discourse](https://discourse.aurelia.io/t/router-questions/6360/3). Thanks @elmt1 for reporting this issue!

### DIFF
--- a/packages/router-lite/src/instructions.ts
+++ b/packages/router-lite/src/instructions.ts
@@ -18,7 +18,7 @@ import {
 import { IRouteViewModel } from './component-agent';
 import { RouteType } from './route';
 import { type $RecognizedRoute, IRouteContext, RouteContext } from './route-context';
-import { expectType, isPartialViewportInstruction, shallowEquals } from './validation';
+import { expectType, isPartialCustomElementDefinition, isPartialViewportInstruction, shallowEquals } from './validation';
 import { INavigationOptions, NavigationOptions, type RouterOptions } from './options';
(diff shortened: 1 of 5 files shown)


packages/router-lite/src/instructions.ts
```diff
@@ -459,6 +459,24 @@ export class TypedNavigationInstruction<TInstruction extends NavigationInstructi
...
+    //   routes: [
+    //     {
+    //       path: 'path',
+    //       component,
+    //     },
+    //   ],
+    // })
+    if (isPartialCustomElementDefinition(instruction)) {
+      const definition = CustomElementDefinition.create(instruction);
+      CustomElement.define(definition);
```
bigopon, line 476: do we need this line?

bigopon (approved):
nice work @Sayan751

packages/router-lite/src/instructions.ts
Sayan751: Yes. Because, `CustomElement.define` defines the custom-element metadata on the type. Later that metadata is needed to determine the path (name of the CE). This can be rewritten as `CustomElement.getDefinition(CustomElement.define(instruction))`; however, not sure if that is any better.

Sayan751:
@bigopon I am merging this PR. In case you notice other issues, let me know.

PR #2127, opened 2025-03-08
doc(multi-root): correct multi root scenario doc and link to example

simonfox:
# Pull Request

## 📖 Description

Correction of documentation for multi root scenario as suggested by @Sayan751 on Discourse https://discourse.aurelia.io/t/aurelia2-reload-root-component/6365

### 🎫 Issues

The solution that was described didn't actually work.

### DIFF
--- a/docs/user-docs/developer-guides/scenarios/multi-root.md
+++ b/docs/user-docs/developer-guides/scenarios/multi-root.md
@@ -4,68 +4,65 @@ In a scenario where you need to manage multiple parts of an application that mig
 
 Below is a step-by-step guide on how to create a multi-root Aurelia 2 application, complete with detailed explanations and code examples.
 
-## Setting up the Initial Root
+## Strategy
 
-In `src/main.ts`, we set up the initial root of the application. This is typically the entry point of your Aurelia app.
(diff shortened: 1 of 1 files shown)


Sayan751:
Hi @simonfox! Thank you for the PR. Just played with the previous strategy a bit. In fairness, I should say that the previous strategy does work. It was missing some of the injection stuffs. Here is a working example of that: https://stackblitz.com/edit/router-lite-load-nav-options-query-9q1n27vs?file=src%2Flogin.ts,package.json,tsconfig.json,src%2Fabout.html,src%2Fmain.ts

My personal preference would still be the event-based approach due to the separation of concerns. IMO, the login page making the decision what the app root should be, seems a bit odd (think about mutli-app roots).

On the other hand, the goal of the documentation should also be to present the idea as simply as possible, lowering the entry level.

simonfox:
@Sayan751 ah I see - I tried to make that work adding `@inject(Aurelia)` but got the DI error in my Discourse post. Changing `Āurelia` to `IAurelia` and it works!

I agree that your event based solution is nicer. Feel free to close this PR if you'd rather keep the docs simple.

bigopon (approved):
lgtm, thanks @simonfox @Sayan751

Vheissu (approved):

PR #2129, opened 2025-03-11
docs(process-content): add missing $host reference and hint for its u…

simonfox:
# Pull Request

## 📖 Description

Correction of documentation for `processContent` describing the need for `$host` scope reference.

### 🎫 Issues

The example doesn't work without `$host` to access props of `Tabs` custom element.


Sayan751:
Thanks @simonfox! 👍

PR #2136, opened 2025-03-29
feat(validation): make caching of validate behavior property info configurable

simonfox:
# Pull Request

## 📖 Description

see [Discord discusson](https://discord.com/channels/448698263508615178/605165596233170944/1354621721361252533) for details.

When using a binding source which is a projection object, syncing validation errors to the view does not work because the binding target element search uses cached property info and a reference equality check between the binding source and the cached source. In the context of redux/reselect, the entire projection object is replaced and so the binding source is different to that which was cached so the target element is skipped.

This change adds a fourth parameter to the validate binding behavior to disable caching of the property info. It also adds the ability to configure the plugin such that the default is to not use caching.

### DIFF
--- a/packages/validation-html/src/validate-binding-behavior.ts
+++ b/packages/validation-html/src/validate-binding-behavior.ts
@@ -61,6 +61,8 @@ export enum ValidationTrigger {
 }
 
 export const IDefaultTrigger = /*@__PURE__*/DI.createInterface<ValidationTrigger>('IDefaultTrigger');
+// eslint-disable-next-line @typescript-eslint/ban-types
+export const IDefaultCacheBindingPropertyInfo = /*@__PURE__*/DI.createInterface<Boolean>('IDefaultCacheBindingPropertyInfo');
 
 const validationConnectorMap = new WeakMap<IBinding, ValidationConnector>();
 const validationTargetSubscriberMap = new WeakMap<PropertyBinding, WithValidationTargetSubscriber>();
@@ -83,6 +85,7 @@ export class ValidateBindingBehavior implements BindingBehaviorInstance {
         this._platform,
         this._observerLocator,
         binding.get(IDefaultTrigger),
+        binding.get(IDefaultCacheBindingPropertyInfo) as boolean,
         binding as BindingWithBehavior,
         binding.get(IContainer)
       ));
@@ -110,7 +113,7 @@ export class ValidateBindingBehavior implements BindingBehaviorInstance {
 }
 BindingBehavior.define('validate', ValidateBindingBehavior);
 
-interface ValidationConnector extends IAstEvaluator, IObserverLocatorBasedConnectable, IConnectable {}
+interface ValidationConnector extends IAstEvaluator, IObserverLocatorBasedConnectable, IConnectable { }
 /**
  * Binding behavior. Indicates the bound property should be validated.
  */
@@ -122,6 +125,7 @@ class ValidationConnector implements ValidationResultsSubscriber {
   private controller!: IValidationController;
   private isChangeTrigger: boolean = false;
   private readonly defaultTrigger: ValidationTrigger;
+  private readonly defaultCachePropertyInfo: boolean;
   public scope?: Scope;
   private isDirty: boolean = false;
   private validatedOnce: boolean = false;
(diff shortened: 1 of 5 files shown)


packages/validation-html/src/validate-binding-behavior.ts
```diff
@@ -61,6 +61,8 @@ export enum ValidationTrigger {
 }
 
 export const IDefaultTrigger = /*@__PURE__*/DI.createInterface<ValidationTrigger>('IDefaultTrigger');
+// eslint-disable-next-line @typescript-eslint/ban-types
```
simonfox, line 64: @Sayan751 have changed to Boolean here as requested - is this safe or should I create a wrapper like `class CacheBindingPropertyInfo extends Boolean { }`?

simonfox:
@Sayan751 tests and doc updates in progress. Are you happy with the naming? Maybe update `cache` to `cachePropertyInfo`?

packages/validation-html/src/validate-binding-behavior.ts
```diff
@@ -61,6 +61,8 @@ export enum ValidationTrigger {
...
+export const IDefaultCacheBindingPropertyInfo = /*@__PURE__*/DI.createInterface<Boolean>('IDefaultCacheBindingPropertyInfo');
```
simonfox, line 65: @Sayan751 have changed to Boolean here as requested - is this safe or should I create a wrapper like `class CacheBindingPropertyInfo extends Boolean { }`?

Sayan751: `DI.createInterface<Boolean>` should do.

packages/validation-html/src/validate-binding-behavior.ts
```diff
@@ -83,6 +85,7 @@ export class ValidateBindingBehavior implements BindingBehaviorInstance {
         this._platform,
         this._observerLocator,
         binding.get(IDefaultTrigger),
+        binding.get(IDefaultCacheBindingPropertyInfo) as boolean,
```
Sayan751, line 88: Just trying to understand why this casting is needed?

packages/validation-html/src/validate-binding-behavior.ts
```diff
@@ -331,6 +348,15 @@ class ValidationConnector implements ValidationResultsSubscriber {
     }
   }
 
+  /** @internal */
+  private _ensureCache(cache: unknown): boolean {
```
Sayan751, line 352: Do this leniency really required? IMO this can be reduced to `cache !== false`.

Sayan751:
Thank you @simonfox for this addition!

packages/validation-html/src/validate-binding-behavior.ts
simonfox: 👍 will change.

simonfox:
@Sayan751 tests added

bigopon (commented):
It feels to me rather than adding more infrastructure, we can add a check in `getPropertyInfo` so that if the old cached object is not the same with the current object anymore, we re-retrieve property info instead.

simonfox:
@bigopon 

> It feels to me rather than adding more infrastructure, we can add a check in `getPropertyInfo` so that if the old cached object is not the same with the current object anymore, we re-retrieve property info instead.

Wouldn't that mean we still do all the evaluation `getPropertyInfo` does to get the object that we need to compare anyway? i.e. does caching become redundant?

bigopon:
> Wouldn't that mean we still do all the evaluation getPropertyInfo does to get the object that we need to compare anyway? i.e. does caching become redundant?

I still don't know the validation and its flow inside out, so the correctness of my understanding, reflected by my comments, may be incorrect. Though I would think the caching should work this way:
- at the bind lifecycle, bindings should already be able to determine whether it's dealing with what type of expression:
    - if the expression is "static", the object is cacheable: it can use a strategy that can skip most of the work when revalidate.
    - if the expression is "dynamic", the object is not safe to cache: it can use a more expensive strategy to calculate the info every time, and caching should be opt-in per binding, not opt-out.
        This is reflected in your issue reported: it's a highly dynamic expression; caching is likely to be incorrect/stale.

Sayan751:
I think that the issue is that the expression looks static, but due to the change of the whole object the property information needs to be recomputed when the binding is validated. However, @bigopon's objection to add more infra, can result in more elegant solution. If I am not much wrong, then we can cache the source object in the `ValidationConnector`, and when the source is updated, make a check if the source object is still the same, if not then clear the cache. A sweet spot to make this check might be here: https://github.com/aurelia/aurelia/blob/0880c7be18f09e5328ce29430d87bedbff8ee3e3/packages/validation-html/src/validate-binding-behavior.ts#L162-L168

Would you be interested in making the change @simonfox? Let me know if you need a hand.

Regarding the tests, please add a test that focuses on asserting the execution of the rules, which is originally the goal of this PR, even if the underlaying object is changed.

simonfox:
closing in favor of aurelia#2138

PR #2137, opened 2025-03-29
feat(router-lite): navigation strategy

Sayan751:
# Pull Request

## 📖 Description

This is a port of navigation strategy of au1 router. An example looks as follows.

```ts
import { customElement } from '@aurelia/runtime-html';
import { route, NavigationStrategy, RecognizedRoute } from '@aurelia/router-lite';
import template from './my-app.html';
import { About } from './about';
import { Home } from './home';

let dataLoaded = false;

@route({
  routes: [
    {
      path: ['', 'home'],
      component: Home,
    },
    {
      path: 'about',
      component: About,
    },
    {
      path: 'foo',
      component: new NavigationStrategy((viewportInstruction: IViewportInstruction, ctx: IRouteContext, node: RouteNode, route: RecognizedRoute<unknown>) => {
        if (dataLoaded) return Home;

        dataLoaded = true;
        return About;
      })
    }
  ],
})
export class MyApp {}
```

cc: @simonfox

### DIFF
--- a/packages/router-lite/src/route.ts
+++ b/packages/router-lite/src/route.ts
@@ -46,12 +46,14 @@ export class RouteConfig implements IRouteConfig, IChildRouteConfig {
     public readonly nav: boolean,
     private readonly navigatingObserver: IObserver | null,
   ) {
-    this.isNavigationStrategy = _component instanceof NavigationStrategy;
-    navigatingObserver?.subscribe({
-      handleChange: (newValue, _previousValue) => {
-        if (newValue === true) this._currentComponent = null;
-      }
-    });
+    // eslint-disable-next-line no-cond-assign
+    if (this.isNavigationStrategy = _component instanceof NavigationStrategy) {
+      navigatingObserver?.subscribe({
+        handleChange: (newValue, _previousValue) => {
+          if (newValue === true) this._currentComponent = null;
+        }
+      });
+    }
   }
 
   /** @internal */
--- a/packages/router-lite/src/route.ts
+++ b/packages/router-lite/src/route.ts
@@ -240,7 +240,7 @@ export class RouteConfig implements IRouteConfig, IChildRouteConfig {
     if (vi == null) {
       if (this._currentComponent != null) return this._currentComponent;
       if (this.isNavigationStrategy) throw new Error(getMessage(Events.rtInvalidOperationNavigationStrategyComponent, this.id));
-      return this._currentComponent ??= this._component;
+      return this._currentComponent = this._component;
     }
     return this._currentComponent ??= this.isNavigationStrategy ? (this._component as NavigationStrategy).getComponent(vi, ctx!, node!) : this._component;
   }
--- a/packages/router-lite/src/route.ts
+++ b/packages/router-lite/src/route.ts
@@ -2,7 +2,7 @@ import { Metadata } from '@aurelia/metadata';
 import { Constructable, emptyArray, onResolve, ResourceType, Writable, getResourceKeyFor } from '@aurelia/kernel';
 
 import { validateRouteConfig, expectType, shallowEquals, isPartialRedirectRouteConfig, isPartialChildRouteConfig, isPartialCustomElementDefinition } from './validation';
-import { defaultViewportName, ITypedNavigationInstruction_Component, ITypedNavigationInstruction_NavigationStrategy, NavigationInstructionType, NavigationStrategy, TypedNavigationInstruction, ViewportInstruction } from './instructions';
+import { defaultViewportName, ITypedNavigationInstruction_Component, ITypedNavigationInstruction_NavigationStrategy, IViewportInstruction, NavigationInstructionType, NavigationStrategy, TypedNavigationInstruction, ViewportInstruction } from './instructions';
 import type { RouteNode } from './route-tree';
 import type { IRouteContext } from './route-context';
 import { CustomElement, CustomElementDefinition, PartialCustomElementDefinition } from '@aurelia/runtime-html';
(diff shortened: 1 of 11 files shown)


simonfox:
getting this when linking my repro project to your branch
image

all I've changed is package.json reference to local build of your branch
image

simonfox:
@Sayan751 possibly an issue with parameterized routes. Trying to sort a minimal repro for you.

simonfox:
> @Sayan751 possibly an issue with parameterized routes. Trying to sort a minimal repro for you.

ok so maybe not an _issue_, but if I have the following route config that uses a navigation strategy
```
{
          id: "feature-load-id",
          path: ":id",
          title: "Load",
          component: new NavigationStrategy((viewportInstruction: IViewportInstruction, ctx: IRouteContext, node: RouteNode) => {
            console.log("nav strategy: ", viewportInstruction.params, ctx);
            if (!this._loaded) {
              this._loaded = true;
              return FeatureLoading;
            } else {
              return FeatureLoad;
            }
          }),
          nav: false,
        },
```
I want to be able to get the value of the id param so I can determine if that id is loaded. `viewportInstruction.params` must be params of the parent?

packages/router-lite/src/route.ts
```diff
@@ -21,9 +22,14 @@ export class RouteConfig implements IRouteConfig, IChildRouteConfig {
   public get path(): string[] {
     const path = this._path;
     if (path.length > 0) return path;
-    const ceDfn = CustomElement.getDefinition(this.component as RouteType);
+    const ceDfn = CustomElement.getDefinition(this._component as RouteType);
     return this._path = [ceDfn.name, ...ceDfn.aliases];
   }
+  private _currentComponent: Routeable | null = null;
```
bigopon, line 28:
```suggestion
  /** @internal */
  private _currentComponent: Routeable | null = null;
```

packages/router-lite/src/route.ts
```diff
@@ -36,14 +42,25 @@ export class RouteConfig implements IRouteConfig, IChildRouteConfig {
...
-  ) { }
+    private readonly navigatingObserver: IObserver | null,
+  ) {
+    // eslint-disable-next-line no-cond-assign
+    if (this.isNavigationStrategy = _component instanceof NavigationStrategy) {
+      navigatingObserver?.subscribe({
+        handleChange: (newValue, _previousValue) => {
+          if (newValue === true) this._currentComponent = null;
+        }
+      });
```
bigopon, line 55: don't we need to unsubscribe?

packages/router-lite/src/route.ts
```diff
@@ -36,14 +42,25 @@ export class RouteConfig implements IRouteConfig, IChildRouteConfig {
     public readonly data: Record<string, unknown>,
     public readonly routes: readonly Routeable[],
     public readonly fallback: Routeable | FallbackFunction | null,
-    public readonly component: Routeable,
+    private readonly _component: Routeable | NavigationStrategy,
     public readonly nav: boolean,
-  ) { }
+    private readonly navigatingObserver: IObserver | null,
```
bigopon, line 47: `navigatingObserver` seems disconnected with a `RouteConfig`, Rather than making `RouteConfig` know about some foreign concepts/ideas, what about we do the subscription somewhere else and let `RouteConfig` expose a method/property to mutate the current component instead?

bigopon (commented):
Nice work @Sayan751 only a few small comments.

packages/router-lite/src/route.ts
Sayan751: The route config objects have a lifetime similar to the application lifetime. Thus, a disposal of those objects are of little use. I can explain this by annotating the code with comments. However, if you think that unsubscribe is needed, let me know, I can surely arrange that.

packages/router-lite/src/route.ts
Sayan751: Refactored the infra.

Sayan751:
@simonfox For easily handling the parameters, I have added a fourth parameter to the navigation strategy factory function. Also added a test as an example. You may have a look if the example in the test case covers your use-case.

simonfox:
Thanks @Sayan751! I'm getting the `rtInvalidOperationNavigationStrategyComponent` error in our app when I try to redirect back using `router.load`. However I've been able to do what I want to when trying to repro. I'll keep trying to get to the bottom of it, but confirm that the extra param does what I need it to.

simonfox:
@Sayan751 this is not the exact issue I'm getting in our app, but could be related - https://github.com/simonfox/router-issue

A child router doesn't seem to get activated after using nav strategy. See `src/feature/load`. Nav strategy is used to choose between `feature-loading` and `feature-loaded`. When redirecting to `feature-loaded` after doing the loading from `feature-loading` the child router activation of loaded doesn't happen i.e. getRouteConfig isnt called and the au-viewport doesn't get registered.

Note that I've used dumber deps property (in `gulpfile.js`) to link to a local build of aurelia. You may want to just drop the example into your own test app.

bigopon (approved):
nice work @Sayan751 , though I think we probably should have private props prefixed with `_`, there's no reason to delay this small optimization.

Sayan751:
@simonfox I have added another test for your scenario. It seems to be working. I would recommend calling the `router.load` in the `attached` hook.

I will go ahead and merge this PR.

simonfox:
@Sayan751  Scenario I was trying to describe would mean `C12` component would describe a child router and template contain a `au-viewport`

In that case the get routes callback wasn't being invoked.

Sayan751:
@simonfox Alright. I will refactor the test.

Sayan751:
@simonfox Seems to work as well.

PR #2138, opened 2025-04-01
fix(validation): invalidate property info cache when source changes

simonfox:
this supercedes aurelia#2136

# Pull Request

## 📖 Description

see [Discord discusson](https://discord.com/channels/448698263508615178/605165596233170944/1354621721361252533) for details.

When using a binding source which is a projection object, syncing validation errors to the view does not work because the binding target element search uses cached property info and a reference equality check between the binding source and the cached source. In the context of redux/reselect, the entire projection object is replaced and so the binding source is different to that which was cached so the target element is skipped.

This change adds an observer to the binding that is used to invalidate the cached property info when the binding source changes.

### DIFF
--- a/packages/__tests__/src/validation-html/validate-binding-behavior.spec.ts
+++ b/packages/__tests__/src/validation-html/validate-binding-behavior.spec.ts
@@ -1785,33 +1785,33 @@ describe('validation-html/validate-binding-behavior.spec.ts', function () {
       await stop(true);
     });
 
-    it('replaced binding source validates', async function () {
-      class MyApp {
-        private _rules: IValidationRules;
-        public person: Person = new Person((void 0)!, (void 0)!);
-        public readonly validationController: IValidationController = resolve(newInstanceForScope(IValidationController));
-        public constructor()  {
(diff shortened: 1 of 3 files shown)


simonfox:
@Sayan751 @bigopon this solution feels a lot better. Let me know if you're happy with the approach and I can add tests.

bigopon (commented):
this feels indeed much better, we are observing both object and the property for changes, so validation will be retriggered automatically, properly. 👍 from me

Sayan751:
LGTM. Just verify with proper tests.

simonfox:
@Sayan751 how's this? Any other cases I should check?

Sayan751 (approved):
@simonfox The tests look good. Thanks!

Do you think the docs needs to be updated?

simonfox:
@Sayan751 public interface hasn't changed so I'm not sure what documentation might be needed? Happy to add something if there is something specific you'd like to see?

Sayan751:
@simonfox In that case I am merging this PR.

PR #2143, opened 2025-04-06
fix(validation): release reference to binding behavior source

simonfox:
# Pull Request

## 📖 Description

Drop reference to the binding source reference added as part of #2138 when the binding behavior is cleaning up.

cc @Sayan751


Sayan751:
Thanks @simonfox!

Issue #2139, opened 2025-04-01
router-lite Navigation Strategy route hierarchy activation


simonfox:
First, thank you for your work on the nav strategy feature @Sayan751 !!

Here is a failing test that describes the situation I was trying to explain...

```
it('load new route and activates child router', async function () {

    let resolver: () => void;
    const promise = new Promise<void>(r => resolver = r);
    let dataLoaded = false;

    @customElement({ name: 'gc-2', template: 'gc2' })
    class GC12 { }

    @customElement({ name: 'c-1', template: 'c1' })
    class C11 implements IRouteViewModel {
      private readonly router: IRouter = resolve(IRouter);
      public canLoad(params: Params, _next: RouteNode, _current: RouteNode | null): boolean {
        void promise.then(() => this.router.load(`p-1/c-12/${params.id}`, { transitionPlan: 'replace' }));
        return true;
      }
    }

    @customElement({ name: 'c-2', template: 'c2 <au-viewport></au-viewport>' })
    @route({
      routes: [
        C11,
        { path: ['', 'gc-12'], component: GC12 }
      ]
    })
    class C12 { }

    @route({
      routes: [
        C11,
        {
          path: ':id',
          component: new NavigationStrategy(() => {
            if (dataLoaded) return C12;

            dataLoaded = true;
            return C11;
          })
        }
      ]
    })
    @customElement({ name: 'p-1', template: 'p1 <au-viewport></au-viewport>' })
    class P1 { }

    @customElement({ name: 'p-2', template: 'p2' })
    class P2 { }

    @route({ routes: [P1, P2] })
    @customElement({ name: 'ro-ot', template: '<au-viewport></au-viewport>' })
    class Root { }

    const { au, container, host } = await start({ appRoot: Root });
    const router = container.get(IRouter);
    const taskQueue = container.get(IPlatform).taskQueue;

    assert.html.textContent(host, '', 'initial');

    await router.load('p-1/1');
    assert.html.textContent(host, 'p1 c1', 'round#1');

    // resolve the promise to trigger the navigation
    resolver!();
    taskQueue.queueTask(() => new Promise(r => setTimeout(r, 0)));
    await taskQueue.yield();
    assert.html.textContent(host, 'p1 c2 gc2', 'initial');

    await au.stop(true);
  });
```

Sayan751:
@simonfox This is just a simple case of misconfiguration. However, I have added the test here: https://github.com/aurelia/aurelia/pull/2141

[...]

PR #2144, opened 2025-04-06
fix(router-lite): cleanup while reusing route context due to nav strategy

Sayan751:
# Pull Request

## 📖 Description

This addresses the issue reported in https://github.com/aurelia/aurelia/issues/2139#issuecomment-2781120624.
Thanks @simonfox for helping to shape this feature!

### DIFF
--- a/packages/router-lite/src/route.ts
+++ b/packages/router-lite/src/route.ts
@@ -222,6 +222,16 @@ export class RouteConfig implements IRouteConfig, IChildRouteConfig {
   }
 
   /** @internal */
+  public _getComponentName(): string {
+    try {
+      return (this._getComponent() as RouteType).name;
+    } catch {
+      // TODO(Sayan): Convert all Errors in router lite to instances of RouterError
+      return 'UNRESOLVED-NAVIGATION-STRATEGY';
+    }
+  }
+
+  /** @internal */
   public _getComponent(): Routeable;
   public _getComponent(vi: IViewportInstruction, ctx: IRouteContext, node: RouteNode, route: RecognizedRoute<unknown>): Routeable;
   public _getComponent(vi?: IViewportInstruction, ctx?: IRouteContext, node?: RouteNode, route?: RecognizedRoute<unknown>): Routeable {
--- a/packages/router-lite/src/route.ts
+++ b/packages/router-lite/src/route.ts
@@ -343,7 +343,7 @@ export function resolveRouteConfiguration(routeable: Routeable, isChild: boolean
     const type = $ceDef.Type;
     const routeConfig = Route.getConfig(type);
 
-    // If the component is used as a child, then apply the child configuration (comping from parent) and return a new RouteConfig with the configuration applied.
+    // If the component is used as a child, then apply the child configuration (coming from parent) and return a new RouteConfig with the configuration applied.
     if (isPartialChildRouteConfig(routeable)) return routeConfig._applyChildRouteConfig(routeable, parent);
 
     // If the component is used as a child, then return a clone.
--- a/packages/router-lite/src/route.ts
+++ b/packages/router-lite/src/route.ts
@@ -27,7 +27,7 @@ export class RouteConfig implements IRouteConfig, IChildRouteConfig {
   /** @internal */ private _currentComponent: Routeable | null = null;
   public get component(): Routeable { return this._getComponent(); }
 
-  /** @internal */ private readonly _isNavigationStrategy: boolean;
+  /** @internal */ public readonly _isNavigationStrategy: boolean;
   /** @internal */ private readonly _component: Routeable | NavigationStrategy;
 
   private constructor(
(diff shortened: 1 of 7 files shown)


simonfox:
Nice @Sayan751!! Confirming that this resolves the issues I was having.

simonfox:
@Sayan751 here's a failing test 😉 

```
it('load new route and activates child viewport - more than once with different ids', async function () {

    const resolvers = {};
    const promises = {
      1: new Promise<void>(r => resolvers[1] = r),
      2: new Promise<void>(r => resolvers[2] = r),
    };

    const dataLoaded = {
      1: false,
      2: false,
    };

    @customElement({ name: 'gc-12', template: 'gc12' })
    class GC12 { }

    @customElement({ name: 'c-11', template: 'c11' })
    class C11 implements IRouteViewModel {
      private readonly router: IRouter = resolve(IRouter);
      public canLoad(params: Params, _next: RouteNode, _current: RouteNode | null): boolean {
        void promises[params.id].then(() => this.router.load(`p-1/${params.id}`, { transitionPlan: 'replace' }));
        return true;
      }
    }

    @customElement({ name: 'c-12', template: 'c12 ${id} <au-viewport></au-viewport>' })
    @route({
      routes: [
        C11,
        { path: ['', 'gc-12'], component: GC12 }
      ]
    })
    class C12 {
      id: string;
      loading(params) { this.id = params.id; }
    }

    @customElement({ name: 'c-13', template: 'c13' })
    class C13 implements IRouteViewModel {
    }

    @route({
      routes: [
        C11,
        C12,
        C13,
        {
          path: ':id',
          component: new NavigationStrategy((_, __, ___, route) => {
            if (dataLoaded[route.params.id]) return C12;

            dataLoaded[route.params.id] = true;
            return C11;
          })
        }
      ]
    })
    @customElement({ name: 'p-1', template: 'p1 <au-viewport></au-viewport>' })
    class P1 { }

    @customElement({ name: 'p-2', template: 'p2' })
    class P2 { }

    @route({ routes: [P1, P2] })
    @customElement({ name: 'ro-ot', template: '<au-viewport></au-viewport>' })
    class Root { }

    const { au, container, host } = await start({ appRoot: Root });
    const router = container.get(IRouter);
    const taskQueue = container.get(IPlatform).taskQueue;

    assert.html.textContent(host, '', 'initial');

    await router.load('p-1/1');
    assert.html.textContent(host, 'p1 c11', 'round#1');

    // resolve the promise to trigger the navigation
    resolvers[1]();
    taskQueue.queueTask(() => new Promise(r => setTimeout(r, 0)));
    await taskQueue.yield();
    assert.html.textContent(host, 'p1 c12 1 gc12', 'post-data-load');

    // reset to some other route
    await router.load('p-1/c-13');
    assert.html.textContent(host, 'p1 c13', 'reset');

    /********************************************************/
    // everything is OK up to here - all assertions pass
    // then hit an infinite loop
    // if we do
    // await router.load('p-2') and leave the hierarchy
    // rooted at P1 before doing await router.load('p-1/2')
    // then the issue doesn't occur
    /********************************************************/

    await router.load('p-1/2');
    assert.html.textContent(host, 'p1 c11', 'round#2');

    // resolve the promise to trigger the navigation
    resolvers[2]();
    taskQueue.queueTask(() => new Promise(r => setTimeout(r, 0)));
    await taskQueue.yield();
    assert.html.textContent(host, 'p1 c12 2 gc12', 'post-data-load 2');

    await au.stop(true);
  });
```

packages/router-lite/src/router.ts
```diff
@@ -379,6 +379,7 @@ export class Router {
         let routeContext = routeConfigLookup.get(rdConfig);
```
bigopon, line 379: quite a lot of caching, do we need to have it at this level?

bigopon (commented):
I'm a bit concerned by the addition of state management inside route context. It feels like it'll get out of hand any time soon.

Sayan751:
> I'm a bit concerned by the addition of state management inside route context. It feels like it'll get out of hand any time soon.

@bigopon I agree. Found an easier solution. Got rid of all those stuffs.

Sayan751:
@simonfox Thanks for the test! Added your test to the suite.

packages/router-lite/src/route.ts
```diff
@@ -314,13 +337,13 @@ export function resolveRouteConfiguration(routeable: Routeable, isChild: boolean
 
   const [instruction, ceDef] = resolveCustomElementDefinition(routeable, context);
   if (instruction.type === NavigationInstructionType.NavigationStrategy)
-    return RouteConfig._create(routeable as IChildRouteConfig, null);
+    return RouteConfig._create({ ...routeable as IChildRouteConfig, nav: false }, null);
```
bigopon, line 340: 👍 looks like a really simple fix. Though I think the router isn't doing so much work so frequently, I'm not sure if we need to cache, and then avoid cache in specific scenarios like this. Wouldn't it be simpler to always create a new config? Caching of promise resolution can still be there though.

bigopon (approved):
nice work @Sayan751

packages/router-lite/src/route.ts
Sayan751: I see the confusion. Note that, here, the route config containing the navigation strategy is resolved only once. So this does not actually speak against the caching.

I am assuming that what might be causing the confusion, is this part: https://github.com/aurelia/aurelia/pull/2144/files#diff-5061a94fd6174829b1bf6f18a45b88b09ece1e3d127d002b299eee92a5e2dd77R363-R369.

That part is done to accommodate the inherent nature of the navigation strategy, where the routable component can change with potentially every call to the factory method. The changes in the linked part is thus required to load the route configuration for the resolved component, as otherwise the child routes configured by the current component will be ignored.

IMO, we should keep the caching infra, till it proves to be counter-productive.

Sayan751: I am going ahead and merging this PR for now.

PR #2145, opened 2025-04-10
feat(dialog): add standard html dialog renderer

### LINKED ISSUE #2053: Dialog should be based on web standards HTML <dialog> element
# 🙋 Feature Request

Simply as in the title: dialog box should be using HTML standard \<dialog> element.
There was a similar ticket in Aurelia V1: https://github.com/aurelia/dialog/issues/335

## 🤔 Expected Behavior

One feature that would be allowed by the HTML element is having a true modal dialog where it's impossible to interact with what's behind the dialog box.

That would also allow the use of the "autofocus" attribute inside the dialog.

## 😯 Current Behavior

Currently, there's a complicated way of styling the overlay (which is a DOM element itself).
Not sure if anyone wants to create a class for that.
Why dealing with a separated DOM element when they just want a popup dialog box.
The standard ::backdrop CSS seems like an elegant solution.

Why reinventing the wheel over HTML standards?
A \<dialog> element already exists in browsers, let's make use of it.

## 💁 Possible Solution

One problem with the HTML \<dialog> element is you can't simply close the dialog by clicking outside of it.
There are solutions discussed here: https://stackoverflow.com/questions/50037663/how-to-close-a-native-html-dialog-when-clicking-outside-with-javascript
An easy one would be:
```javascript
myDialog.addEventListener('click', event => {
    if(event.target === myDialog) {
        myDialog.close();
    }
});
```

## 🔦 Context

With the current dialog package, the resulting dialog box isn't really "modal".
There is indeed a visual overlay to cover what's behind the dialog, but you can still interact with it.
You can reach all background inputs navigating with the [TAB] key.

bigopon:
## 📖 Description

As per request from #2053 , add another renderer implementation for the dialog plugin that will renders the `<dialog>` element instead of a div with dialog like styling.
Also add a config `asModal` in the dialog open settings to support `modal` behavior for renderers.

BREAKING CHANGES:
- Dialog dom property `overlay` is now optional, since not all dialog needs an overlay, having it optional allows renderer to skip populating this altogether.

### 🎫 Issues

Close #2053

cc @ekzobrain @stopi

### DIFF
--- a/packages/dialog/src/dialog-interfaces.ts
+++ b/packages/dialog/src/dialog-interfaces.ts
@@ -54,7 +54,7 @@ export interface IDialogDomRenderer {
  */
 export const IDialogDom = /*@__PURE__*/createInterface<IDialogDom>('IDialogDom');
 export interface IDialogDom extends IDisposable {
-  readonly overlay: HTMLElement;
+  readonly overlay: HTMLElement | null;
   readonly contentHost: HTMLElement;
   /**
    * Called when the dialog should be shown. Application can use this for animations
--- a/packages/dialog/src/dialog-interfaces.ts
+++ b/packages/dialog/src/dialog-interfaces.ts
@@ -152,6 +152,19 @@ export interface IDialogSettings<
   lock?: boolean;
 
   /**
+   * When set to "true" the dialog will be modal.
+   * This means that the dialog will be displayed as a modal dialog.
+   * The default value is "false".
+   *
+   * Note that this depends on the renderer,
+   * Some renderers may not support this feature.
+   *
+   * Readmore on the modal behavior of dialogs on MDN
+   * https://developer.mozilla.org/en-US/docs/Web/API/HTMLDialogElement#opening_a_modal_dialog
+   */
+  asModal?: boolean;
+
+  /**
    * Allows for closing the top most dialog via the keyboard.
    * When set to "false" no action will be taken.
    * If set to "true", "Escape" or an array containing "Escape"
--- a/docs/user-docs/aurelia-packages/dialog.md
+++ b/docs/user-docs/aurelia-packages/dialog.md
@@ -511,6 +511,27 @@ There are two ways to use your own dialog renderer: register your own default di
     }
     ```
 
+### The standard `<dialog>` element renderer
+
+Beside the default dialog renderer, Aurelia also provides an out-of-the-box renderer that uses the `<dialog>` element.
+
+Compared to the default renderer, this dialog renderer has the following differences:
+- the dialog wrapper doesn't come with any default style, so it will take the default user agent style, which could come with the strong dark border around the content.
+- the dialog dom does not have an overlay. Since only modal dialog has a backdrop for overlay purpose, the overlay isn't always needed.
+- the overlay, can be styled with a helper via a method on the dialog dom (`HtmlDialogDom`), like the following example:
+    ```ts
+    import { IDialogDom, HtmlDialogDom } from '@aurelia/dialog';
+
+    class MyDialogComponent {
+      dom = resolve(IDialogDom) as HtmlDialogDom;
+
+      binding() {
+        dom.setOverlayStyle('background: pink');
+      }
+    }
+    ```
+    Note that the backdrop of a `<dialog>` is a pseudo element, and thus its style cannot be modified like a normal element.
+
 ### Animation
 
 #### Using the `IDialogDomAnimator`
--- a/packages/__tests__/src/3-runtime-html/dialog/dialog-html-dom.spec.ts
+++ b/packages/__tests__/src/3-runtime-html/dialog/dialog-html-dom.spec.ts
@@ -1,0 +1,57 @@
+import { delegateSyntax } from '@aurelia/compat-v1';
+import { Registration, noop, resolve } from '@aurelia/kernel';
+import {
+  INode,
+  customElement,
+  CustomElement,
+} from '@aurelia/runtime-html';
+import {
+  IDialogService,
+  IDialogSettings,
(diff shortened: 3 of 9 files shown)


ekzobrain:
@bigopon it a great feature. give me several days to look into it and give feedback, please

bigopon:
Hi @ekzobrain , what definition of several did you use? I normally use the one from heroes3, which is about 5-9 days 😀
Dont feel ruahed though, this can go and be out for test in a dev version before the next beta.

ekzobrain:
I will do it today within several hours (in your definition :)) I just had a lot of headache last weeks and forgot about it.

ekzobrain:
From Discord:

**bigopon — 06.05.2025, 12:44**
we need a good default implementation your suggested interface is over simplified and I think we need more than that
so the current .open call can be changed to
from
`.open<TModal>`
to
`.open<TModel, TSettings>`
or  something like this, regardless what we do we cant make app supply the renderer for each .open call
**Dmitry Arkhipov — 06.05.2025, 12:47**
Sure, it is possible to configure default renderer globally... I did dot suggest to remove this feature )
**Dmitry Arkhipov — 06.05.2025, 12:47**
This is exactly what I ment
I just left away some already existing features that don't need to be changed
**Dmitry Arkhipov — 06.05.2025, 12:50**
`<dialog>` DOM element is a great default implementation. It may be configured via CSS mostly, but that is even better - less code, more power and stability 🙂 
I always stay for: write less, do more
**bigopon — 06.05.2025, 12:51**
but even with that TSettings, it needs to kind of extend some basic interface, like this
```ts
DialogService.open<TModel, TSettings extends TDefaultSettings>

interface TBaseSettings {
  model?:
  component?:
  template?:
  renderer?:
}
```
maybe?
**Dmitry Arkhipov — 06.05.2025, 12:53**
I think we have a little misunderstanding about settings term. Let me clarify a bit
**bigopon — 06.05.2025, 12:53**
actually
**Dmitry Arkhipov — 06.05.2025, 12:55**
I would better call this IDialogParameters or smth like that. Because these are actually just params for the open() call.
And under settings I mean properties, that are passed through to the renderer and take effect on rendered dialog behavior/visualization. 
So I would change this:
```ts
interface TBaseSettings {
  model?:
  component?:
  template?:
  renderer?:
}
```
To this:
```ts
interface TDialogParameters<TRendererSettings> {
  model?:
  component?:
  template?:
  renderer?:
  settings?: TRendererSettings
}
 ```
And `TRendererSettings` would be like this for native DOM `<dialog>` (because this implementation does not have any other JS settings, anything else is congifured via CSS):
```ts
 interface INativeRendererSettings {
  asModal: boolean
}
 ```
**Dmitry Arkhipov — 06.05.2025, 13:03**
And for old Aurelia renderer:
```ts
interface IAureliaRendererSettings {
  lock?: boolean;
  startingZIndex?: number;
  // and all other setting that affect current default renderer https://docs.aurelia.io/aurelia-packages/dialog#the-dialog-settings
}
 ```
And for my Bootstrap renderer (https://github.com/ekzo-dev/aurelia-components/blob/main/packages-adapters/bootstrap/src/components/modal/dialog-dom-renderer.ts):
```ts
interface IBootstrapRendererSettings {
  animation?: boolean;
  centered?: boolean;
  scrollable?: boolean;
  // and all settings that are supported by Bootstrap modals https://getbootstrap.com/docs/5.3/components/modal/
}
```
**Dmitry Arkhipov — 06.05.2025, 13:16**
You have this example of .open() call in docs:
```ts
dialogService.open({
  component: () => Alert,
  lock: false,
  startingZIndex: 10,
})
```
you see that lock/startingZIndex settings are mixed together with open parameters (model/component/template). So in the concept of multiple configurable renderers I would change this exact example to:
```ts
dialogService.open({
  component: () => Alert,
  settings: {
    lock: false,
    startingIIndex: 10
  },
})
```
**bigopon — 06.05.2025, 13:41**
yeah I think this is better
I actually sometimes felt uncomfortable with the current mix as well just the tendency not to cause a huge break in the dialog API stopped me
but maybe this is better
**Dmitry Arkhipov — 06.05.2025, 13:43**
I think V2 is a good point to make needed braking changes. And compat plugin may cover that if desired.
**bigopon — 06.05.2025, 13:45**
yeah ok lets do it
probably I didn't do it because I never laid out the API the way we are discussing as well, so your example made it clearer to me what's better
**Dmitry Arkhipov — 06.05.2025, 13:47**
Also, could you fix that? It's annoying 🙂
`component: () => Alert`
So that I could do this:
`component: Alert`
Remember we talked about generic syntax for passing dynamic composition components to dialog/au-compose/router?
**bigopon — 06.05.2025, 13:49**
yeah it should be possible to distinguish a fn from a custom element
yup, it should be quite similar now

packages/dialog/src/dialog-interfaces.ts
```diff
@@ -15,7 +15,8 @@ export interface IDialogService {
    * @param settings - Dialog settings for this dialog instance.
    * @returns Promise A promise that settles when the dialog is closed.
    */
-  open(settings?: IDialogSettings): DialogOpenPromise;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  open<TOptions, TModel = any, TVm extends object = any>(settings: IDialogSettings<TOptions, TModel, TVm>): DialogOpenPromise;
```
bigopon, line 19: @ekzobrain maybe we shouldn't be enforcing a type parameter in these open calls, it seems cumbersome. The type parameter can be specified if one wants to enforce the type of the `options` object. Also, I'm thinking of splitting render options out of the main settings, so we don't have to pass the types around unnecessarily, quite a bit cleanup too. any thoughts?

Issue #2135, opened 2025-03-27
[Feature request] Add support for multiple classes in attribute class binding.


dtaalbers:
# 🙋 Feature Request

As requested by @bigopon. At the moment I have to use this with Aurelia 2:

```html
<div flex.class="myValue === 'desiredvalue'" flex-col.class="myValue === 'desiredvalue'"></div>
```

I'd like to be able to add multiple classes.

```html
<div flex,flex-col.class="myValue === 'desiredvalue'"></div>
```

This will save me a lot of overhead in my projects .

bigopon:
resolved in #2146

PR #2146, opened 2025-04-14
feat(template-compiler): add support for multiple .class values

Vheissu:
Enables binding to multiple CSS classes using a comma-separated list in the `.class` binding syntax.

This allows toggling a set of related classes together based on a single boolean expression and simplifies dynamic class binding.

Adds unit tests to verify the functionality and updated documentation.

Closes issue: https://github.com/aurelia/aurelia/issues/2135


Vheissu:
@bigopon Thank you. I believe I've addressed everything, the biggest one was refactoring the tests and breaking those out.

packages/template-compiler/src/binding-command.ts
```diff
@@ -449,7 +449,23 @@ export class ClassBindingCommand implements BindingCommandInstance {
...
+    const target = info.attr.target;
+    const rawValue = info.attr.rawValue;
+
+    // Attempt to parse as multiple classes only if a comma is present
+    if (target.includes(",")) {
+      const classes = target
+        .split(",")
+        .filter(c => c.length > 0);
+
+      if (classes.length > 1) {
```
bigopon, line 461: by handling only `classes.length` > 1, we are allowing this scenario:
```html
<div ,.class="condition">
```
I don't think it should be allowed if we are supporting `,` syntax. `,` is not a common class name, also it can be done via normal syntax. We should just throw if classes.length === 0 I guess.

Vheissu:
Spot on with the last comment @bigopon I've made that change for the `,.class` scenario and also added in an error as well, as well as added in a test case covering the error logic.

bigopon (approved):
nice work @Vheissu , I've added some tweak so that we can keep using Attribute binding. This is suitable because we only have static classes with `.class` syntax, unlike `class.bind="dynamicClasses"` syntax. And nice work on the tests & doc 👍

bigopon:
> the biggest one was refactoring the tests and breaking those out.

while tweaking something in my local, some tests were also failing and this heavily parameterised style with loops made it quite difficult/time consuming to pinpoint the failure. I probably should have a task for myself to unloop all the tests.

Vheissu:
Awesome @bigopon

Issue #2140, opened 2025-04-02
could Vite be upgraded to v6 in Aurelia-Vite plugin


ghiscoding:
# 🙋 Feature Request

Currently the Aurelia-Vite plugin is using an older version of Vite v5.3.0 and Vite is now at v6.2.4, so can it be updated to latest Vite version?

https://github.com/aurelia/aurelia/blob/6a7757fc7291e361d4db22416452f8cb311d4365/packages-tooling/vite-plugin/package.json#L54

## 🤔 Expected Behavior

be on latest Vite version (v6)

## 😯 Current Behavior

uses an older Vite version (v5)

## 💁 Possible Solution

upgrade Vite so that we got all recent security fixes from Vite and esbuild (there are open CVE on older version of Vite & esbuild)

Alexander-Taran:
@ghiscoding did you try bumping up the version yourself and checking out if it still works?

ghiscoding:
I did not, I just assumed that there were possibly breaking changes!? I don't personally use Aurelia, I just maintain Aurelia-Slickgrid and a security issue came up because of Vite 5. But anyway, the thing is that it would be better if the aurelia-plugin is updated anyway, adding npm override is not ideal

Alexander-Taran:
@ghiscoding could you make a pull request, please?

Vheissu:
I've created a PR to update this here: https://github.com/aurelia/aurelia/pull/2147

bigopon:
resolved by #2147

PR #2147, opened 2025-04-15
refactor(vite-plugin): update vite package to latest v6

Vheissu:
This addresses issue #2140 and updates to the latest Vite package version 6, which addresses possible security related issues in the existing package.


bigopon:
We have a few nested deps but it shouldn't be an issue for now.

PR #2149, opened 2025-04-22
docs(template-compiler): extensive docs for template compiler errors

Vheissu:
### DO NOT MERGE BEFORE #2148 PR is merged first. 

This was branched off of the previous branch, so it will need to be rebased before being merged.

----

Documenting template compiler error message codes into dedicated error doc pages.

This continues the work that was begun in a previous PR: https://github.com/aurelia/aurelia/pull/2148

Furthermore, the `createMappedError` method has been modified to link to the official Aurelia docs when the error is thrown in the console (allowing easier access to errors).

Future work will continue to be done in separate PR's to make changes smaller and more containable.


packages/kernel/src/errors.ts
```diff
@@ -4,8 +4,17 @@ import { safeString } from './utilities';
...
+  ? (code: ErrorNames, ...details: unknown[]) => {
+    const paddedCode = safeString(code).padStart(4, '0');
+    const message = getMessageByCode(code, ...details);
+    const link = `https://docs.aurelia.io/developer-guides/error-messages/0001-to-0023/aur${paddedCode}`;
+    return new Error(`AUR${paddedCode}: ${message}\\n\\nFor more information, see: ${link}`);
+  }
+  : (code: ErrorNames, ...details: unknown[]) => {
+    const paddedCode = safeString(code).padStart(4, '0');
+    const link = `https://docs.aurelia.io/developer-guides/error-messages/0001-to-0023/aur${paddedCode}`;
+    return new Error(`AUR${paddedCode}:${details.map(safeString)}\\n\\nFor more information, see: ${link}`);
```
bigopon, line 16: I don't think we want this for production build. There should only be only necessary information.

bigopon (commented):
the specific error pages look good, nice work👍 
for the production error change, I'm not sure it should be there. In a production build, have a single link will be likely not enough to fix some bug(s).

packages/kernel/src/errors.ts
Vheissu: Agreed. The thought process was in prod you want to equally link, but you'd ideally catch these things and deal with them in a dev environment. I refactored this in this PR and the other one.

Issue #2150, opened 2025-04-22
Localization of conditional projected content does not work correctly when only the t-params changes


Sayan751:
# 🐛 Bug Report

Repro link:

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Aurelia          |
| Language         |
| Browser          |
| Bundler          |
| Operating System |
| NPM/Node/Yarn    |

PR #2152, opened 2025-04-23
fix(i18n): reactive translation of conditional projected content - t-params

Sayan751:
# Pull Request

## 📖 Description

This handles the residual fix of reactive translation on change of t-params for a projcted content.

#2150


packages/i18n/src/t/translation-binding.ts
```diff
@@ -441,6 +441,7 @@ class ParameterBinding implements IBinding {
 
     this._scope = (void 0)!;
     this.obs.clearAll();
+    this.isBound = false;
```
bigopon, line 444: Well... 👍

bigopon (approved):

PR #2159, opened 2025-05-20
feat(runtime-html): add value converter context

Vheissu:
Introduces a mechanism for value converters to access contextual information, such as the target element, component, and binding instance like the following example:

```ts
class MyConverter {
  withContext = true;
  toView(value, context, ...args) {
    console.log('component calling this converter:', context.source);
    console.log('binding owning this converter:', context.binding);
  }
}
```

This enables more sophisticated value converter logic that can adapt based on the environment in which it is used, allowing for more flexible and reusable converters, but primarily motivated by the desire to be able to debug where value converters are being run.

Maybe mistakes and inefficiencies in the design here, but it's a preliminary implementation and this all works.


bigopon (commented):
nice work @Vheissu , I think it's a nice to have but the implementation needs to be more generic:
- the evaluation of an ast isn't always with a binding, and in our ast evaluation, it's called evaluator
- we are passing way more information than necessary while most of the time it's likely the three:
    1. target: target (dom/view model)
    2. bridge: binding (bridge)
    3. source: view model (most of the time)

    that matter.

After looking at the original request in the discourse thread again, I feel like the application should be able to know when something is invoked, and start the tracking there, instead of relying on the value converter from the framework to do it. It doesn't feel like we need to support this just yet, especially if we have to bring in quite a bit of extra weight on the simplicity of the value converter.

cc @fkleuver @Sayan751

packages/runtime-html/src/binding/binding-utils.ts
```diff
@@ -165,11 +167,74 @@ export const mixinAstEvaluator = /*@__PURE__*/(() => {
...
-      case 'fromView':
-        return 'fromView' in vc ? vc.fromView?.(value, ...args) : value;
+
+    let component;
+    let element;
+    let controller;
+
+    try {
+      // Prefer to get the component from the binding's controller if available
+      const maybeController = (this as any)._controller;
```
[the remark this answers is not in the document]
bigopon: another issue around this is we never know exactly what we are dealing with here, since binding could be from our implementation, or any custom implementation an application may have. If we are to support only the shape of the bindings we know then we should turn `_controller` into some officially public API so that custom bindings can also provide caller context properly.

Vheissu:
I've pushed up some additional changes to clean things up. Notably, a lot of those extra properties are gone. I was being too flexible in that I wanted to expose everything when we mostly just want the source and in some instances, the target too. Also, adjust some of the other stuff around resolving.

bigopon:
@Vheissu Another issue with resolver approach is it's always some extra unnecessary work in the evaluation. Because the resolver doesn't know whether it's needed or not, it always needs to prepare the value. While with the config property approach, during evaluation it's determined clearly whether we need to compute those values to pass onto the toview/fromview or not. I think this is the main difference and the 2nd reason why we wouldn't want to go with the resolver approach.

Vheissu:
> @Vheissu Another issue with resolver approach is it's always some extra unnecessary work in the evaluation. Because the resolver doesn't know whether it's needed or not, it always needs to prepare the value. While with the config property approach, during evaluation it's determined clearly whether we need to compute those values to pass onto the toview/fromview or not. I think this is the main difference and the 2nd reason why we wouldn't want to go with the resolver approach.

I think those are valid reasons. If we want to negate the need to add overhead at runtime and make this strictly opt-in and change order on fromView and toView, let's circle back.

Vheissu:
Okay @bigopon we now pass the context as the second argument if `useCallerContext` is specified in the value converter. It's purely opt-in now.

packages/runtime-html/src/binding/binding-utils.ts
```diff
@@ -165,11 +167,62 @@ export const mixinAstEvaluator = /*@__PURE__*/(() => {
...
+            const ceController = findElementControllerFor(searchNode, { optional: true });
+            component = ceController?.viewModel;
+          } else if (node && node.element) {
+            const ceController = findElementControllerFor(node.element, { optional: true });
+            component = ceController?.viewModel;
+          }
+        }
+      } catch (e) {
+        // ignore
+      }
```
bigopon, line 200: why are we swallowing the error here?

packages/runtime-html/src/binding/binding-utils.ts
```diff
@@ -165,11 +167,62 @@ export const mixinAstEvaluator = /*@__PURE__*/(() => {
...
+          if (node && node.nodeType === 3 && node.parentNode) {
+            searchNode = node.parentNode;
+          }
+          if (searchNode && searchNode.nodeType !== undefined) {
+            const ceController = findElementControllerFor(searchNode, { optional: true });
+            component = ceController?.viewModel;
+          } else if (node && node.element) {
+            const ceController = findElementControllerFor(node.element, { optional: true });
+            component = ceController?.viewModel;
+          }
```
bigopon, line 196: Please avoid code that looks like this:
```
a && b
```
Our condition should most of the time be very clear, unless a variable is to have so many different types of values and we only care about either the truthy or falsy ones. Most of the time in our code, we know we are only comparing against `null`/`undefined` so it'll be like this
```ts
a != null && b != null
```
for example.

---
For this part of the change, casting to any and accessing a private property like this is a big no, since it creates a blind spot that we likely will forget in the future. The tests may help but this still is something that we should avoid. I believe in no instance that we can afford a cast to any, especially when it comes to bindings/controllers.
There's a better way to resolve the context of a binding, like this:
```ts
const hydrationContext = this.l.get(IHydrationContext);
const controller = hydrationContext.controller;
const vm = controller.viewModel;
// etc ...
```

But this still leaves a chance for other custom bindings to break if they don't have a locator at the property `l`. Maybe we'll need to have a note that caller context is only available with standard bindings from Aurelia somewhere in the doc of this feature.

packages/runtime-html/src/resources/value-converter.ts
```diff
@@ -31,6 +31,7 @@ export type ValueConverterStaticAuDefinition = PartialValueConverterDefinition &
 export type ValueConverterType<T extends Constructable = Constructable> = ResourceType<T, ValueConverterInstance>;
 export type ValueConverterInstance<T extends {} = {}> = {
   signals?: string[];
+  useCallerContext?: boolean;
```
bigopon, line 34: maybe let's try to come up with some alternatives? the name feels quite a big long.

packages/runtime-html/src/binding/binding-utils.ts
```diff
@@ -165,11 +167,62 @@ export const mixinAstEvaluator = /*@__PURE__*/(() => {
     if (vc == null) {
       throw createMappedError(ErrorNames.ast_converter_not_found, name);
     }
+    // Check for instance useCallerContext property
+    const useCallerContext = vc.useCallerContext === true;
+    // Compose caller context
+    let callerContext: any = null;
```
bigopon, line 173: we should know the type of this variable, having an any is likely always not a good choice. If it needs to start with null then it can be
```ts
let callerContext: ICallerContext | null = null
```
The interface `ICallerContext` can be grouped with where the current interface for `ValueConverterInstance` is.

packages/runtime-html/src/binding/binding-utils.ts
```diff
@@ -165,11 +167,62 @@ export const mixinAstEvaluator = /*@__PURE__*/(() => {
...
+          }
+        }
+      } catch (e) {
+        // ignore
+      }
+      callerContext = {
+        target: (this as any).target,
+        bridge: this,
+        source: component,
+      };
```
bigopon, line 205: target is not yet standardized among our bindings, so this doesn't feel right. Also the cast to any is something we cannot employ. I don't think we need to be concerned about target right now, it can be just source and binding probably.

packages/__tests__/src/3-runtime-html/value-converters.spec.ts
```diff
@@ -59,51 +63,181 @@ describe('3-runtime-html/value-converters.spec.ts', function () {
...
+      await options.stop(true);
+    });
+
+    it('provides caller.target for custom attribute binding', async function () {
+      let capturedTarget: any = null;
+      let capturedBridge: any = null;
+      @valueConverter('attrCaller')
+      class AttrCallerConverter {
+        public useCallerContext = true;
+        public toView(v: any, caller: { target: unknown; bridge: unknown; source?: unknown }) {
```
bigopon, line 188: once we have an interface for caller context, we can replace the types in these to view with it.

packages/__tests__/src/3-runtime-html/value-converters.spec.ts
```diff
@@ -59,51 +63,181 @@ describe('3-runtime-html/value-converters.spec.ts', function () {
...
       const options = createFixture('<template> <div foo4.bind="value | woot2"></div> </template>', app, resources);
       assert.strictEqual(options.appHost.firstElementChild.getAttribute('test'), 'wOOt1');
-      await options.tearDown();
+      await options.stop(true);
     });
 
   });
 
+  describe('02. Caller Context', function () {
+    it('passes the binding as the second argument if useCallerContext is true', async function () {
```
bigopon, line 114: this seems to be the only test that deals with `fromView`, but it doesn't really assert it. We'll need to assert that too.

bigopon (commented):
Looking good @Vheissu , let's try the `IHydrationContext`, it should work but if it doesn't work then we will have to revise the whole thing, since we cannot afford to have code with uncertainty in the framework, especially around bindings/controller.

packages/runtime-html/src/binding/binding-utils.ts
```diff
@@ -165,11 +167,62 @@ export const mixinAstEvaluator = /*@__PURE__*/(() => {
...
+    }
     switch (mode) {
       case 'toView':
-        return 'toView' in vc ? vc.toView(value, ...args) : value;
+        if ('toView' in vc) {
+          if (useCallerContext) {
+            return vc.toView(value, callerContext, ...args);
```
Sayan751, line 211: I think we also need to have test for this case where we have both the context and the arguments.

packages/runtime-html/src/resources/value-converter.ts
Sayan751: Q: Isn't this information be static? Like part of the VC definition? Example: `@valueConverter({ name: 'fooBar', contextual: true })`?

packages/runtime-html/src/binding/binding-utils.ts
```diff
@@ -165,11 +167,62 @@ export const mixinAstEvaluator = /*@__PURE__*/(() => {
...
+          } else {
+            return vc.toView(value, ...args);
+          }
+        }
+        return value;
       case 'fromView':
-        return 'fromView' in vc ? vc.fromView?.(value, ...args) : value;
+        if ('fromView' in vc) {
```
Sayan751, line 218: A bit of nitpicking, I think we can reduce the nesting by returning early. Also, applicable for calling `toView`.

packages/runtime-html/src/binding/binding-utils.ts
```diff
@@ -165,11 +167,62 @@ export const mixinAstEvaluator = /*@__PURE__*/(() => {
...
+    let callerContext: any = null;
+    if (useCallerContext) {
+      // this is the binding instance
+      // .target is the element/attribute
+      // .component is the closest custom element view-model (if any)
+      let component;
+      try {
+        // Prefer to get the component from the binding's controller if available
+        const maybeController = (this as any)._controller;
+        if (maybeController instanceof Controller) {
```
Sayan751, line 182: Are there cases where the `_controller` is not an instance of `Controller`? If this is supposed to be null-check then probably we should stick to just that.

packages/runtime-html/src/binding/binding-utils.ts
```diff
@@ -165,11 +167,62 @@ export const mixinAstEvaluator = /*@__PURE__*/(() => {
...
+          component = maybeController.viewModel;
+        } else {
+          const node = (this as any).target;
+          let searchNode = node;
+          if (node && node.nodeType === 3 && node.parentNode) {
```
Sayan751, line 187: I think a comment denoting what the nodeType 3 means would be nice. I personally had to look up.

Sayan751 (commented):
Thanks for the awesome work @Vheissu! Left a few comments and questions.

packages/runtime-html/src/resources/value-converter.ts
Vheissu: @bigopon @Sayan751 The contextual property works nicely.

packages/runtime-html/src/binding/binding-utils.ts
Vheissu: @bigopon Agreed. I think it's fine to remove this for now as the original request was to get the context, not the target element. Devs can still employ other ways to get the target if they want it.

packages/runtime-html/src/binding/binding-utils.ts
Vheissu: I think that's a lot cleaner. My initial approach was verbose because I didn't want to get confused during implementation.

packages/runtime-html/src/binding/binding-utils.ts
Vheissu: @Sayan751 This is no longer relevant now we're using the hydration context.

packages/runtime-html/src/binding/binding-utils.ts
Vheissu: @Sayan751 I did have nodeType 3 used elsewhere with a comment, but not here. This code has been removed because of the addition of hydration context anyways.

packages/runtime-html/src/binding/binding-utils.ts
Vheissu: I didn't know what kind of error to throw here or if this code would remain, so it was placeheld until ready for merge.

packages/runtime-html/src/binding/binding-utils.ts
```diff
@@ -165,11 +167,41 @@ export const mixinAstEvaluator = /*@__PURE__*/(() => {
     if (vc == null) {
       throw createMappedError(ErrorNames.ast_converter_not_found, name);
     }
+    // Get the value converter definition to check for contextual
+    const def = ValueConverter.find(this.l as IContainer, name);
+    const contextual = def?.contextual === true;
```
bigopon, line 172: @Sayan751 @Vheissu we already have `signals` being an instance property, we should have this property as an instance property too.
Though I do feel the annoying aspect of the mix here. Something like signals and context should be static but it'll require us to do extra work here, as you can see from the above 3 lines of code: we have to do an extra look up for the value converter definition beside the instance. This is not really desirable I believe. For now I think it's best to just leave everything on the instance.

packages/runtime-html/src/resources/value-converter.ts
```diff
@@ -23,7 +23,9 @@ import type {
 import { ErrorNames, createMappedError } from '../errors';
 import { getDefinitionFromStaticAu, type IResourceKind } from './resources-shared';
 
-export type PartialValueConverterDefinition = PartialResourceDefinition;
+export type PartialValueConverterDefinition = PartialResourceDefinition & {
+  contextual?: boolean;
```
bigopon, line 27: I'm not a native so I don't know how it would sound. But to me, this is quite confusing how it's related to a parameter being passed into the toView/fromView. What about `withContext`?

packages/__tests__/src/3-runtime-html/value-converters.spec.ts
```diff
@@ -59,51 +62,205 @@ describe('3-runtime-html/value-converters.spec.ts', function () {
...
+      const app = class { public value = 'foo'; };
+      const options = createFixture('<template> <input value.bind="value | callerAware"></template>', app, resources);
+      // toView assertion
+      const input = options.appHost.querySelector('input');
+      assert.strictEqual(input.value, 'foo-called');
+
+      input.value = 'bar';
+      input.dispatchEvent(new Event('input', { bubbles: true }));
+
+      await Promise.resolve();
```
bigopon, line 133: target updating source is immediate, you don't need to wait for a promise resolve. Also options.stop(true) is immediate, you don't need to wait. If we ever change those to asynchronous, we can do it all together in that future PR instead.

bigopon (commented):
nice work @Vheissu 👍

bigopon:
I think there should also be some tests related to au-slot + context beside the set of tests we currently have in this PR.

packages/runtime-html/src/binding/binding-utils.ts
Vheissu: Happy to put this onto the value converter instance itself. It does remove the need to do the extra lookup which simplifies things and reduces the surface area.

packages/runtime-html/src/resources/value-converter.ts
Vheissu: I'm happy with `withContext`

packages/__tests__/src/3-runtime-html/value-converters.spec.ts
Vheissu: @bigopon It won't let me call a promise method like `stop` without awaiting it and this seems to be a pattern we use elsewhere, also in instances where we were using `tearDown`. But the `Promise.resolve` has been removed.

bigopon: I just remembered that we hooked into fixture creation and automatically dispose of them. You won't have to deal with disposal if it's not part of your tests. Let's remove all of `fixture.stop()` (where fixture comes from `createFixture`)

packages/runtime-html/src/index.ts
```diff
@@ -260,6 +260,7 @@ export {
   type ValueConverterDecorator,
   type ValueConverterType,
   type ValueConverterInstance,
+  type ICallerContext,
```
bigopon, line 263: this should likely be reexported in the aurelia package, but it's fine for now.

Sayan751 (approved):
LGTM! The usage of IHydrationContext brings clarity. Awesome work @Vheissu!

Issue #2161, opened 2025-05-22
Recognizing router path leads to double-decoding of a URI


PetrMotlicek:
# 🐛 Bug Report

Having 

defined route path `/tag/*parameter`

and

the HTML template

```html
<a load="/tag/26%25>26%</a>
```

or navigating via the URL bar in a browser

`http://XYZ:5555/#/tag/26%25`

leads to runtime error **"URIError: URI malformed"** in a browser.

## 🤔 Expected Behavior

There is no runtime error, just because URI component is encoded. 
URI component with the `%HEX` sequences should  not  be decoded TWO times.

A developer can use `enocdeURIComponent('%/?=&')` without side effects, after recognizing route, even with parameters: e.g. the slash character does not create another sub-path for the '*parameter'.

The **route recognizer** should not decode whole URI at once, but just each single URI component/fragment.

## 😯 Current Behavior

The `'%'`is encoded as `%25` (using `encodeURIComponent`) and put into the load attribute as `/tag/*parameter` for the router.

The load attribute tries to recognize a route, and it calls first

```TypeScript

// @aurelia/router-recongizer/src/index.ts

 private $recognize(path: string): RecognizedRoute<T> | null {
    path = decodeURI(path); // THE FIRST DECODE of WHOLE URI

    if (!path.startsWith('/')) {
      path = `/${path}`;
    }

    if (path.length > 1 && path.endsWith('/')) {
      path = path.slice(0, -1);
    }

    const result = new RecognizeResult(this.rootState);
    for (let i = 0, ii = path.length; i < ii; ++i) {
      const ch = path.charAt(i);
      result.advance(ch);

      if (result.isEmpty) {
        return null;
      }
    }

    const candidate = result.getSolution();
    if (candidate === null) {
      return null;
    }

    const { endpoint } = candidate;
    const params = candidate._getParams();

    return new RecognizedRoute<T>(endpoint, params);
  }
```

Then, a route is correctly recognized as `/tag/*parameter`, but parameter is DECODED SECOND TIME
```TypeScript

// @aurelia/router-recongizer/src/index.ts

export class RecognizedRoute<T> {
  public readonly params: Readonly<Record<string, string | undefined>>;
  public constructor(
    public readonly endpoint: Endpoint<T>,
    params: Readonly<Record<string, string | undefined>>,
  ) {
    const $params: Record<string, string | undefined> = Object.create(null);
    for (const key in params) {
      const value = params[key];
[...]

PR #2162, opened 2025-05-22
fix(route-recognizer): uri component decoding

Sayan751:
#2161

# Pull Request

### DIFF
--- a/packages/__tests__/src/router-lite/resources/load.spec.ts
+++ b/packages/__tests__/src/router-lite/resources/load.spec.ts
@@ -1286,4 +1286,56 @@ describe('router-lite/resources/load.spec.ts', function () {
 
     await au.stop(true);
   });
+
+  it('encoded path segment works correctly', async function () {
+    @route('p1/*rest')
+    @customElement({ name: 'p-1', template: `p1 \${rest}` })
+    class P1 {
+      public rest: string;
+      public loading(params: Params, _next: RouteNode, _current: RouteNode): void {
+        this.rest = params.rest;
+      }
+    }
+
+    @route({ routes: [P1] })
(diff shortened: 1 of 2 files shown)


Vheissu:
Looking good @Sayan751 

Just one nitpicky thing. The tests seem to use a repetitive code pattern. I'm wondering if maybe you use a data-driven approach if it would be cleaner. It makes for adding future test cases a little easier and also easier to read. Not a biggy, but just thought it cleaned things up a little.

**example**
```javascript
const testCases = [
  { input: 'foo/bar', encoded: 'foo/bar', description: 'slash' },
  { input: 'foo%baz', encoded: 'foo%25baz', description: 'percent' },
  { input: 'foo?baz', encoded: 'foo%3Fbaz', description: 'question mark' },
  { input: 'foo#baz', encoded: 'foo%23baz', description: 'hash' },
  { input: 'foo&baz', encoded: 'foo%26baz', description: 'ampersand' },
  { input: 'foo😀baz', encoded: 'foo%F0%9F%98%80baz', description: 'emoji' }
];

for (const [index, testCase] of testCases.entries()) {
  const round = index + 1;
  rootVm.route = `p1/${encodeURIComponent(testCase.input)}`;
  await queue.yield();
  
  assert.strictEqual(
    anchor.getAttribute('href').endsWith(`p1/${testCase.encoded}`), 
    true, 
    `round#${round} (${testCase.description}) - href encoding`
  );
  
  anchor.click();
  await queue.yield();
  assert.html.textContent(host, `p1 ${testCase.input}`, `round#${round} (${testCase.description}) - content`);
}
```

bigopon (approved):
nice work @Sayan751

Sayan751:
@Vheissu Thanks! Made the changes.

PR #2171, opened 2025-06-07
fix(i18n): Add isBound check in TranslationBinding.handleChange to prevent AUR0203

### LINKED ISSUE #2170: [i18n] TranslationBinding.handleChange() lacks isBound check, leading to AUR0203 error when unbound
# 🐛 Bug Report
The `handleChange()` method in the `TranslationBinding` class of the `@aurelia/i18n` package does not check if the binding is unbound (`isBound` flag) before processing changes. This leads to an `AUR0203` error when the binding’s `_scope` becomes undefined after being unbound, typically in scenarios involving conditional rendering with `if.bind`.

## 🤔 Expected Behavior

The `TranslationBinding.handleChange()` method should check the `isBound` flag and return early if the binding is unbound, preventing attempts to evaluate expressions with an undefined `_scope`. This aligns with the behavior in `PropertyBinding.handleChange()` and the comment in `SubscriberRecord`:
```Typescript
/**
 * Note: change handlers may have the side-effect of adding/removing subscribers to this collection during this
 * callSubscribers invocation, so we're caching them all before invoking any.
 * Subscribers added during this invocation are not invoked (and they shouldn't be).
 * Subscribers removed during this invocation will still be invoked (and they also shouldn't be,
 * however this is accounted for via $isBound and similar flags on the subscriber objects)
 */
```

## 😯 Current Behavior

The `TranslationBinding.handleChange()` method does not check isBound, causing it to process changes after the binding is unbound. This results in `astEvaluate()` calls with an undefined `scope`, triggering an AUR0203 error.

## 💁 Possible Solution

Add the following check at the start of TranslationBinding.handleChange()
https://github.com/aurelia/aurelia/blob/23d83f0e59af6cfa80169b85de0c95179e11e12d/packages/i18n/src/t/translation-binding.ts#L198
```Typescript
if (!this.isBound) {
    return;
}
```

## 🔦 Context

[Strange behavior of t.bind in Aurelia 2](https://discourse.aurelia.io/t/strange-behavior-of-t-bind-in-aurelia-2/6443/5?u=graycrow)

## 🌍 Your Environment

| Aurelia          | 2.0.0-beta.24 |
| Language         | Typescript 5.8.3 |
| Browser          | Mozilla Firefox 139.0b9 (64-bit) |
| Bundler          | Webpack 5.99 |
| Operating System | Manjaro Linux |
| NPM/Node/Yarn    | 11.4.1/v23.11.1/4.9.2 |

graycrow:
# Pull Request

## 📖 Description
Fixes #2170 by adding an `isBound` check to `TranslationBinding.handleChange()` in `@aurelia/i18n`. This prevents an `AUR0203` error when the binding is unbound, such as in scenarios with `if.bind` and `t.bind`.

### 🎫 Issues

#2170

## 📑 Test Plan

- Ran `npm run dev -- --dev i18n --test i18n` to ensure no regressions.

### DIFF
--- a/packages/__tests__/src/i18n/t/translation-integration.spec.ts
+++ b/packages/__tests__/src/i18n/t/translation-integration.spec.ts
@@ -322,6 +322,28 @@ describe('i18n/t/translation-integration.spec.ts', function () {
       assertTextContent(host, 'span', translation.simple.text);
     }, { component: App });
   }
+  {
+    @customElement({ name: 'app', template: `<span if.bind='obj.condition'><span t.bind='obj.key'></span></span>` })
+    class App {
+      public obj: { key: string; condition: boolean } = { key: 'simple.text', condition: true };
+
+      public changeCondition() {
+        this.obj = { key: 'simple.text', condition: false };
+      }
+    }
+    $it('does not throw AUR0203 when handleChange is called after unbind in if.bind/t.bind scenario', function ({ host, app, ctx, en: translation }: I18nIntegrationTestContext<App>) {
+      assertTextContent(host, 'span > span', translation.simple.text, 'initial rendering');
+
+      app.changeCondition();
+      ctx.platform.domQueue.flush();
+      assert.equal((host as Element).querySelector('span > span'), null, 'inner span removed after unbind');
+
+      assert.doesNotThrow(() => {
+            app.obj.key = 'simple.attr';
+            ctx.platform.domQueue.flush();
+        }, 'AUR0203 error should not be thrown');
+    }, { component: App });
+  }
 
   describe('translation can be manipulated by using t-params', function () {
     {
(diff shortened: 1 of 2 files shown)


Sayan751:
Thanks @graycrow for the PR! Can you please also add a test asserting the change?

graycrow:
> @graycrow Is it possible that the non-webpack bundles are failing silently.? Can you try the following in your test to check if your change is covered?
> 
> ```ts
> public changeCondition() {
>   this.obj = { condition: false };
> }
> ```

Yes, in this case, when the whole object is re-assigned, the `handleChange()` method is being called, so I assume the PR change will be covered. The changed part:
```typescript
    class App {
      public obj: { key: string; condition: boolean } = { key: 'simple.text', condition: true };

      public changeCondition() {
        this.obj = { key: 'simple.text', condition: false };
      }
    }
```
The test passes when `isBound` is checked and fails when it is not.

Sayan751 (changes requested):
LGTM. Suggested some minor changes.

Sayan751 (approved):
Thanks @graycrow!

Issue #2177, opened 2025-06-16
Complete ui-virtualization plugin


Vheissu:
We currently have a very basic ui-virtualization plugin which offers virtual repeat functionality, but is lacking numerous features which are currently documented as to implement. For release candidate we ideally want to avoid todos in the codebase, especially for improvements like these which are not massive, but make the world of difference for end developers using the plugin.

- [ ] Support for configurable properties such as buffer size, item height, min number of views
- [ ] Support for infinite scroll via event callbacks with directional and list data (allows listening to scroll events and populating accordingly)
- [ ] Support for horizontal scrolling (currently only vertical is supported)
- [ ] Support for variable sizes in width and height (we will want to document this is expensive and not recommended as it will require cumulative size arrays for efficient position calculations, binary search for finding items by scroll position and modified buffer calculations to use actual cumulative sizes instead of fixed multiplations). I've already written down notes on how this would work.
- [ ] Will require documentation updates
- [ ] Test cases covering new functionality whilst ensuring existing functionality does NOT break. None of the new changes should affect anything we already have.

Estimated development time is 10 hours (work has already commenced prior to this issue)

bigopon:
resolved by #2178

PR #2178, opened 2025-06-16
WIP feat(ui-virtualization): 

Vheissu:
This completes the todo items for the UI virtualization plugin.

- Support for variable item heights (expensive from a layout calculation perspective and documented to use with caution)
- Horizontal scrolling support for horizontal layouts
- Infinite scrolling functionality using event driven approach
- Better configurability options

Aforementioned issue here: https://github.com/aurelia/aurelia/issues/2177

### DIFF
--- a/packages/ui-virtualization/src/virtual-repeat-dom-renderer.ts
+++ b/packages/ui-virtualization/src/virtual-repeat-dom-renderer.ts
@@ -15,7 +15,7 @@ export class DefaultDomRenderer implements IDomRenderer {
     protected p: IPlatform,
   ) { }
 
-  public render(target: HTMLElement | IRenderLocation): IVirtualRepeatDom {
+  public render(target: HTMLElement | IRenderLocation, layout: 'vertical' | 'horizontal' = 'vertical'): IVirtualRepeatDom {
     const doc = this.p.document;
     const parent = target.parentNode as Element;
     // Todo: should this ever happen?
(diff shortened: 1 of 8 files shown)


packages/ui-virtualization/src/virtual-repeat-dom-renderer.ts
```diff
@@ -60,8 +61,23 @@ class DefaultDom implements IVirtualRepeatDom {
...
-    this.top.style.height = `${this.tH = top}px`;
-    this.bottom.style.height = `${this.bH = bot}px`;
+    if (this.layout === 'horizontal') {
+      this.top.style.width = `${this.tH = top}px`;
+      this.bottom.style.width = `${this.bH = bot}px`;
+      // Reset height and set display to inline-block for horizontal layout
+      this.top.style.height = '100%';
+      this.bottom.style.height = '100%';
+      this.top.style.display = 'inline-block';
+      this.bottom.style.display = 'inline-block';
```
bigopon, line 71: I don't think we can just support table this way. Having a `<tr>` with `inline-block` display and fixed width probably won't work. We better throw when detecting horizontal layout is being used with table instead.

bigopon (commented):
Looks good at a glance @Vheissu , can you test the changes with some real browser-based rendering?
There's a local e2e test setup at `packages/__e2e__/8-ui-virtualisation` that you can run like this:
```
npm run dev -- -d ui-virtualization -e 8-ui-virtualization
```
then you can head over localhost:9008 to add the horizontal scrolling test apps.

ekzobrain:
@bigopon this one is ready, you may review

PR #2182, opened 2025-06-17
feat(state): adds memoized selector creator

Vheissu:
Implements `createSelector` to memoize derived state, preventing unnecessary recomputations. This improves performance by recalculating values only when their dependencies change.

Provides a clear and centralised way to define and share complex derived logic across components.

This is a PR branched from the middleware PR and will need to be rebased on main once the other PR is merged first.

### DIFF
--- a/docs/user-docs/aurelia-packages/state.md
+++ b/docs/user-docs/aurelia-packages/state.md
@@ -162,6 +162,51 @@ export class AutoSuggest {
 
 With the above, whenever the state changes, it will ensure the `keywords` property of the view model stays in sync with the `keywords` property on the global state.
 
+### Memoizing derived state
+
+Expensive computations in `@fromState` selectors will run on every state change by default. To avoid unnecessary work, the `createSelector` helper allows you to memoize derived values so they are recomputed only when their dependencies actually change.
+
+```ts
+import { fromState, createSelector } from '@aurelia/state';
+
+interface State { items: number[]; }
+
+const selectTotal = createSelector(
+  (s: State) => s.items,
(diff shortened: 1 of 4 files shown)


Vheissu:
@bigopon 

The current implementation supports multiple dependencies through a multi-stage approach. I stole the approach that's taken in other libraries.

## 1. Dependency Extraction
Each selector function extracts its own slice of state:
```typescript
const inputs = selectors.map(fn => fn(state));
// Example: [userSlice, settingsSlice, themeSlice]
```

## 2. Memoisation Strategy
We memoise based on reference equality of all extracted values:
```typescript
if (
  lastInputs !== undefined &&
  inputs.length === lastInputs.length &&
  inputs.every((v, i) => v === lastInputs![i])  // Reference comparison
) {
  return lastResult; // Skip recomputation
}
```

## 3. Final Value Derivation
The result function receives the extracted values (not the validation results):
```typescript
lastInputs = inputs;
return (lastResult = resultFn(...inputs));
```

TLDR;
- Memo validation checks if any dependency changed by reference
- Result computation uses the actual extracted values, not validation results
- Multiple dependencies are tracked by comparing arrays of extracted values
- Recomputation only happens when at least one dependency reference changes

This approach ensures that expensive computations only run when their actual inputs change, regardless of how many dependencies are involved. The approach seemed quite sound. I also made sure we had tests around multi-dependency usage too.

bigopon:
I see, its a good addition. Only the name needs to be changed to make it more obvious what the fn is about. Ill do a review again later today. Thanks @Vheissu

Issue #2089, opened 2025-01-10
Au2 repeat.for removes the first item from DOM when adding the second.


haltesh:
# 🐛 Bug Report

When the repeat component has a single item, and we add a second one, the first row is is removed from DOM and immediately re-added. This happens regardless of whether we're using keyed iteration or not.

Since the item is removed from DOM, focus is also lost.

As long as we have two or more items in the array, we can add additional items without the problem occurring.

## 🤔 Expected Behavior

As long as the item's new position in the DOM is the same as before, it should not be removed.

## 💻 Code Sample

https://stackblitz.com/edit/au2-repeat-for-keyed-mnw5drd9-druwxluu?file=src%2Fmy-app.html

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Aurelia          |  2.1.0-dev.202501092330

bigopon:
Thanks @haltesh for this issue, the fix has been merged. Will be out the next beta. We originally didn't have this bug, hence no tests and it slipped in a refactoring a while ago, probably.

PR #2183, opened 2025-06-18
fix(runtime-html): prevent unnecessary DOM manipulation in repeater

Vheissu:
This one led me down a few false paths and console.log statements before I realised what was causing the aforementioned bug in issue #2089 (was a fun little debugging exploration effort).

The problematic code was calling `view.nodes.link()`unconditionally for ALL views:

```ts
for (; i >= 0; --i) {
  view = views[i];
  next = views[i + 1];

  view.nodes.link(next?.nodes ?? _location); // Called for ALL views

  if (indexMap[i] === -2) {
    // Handle new views
  } else if (j < 0 || i !== seq[j]) {
    // Handle views that need to move
  } else {
    // Handle views that don't need to move
  }
}
```

The `view.nodes.link()` method causes DOM nodes to be removed from their current position and re-linked, even for views that don't need to move. This was happening to the first item when adding a second item because all views were being processed through this linking operation.

To fix this I moved the `view.nodes.link()` call inside the appropriate conditional branches so it's only executed when actually needed.

TL;DR

- Views that are already in the correct position (else branch) no longer have their DOM nodes manipulated
- Only new views `(indexMap[i] === -2)` and views that need repositioning `(j < 0 || i !== seq[j])` have their nodes linked
- Preserves the existing logic for the longest increasing subsequence algorithm that minimises DOM operations
- Added test cases for this issue to confirm the fix (I made them fail first, then wrote the fix and confirmed they passed) and also confirmed existing repeater tests continued to pass


bigopon:
Thanks @Vheissu , its nice to see we are looking at the repeater with fresh eyes. I'm not sure about `link()` removing dom part, since this is what `link()` does:
image

It doesn't remove dom, or modify it.

All the tests passed so i think it's a potentially good change, though I think we can take this opportunity and see if we can remove this link altogether.

cc @fkleuver

packages/__tests__/src/3-runtime-html/repeat.single-item.regression.spec.ts
```diff
@@ -0,0 +1,47 @@
+import { assert, createFixture } from '@aurelia/testing';
+
+describe('3-runtime-html/repeat.single-item.regression.spec.ts', function () {
```
bigopon, line 3: in my local, these two tests aren't failing with the existing linkage code. Can you verify again @Vheissu ?

bigopon:
typical issue with codecov, -5% coverage even though there's nothing that could make that big impact. We'll just merge.

Issue #2111, opened 2025-01-30
Pass custom element directly to binding `<board left.bind="<card/>" />`


tamagosante:
# 🙋 Feature Request
Pass custom element directly to binding, eg. `<board left.bind="<card/>" />`

## 🤔 Expected Behavior

- passed custom element will be rendered as expected (cf. React's way `<Board left={<Card/>} />`
 

## 😯 Current Behavior

A version like the request is possible via
```html
parent
<board left="${myComponent}">

board.html
<au-compose component.bind="left">
```

```ts
class Board {
    myComponent = { }
    myComponent = class { }
    myComponent = CustomElement.define(...)
    myComponent = myComponent = somePromises
}
```

Repro link:

## 💁 Possible Solution

- be able to bind directly to custom elements

## 🔦 Context

- While `<au-compose>` or `<au-slot>` are powerful alternatives, I find the proposed syntax to be both convenient and descriptive
- Looking at the "Current Behavior", just a single place can be modified

## 💻 Examples

```html
parent
<board left.bind="<card/>" />

board.html
<template>
     ${left} <section>main stuff<section/>
</template>

<template>
    <template if.bind="left">${left || default}<template/>
     <template else>Another thing<template/>
     <section>main stuff<section/>
</template>
```

bigopon:
The benefit of this syntax in JS templating is `<card />` is understood via its props. In our template, `<card />` sits inside a string making it harder/less beneficial.
This request should be broken down into 2 parts:
1. the ability to parse `.bind="<card />"` into creating an instance of `<card />` on the spot for the binding (with all of the observation and new instance whenever props of `<card />` changes, assumed we support `.bind="<card prop.bind="..." />"`
2. the ability for content binding to support adding/removing an instance of a custom element:
    ```html
    <div>${left}</div>
    ```
    if `left` was a custom element instance, then the binding should somehow understand how to retrieve the view of the component and render it there.

- For (1), I think is not as good as it sounds when we start adding props:
    ```html
    <board left.bind="<card title="...">">
    ```
    The `<au-slot/>` should be better:
    ```html
    <board>
      <card au-slot="left" title="...">...</card>
    </board>
    ```
- For (2), maybe it could be an enhancement where we automatically compose & render, making content binding work similarly like `<au-compose/>`:
    ```html
    <div>
        ${someComponent}
    </div>
    ```
    kind of roughly translate as
    ```html
    <div>
        <au-compose component.bind="someComponent"></au-compose>
[...]

PR #2185, opened 2025-06-22
feat(runtime-html): allow components in interpolation

Vheissu:
This allows rendering components in views using interpolation as discussed in issue #2111 - I did quickly look at passing the component using the React style syntax, until the caveats listed in the linked issue became immediately obvious if you wanted bindable properties.

This was a new part of the codebase for me, so it might not be 100% concrete. But it works and follows similar rendering location logic that we use in `au-compose` - it's funny, looking at the PR, the changes don't look that dramatic, but there was a bit of testing and thinking to get to this point.


packages/runtime-html/src/binding/content-binding.ts
```diff
@@ -219,4 +275,138 @@ export class ContentBinding implements IBinding, ISubscriber, ICollectionSubscri
     }, queueTaskOptions);
     task?.cancel();
   }
+
+  /** @internal */
+  private _isCustomElementValue(value: unknown): boolean {
```
bigopon, line 280: if we are to ever do this, we likely will have only one possible value, it must be a custom element constructor. It's not easy to guarantee we have comprehensive checks for other types of values.

bigopon (changes requested):
I think we need to have a clear RFC with some basic spec'd out so we know what we need. This PR can serve as a POC, but we won't be able to go with such a big API surface.

fkleuver:
Closing this proof of concept. The proposal moved to #2199, and that discussion did not establish component interpolation as a supported direction.

PR #2186, opened 2025-06-26
chore: renamed the routers 

Sayan751:
# Pull Request

### BREAKING CHANGE 

The routers have been renamed. The `@aurelia/router` is renamed to `@aurelia/router-direct` and the `@aurelia/router-lite` is renamed to `@aurelia/router`. No change of functionalities have been introduced for either of those routers. Hence, updating to the next version of Aurelia2 would require the following changes.

#### For `@aurelia/router-lite` users

```diff
- import { ... } from '@aurelia/router-lite';
+ import { ... } from '@aurelia/router';
```

#### For `@aurelia/router` users

```diff
- import { ... } from '@aurelia/router';
+ import { ... } from '@aurelia/router-direct';
```

## 👩‍💻 Reviewer Notes

- The docs are updated as well.
- The two failing pipeline tasks, involving working with git merge, should be understandable. Not sure how to "fix" those or if it is necessary.


Vheissu (approved):
This looks good to me!

bigopon (approved):
only a few notes but we can go with this first. I'm able to review all of the files, the UI kind of froze, and most are about renaming so I think it's alright.

ghiscoding:
@Sayan751 @bigopon This PR did not show up at all in the new release Beta 25 even though it's a breaking change. I've only tried the 2.1-dev yesterday (as per @bigopon ping to test v2.1-dev), is that supposed to be included in Beta 25 though? If so, you should definitely put more highlight in the release for this breaking change. The PR is tagged as "chore" which might be why it didn't show in the release changes, but it's a rather important breaking change

bigopon:
@ghiscoding indeed, this PR happened a bit while ago and it completely went off my radar when i was composing the changelog. Quite an oops. I've added it to the change log https://github.com/aurelia/aurelia/releases/tag/v2.0.0-beta.25, thanks @ghiscoding

PR #2191, opened 2025-07-04
fix(runtime): don't wrap non-configurable and non-writable object properties

ekzobrain:
Currently proxy observer violates Proxy spec (see Notes section): https://262.ecma-international.org/8.0/#sec-proxy-object-internal-methods-and-internal-slots-get-p-receiver and browsers throw an exception "TypeError: proxy must report the same value for the non-writable, non-configurable property 'xxx'"
This PR is intended to fix it so that a proxy observer does not wrap such properties.

# Pull Request

### DIFF
--- a/packages/runtime/src/proxy-observation.ts
+++ b/packages/runtime/src/proxy-observation.ts
@@ -49,7 +49,7 @@ export function unwrap<T>(v: T): T {
 }
 
 function doNotCollect(object: object, key: PropertyKey): boolean {
-  return key === 'constructor'
+  if (key === 'constructor'
     || key === '__proto__'
     // probably should revert to v1 naming style for consistency with builtin?
     // __o__ is shorters & less chance of conflict with other libs as well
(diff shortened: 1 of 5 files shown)


fkleuver:
@bigopon On his request I also took a look at it and my recommendation was to move the descriptor check inside `doNotCollect` to prevent immutable properties from going through `connectable.observe` as well. Functionally the end result is probably the same but in this way, immutable properties won't cause "dead" observers to be created in the first place, so this seems cleaner. Current solution LGTM

packages/runtime/src/proxy-observation.ts
```diff
@@ -59,7 +59,15 @@ function doNotCollect(object: object, key: PropertyKey): boolean {
...
-    || (object.constructor as IIndexable<() => unknown>)[`${nowrapPropKey}_${rtSafeString(key)}__`] === true;
+    || (object.constructor as IIndexable<() => unknown>)[`${nowrapPropKey}_${rtSafeString(key)}__`] === true) {
+    return true;
+  }
+
+  // ensure Proxy spec compliance, value of non-configurable and non-writable property MUST be returned as is
+  // https://262.ecma-international.org/8.0/#sec-proxy-object-internal-methods-and-internal-slots-get-p-receiver
+  const descriptor = Reflect.getOwnPropertyDescriptor(object, key);
+
+  return descriptor?.configurable === false && descriptor.writable === false;
```
bigopon, line 70: thanks @fkleuver @ekzobrain I think this is the good place to check as well.

bigopon (approved):
lgtm, thanks @fkleuver @ekzobrain

PR #2195, opened 2025-07-08
feat(queue): impl queueRecurringTask

fkleuver:
## 📖 Description

In queue replacement in #2188, we left out a persistent queue feature, leaving features like dirty checker and others without a clear way to move forward. This PR adds an equivalent of persistent queue (similar to `setInterval`), and move dirty checker to use this new queue fn.

Example:

```ts
import { queueRecurringTask } from 'aurelia';

queueRecurringTask(() => {
  console.log('Called about every 1000ms');
}, { interval: 1000 })
```

The result of a `queueRecurringTask` call is a `RecurringTask`. Unlike `setInterval`, the callback of this task is influenced by the `runTask` flush, which immediately calls the callback of a recurring task, like the following example:

```ts
import { runTasks, queueRecurringTask } from 'aurelia';

queueRecurringTask(() => {
  console.log('called');
}, { interval: 1000 });

runTask(); // "called"
await new Promise(r => setTimeout(r, 1000));
// no more log, already flushed
await new Promise(r => setTimeout(r, 1000));
// 2nd called logged again
```


bigopon (approved):
lgtm, nice work @fkleuver

PR #2217, opened 2025-08-03
[wip] feat(validation): ensureGroup

Sayan751:
# Pull Request

### DIFF
--- a/packages/validation-html/src/validation-controller.ts
+++ b/packages/validation-html/src/validation-controller.ts
@@ -29,6 +29,7 @@ import {
   ValidationResult,
   IValidator,
   ValidateInstruction,
+  isGroupRule,
   type IValidationRule,
   type IValidateable,
 } from '@aurelia/validation';
@@ -405,7 +406,11 @@ export class ValidationController implements IValidationController {
         ));
         const newResults = results.reduce(
           (acc, resultSet) => {
-            acc.push(...resultSet);
+            for (const result of resultSet) {
+              if (acc.findIndex(x => x.propertyName === result.propertyName && x.rule === result.rule) === -1) {
+                acc.push(result);
+              }
+            }
             return acc;
           },
           []);
@@ -500,7 +505,7 @@ export class ValidationController implements IValidationController {
 
     return x => !x.isManual
       && x.object === instruction.object
-      && (propertyName === void 0 || x.propertyName === propertyName)
+      && (propertyName === void 0 || x.propertyName === propertyName || isGroupRule(x.rule!))
       && (
         rules === void 0
         || rules.includes(x.propertyRule!)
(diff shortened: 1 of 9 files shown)


packages/validation-html/src/validation-controller.ts
```diff
@@ -405,7 +406,11 @@ export class ValidationController implements IValidationController {
         ));
         const newResults = results.reduce(
           (acc, resultSet) => {
-            acc.push(...resultSet);
+            for (const result of resultSet) {
+              if (acc.findIndex(x => x.propertyName === result.propertyName && x.rule === result.rule) === -1) {
+                acc.push(result);
+              }
+            }
```
bigopon, line 413: hmm i find this addition confusing, can you add use some variable to explain, or some comments? or maybe replying to this comment is ok too i guess.

packages/validation/src/rule-provider.ts
```diff
@@ -168,36 +170,56 @@ export class PropertyRule<TObject extends IValidateable = IValidateable, TValue
 
     let isValid = true;
     const validateRuleset = async (rules: IValidationRule[]) => {
-      const validateRule = async (rule: IValidationRule) => {
-        let isValidOrPromise = rule.execute(value, object);
+      async function validateRule(this: PropertyRule, rule: IValidationRule) {
```
bigopon, line 173: having a `function` with `this` feels a bit uncomfortable compared to an arrow fn, it also forces us to `.call(this)`, which I think is distracting. Is this for perf reason?

packages/validation/src/rule-provider.ts
```diff
@@ -502,6 +537,47 @@ export class ValidationRules<TObject extends IValidateable = IValidateable> impl
...
+    const rules: PropertyRule[] = new Array(numProperties);
+
+    for (let i = 0; i < numProperties; ++i) {
+      const [name, expression] = parsePropertyName(properties[i] as any, this.parser);
+      // eslint-disable-next-line eqeqeq
+      let rule = this.rules.find((r) => r.property.name == name);
+      let ruleProperty: IRuleProperty;
+      if (rule == null) {
+        ruleProperty = new RuleProperty(expression, name);
+        rule = new PropertyRule(this.locator, this, this.messageProvider, ruleProperty);
```
bigopon, line 563: sometimes reading `RuleProperty` and `PropertyRule` together gets really confusing
If it's not too late, can we try again 😬 ?

bigopon (commented):
nice work @Sayan751 , only did a quick check so only a few cleanup/clarifying comments.

packages/validation-html/src/validation-controller.ts
Sayan751: Good point! Added a comment explaining the need for the filter.

packages/validation/src/rule-provider.ts
Sayan751: Not sure why I did that. Reverted to arrow function. Thanks for pointing it out.

packages/validation/src/rule-provider.ts
Sayan751: Renamed `RuleProperty` to `Property`.

PR #2224, opened 2025-08-13
feat(plugin-conventions): add 'with' support to template type-checker

fkleuver:
# Pull Request

## 📖 Description

Adds `with` support to the experimental template type-checker in `plugin-conventions`.

- Inside a `with` subtree, unqualified identifiers (e.g., `name`) are treated as properties of the RHS object. Outside the subtree, nothing leaks.
- Precedence: `let/repeat/promise locals` > `with` scope > VM members > `$parent` chain.
- `${this.x}` still refers to the VM, not the `with` scope 

Example:

```html
<template with.bind="person">
  ${name.toUpperCase()}   <!-- typed as person.name -->
</template>
````

## 👩‍💻 Reviewer Notes

When the RHS base is a VM member we index using the class union (`(ClassUnion)['prop']…`) instead of the access accumulator to avoid self-reference in the generated type. When the base is already synthetic we index the current access type.

## 📑 Test Plan

- Basic pass/fail for `<template with.bind="…">` and `with="…"`
- Precedence: locals (`let`/`repeat`/`promise`) > `with` overlay > VM > `$parent`
- Leak guard: overlay is branch-local; no visibility outside the `with` subtree
- `${this.x}` reads VM
- Primitive RHS (e.g., `42`) has no overlay (unknown names)
- Nested `with`: inner overlay wins on conflicts.
- `with` + `repeat`: verify overlay from `item` (names are not visible after the block)
- Stable diagnostics for unknown iterable roots (repeat) via synthetic throw-away identifier

### DIFF
--- a/packages-tooling/plugin-conventions/src/template-typechecking.ts
+++ b/packages-tooling/plugin-conventions/src/template-typechecking.ts
@@ -79,7 +79,7 @@ class IdentifierScope {
   private readonly decIdentMap: Map<string, number> | null;
   private readonly overriddenIdentMap: Map<string, string> = new Map();
   private readonly root: IdentifierScope;
-  public get isRoot(): boolean { return this.parent === null; }
+  public get isRoot(): boolean { return this.parent == null; }
   private promiseProperty: string | null = null;
 
   public constructor(
@@ -87,8 +87,8 @@ class IdentifierScope {
     private readonly classes: ClassMetadata[],
     public readonly parent: IdentifierScope | null = null,
   ) {
-    this.decIdentMap = parent === null ? new Map() : null;
-    this.root = parent === null ? this : parent.root;
+    this.decIdentMap = parent == null ? new Map() : null;
+    this.root = parent == null ? this : parent.root;
   }
 
   public createChild(type: ScopeType): IdentifierScope { return new IdentifierScope(type, this.classes, this); }
@@ -96,7 +96,7 @@ class IdentifierScope {
   public getOverriddenIdentifier(ident: string): string | null {
     // eslint-disable-next-line @typescript-eslint/no-this-alias
     let current: IdentifierScope | null = this;
-    while (current !== null) {
+    while (current != null) {
       const lookup = current.overriddenIdentMap;
       if (lookup.has(ident)) return lookup.get(ident)!;
       current = current.parent;
@@ -118,7 +118,7 @@ class IdentifierScope {
     }
     // eslint-disable-next-line @typescript-eslint/no-this-alias
     let current: IdentifierScope | null = this;
-    while (current !== null) {
+    while (current != null) {
       const lookup = current.overriddenIdentMap;
       if (lookup.has(ident)) return returnIdentifier(lookup.get(ident)!);
       current = current.parent;
@@ -148,8 +148,8 @@ class IdentifierScope {
   public getPromiseProperty(): string | null {
     // eslint-disable-next-line @typescript-eslint/no-this-alias
     let current: IdentifierScope | null = this;
-    while (current !== null) {
-      if (current.promiseProperty !== null) return current.promiseProperty;
+    while (current != null) {
+      if (current.promiseProperty != null) return current.promiseProperty;
       current = current.parent;
     }
     return null;
@@ -326,7 +326,7 @@ function __typecheck_template_${classes.map(x => x.name).join('_')}__() {
 
   public addPromiseResolvedProperty(identifier: string): void {
     const promiseProp = this.scope.getPromiseProperty();
-    if (promiseProp === null) throw new Error('Promise property not found');
+    if (promiseProp == null) throw new Error('Promise property not found');
     this.accessTypeParts.push((acc) => `${acc} & { ${identifier}: Awaited<${this.accessTypeIdentifier}['${promiseProp}']> }`);
   }
 }
@@ -443,7 +443,7 @@ function tryProcessPromise(syntax: AttrSyntax, attr: Token.Attribute, node: Defa
       const value = attr.value.length === 0 ? target : attr.value;
       const expr = ctx.exprParser.parse(value, 'None');
       // TODO: handle primitive literal
-      if (expr !== null) {
+      if (expr != null) {
         const [_expr] = mutateAccessScope(expr, ctx, member => ctx.getIdentifier(member, IdentifierInstruction.SkipGeneration | IdentifierInstruction.Promise) ?? member);
         addReplaceParts(_expr, value);
       }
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: packages-tooling/plugin-conventions/src/template-typechecking.ts [hunk 1]
```diff
@@ -238,12 +279,23 @@ function __typecheck_template_${classes.map(x => x.name).join('_')}__() {
...
-        default: return `${TypeCheckingContext._o} => (${args.map(arg => `${TypeCheckingContext._o}.${arg}`).join(',')})`;
+        case 1: {
+          // Strip a leading "this." so `${o.this.x}` type-checks as VM access `${o.x}`
+          const arg0 = args[0].startsWith('this.') ? args[0].slice(5) : args[0];
+          return `${TypeCheckingContext._o} => ${TypeCheckingContext._o}.${arg0}`;
+        }
+        default: {
+          const parts = args.map(arg => arg.startsWith('this.') ? arg.slice(5) : arg);
+          return `${TypeCheckingContext._o} => (${parts.map(arg => `${TypeCheckingContext._o}.${arg}`).join(',')})`;
+        }
```
bigopon, line 290: I think once we start having a better feature coverage, some "smoke" tests on the generated code would be nice. I have to use so much imagination every time to review the type checker code 😃 High level review will be easier with it too: skimming the tests and looking at the generated code give quite good overview on changes. Not insisting any changes in this PR though,
cc @Sayan751

packages-tooling/plugin-conventions/src/template-typechecking.ts
```diff
@@ -323,11 +375,15 @@ function tryProcessRepeat(syntax: AttrSyntax, attr: Token.Attribute, node: Defau
     propAccExpr = `${ctx.accessTypeIdentifier}['${identifier}']`;
   } else {
     const rawIterIdentifier = Unparser.unparse(expr.iterable);
-    const [iterIdent, path] = mutateAccessScope(expr.iterable, ctx, member => {
-      const newName = ctx.getIdentifier(member, IdentifierInstruction.AddToOverrides)!;
-      return newName;
-    }, true);
-    propAccExpr = `${ctx.accessTypeIdentifier}${path.map(p => `['${p}']`).join('')}`;
+    const [iterIdent, path] = mutateAccessScope(expr.iterable, ctx, member => ctx.getIdentifier(member, IdentifierInstruction.SkipGeneration) ?? ctx.getIdentifier(member)!, true);
```
bigopon, line 378: you probably have a very big screen 😆 
```suggestion
    const [iterIdent, path] = mutateAccessScope(
      expr.iterable,
      ctx,
      member => ctx.getIdentifier(member, IdentifierInstruction.SkipGeneration) ?? ctx.getIdentifier(member)!,
      true
    );
```

packages-tooling/plugin-conventions/src/template-typechecking.ts
```diff
@@ -99,25 +116,33 @@ class IdentifierScope {
...
       if (lookup.has(ident)) return returnIdentifier(lookup.get(ident)!);
       current = current.parent;
     }
-    if ((instruction & IdentifierInstruction.SkipGeneration) > 0) return null;
 
-    const lookup = this.root.decIdentMap!;
-    let count = lookup.get(ident) ?? 0;
-    lookup.set(ident, ++count);
-    const newName = `${identifierPrefix}${ident}${count}`;
+    const allocate = (): string => {
```
bigopon, line 127: 👍

packages-tooling/plugin-conventions/src/template-typechecking.ts
```diff
@@ -448,7 +554,7 @@ function processNode(node: DefaultTreeElement | DefaultTreeTextNode, ctx: TypeCh
 
         const value = attr.value.length === 0 ? target : attr.value;
         const ast = ctx.exprParser.parse(value, 'None');
-        if (ast == null) continue;
+        if (ast === null) continue;
```
bigopon, line 557: generally `== null` gives us a better insurance against nullable values I feel. It allows us to ignore the question "what happens if `undefined`/what happens if `null`" when we use `===`
For that reason, we should use `== null` instead. It's consistent with the code in the runtime packages as well.

packages-tooling/plugin-conventions/src/template-typechecking.ts
```diff
@@ -431,6 +487,56 @@ function tryProcessPromise(syntax: AttrSyntax, attr: Token.Attribute, node: Defa
...
+    const [rewritten, path] = mutateAccessScope(expr, ctx, m => ctx.getIdentifier(m, IdentifierInstruction.SkipGeneration) ?? m, true);
+    ctx.toReplace.push({
+      loc: node.sourceCodeLocation!.attrs![attr.name],
+      modifiedContent: () => `${attr.name}="\${${ctx.accessIdentifier}(${ctx.createLambdaExpression(rewritten)}, '${escape(value)}')}"`
+    });
+
+    const baseExpr = path?.length ? `${ctx.accessTypeIdentifier}${path.map(p => `['${p}']`).join('')}` : null;
+    const first = path?.[0] ?? null;
+    const env: OverlayEnv | null = baseExpr === null ? null : { mode: first?.startsWith(identifierPrefix) ? 'synthetic' : 'vm', expr: baseExpr };
+    if (env !== null) ctx.pushEnv(env);
```
bigopon, line 507: if we are to add this new mechanism, I think it should be consistently used for `promise`/`repeat` too, no?

bigopon (commented):
nice work @fkleuver , I did a high level overview mostly, look like quite a good way to handle scope layering 👍

packages-tooling/plugin-conventions/src/template-typechecking.ts
fkleuver: You may be right for `repeat` yeah. For `promise` I don't think it's necessary because `promise`/`then`/`catch` introduce explicit locals, so bare names don't implicitly resolve to their members. The overlay is only needed there are different possibilities for what root to index for types (either a VM member or a synthetic local created by repeat/let)

bigopon:
> Precedence: let/repeat/promise locals > with scope > VM members > $parent chain.

For a template that looks like this
```ts
<div with.bind="{ a: [1, 2, 3] }" repeat.for="i of a" promise.bind="fetch(i)">
```
does the spec above apply?

fkleuver:
> > Precedence: let/repeat/promise locals > with scope > VM members > $parent chain.
> 
> For a template that looks like this
> 
> ```ts
> <div with.bind="{ a: [1, 2, 3] }" repeat.for="i of a" promise.bind="fetch(i)">
> ```
> 
> does the spec above apply?

The precedence does apply, but it's evaluated per nested scope.

So what you just said, the runtime rewrites that into this:
```html
<template with.bind="{ a: [1, 2, 3] }">
  <template repeat.for="i of a">
    <template promise.bind="fetch(i)">
      <div> ... </div>
    </template>
  </template>
</template>
```

Inside the innermost template, name resolution still follows the same rule: locals (`i`) > with (`a`) > VM > $parent

Or am I misunderstanding your question?

bigopon:
> Inside the innermost template, name resolution still follows the same rule: locals (i) > with (a) > VM > $parent

if so then it's all good. I was only clarifying what was said previously, since it almost sounded like we ignore attr order.

Issue #2225, opened 2025-08-15
CSS module transformation broken for static class="" syntax inside <template>


deskjet:
# 🐛 Bug Report
Static CSS classes fail to be processed correctly inside template repeat loops when using CSS modules

## 🤔 Expected Behavior
CSS classes defined with static `class="class-name"` syntax should be processed and transformed correctly by CSS modules regardless of whether they are used inside or outside of `<template>` elements.

## 😯 Current Behavior
When using CSS modules (`.module.css` files), static CSS classes defined with `class="class-name"` syntax fail to be processed correctly when used inside `<template>` elements. 

- **Inside templates**: `class="my-li"` → CSS module transformation fails
- **Outside templates**: `class="my-li"` → CSS module transformation works correctly  
- **Workaround**: `class.bind="'my-li'"` and `my-li.class="true"` work in both contexts

## 🔦 Context
We're beginning to use `.module.css` for all our reusable custom components. Using `<template>` is quite common for repeating elements snippets that don't require a full custom element.

## 💻 Code Sample

**my-app.html:**
```html
<h1>with &lt;template&gt;</h1>
<ul>
    <template repeat.for="i of 1">
        <li my-li.class="true">my-li.class="true"</li>        <!-- works -->
        <li class.bind="'my-li'">class.bind="'my-li'"</li>        <!-- works -->
        <li class="my-li">class="my-li"</li>        <!-- fails! -->
    </template>
</ul>

<h1>without &lt;template&gt;</h1>
<ul>
    <li my-li.class="true">my-li.class="true"</li>    <!-- works -->
    <li class.bind="'my-li'">class.bind="'my-li'"</li>    <!-- works -->
    <li class="my-li">class="my-li"</li>    <!-- works -->
</ul>
```

**my-app.module.css:**
```css
.my-li::before {
    content: "✔ works ";
    color: green;
}
```

**my-app.ts:**
generated from `npx makes aurelia`

**vite.config.ts:**
generated from `npx makes aurelia`

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Aurelia          | 2.0.0-beta.25        |
| Language         | TypeScript |
| Browser          | Firefox 141 |
| Bundler          | Vite with @aurelia/vite-plugin |
| Operating System | Fedora 42 |
| NPM/Node/Yarn    | npm 10.9.2 node v22.13.1 |

Vheissu:
I've got a pull request up for this now.

PR #2226, opened 2025-08-16
fix(runtime-html): map static classes inside nested <template

Vheissu:
This fixes the issue reported here: https://github.com/aurelia/aurelia/issues/2225

- Fix CSS Modules processing so static class="…" attributes are transformed even when they appear within nested <template> elements generated by structural directives (repeat.for, if/else, switch/case).
- Replace shallow [class] selection with a recursive DOM traversal that:
- visits every element
- explicitly descends into <template>.content childNodes
- descends normal childNodes for other nodes
- Keep the original token mapping behavioir (split by whitespace, replace tokens using ICssClassMapping, preserve unknown tokens).

### DIFF
--- a/packages/runtime-html/src/templating/styles.ts
+++ b/packages/runtime-html/src/templating/styles.ts
@@ -66,34 +66,31 @@ export class CSSModulesProcessorRegistry implements IRegistry {
           }
         };
 
-        const walk = (node: Node): void => {
-          // Element node
-          if ((node as Element).tagName != null) {
-            const el = node as Element;
-            // process current element
+        const processContainer = (container: Element | DocumentFragment): void => {
+          if ((container as Element).tagName != null) {
+            const el = container as Element;
             processElement(el);
-            // if this is a <template>, also traverse into its content
+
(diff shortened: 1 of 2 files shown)


packages/runtime-html/src/templating/styles.ts
```diff
@@ -55,20 +55,45 @@ export class CSSModulesProcessorRegistry implements IRegistry {
...
+              .map(x => existingMapping![x] || x)
+              .join(' ');
+            classes.value = newClasses;
           }
-          const newClasses = classes.value.split(/\s+/g).map(x => existingMapping![x] || x).join(' ');
-          classes.value = newClasses;
-        }
+        };
+
+        const walk = (node: Node): void => {
```
bigopon, line 69: we don't need to walk all the nodes, we know that we only care about template elements, i think we can do something like this
```
1. get all elements with `[class]` + process
2. get all `template` elements + repeat step 1 for `content`
```

bigopon (commented):
looks good @Vheissu i think we can improve the search steps a bit 👍

PR #2241, opened 2025-08-27
fix(bindable): dont transform underscore in attribute names

bigopon:
## 📖 Description

Currently we treat underscores (`_`) in custom element/custom attribute props as a valid character to be replaced by `-` when transforming to kebab case, this leads to the following transformation:
```
_my_prop_    -> my-prop
my_prop    -> my-prop
```

This isn't how v1 works, and it also dismisses a valid usage of `_`. This PR changes the way attribute is parsed and converting element property names to attribute names will correctly leave `_` alone.


3cp:
LGTM

PR #2243, opened 2025-09-08
feat(router): history navigation direction marker

Sayan751:
# Pull Request

## 📖 Description

This pull request implements a history navigation direction marker for `@urelia/router`.  To detect the direction of navigation, you can use the `isBack` property from the `options: INavigationOptions` parameter in the lifecycle hooks.
The `isBack` property will be `true` if the navigation is a backward navigation (e.g., user clicked the browser's back button) and `false` for forward navigation.

```typescript
import { IRouteViewModel, INavigationOptions, Params, RouteNode } from '@aurelia/router';
import { customElement } from '@aurelia/runtime-html';

@customElement({
  name: 'my-component',
  template: `<div class="page">My Component Content</div>`,
})
export class MyComponent implements IRouteViewModel {
  public loading(
    params: Params,
    next: RouteNode,
    current: RouteNode | null,
    options: INavigationOptions
  ): void {
    if (options.isBack) {
      console.log('User navigated back to this component');
      // Apply different logic for back navigation
    } else {
      console.log('User navigated forward to this component');
      // Apply logic for forward navigation
    }
  }

  public unloading(
    next: RouteNode | null,
    current: RouteNode,
    options: INavigationOptions
  ): void {
    if (options.isBack) {
      console.log('User is navigating back from this component');
    } else {
      console.log('User is navigating forward from this component');
    }
  }
}
```

Refer to the docs for an example of leveraging this property to animate the page transitions. Following is a visual example of the same.

router-animation

cc: @davidsk

### DIFF
--- a/packages/__e2e__/6-router/playwright.config.js
+++ b/packages/__e2e__/6-router/playwright.config.js
@@ -1,2 +1,2 @@
 // @ts-check
-module.exports = require('../playwright-util')(require('./package.json'));
+module.exports = { ...require('../playwright-util')(require('./package.json')), workers: 1 };
--- a/packages/__e2e__/6-router/src/animation-hooks.ts
+++ b/packages/__e2e__/6-router/src/animation-hooks.ts
@@ -1,0 +1,58 @@
+/* eslint-disable no-console */
+import { INavigationOptions, IRouteViewModel, Params, RouteNode } from '@aurelia/router';
+import { animate, eases } from 'animejs';
+import { IController, lifecycleHooks } from 'aurelia';
+
+const durationMs = 900;
+
+// Page entering animation
+const enterFromRight = (el: HTMLElement) =>
+  animate(el, {
+    translateX: ['150%', '0%'],
+    duration: durationMs,
+    ease: eases.outCubic,
+  });
+
+const enterFromLeft = (el: HTMLElement) =>
+  animate(el, {
+    translateX: ['-150%', '0%'],
+    duration: durationMs,
+    ease: eases.outCubic,
+  });
+
+// Page exiting animation
+const exitToRight = (el: HTMLElement) =>
+  animate(el, {
+    translateX: ['0%', '150%'],
+    duration: durationMs,
+    ease: eases.inCubic,
+  });
+
+const exitToLeft = (el: HTMLElement) =>
+  animate(el, {
+    translateX: ['0%', '-150%'],
+    duration: durationMs,
+    ease: eases.inCubic,
+  });
+
+@lifecycleHooks()
+export class AnimationHooks {
+  private element: HTMLElement;
+  private isBack: boolean = false;
+
+  public created(vm, controller: IController): void {
+    this.element = controller.host;
+  }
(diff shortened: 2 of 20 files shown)


packages/__e2e__/6-router/playwright.config.js
```diff
@@ -1,2 +1,2 @@
 // @ts-check
-module.exports = require('../playwright-util')(require('./package.json'));
+module.exports = { ...require('../playwright-util')(require('./package.json')), workers: 1 };
```
bigopon, line 2: ah, this is also a way to do it, though we can also do the `test.sequential()` (or something like that, i dont remember the API exactly), so tests in one file won't have to wait for the others in another.
p/s: could be totally wrong and there's no such thing :p

packages/__e2e__/6-router/src/animation-hooks.ts
```diff
@@ -0,0 +1,58 @@
...
+  public created(vm, controller: IController): void {
+    this.element = controller.host;
+  }
+  public loading(vm: IRouteViewModel, params: Params, next: RouteNode, current: RouteNode, options: INavigationOptions): void | Promise<void> {
+    this.isBack = options.isBack;
+  }
+
+  public async unloading(vm: IRouteViewModel, next: RouteNode, current: RouteNode, options: INavigationOptions): Promise<void | Promise<void>> {
+    // the typings are weird here
+    await (options.isBack ? exitToRight(this.element).then() : exitToLeft(this.element).then());
```
bigopon, line 52: missed opportunity for some fancy visual in the PR description :D

packages/router/src/router.ts
```diff
@@ -228,7 +230,10 @@ export class Router {
         const state = isManagedState(e.state) ? e.state : null;
 
         const routerOptions = this.options;
-        const options = NavigationOptions.create(routerOptions, { historyStrategy: 'replace' });
+        const auNavId = state?.[AuNavId] ?? 0;
+        const isBack = auNavId <= this._lastLocationChangeStateId || this._lastLocationChangeStateId === 0;
```
bigopon, line 234: hmm the `this._lastLocationChangeStateId === 0` part isn't that obvious to me why. Can you elaborate?

packages/__e2e__/6-router/playwright.config.js
Sayan751: As we have multiple suites, all are working with the same URL, a parallel run of all tests create interference for the individual tests. Thus, we need serial execution.

You are right; there is [`test.describe.serial`](https://playwright.dev/docs/api/class-test#test-describe-serial). Tried that already, and that makes the tests run sequentially in that suite, as the documentation says. Even making every suite serial might not work (not tried this one though), as if the suites are run in parallel, it will cause interference for the tests in other suites.

packages/router/src/router.ts
Sayan751: Thanks for pointing it out. I thought about writing some comments there in code, but missed it. Now the comments are added. Please have another look to check if the comments can explain the reason sufficiently.

packages/__e2e__/6-router/src/animation-hooks.ts
Sayan751: Now corrected :P

PR #2260, opened 2025-10-02
feat(observation): ability to declare dependencies of a getter manually

### LINKED ISSUE #2244: Offer a Deep Computed option for Aurelia 2
# 🙋 Feature Request

Add `deepComputedFrom` and `shallowComputedFrom` attributes, as we had on Aurelia v1:
[https://www.npmjs.com/package/aurelia-deep-computed](https://www.npmjs.com/package/aurelia-deep-computed)

## 🤔 Expected Behavior

When a model property is an object with nested fields, changing the nested properties doesn't trigger change events. This means when you (for example) retrieve a nested structure from an API call, then start allowing the user to modify individual fields, any function or binding relating to the base object doesn't trigger.

## 🔦 Context

A very common example of this is when someone uses
`${myNestedObject | json}` for debugging, which will not update on-screen unless you perform something like `myNestedObject = [...myNestedObject]` each time a nested item is updated (or uses a semaphore), or does this on a timer.

bigopon:
## 📖 Description

In v1, we provide `@computedFrom` as a way to declare dependencies of a getter manually to avoid having to dirty check when observing a getter. In v2, getters are automatically observed with proxy, removing the need to provide a way to avoid dirty checking. Though sometimes it's still desirable to specify the dependencies manually, in case an application wants to control precisely, for performance as an example, or for some migration purposes.

This PR enhances the `@computed` decorator with some extra options to make it more useful, like the following examples:

### Basic usage:

```ts
import { computed } from 'aurelia';

export class ShoppingCart {
  items: CartItem[] = [];

  // we only care when there's a change in the number of items
  // but not when the price or quantity of each item changes
  @computed('items.length')
  get total() {
    // This computed property automatically updates when items change
    return this.items.reduce((sum, item) => sum + (item.price * item.quantity), 0);
  }
  // other code ...
}
```

### multiple dependencies

```ts
import { computed} from 'aurelia';

export class ShoppingCart {
  items: CartItem[] = [];
  gst = .1;

  // we only care when there's a change in the number of items, or the gst
  // but not when the price or quantity of each item changes
  @computed('items.length', 'gst')
  get total() {
    // This computed property automatically updates when items change
    return this.items.reduce((sum, item) => sum + (item.price * this.tax * item.quantity), 0);
  }

  get tax() {
    return 1 + this.gst;
  }

  // other code ...
}
```

### Flush timing with `flush`

Like how you can specify flush mode of computed getter with `@computed({ flush: 'sync' })`, flush mode of `@computedFrom` can also be done in a similar way,
like the following example:

```ts
import { computed } from 'aurelia';

export class ShoppingCart {
  items: CartItem[] = [];
  gst = .1;

  // immediately notify changes whenever the number of items or gst changes
  // instead of waiting until the next tick
  @computed({
    deps: ['items.length', 'gst'],
    flush: 'sync'
  })
  get total() {
    // This computed property automatically updates when items change
    return this.items.reduce((sum, item) => sum + (item.price * this.tax * item.quantity), 0);
  }

  get tax() {
    return 1 + this.gst;
  }

  // other code ...
}
```

### Deep observation with `deep`

Sometimes you also want to automatically observe all properties of an object recursively, regardless at what level, `deep` option on the `@computed` decorator
can be used to achieve this goal, like the following example:

```ts
import { computed } from 'aurelia';

export class ShoppingCart {
  _cart = {
    items = [],
    gst = .1,
  }

  // notify changes whenever properties of any items in the cart,
  // or the number of items,
  // or the gst changes
  @computed({
    deps: ['_cart'],
    deep: true,
  })
  get total() {
    // This computed property automatically updates when items change
    return this._cart.items.reduce((sum, item) => sum + (item.price * this.tax * item.quantity), 0);
  }

  get tax() {
    return 1 + this._cart.gst;
  }

  // other code ...
}
```

> [!WARNING]
> `deep` observation doesn't observe non-existent properties, which means newly added properties won't trigger any changes notification. Replace the entire object instead.

### 🎫 Issues

Close #2244

cc @fkleuver @Sayan751 @Vheissu @jsobell @3cp

### DIFF
--- a/docs/user-docs/essentials/reactivity.md
+++ b/docs/user-docs/essentials/reactivity.md
@@ -190,6 +190,9 @@ export class ShoppingCart {
 Now whenever `_cart.items[].price` or `_cart.items[].quantity` (or whatever else properties on each element in the `items` array),
 or `_cart.gst` changes, the `total` is considered dirty.
 
+> [!WARNING]
+> `deep` observation doesn't observe non-existent properties, which means newly added properties won't trigger any changes notification. Replace the entire object instead.
+
 ## Deep Observation
 
 Aurelia can observe nested object changes:
--- a/packages/__tests__/src/3-runtime-html/computed-decorator.spec.ts
+++ b/packages/__tests__/src/3-runtime-html/computed-decorator.spec.ts
@@ -2,7 +2,7 @@ import { computed, ICoercionConfiguration, runTasks } from '@aurelia/runtime';
 import { AppTask, bindable } from '@aurelia/runtime-html';
 import { assert, createFixture } from '@aurelia/testing';
 
-describe('3-runtime-html/computed-from.spec.ts', function () {
+describe('3-runtime-html/computed-decorator.spec.ts', function () {
   describe('async', function () {
     it('works with normal property dependency', async function () {
       let i = 0;


bigopon:
@ekzobrain in this PR, we get `getExpressionObserver` as an official API on the `IObserverLocator` interface, so `watchExpression` can reuse this, and I'd imagine sometimes it's easier for application to do that too.

ekzobrain:
Why do we need two decorators - `computed` and `computedFrom`? Lets just use new name `computed`, which will have all options: list of fields to observe/flush/deep. And create `computedFrom` in compat plugin for migration?

bigopon:
Thanks @ekzobrain , i've addressed your feedback, we now only use a single `@computed` decorator, and a `@computedFrom` is added in the compat package, with a simpler API, for migration purposes.

Issue #2245, opened 2025-09-10
Provide a `$previous` variable inside loops


jsobell:
# 🙋 Feature Request

It's not uncommon to want to render a divider or heading while looping through a collection of items.
Traditionally we save the previous item in the collection, and on each loop we check to see `if previous.whatever = current.whatever`.
In Aurelia 2 it looks like the `<let>` feature could be used to hold the previous iteration, but that doesn't work, as the bound variable doesn't behave as expected (which also need to be clarified in the doco).
Given that the loop is pre-calculated into an array before the repeat occurs, why not have a $previous that points to the previous iteration's item, to avoid having to write code referring back to the original collection with an index offset like this:

```html
<let items.bind="[{'L':'A',v:9},{'L':'B',v:8},{'L':'C',v:6},{'L':'C',v:5},{'L':'D',v:4}]"></let>

<div repeat.for="item of items">
  <div if.bind="item.L != items[$index-1].L">New section ${item.L}</div>
  <div>${item.L}, ${item.v}</div>
</div>
```

## 💁 Possible Solution

Introduce a $previous, that is null for the first iteration, then lags the current iteration by one

## 💻 Examples

```html
<let items.bind="[{'L':'A',v:9},{'L':'B',v:8},{'L':'C',v:6},{'L':'C',v:5},{'L':'D',v:4}]"></let>

<div repeat.for="item of items">
  <div if.bind="item.L != $previous.L">New section ${item.L}</div>
  <div>${item.L}, ${item.v}</div>
</div>
```

PR #2261, opened 2025-10-04
feat(repeat): add opt-in $previous contextual property for repeats

Vheissu:
This PR introduces a new opt-in `$previous` contextual property for `repeat.for`, enabling developers to access the previous item in an iteration. This relates to this user submitted request here https://github.com/aurelia/aurelia/issues/2245 — initially there were concerns that adding this property would introduce overhead for those who don't need it, so the decision was made to make this an opt-in feature.

# Motivation

Developers frequently need to compare an item with the previous item in a list to implement common UI patterns:

- **Section headers** that appear only when a grouping property changes
- **Add dividers** depending on the previous item, we might want to add a divider (for example, menus with child items)
- **Timeline/changelog interfaces** showing what changed between consecutive entries  
- **Diff-style displays** highlighting changes between adjacent items
- **Conditional styling** based on the previous item's state

Previously, this required maintaining external state or workarounds. The `$previous` property provides a clean, declarative solution.

### 1. Repeat Implementation (`packages/runtime-html/src/resources/template-controllers/repeat.ts`)

Added support for the `previous` iterator option with three key capabilities:

**Static enabling:**
```html
<div repeat.for="item of items; previous: true">
  ${$previous ?? 'null'} → ${item}
</div>
```

**Dynamic enabling via binding:**
```html
<div repeat.for="item of items; previous.bind: showPrevious">
  <!-- $previous only available when showPrevious is truthy -->
</div>
```

**Key implementation details:**
- The `_enablePrevious` flag controls whether `$previous` is set (defaults to `false` for zero overhead)
- Expression-based enabling (`previous.bind`) is evaluated once at binding time
- `$previous` is set to `null` for the first item, `undefined` when disabled
- Works seamlessly with all collection types (Array, Map, Set, Number)
- **Compatible with keyed repeats** — `$previous` correctly updates during view recycling and reordering

### 2. Template Compiler Enhancement (`packages/template-compiler/src/binding-command.ts`)

@bigopon this is worth highlighting as I couldn't get this to work without modifying the `ForBindingCommand` as it wasn't previously built in a way where multiple semicolons could be used.

**Before:** The `ForBindingCommand` could only parse a single iterator property after the semicolon:
```typescript
// Only supported: repeat.for="item of items; key: id"
const attr = info.attr.rawValue.slice(forOf.semiIdx + 1);
const i = attr.indexOf(':');
// ... parsed single property
```

**After:** Enhanced to parse multiple semicolon-separated properties:
```typescript
// Now supports: repeat.for="item of items; key: id; previous: true"
const attrsString = info.attr.rawValue.slice(forOf.semiIdx + 1);
const attrParts = attrsString.split(';');  // Split on semicolons
const parsedProps: MultiAttrInstruction[] = [];

for (let j = 0, jj = attrParts.length; j < jj; j++) {
  // Parse each property individually
  const colonIdx = attrPart.indexOf(':');
  // ... create MultiAttrInstruction for each
}
```

**Why this change was necessary:**

The original implementation only parsed the first property after the semicolon. To support `previous` alongside `key`, we needed to extend the parser to handle multiple properties. This change is backward compatible. Single-property syntax continues to work exactly as before.

If `previous` is not specified, no additional memory or processing occurs. When enabled, the overhead is minimal (one additional reference per view).

### Section Headers Pattern
```html
<div repeat.for="item of items; previous: true">
  <h2 if.bind="item.category !== $previous?.category">
    ${item.category}
  </h2>
  <div>${item.name}</div>
</div>
```

### Diff Display
```html
<div repeat.for="entry of changelog; previous.bind: true">
  <span class="${entry.value !== $previous?.value ? 'changed' : ''}">
    ${entry.value}
  </span>
</div>
```

### With Keyed Repeats
```html
<div repeat.for="item of items; key: id; previous: true">
  Previous ID: ${$previous?.id ?? 'none'} → Current: ${item.id}
</div>
```

## Technical Notes

1. **Template Compiler Change**: The enhancement to parse multiple iterator properties is backward compatible and follows the existing parsing pattern. It simply loops through semicolon-separated segments instead of processing only the first one.

2. **Keyed Repeat Compatibility**: `$previous` is set during view activation in `_activateAllViews()`, which occurs after scope creation and view recycling. This ensures correctness even when views are reused.

3. **Memory Management**: The `$previous` reference is managed through the override context and is automatically cleaned up when views are deactivated.

4. **Expression Evaluation**: `previous.bind` expressions are evaluated once at binding time (not reactive), which is consistent with how `key.bind` works.

## Review Questions

- Is the template compiler enhancement approach correct, or would we prefer a different parsing strategy? This seemed to be the easiest means of adding support, but maybe I am unaware of a different means of adding this functionality. It seems `repeat.for` is handled in a specific way compared to other attributes and template syntax.
- Should we consider making `previous.bind` reactive (re-evaluate on changes), or is one-time evaluation sufficient? I personally think making it one-time is sufficient and probably best for optimising.

### DIFF
--- a/docs/user-docs/templates/repeats-and-list-rendering.md
+++ b/docs/user-docs/templates/repeats-and-list-rendering.md
@@ -137,6 +137,7 @@ Every repeat iteration provides rich contextual information:
 | `$even` | `boolean` | `true` for even indices (0, 2, 4...) |
 | `$odd` | `boolean` | `true` for odd indices (1, 3, 5...) |
 | `$length` | `number` | Total number of items |
+| `$previous` | `any \| null` | Previous iteration's item (opt-in, see below) |
 | `$parent` | `object` | Parent binding context |
 
 ### Nested Repeats and $parent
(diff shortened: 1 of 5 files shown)


Vheissu:
Ooh yes, computed is way better and making it opt-out is probably better. I actually had a use-case for this feature which was the catalyst for taking a stab at it. It was a deeply nested menu (4 levels deep, don't ask) and this actually means not having to track levels in the template and makes things a lot cleaner. I love the idea of future additions such as `$next` I don't think I've ever seen these features in other frameworks or libraries before and it makes so much sense, it's crazy.

Issue #2179, opened 2025-06-16
Improve @aurelia/state plugin


Vheissu:
# Enhanced State Management Features for @aurelia/state

The `@aurelia/state` plugin does a great job covering basic state management needs, but there are some gaps that make it challenging to adopt for larger applications or teams coming from other state management solutions.

### 1. Middleware Support
**Why this matters:** Right now, there's no way to intercept actions before they hit action handlers. This makes it difficult to implement common patterns that teams expect:

- **Logging**: Can't easily track what actions are being dispatched and when
- **Analytics**: No clean way to send user interaction data to analytics services
- **Validation**: Can't validate actions before they modify state
- **Persistence**: Have to manually wire up localStorage/sessionStorage in every relevant action handler
- **Error handling**: No centralized place to catch and handle action processing errors

Other state libraries solve this with middleware systems, and developers migrating from Redux or Vuex expect this functionality. Without it, teams end up with scattered, duplicated logic across action handlers.

### 2. Selectors and Computed State
**The performance problem:** Currently, any derived state calculation happens every time a component re-renders, even when the underlying data hasn't changed. Think filtering large lists or calculating totals from arrays.

**Developer experience issue:** Teams struggle with where to put complex state derivation logic. Should it go in components? Action handlers? There's no clear pattern, leading to inconsistent code organization.

Memoized selectors would solve both issues by only recalculating when dependencies actually change and providing a clear place for derived state logic.

### 3. Multiple Store Support
**Scalability concern:** The single global store works fine for small apps, but becomes unwieldy as applications grow. Issues that arise:

- Different feature teams stepping on each other's state properties
- The state object becomes a massive nested structure that's hard to reason about
- Testing becomes difficult because you have to mock the entire application state even for isolated features
- Code splitting is hampered because everything depends on the global state shape

**Module boundaries:** Large applications naturally organize around feature modules (user management, product catalog, shopping cart, etc.). Having separate stores that can be composed makes the architecture much cleaner and more maintainable.

### 4. State Persistence
**Common requirement:** Most real applications need to persist some state between sessions - user preferences, authentication tokens, draft form data, etc. 

Currently, this requires manually wiring up localStorage/sessionStorage logic in every relevant action handler, which leads to:
- Inconsistent persistence patterns across the app
- Easy to forget to persist new state properties
- No standard way to handle persistence errors or data migration
- Duplicate serialization/deserialization logic

### 5. Action Creators and Enhanced Type Safety
[...]

PR #2262, opened 2025-10-05
feat(state): add support for multiple named and keyed stores

Vheissu:
Enables registering and managing multiple stores identified by names or DI keys alongside the default store. Extends bindings, decorators, and dispatch commands to target specific stores via identifiers or store instances. Introduces a store manager to centralise store registration and retrieval, ensuring backward compatibility with default store resolution. Enhances template syntax and decorators to allow fine-grained control over which store to interact with.

This completes one of the longstanding features from this issue: https://github.com/aurelia/aurelia/issues/2179

Functional demo of multiple stores here: https://stackblitz.com/edit/fmkd23tj

### DIFF
--- a/docs/user-docs/aurelia-packages/state.md
+++ b/docs/user-docs/aurelia-packages/state.md
@@ -120,7 +120,7 @@ If you are familiar with Redux, you'll find this plugin familiar. The most obvio
 
 #### Multiple stores
 
-Large applications often need more than a single global store. You can register additional stores by providing either a `storeName` (string alias) or a `storeKey` (DI key) when calling `StateDefaultConfiguration.init`:
+Large applications often need more than a single global store. You can register additional stores by providing a `storeName` (string) when calling `StateDefaultConfiguration.init`:
 
 ```ts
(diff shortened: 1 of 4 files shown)


bigopon (commented):
I don't think we should be introducing another binding behavior, since it will create conflict with the existing one. Imagine `& state & store` vs `& store & state:'users'` etc...
It seems like trouble.

Vheissu:
> I don't think we should be introducing another binding behavior, since it will create conflict with the existing one. Imagine `& state & store` vs `& store & state:'users'` etc... It seems like trouble.

Agreed. I think initially I was trying to avoid overloading the existing behaviour, but it makes sense we parameterise it and that way nothing changes. I saw you made the change, thank you!

bigopon:
So after the child dialog service feature for the dialog plugin, we kind of have a precedence, I guess we can go ahead with this. One thing that seems like a gap to me is the ability to connect stores together. But we don't need to do it now. I'll get this PR ready to go.

bigopon:
@Vheissu @ekzobrain this is now ready to go, when you have some time pls help review.

PR #2264, opened 2025-10-06
feat(router): expose routeParameters helper for aggregated params

### LINKED ISSUE #2257: Router path parent params access
I have page `A/:id1`
That has a subpage `X/:id2`
In component X I want to access the parameter id1, how would I do this?
In the V1 router I could walk back up the tree and read the params from each earlier level. It was messy, and sometimes I used the dependency injection with transient variables to pass variables up the tree.

There is a very sound reason for all of this. In my apps in the past I use the structure:
`/company/[companyid]/project/[projectid]/user/[userid]/whatever`
So we have routers at the root, at company, at project, and at user components.
This gives lovely modular hierarchies, as each level builds (and caches) data used by nested components.

But, the component `whatever` needs the values of all the preceding variables for its functionality.
I don't think we should pollute the `params` in child routes with earlier variables, as this could cause issues with duplicate names, but an easy way to access the params of the parent routes would be very helpful.

For example, could we have a function on the ICurrentRoute that takes a route definition Id and returns the params?
We have a field 'Id' in the routes that I haven't noticed being used, but this could be passed in and the params object returned.
Alternatively, an array of params objects for the current route tree, and let the user search it for the appropriate fields?

In the current implementation, digging through the resolved `ICurrentRoute` object doesn't help much. There is a `parameterInformation` array, but this is always empty in the `attached` event. I suspect this may be a bug, as I would have expected the route information to be available when the `loading` lifecycle is called?

Vheissu:
## Summary

- add `routeParameters()` to `RouteContext` so nested views can access ancestor params without walking parent chains
- reuse the helper in docs/examples and export `RouteParametersOptions`
- cover duplicates, query inclusion, immutability, and empty-state behaviour with focused router tests

## Why

Fixes #2257 — developers requested an ergonomic way for deeply nested components (e.g. `company/:companyId/project/:projectId/...`) to read upstream parameters. The helper provides the aggregated snapshot
  without polluting child route configs or re-implementing traversal logic.

### DIFF
--- a/packages/__tests__/src/router/route-parameters.spec.ts
+++ b/packages/__tests__/src/router/route-parameters.spec.ts
@@ -1,0 +1,225 @@
+import { resolve } from '@aurelia/kernel';
+import { customElement } from '@aurelia/runtime-html';
+import { assert } from '@aurelia/testing';
+import { IRouteContext, IRouter, route } from '@aurelia/router';
+
+import { start } from './_shared/create-fixture.js';
+
+describe('router/route-parameters.spec.ts', function () {
+  describe('RouteContext.routeParameters', function () {
+    it('aggregates parameters from ancestor contexts', async function () {
+      @customElement({ name: 'details-view', template: `<div>company:\${params.companyId};project:\${params.projectId};user:\${params.userId};detail:\${params.detailId}</div>` })
+      class DetailsView {
+        public readonly params = resolve(IRouteContext)
+          .routeParameters<{ companyId: string; projectId: string; userId: string; detailId: string }>();
+      }
+
+      @route({ routes: [{ path: 'details/:detailId', component: DetailsView }] })
+      @customElement({ name: 'user-view', template: `<au-viewport></au-viewport>` })
+      class UserView { }
+
+      @route({ routes: [{ path: 'user/:userId', component: UserView }] })
+      @customElement({ name: 'project-view', template: `<au-viewport></au-viewport>` })
+      class ProjectView { }
+
+      @route({ routes: [{ path: 'project/:projectId', component: ProjectView }] })
+      @customElement({ name: 'company-view', template: `<au-viewport></au-viewport>` })
+      class CompanyView { }
+
+      @route({
+        routes: [
+          { path: '', redirectTo: 'company/placeholder/project/placeholder/user/placeholder/details/placeholder' },
+          { path: 'company/:companyId', component: CompanyView },
+        ],
+      })
+      @customElement({ name: 'app-root', template: `<au-viewport></au-viewport>` })
+      class AppRoot { }
+
+      const { host, au, container } = await start({ appRoot: AppRoot });
+      const router = container.get(IRouter);
+
+      await router.load('company/10/project/20/user/30/details/40');
+
+      assert.html.textContent(host, 'company:10;project:20;user:30;detail:40');
+
+      await au.stop(true);
+    });
+
+    it('prefers the closest route parameters when keys overlap', async function () {
+      @customElement({ name: 'leaf-view', template: `<div>id:\${params.id}</div>` })
+      class LeafView {
+        public readonly params = resolve(IRouteContext)
+          .routeParameters<{ id: string }>();
+      }
+
+      @route({ routes: [{ path: 'leaf/:id', component: LeafView }] })
+      @customElement({ name: 'user-node', template: `<au-viewport></au-viewport>` })
+      class UserNode { }
+
+      @route({ routes: [{ path: 'user/:id', component: UserNode }] })
+      @customElement({ name: 'project-node', template: `<au-viewport></au-viewport>` })
+      class ProjectNode { }
+
+      @route({ routes: [{ path: 'project/:id', component: ProjectNode }] })
+      @customElement({ name: 'company-node', template: `<au-viewport></au-viewport>` })
+      class CompanyNode { }
+
+      @route({
+        routes: [
+          { path: '', redirectTo: 'company/root/project/middle/user/account/leaf/final' },
+          { path: 'company/:id', component: CompanyNode },
+        ],
+      })
+      @customElement({ name: 'app-root-overlap', template: `<au-viewport></au-viewport>` })
+      class AppRootOverlap { }
+
+      const { host, au, container } = await start({ appRoot: AppRootOverlap });
+      const router = container.get(IRouter);
+
+      await router.load('company/root/project/middle/user/account/leaf/final');
+
+      assert.html.textContent(host, 'id:final');
+
+      await au.stop(true);
+    });
+
+    it('optionally includes query parameters', async function () {
+      @customElement({ name: 'query-leaf', template: `<div>ready</div>` })
+      class QueryLeaf {
+        public static instance: QueryLeaf | null = null;
+        public readonly params = resolve(IRouteContext)
+          .routeParameters<{
+            companyId: string;
+            projectId: string;
+            userId: string;
+            detailId: string;
+            filter: readonly string[];
+            mode: string;
+          }>({ includeQueryParams: true });
+
+        public constructor() {
+          QueryLeaf.instance = this;
+        }
+      }
+
+      @route({ routes: [{ path: 'details/:detailId', component: QueryLeaf }] })
+      @customElement({ name: 'query-user', template: `<au-viewport></au-viewport>` })
+      class QueryUser { }
+
+      @route({ routes: [{ path: 'user/:userId', component: QueryUser }] })
+      @customElement({ name: 'query-project', template: `<au-viewport></au-viewport>` })
+      class QueryProject { }
+
+      @route({ routes: [{ path: 'project/:projectId', component: QueryProject }] })
+      @customElement({ name: 'query-company', template: `<au-viewport></au-viewport>` })
+      class QueryCompany { }
+
+      @route({
+        routes: [
+          { path: '', redirectTo: 'company/placeholder/project/placeholder/user/placeholder/details/placeholder' },
+          { path: 'company/:companyId', component: QueryCompany },
+        ],
+      })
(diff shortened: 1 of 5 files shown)


packages/__tests__/src/router/route-parameters.spec.ts
```diff
@@ -0,0 +1,372 @@
...
+      const router = container.get(IRouter);
+
+      await router.load('company/10/project/20/user/30/details/40');
+
+      assert.html.textContent(host, 'company:10;project:20;user:30;detail:40');
+
+      await au.stop(true);
+    });
+
+    it('prefers the closest route parameters when keys overlap', async function () {
```
Sayan751, line 48:
```suggestion
    it('prefers the closest route parameters by default when the parameter names collide', async function () {
```

packages/__tests__/src/router/route-parameters.spec.ts
```diff
@@ -0,0 +1,372 @@
...
+      const router = container.get(IRouter);
+
+      await router.load('company/root/project/middle/user/account/leaf/final');
+
+      assert.html.textContent(host, 'id:final');
+
+      await au.stop(true);
+    });
+
+    it('allows selecting which duplicate values win', async function () {
```
Sayan751, line 86:
```suggestion
    it('allows resolving parameter name collisions', async function () {
```

packages/__tests__/src/router/route-parameters.spec.ts
```diff
@@ -0,0 +1,372 @@
...
+      @customElement({ name: 'strategy-leaf', template: `<div></div>` })
+      class StrategyLeaf {
+        public static instance: StrategyLeaf | null = null;
+        public readonly nearest = resolve(IRouteContext)
+          .routeParameters<{ id: string }>();
+        public readonly furthest = resolve(IRouteContext)
```
Sayan751, line 92: A bit of a nitpicking. You do the `resolve(IRouteContext)` only once.

packages/__tests__/src/router/route-parameters.spec.ts
```diff
@@ -0,0 +1,372 @@
...
+
+      assert.notStrictEqual(StrategyLeaf.instance, null);
+      assert.strictEqual(StrategyLeaf.instance!.nearest.id, 'final');
+      assert.strictEqual(StrategyLeaf.instance!.furthest.id, 'root');
+
+      await au.stop(true);
+      StrategyLeaf.instance = null;
+    });
+
+    it('appends duplicate values as arrays when requested', async function () {
```
Sayan751, line 134:
```suggestion
    it('returns the values as arrays for the collided paramter names with the \'append\' merge strategy', async function () {
```

packages/__tests__/src/router/route-parameters.spec.ts
```diff
@@ -0,0 +1,372 @@
...
+      assert.notStrictEqual(AppendLeaf.instance, null);
+      const appended = AppendLeaf.instance!.appended;
+
+      assert.deepStrictEqual(appended.id, ['root', 'middle', 'account', 'final']);
+
+      await au.stop(true);
+      AppendLeaf.instance = null;
+    });
+
+    it('can return duplicate values keyed by route id', async function () {
```
Sayan751, line 181:
```suggestion
    it('returns the values as record/object, keyed by route id, for the collided paramter names with the \'append-by-route\' merge strategy', async function () {
```

packages/router/src/route-context.ts
```diff
@@ -65,6 +65,26 @@ import { ILocationManager } from './location-manager';
...
+export interface RouteParametersOptions<TStrategy extends RouteParameterMergeStrategy = 'nearest'> {
+  /**
+   * Include query-string values from each active route segment. Defaults to the router option `treatQueryAsParameters`.
+   */
+  includeQueryParams?: boolean;
+  /**
+   * Decide how to merge duplicate parameter names. Defaults to `'nearest'` (the current segment overrides ancestors).
+   */
+  mergeStrategy?: TStrategy;
+  /** @deprecated Use {@link mergeStrategy} */
```
Sayan751, line 84: Missing cleanup?

packages/router/src/route-context.ts
```diff
@@ -65,6 +65,26 @@ import { ILocationManager } from './location-manager';
 export interface IRouteContext extends RouteContext { }
 export const IRouteContext = /*@__PURE__*/DI.createInterface<IRouteContext>('IRouteContext');
 
+export type RouteParameterValue = string | readonly string[];
+
+export type RouteParameterMergeStrategy = 'nearest' | 'furthest' | 'append' | 'append-by-route';
```
Sayan751, line 70: `append-by-route` sounds a bit odd. Since I am not better with naming either, poking you @bigopon for some alternatives/suggestions, if you happen to have any.

packages/router/src/route-context.ts
```diff
@@ -126,6 +155,9 @@ export function createEagerInstructions(instructionOrInstructions: NavigationIns
  * - Different components (with different `RenderContext`s) reference the same component via a child route config
  */
 export class RouteContext {
+  private static readonly _emptyRouteParameters = Object.freeze(Object.create(null)) as Readonly<RouteParameterAccumulator>;
```
Sayan751, line 158: The kernel already exposes a frozen empty object named `emptyObject`. Remove these 3 properties and use the `emptyObject` directly in the methods when you want to return an empty object.

packages/router/src/route-context.ts
```diff
@@ -211,6 +243,141 @@ export class RouteContext {
...
+    const mergeStrategy = (options?.mergeStrategy ?? options?.collisionStrategy ?? 'nearest') as TStrategy;
+
+    switch (mergeStrategy) {
+      case 'append':
+        return this._collectAppend(includeQueryParams) as RouteParametersResult<TStrategy, TParams>;
+      case 'append-by-route':
+        return this._collectAppendByRoute(includeQueryParams) as RouteParametersResult<TStrategy, TParams>;
+      case 'furthest':
+        return this._collectFlat(includeQueryParams, false) as RouteParametersResult<TStrategy, TParams>;
+      case 'nearest':
```
Sayan751, line 260: These four methods as well as `_forEachNodeValue` can all be local functions inside the current method.

packages/router/src/route-context.ts
```diff
@@ -211,6 +243,141 @@ export class RouteContext {
     }
   }
 
+  public routeParameters<
```
Sayan751, line 246:
```suggestion
  public getRouteParameters<
```

packages/router/src/route-context.ts
```diff
@@ -211,6 +243,141 @@ export class RouteContext {
...
+        aggregated[key] = RouteContext._freezeValue(value);
+        mutated = true;
+      });
+    }
+
+    return mutated ? Object.freeze(aggregated) : RouteContext._emptyRouteParameters;
+  }
+
+  private _collectAppend(includeQueryParams: boolean): Readonly<Record<string, readonly RouteParameterValue[]>> {
+    const contexts: RouteNode[] = [];
```
Sayan751, line 285: The naming is a bit confusing. You are collecting nodes and the variable is named `contexts`. Moreover, I am not sure why 2 loops are needed. I think that the work can be done in a single loop.

packages/router/src/route-context.ts
```diff
@@ -211,6 +243,141 @@ export class RouteContext {
...
+    const result = Object.create(null) as Record<string, readonly RouteParameterValue[]>;
+    for (const key of Object.keys(accumulator)) {
+      result[key] = Object.freeze(accumulator[key]) as readonly RouteParameterValue[];
+    }
+    return Object.freeze(result);
+  }
+
+  private _collectAppendByRoute(includeQueryParams: boolean): Readonly<Record<string, Readonly<Record<string, RouteParameterValue>>>> {
+    const contexts: { node: RouteNode; routeId: string }[] = [];
+    for (let current: RouteContext | null = this; current !== null; current = current.parent as RouteContext | null) {
```
Sayan751, line 319: Please consider using one single loop.

packages/router/src/route-context.ts
```diff
@@ -211,6 +243,141 @@ export class RouteContext {
...
+      const node = current._node;
+      if (node === null) continue;
+      contexts.push({ node, routeId: RouteContext._resolveRouteIdentifier(node) });
```
Sayan751, line 322: `_resolveRouteIdentifier` should be a local function inside this function. I think that we don't need to crowd the class with methods, those are used only once for the specific case.

packages/router/src/route-context.ts
```diff
@@ -211,6 +243,141 @@ export class RouteContext {
...
+      }
+    }
+
+    for (const [key, value] of Object.entries(node.params)) {
+      if (value === undefined) continue;
+      visit(key, value);
+    }
+  }
+
+  private static _freezeValue(value: RouteParameterValue): RouteParameterValue {
```
Sayan751, line 369: `_freezeValue` should be a local function inside `getRouteParameters`.

packages/router/src/route-context.ts
Vheissu: @Sayan751 what about `per-route`? I made this change because I agree `append-by-route` is too wordy. But happy for any other suggestions, it's quick to change.

packages/router/src/route-context.ts
```diff
@@ -65,6 +65,25 @@ import { ILocationManager } from './location-manager';
 export interface IRouteContext extends RouteContext { }
 export const IRouteContext = /*@__PURE__*/DI.createInterface<IRouteContext>('IRouteContext');
 
+export type RouteParameterValue = string | readonly string[];
+
+export type RouteParameterMergeStrategy = 'nearest' | 'furthest' | 'append' | 'per-route';
```
bigopon, line 70: I dont feel like `furthest` or `nearest` can tell clearly. What about `parent-first` or `child-first`?
About `per-route`, is it `original`?

PR #2272, opened 2025-10-09
fix(au-compose): apply surrogate attributes to host

### LINKED ISSUE #2271: Surrogate behaviors do not work with <au-compose> for template only composition
# 🐛 Bug Report

Template class attribute is not added to the `au-compose` host element:
```html
<au-compose tag="div" template="
  <template class='message'>
    <div>${message}</div>
  </template>
" message.bind></au-compose>
```

Repro link: https://stackblitz.com/edit/au2-conventions-talfwx57?file=src%2Fmy-app.html

## 🤔 Expected Behavior

If `tag` attribute is present in `<au-compose>` then html template attributes should be added to composed host element.

## 😯 Current Behavior

The above example produces the following html:
```html
<div>
    <div>Hello World!</div>
</div>
```

The above example should produce the following html:
```html
<div class="message">
    <div>Hello World!</div>
</div>
```

## 🌍 Your Environment

| Aurelia          |
| Language         |
| Browser          |
| Bundler          |
| Operating System |
| NPM/Node/Yarn    |

Vheissu:
Compiles the target definition to expose surrogates and applies them to the host when no explicit location is provided; uses the compiled definition for view factory and for collecting bindings. Enables correct application of inline surrogate attributes when a tag is supplied. 

- Inline template compositions now run through the template compiler before hydration so surrogate instructions (e.g., class, style) are replayed onto the generated host when tag is used. We switched to the compiled definition when creating spread bindings for the same scenario.
- Added a regression test under 3-runtime-html/au-compose.spec.ts that composes an inline <template class='message'>…</template> with tag="div" to ensure the host retains surrogate attributes.

Look, I'm gonna be honest. I feeled my way around this stuff. It looked somewhat familiar to me because of the template controller render instruction stuff I did a while back. So, if something I've done doesn't seem right, let me know so I can keep learning.

Working example that uses the same provided repo from the linked issue: https://stackblitz.com/edit/8iaevjzf

Closes #2271

### DIFF
--- a/packages/__tests__/src/3-runtime-html/au-compose.spec.ts
+++ b/packages/__tests__/src/3-runtime-html/au-compose.spec.ts
@@ -7,7 +7,7 @@ import {
   IRenderLocation,
   bindable,
 } from '@aurelia/runtime-html';
-import { tasksSettled } from '@aurelia/runtime';
+import { runTasks, tasksSettled } from '@aurelia/runtime';
 import { ICompositionController } from '@aurelia/runtime-html/dist/types/resources/custom-elements/au-compose';
 import {
   assert,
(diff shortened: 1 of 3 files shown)


packages/__tests__/src/3-runtime-html/au-compose.spec.ts
```diff
@@ -30,6 +30,24 @@ describe('3-runtime-html/au-compose.spec.ts', function () {
...
+        `<au-compose tag="div" template="<template class='message'><div>\${message}</div></template>">`,
+        class App {
+          public message = 'Hello World!';
+        }
+      ).started;
+
+      const host = appHost.children[0] as HTMLElement | undefined;
+      assert.notStrictEqual(host, void 0);
+      assert.strictEqual(host!.className, 'message');
+      assert.strictEqual(host!.textContent?.trim(), 'Hello World!');
```
bigopon:
```suggestion
      const { appHost, assertClass, assertText } = await createFixture(
        `<au-compose tag="div" template="<template class='message'><div>\${message}</div></template>">`,
        class App {
          public message = 'Hello World!';
        }
      ) // .started; we dont need this, it's a thennable return for createFixture
      // we also don't need to await, since there's no promise in the app
      // au compose doesn't make an app async if its composition doesn't return a promise

      assertClass('div', 'message');
      assertText('Hello World!');
```

PR #2279, opened 2025-10-13
fix(router): make string navigation resolve siblings/root intuitively

### LINKED ISSUE #2256: Router path interpretation changes
# 💬 RFC

The current router implementation is nice, but has some rather unintuitive and confusing behavior regarding navigation between pages.

## 🔦 Context

In most cases, when we have multiple pages we assume that we can navigate between them by using the path of the sibling page.
For example
`/home`
`/about`
`/users` with child routes of:
`   /members`
`   /admins`

If I have a link in my `about` component to return to home, I currently have to use `load='../home'`
However, given that I'm not in a child-router, it's a bit confusing. Why do I have to go back 'up' to a sibling?
If I'm in child router such as `users/members`, I need `../../about`, which doesn't seem correct as it gives the impression of being one above the root node.
This is not terrible, although it is unintuitive, and is at odds with all filing systems, which is a familiar analogy to use in documentation.

The most confusing thing is that I can't use `load='/home'`
If I prefix my path with a backslash I would expect it to be from the root router, but instead it currently looks for a child route.

This is particularly important when re-using a component at different depths of the router tree. If my component is one router deep, I have to refer to `../about`, but if I'm 3 deep I have to refer to `../../../about`. 
This breaks for dynamically added components, and is just unintuitive.

So there are two things I'd suggest we change:
1. I'd suggest we have paths using the traditional filing-system syntax:
  `/path` refers to a path from the root
  `about` refers to a sibling (also available as `./about`)
  `../about` refers to a parent router of current router

2. We also consider making the path references apply to the router 'above' the page we are on.
  simple path statements refer to sibling paths (so `/home` component can `load='about'`, and `members` can link using `load='admins'`)
  statements such as `../about` refer to the parent router, so `members` component can say `load='../about'`

I can't think of any scenarios where this doesn't make sense. I seem to remember that we worked this way in Aurelia v1.

Anyone got any suggestions or comments?

Vheissu:
- make `RouteContext` treat string instructions like `/foo`, `foo`, and `../foo` as expected by walking to the right router instead of blindly climbing the tree
- adjust the docs to explain the new sibling/parent/root rules and refresh the examples that used to warn about “doesn’t work”
- add regression coverage in the router load/href specs so both anchor clicks and imperative `router.load()` calls exercise the new behaviour

So I wrote some tests for this problem first knowing they would fail and then I made the tests pass. Like other changes I've made, I don't necessarily know the entire context of the router, so if I did something we do differently elsewhere, just let me know.

Also, special thanks to @Sayan751 for pointing me at the exact code that handles the context stuff so I didn't have to dig too far.

Closes #2256


packages/router/src/route-context.ts
```diff
@@ -646,6 +648,199 @@ export class RouteContext {
...
+
+  return { instruction, context, contextChanged };
+}
+
+function stripLeadingSlashes(value: string): string {
+  let index = 0;
+  while (index < value.length && value.charCodeAt(index) === 47 /* / */) {
+    index++;
+  }
+  return value.slice(index);
```
Sayan751: I think that this function can be rewritten as:

```ts
  return value.replace(/^\/+/, '');
```

packages/router/src/route-context.ts
```diff
@@ -646,6 +648,199 @@ export class RouteContext {
...
+}
+
+function consumeLeadingDotSegments(path: string, startContext: RouteContext): { path: string; context: RouteContext; changed: boolean } {
```
Sayan751: I think this function needs to be simplified.

- Early termination cases (without staring a loop): if the path is `./foo` (-> `foo`)  or `.` (-> empty string).
- Reserve the loop only for the paths starting with `..`. Even in that case, you can make an early return from the loop if the path becomes empty, without `continue`ing.

Sayan751 (commented):
@Vheissu Thank you for tackling this! Here is the first round of review. I ahve sort of ignored the changes related to `data-testid` :)

Vheissu:
@Sayan751 Thanks for the sharp feedback. I dropped the stale `contextChanged` flag in `createViewportInstructions` so we still normalize string instructions without carrying redundant state. 

`stripLeadingSlashes` now uses `value.replace(/^/+/, '')`. `consumeLeadingDotSegments` handles the simple `.` and `./` cases up front, loops only for `..`, and exits early once the path is exhausted, so `/.foo` stays literal. I also added a short doc
  comment above `resolveContextThroughRecognition` to explain why that lightweight lookup remains. It lets deeply nested components load sibling or root routes without needing `../` prefixes. 

Let me know if you’d like any other tweaks or extra coverage.

fkleuver:
Closing this because #2369 superseded the implementation and has merged. #2256 remains open for the separate URL-first API decision.

PR #2292, opened 2025-10-16
fix(router): honor explicit transition plan for query-only loads

### LINKED ISSUE #2291: Router ignores replace transition when only query params change
User on Discord [reported this issue](https://discord.com/channels/448698263508615178/1243519563283435520/1427940294506319904). Navigating to the same route with new query-string values and transitionPlan: 'replace' leaves the current component mounted without re-running lifecycle hooks. Navigation events fire, the URL updates, but the view-model never reloads.

## Steps to reproduce:

1. Configure a route with transitionPlan: 'replace' in @route.
2. From that component call router.load with the same route id/path but different queryParams.
3. Observe that loading, loaded, and other lifecycle hooks do not run; the component instance remains unchanged.

**Root cause:** In `RouteConfig._getTransitionPlan` the router short-circuits to 'none' whenever the path and path params match, ignoring both query params and an explicit transitionPlan override.

**Expected behaviour:** Setting `transitionPlan: 'replace'` (or returning 'replace' from the transition-plan function) should force a reload even when only query params differ.

Vheissu:
- Keep query-only navigations non-reloading by default, matching router design that query strings are not part of route configuration matching.
- Ensure an explicit `Router#load(..., { transitionPlan })` still overrides the default same-route behavior for query-only loads.
- Add focused coverage for explicit navigation-option replacement and route-config query-only non-reload behavior.

Closes #2291

### DIFF
--- a/packages/__tests__/src/router/transition-plan.spec.ts
+++ b/packages/__tests__/src/router/transition-plan.spec.ts
@@ -1,0 +1,82 @@
+import { resolve } from '@aurelia/kernel';
+import { IRouteContext, IRouter, route } from '@aurelia/router';
+import { customElement } from '@aurelia/runtime-html';
+import { assert } from '@aurelia/testing';
+import { start } from './_shared/create-fixture.js';
+
+describe('router/transition-plan.spec.ts', function () {
+
+  it('reloads component when navigation options set transitionPlan to replace and only query params change', async function () {
+    @customElement({ name: 'qp-option-view', template: '' })
+    class QueryOptionView {
+      private readonly ctx = resolve(IRouteContext);
+      public static messages: string[] = [];
+      public loading(): void {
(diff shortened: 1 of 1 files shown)


Sayan751 (changes requested):
Thanks @Vheissu for the changes. I think that it suffices when any transition plan set via `Router#load` overrides the default configuration. In that case it is a conscious decision from the developers to replace the components or invoke lifecycle events on the components.

I think that a difference in the query string should not trigger a new navigation by default. As the query strings are not a part of the routing configurations, they do not trigger a navigation in case of difference, by design.

In a nutshell, I think we don't need the `hasSameQuery` function and just need to ensure that the users' wish for an explicitly chosen transition plan is respected. IMO that would result in even simpler solution of this problem.

PR #2295, opened 2025-10-29
fix: try remove ctor registration

bigopon:
## 📖 Description

As per the discord discussion https://discord.com/channels/448698263508615178/1425247071820320790, registering constructor unnecessarily could disable a scenario where optional injection of a custom element/attribute is used to detect a usage of that resource. We probably don't need to register that.

### DIFF
--- a/packages/runtime-html/src/resources/custom-attribute.ts
+++ b/packages/runtime-html/src/resources/custom-attribute.ts
@@ -1,4 +1,4 @@
-import { mergeArrays, firstDefined, Key, resourceBaseName, getResourceKeyFor, isFunction, isString, ILogger } from '@aurelia/kernel';
+import { mergeArrays, firstDefined, Key, resourceBaseName, getResourceKeyFor, isFunction, isString } from '@aurelia/kernel';
 import { Bindable } from '../bindable';
 import { Watch } from '../watch';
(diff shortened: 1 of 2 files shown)


bigopon:
@simonfox maybe try install the package released by the bot in this PR to see if your scenario works?

simonfox:
@bigopon repro of what we are trying to do https://stackblitz.com/edit/e2xaaaux?file=src%2Fmy-app.html

bigopon:
I see, thanks @simonfox . The use case you have above is one of the first thing that was fixed in v2. In v1, any time a behavior (attribute/element) is used, it starts a new container hierarchy. This causes an unnecessarily nested/complex tree, and sometimes doesn't make sense. For example attributes are supposed to annotate element, not contain.
So in v1, you get this container tree:
```
app
  my-disabled
    my-input
    my-input
    my-input
```
but in v2, you'll get this container tree:
```
app
  my-disabled
  my-input
  my-input
  my-input
```
hence the container of `my-input` while walking up will not see any `my-disabled` registered.

For this case, we can only walk the dom to figure it out. Or another case is to turn `my-disabled` into a template controller and manage the view yourself. The later is involved.
I guess we will update `findAttributeControllerFor` to make it walk up the dom.

PR #2311, opened 2025-12-01
feat(runtime-html,router): add SSR hydration support

fkleuver:
# Pull Request

## 📖 Description

Add SSR hydration support for Aurelia 2, enabling server-rendered HTML to be adopted and made reactive on the client without re-rendering.

**Core features:**
- `Aurelia.hydrate()` API for client-side hydration of server-rendered DOM
- Tree-based manifest system (`ISSRScope`, `ISSRManifest`, `ISSRTemplateController`) for recording/replaying template controller state
- `<!--au-->` comment markers for hydration target identification
- `adoptSSRView` / `adoptSSRViews` helpers for view adoption
- Router SSR: `ServerLocationManager`, viewport adoption, skip-append mode
- `CustomElement.clearDefinition()` for SSR cache invalidation between requests

**Template controller hydration support:**

| TC | Hydration | Notes |
|----|-----------|-------|
| `repeat` | ✅ | Batch-adopts all views using `nodeCount` from manifest |
| `if` / `else` | ✅ | Manifest records which branch was rendered (`state.value`) |
| `switch` / `case` / `default-case` | ✅ | Switch stores `_ssrViewScopes`; cases consume them |
| `with` | ✅ | Single view adoption, same pattern as `if` |
| `promise` / `pending` / `then` / `catch` | ❌ | Re-renders on client. Server state (pending/fulfilled/rejected) may not match client state when hydration runs. |
| `portal` | ❌ | Content renders to a different DOM location; not meaningful to hydrate in-place. |

### 🎫 Issues

Related: https://github.com/aurelia/aurelia-ls/pull/6 (AOT compiler + Vite SSR plugin that consumes these runtime primitives)

Split-off PRs (already merged):
- #2309 - AST classes to interfaces
- #2310 - Template compiler refactor
- #2329 - Numeric discriminants for instruction types

## 👩‍💻 Reviewer Notes

**New public API:**
- `Aurelia.hydrate(host, ssrScope)` - hydrate server-rendered DOM
- `ISSRScope`, `ISSRManifest`, `ISSRTemplateController` - manifest types
- `adoptSSRView`, `adoptSSRViews` - view adoption helpers
- `CustomElement.clearDefinition(Type)` - cache invalidation
- `ServerLocationManager` - router location for SSR

**Internal changes:**
- Markers changed from `<au-m>` elements to `<!--au-->` comments
- Template controllers check `ssrScope` to adopt vs create views

**Why `<!--au-->` comments instead of `<au-m>` elements:**
- **Unified marker type**: One comment format for all target types (elements, text interpolations, template controllers) instead of a mixed approach requiring multiple code paths
- **Semantically invisible**: Comments don't affect CSS selectors, accessibility tree, or layout
- **Efficient collection**: Native `TreeWalker` API with `NodeFilter.SHOW_COMMENT` (browser C++ implementation) instead of DOM queries

## 📑 Test Plan

- **New:** `ssr-hydration.integration.spec.ts` - comprehensive integration tests
  - All TC types (repeat, if, switch, with)
  - Nested combinations (repeat>if, if>repeat, switch>repeat, etc.)
  - CE boundaries with TCs
  - Failure modes and edge cases
- **New:** `ssr-hydration.spec.ts` - unit tests for adoption helpers
- **Updated:** `dom.spec.ts` - 10 new tests for `<!--au-->` marker system
- **Updated:** `template-compiler.spec.ts` - marker format assertions

## ⏭ Next Steps

Complete https://github.com/aurelia/aurelia-ls/pull/6 (AOT/SSR/SSG tooling that consumes these primitives)

### DIFF
--- a/.changeset/fast-pens-dance.md
+++ b/.changeset/fast-pens-dance.md
@@ -1,0 +1,5 @@
+---
+"@aurelia/runtime-html": patch
+---
+
+Optimize repeat's indexMap computation from O(n²) to O(n) using scope identity map lookup instead of Array.includes/indexOf.
--- a/.changeset/silver-clouds-swim.md
+++ b/.changeset/silver-clouds-swim.md
@@ -1,0 +1,5 @@
+---
+"@aurelia/template-compiler": minor
+---
+
+Add SSR hydration support with manifest-based marker insertion for template controllers and hydration target elements.


packages/runtime-html/src/templating/controller.ts
```diff
@@ -446,9 +517,11 @@ export class Controller<C extends IViewModel = IViewModel> implements IControlle
 
   /** @internal */
   public _hydrateChildren(): void {
+    const targets = this.nodes!.findTargets();
+
     this._rendering.render(
       /* controller */this as ICustomElementController,
-      /* targets    */this.nodes!.findTargets(),
+      /* targets    */targets,
```
bigopon: well, this is the job of minifier, but generally not having a local variable is better for perf, not sure if it's relevant after the JIT though.

packages/runtime-html/src/templating/ssr.ts
```diff
@@ -0,0 +1,336 @@
...
+          const expr = getExpr(ctx.exprMap, aux.exprId);
+          props.push(new MultiAttrInstruction(expr as IsBindingBehavior, aux.name, 'bind'));
+        }
+      }
+      return new IteratorBindingInstruction(forOf, ins.to, props);
+    }
+
+    default:
+      throw new Error(`Unknown instruction type: ${(ins as SerializedInstruction).type}`);
+  }
```
bigopon: i guess we will have a way to hook into the deserialization of expressions here in phase 2?

bigopon (commented):
nice work @fkleuver 👍 , I've only had some preliminary comments, mostly around the concerns of stamping during compilation & rendering.

packages/runtime-html/src/templating/ssr.ts
fkleuver: This is just an intermediate version.

But to answer your question more broadly: my idea is to remove the instruction classes from the runtime (replace with interfaces) and making the runtime-side format a bit more readable (using the full `kind` names instead of two-letter codes) and then creating a specific efficient wire format for both instructions and the AST.
 
And then we can put a small codec in the runtime for deserializing the wire format. If you're on board with that I can first do a separate PR for the instructions and then do the efficient wire format + codec at a later stage.

packages/runtime-html/src/dom.ts
```diff
@@ -237,47 +295,326 @@ export class FragmentNodeSequence implements INodeSequence {
...
+        const el = node as Element;
+
+        // Collect au-hid element target
+        if (el.hasAttribute('au-hid')) {
+          const targetId = parseInt(el.getAttribute('au-hid')!, 10);
+          targets[targetId] = el;
+          if (!preserveMarkers) {
+            el.removeAttribute('au-hid');
+          }
+        }
```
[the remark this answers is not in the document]
fkleuver: Updated it

packages/runtime-html/src/templating/ssr.ts
fkleuver: I did a bit more research and it seems to me the industry standard leans towards using number-based type discriminators all the way for things like instructions and AST. It seems beneficial because then no codec is needed either. Can give that a try. Then we don't need any translator function at all, or at least a less flaky one

packages/__tests__/src/3-runtime-html/dom.spec.ts
```diff
@@ -145,3 +271,302 @@ function appendChildren(parent: HTMLElement, child: HTMLElement, count: number):
...
+    const innerEnd = ctx.doc.createComment('au-end');
+    const outerContent = ctx.doc.createElement('div');
+    const outerEnd = ctx.doc.createComment('au-end');
+
+    parent.appendChild(outerStart);
+    parent.appendChild(innerStart);
+    parent.appendChild(innerContent);
+    parent.appendChild(innerEnd);
+    parent.appendChild(outerContent);
+    parent.appendChild(outerEnd);
```
bigopon: I think for tests like this (it seems to apply to all the tests in this file), we better use innerHTML, can tell the structure of the dom much better.

packages/__tests__/src/3-runtime-html/template-compiler.au-slot.spec.ts
```diff
@@ -57,7 +57,7 @@ describe('3-runtime-html/template-compiler.au-slot.spec.ts', function () {
 
   it('compiles default <au-slot> as the only child', function () {
     const { template, instructions } = compileTemplate('<au-slot></au-slot>');
-    assertTemplateEqual(template, '<!--au*--><!--au-start--><!--au-end-->');
+    assertTemplateEqual(template, '<!--au--><!--au-start--><!--au-end-->');
```
bigopon: with `au*`, I kind of allowed us to have a small relief that we will not accidentally pick something user land generated. Not that I think au will, and it can be fixed. Just explaining why there's an *

packages/runtime-html/src/dom.ts
```diff
@@ -237,30 +312,40 @@ export class FragmentNodeSequence implements INodeSequence {
...
     fragment: DocumentFragment,
+    /**
+     * When true, preserves `<!--au-->` markers in the DOM.
+     * Used during SSR rendering to keep markers for client hydration.
+     * @internal
+     */
+    preserveMarkers?: boolean,
   ) {
-    const targetNodeList = (this.f = fragment).querySelectorAll('au-m');
+    const markers = collectMarkerComments(this.f = fragment);
```
bigopon: With the old processes, we turn markers into au-m, and collect using .querySelectorAll, the process of turning au -> au-m was one off, so the collection is pretty efficient with .querySelectorAll. With this new process, how will it be in comparison?

packages/runtime-html/src/resources/template-controllers/with.ts
```diff
@@ -36,12 +41,36 @@ export class With implements ICustomAttributeViewModel {
...
   public attaching(
-    initiator: IHydratedController,
+    _initiator: IHydratedController,
     _parent: IHydratedParentController,
   ): void | Promise<void> {
     const { $controller, value } = this;
+    const ssrScope = $controller.ssrScope;
+    if (ssrScope != null && isSSRTemplateController(ssrScope) && ssrScope.type === 'with') {
+      return this._hydrateView(ssrScope);
+    }
```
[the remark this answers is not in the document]
fkleuver: The guard is implicit because `_hydrateView()` clears the `ssrScope` which is checked at the start of `attaching()`. I added some clarifying comments about that

packages/template-compiler/src/template-compiler.ts
```diff
@@ -551,24 +551,14 @@ export class TemplateCompiler implements ITemplateCompiler {
...
+      this._replaceByMarker(el, context);
       let tcTemplate: HTMLTemplateElement;
-      if (isMarker(el)) {
-        // Element is already a marker (edge case: nested TC on marker)
-        tcTemplate = context.t();
-        appendManyToTemplate(tcTemplate, [
-          context._marker(),
-          context._comment(auLocationStart),
-          context._comment(auLocationEnd),
-        ]);
```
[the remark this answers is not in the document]
fkleuver: The guard moved into _replaceByMarker itself - it returns early if the node is already a marker, so the explicit check was no longer needed at the call site

packages/runtime-html/src/dom.ts
fkleuver: Both are native implementations operating on DocumentFragments. TreeWalker with SHOW_COMMENT is native C++ that skips all non-comment nodes without JS involvement, so performance is difference is negligible for typical templates. The main win here is simplicity (no transformation step) and SSR compatibility (comments serialize cleanly in HTML)

packages/__tests__/src/3-runtime-html/ssr-hydration.spec.ts
```diff
@@ -0,0 +1,342 @@
...
+      assert.notStrictEqual(result, null, 'should return a result');
+      assert.notStrictEqual(result!.view, null, 'should have view');
+      assert.strictEqual(result!.viewScope.nodeCount, 1, 'should have correct nodeCount');
+    });
+
+    it('creates view wrapping multiple nodes when nodeCount > 1', function () {
+      const { container, platform, doc } = TestContext.create();
+
+      const parent = doc.createElement('div');
+      const { location, nodes } = createLocationWithNodes(doc, parent, ['A', 'B', 'C']);
```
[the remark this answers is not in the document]
fkleuver: Nope, good catch, I fixed it

PR #2322, opened 2025-12-22
fix(tests): use development builds in karma test runner

fkleuver:
# Pull Request

## 📖 Description

Tests were running against production builds (`index.mjs`) which are minified with `__DEV__: false`. This made debugging test failures difficult due to mangled variable names and stripped development assertions.

Now tests use development builds (`index.dev.mjs`) which provide:
- Unminified code for readable stack traces
- `__DEV__: true` for development-only warnings/assertions
- Full variable names for easier debugging

Also removed the dead static `importmap.js` file that was superseded by dynamic import map generation in the karma middleware (Oct 2022).

### Test Updates Required

Switching to dev builds exposed error message format differences between prod and dev. Production builds use compact error codes (`AUR0167:{expr}`) while dev builds include full descriptions (`AUR0167: Expression error: missing expected token...`).

Updated assertions in:
- `di.exception.spec.ts` - regex now matches error code + context loosely
- `eventaggregator.spec.ts` - same pattern
- `dirty-checker.spec.ts` - same pattern
- `computed-observer.spec.ts` - switched from `strictEqual` to `match`
- `expression-parser.spec.ts` - updated `verifyResultOrError()` to check error code prefix and context separately
- `template-compiler.spec.ts` - same pattern
- `dialog/dialog-service.spec.ts` - same pattern

### Controller Lifecycle Fix

Fixed controller lifecycle state machine to handle async activation failures:

1. **Recovery from `deactivating` state:** When async activation fails mid-lifecycle, the controller could get stuck in `deactivating` state. Added handler in `activate()` to reset internal flags and allow re-activation.

2. **Idempotent deactivation:** Added `deactivating` case in `deactivate()` to return existing promise instead of throwing.

3. **If template controller cleanup:** When view activation fails, deactivate the failed view to reset its state for subsequent swaps.

4. **Switch/Case template controller cleanup:** Same fix applied to Case.activate() for switch/case template controllers.

### Dev Script Fix

Removed stale reference to deleted `au` package in `scripts/dev.ts` that was causing `npm run dev` to fail.

### 🎫 Issues

Related: https://github.com/aurelia/aurelia-ls/pull/6
Related: https://github.com/aurelia/aurelia/pull/2311

## 👩‍💻 Reviewer Notes

The static `importmap.js` was never actually served - the karma middleware intercepts requests and serves a dynamically generated import map instead. The karma config entry is kept with `served: true` so the middleware can intercept the request.

### DIFF
--- a/scripts/dev.ts
+++ b/scripts/dev.ts
@@ -123,7 +123,6 @@ const validToolingPackages =  [
   'vite-plugin',
   'webpack-loader',
   'http-server',
-  'au',
 ];
 
 validToolingPackages
@@ -135,7 +134,8 @@ validToolingPackages
     try {
       execSync(buildCmd, { cwd: `packages-tooling/${pkgName}` });
     } catch (ex) {
-      process.stdout.write(ex.stdout);
+      if (ex.stdout) process.stdout.write(ex.stdout);
+      if (ex.stderr) process.stderr.write(ex.stderr);
       process.exit(1);
     }
     console.log(`${pkgDisplay} built in ${getElapsed(Date.now(), start)}s`);


bigopon:
The importmal was used as a fallback, in the case where the request didnt go through karma interceptor. For some reason it was needed but if the ci and local are running fine lets retire it.

PR #2337, opened 2025-12-31
fix(kernel): use cross-realm safe type checks in utilities

fkleuver:
# Pull Request

## 📖 Description

Replace `instanceof` checks in type utilities with cross-realm safe alternatives.

The previous implementations used `instanceof` (e.g., `v instanceof Set`) which fails when objects originate from different JavaScript realms such as:
- iframes (accessing objects via `window.top`)
- JSDOM (Node.js testing with vitest/jest)
- Web Workers

This caused issues like `PrimitiveObserver` being incorrectly selected for DOM elements in JSDOM tests, because `element instanceof Object` returned `false` for cross-realm objects.

### Changes

| Utility | Before | After |
|---------|--------|-------|
| `isPromise` | `v instanceof Promise` | `v != null && typeof v.then === 'function'` |
| `isArray` | `v instanceof Array` | `Array.isArray(v)` |
| `isSet` | `v instanceof Set` | `v != null && v[Symbol.toStringTag] === 'Set'` |
| `isMap` | `v instanceof Map` | `v != null && v[Symbol.toStringTag] === 'Map'` |
| `isObject` | `v instanceof Object` | `typeof v === 'object' && v !== null` |

### Performance considerations

- `isArray`: `Array.isArray` is a built-in specifically designed to be cross-realm safe; no performance penalty
- `isSet`/`isMap`: Uses `Symbol.toStringTag` directly instead of `Object.prototype.toString.call()` for better performance
- `isPromise`: Now Promise/A+ spec compliant (thenables can be objects or functions with a `.then` method)
- `isObject`: Simple `typeof` check, minimal overhead

### ES compatibility

`Symbol.toStringTag` requires ES2015+. This is safe since Aurelia targets ES2021.

### 🎫 Related PRs

- #1975 — Added `isArray`, `isSet`, `isMap`, `isObject` with `instanceof` (with documented cross-realm caveats)
- #2044 — Moved utilities to kernel, added cross-realm safe `isObjectOrFunction`

This PR addresses the limitations documented in those PRs.

### 🎫 Issues

Discovered while writing JSDOM-based SSR tests for [aurelia-ls](https://github.com/aurelia/aurelia-ls). Form binding tests failed because `isObject(htmlInputElement)` returned `false`, causing `PrimitiveObserver` to be used instead of `ValueAttributeObserver`.

## 👩‍💻 Reviewer Notes

- The `isObjectOrFunction` function was already cross-realm safe; its documentation has been simplified since `isObject` now shares the same safety guarantees
- `Symbol.toStringTag` can technically be spoofed, but this is an acceptable trade-off for performance vs using `Object.prototype.toString.call()`
- The `isPromise` change is slightly more permissive (now includes functions with `.then`), which aligns with Promise/A+ spec

## 📑 Test Plan

- Existing Aurelia test suite should pass unchanged (these utilities maintain the same behavior for same-realm objects)
- Cross-realm scenarios (vitest+JSDOM, iframes) will now work correctly
- Verified with aurelia-ls SSR test suite using vitest+JSDOM

## 📦 Changeset

## ⏭ Next Steps

None - this is a self-contained fix.

### DIFF
--- a/packages/kernel/src/utilities.ts
+++ b/packages/kernel/src/utilities.ts
@@ -15,7 +15,7 @@ import { Metadata } from '@aurelia/metadata';
  * Per Promise/A+ spec, a thenable is any object or function with a `.then` method.
  */
 export const isPromise = <T>(v: unknown): v is Promise<T> =>
-  v != null && typeof (v as Promise<T>).then === 'function';
+  typeof (v as Promise<T> | null | undefined)?.then === 'function';
 
 /**
  * Returns true if the value is an Array.
@@ -23,7 +23,7 @@ export const isPromise = <T>(v: unknown): v is Promise<T> =>
  *
  * Cross-realm safe via Array.isArray (works with iframes, JSDOM, vitest, etc.).
  */
-export const isArray = <T>(v: unknown): v is T[] => Array.isArray(v);
+export const isArray = Array.isArray;
 
 /**
  * Returns true if the value is a Set.
@@ -33,7 +33,7 @@ export const isArray = <T>(v: unknown): v is T[] => Array.isArray(v);
  * Uses Symbol.toStringTag directly instead of Object.prototype.toString.call() for performance.
  */
 export const isSet = <T>(v: unknown): v is Set<T> =>
-  v != null && (v as Set<T>)[Symbol.toStringTag] === 'Set';
+  (v as Set<T> | null | undefined)?.[Symbol.toStringTag] === 'Set';
 
 /**
  * Returns true if the value is a Map.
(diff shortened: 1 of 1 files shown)


packages/kernel/src/utilities.ts
```diff
@@ -8,94 +8,69 @@ import { Metadata } from '@aurelia/metadata';
...
- * Returns true if the value is an Array via checking if it's an instance of Array.
- * This does not work for objects across different realms (e.g., iframes).
- * An utility to be shared among core packages for better size optimization
+ * Returns true if the value is an Array.
+ * An utility to be shared among core packages for better size optimization.
+ *
+ * Cross-realm safe via Array.isArray (works with iframes, JSDOM, vitest, etc.).
  */
-export const isArray = <T>(v: unknown): v is T[] => v instanceof Array;
+export const isArray = <T>(v: unknown): v is T[] => Array.isArray(v);
```
bigopon: what about `const isArray = Array.isArray`? looks a bit more direct.

bigopon (approved):
nice work @fkleuver :+

PR #2344, opened 2026-01-08
feat(runtime-html): replace bindable primary with defaultProperty on custom attributes

### LINKED ISSUE #2343: Move `primary` property from bindable to custom attribute definition
# 🙋 Feature Request

Currently we have this algo for detecting primary CA property:
1. Look for a bindable properties with `primary: true` and take the first of them (current API allows specifiing more than one `primary: true`)
2. If there's no bindable property with `primary: true` - then just take the first defined bindable property
3. Else use `value` property

This seems too complicated and confusing for me and it is NOT described in the docs correctly. Second step is missing there.

I suggest making it much cleaner:
1. Remove `primary` option from bindable definition
2. Introduce CA option like `defaultProperty: string = 'value'`, which will allow to specify name of the default property.
 
This, in my opinion, will solve all ambiguity about this configuration and will be even more flexible than now:
1. Remove ambiguous `primary` property, which could be set for many bindables without any effect for CA and has absolutely no effect for CE.
2. Allow to change the name of default property even without making it bindable. I had cases where I wanted to use other name than `value` according to naming conventions, but did not have such possibility.

Then an implicit algo of detecting a primary property will not be needed at all, it will be just that property specified by `defaultProperty`.

## 🤔 Expected Behavior

```ts

// same as now, defaults to 'value'
@customAttribute({
  name: 'a'
})
class ca {
  // or with @bindable to add interactive updates, binding mode, coercion, etc
  value = 'val';
}

// change name of primary property
@customAttribute({
  name: 'a',
  defaultProperty: 'prop'
})
class ca {
  // or with @bindable to add interactive updates, binding mode, coercion, etc
  prop = 'val';
}
```

## Migration notice

Migration from V1 may be provided several ways:
1. Deprecate `primary` bindable option in v2.0, make current algo to compute actual value for `defaultProperty` unless it is explicitely provided. Then remove `primary` option totally in v2.1. This will also not break existing v2 apps.
2. Handle this in `v1-compat` plugin for v1->v2 migration

@bigopon @Vheissu @fkleuver @Sayan751

fkleuver:
# Pull Request

## 📖 Description

Replaces the `primary` property on bindable definitions with `defaultProperty` on custom attribute definitions.

**Before (no longer supported):**
```typescript
@customAttribute('tooltip')
export class TooltipAttribute {
  @bindable({ primary: true }) message: string;
  @bindable position: string;
}
```

**After:**
```typescript
@customAttribute({ name: 'tooltip', defaultProperty: 'message' })
export class TooltipAttribute {
  @bindable message: string;
  @bindable position: string;
}
```

This simplifies the API:
- The default property is explicitly declared in one place, not scattered across bindables
- No possibility of conflicting `primary: true` on multiple bindables
- The template compiler no longer needs to search through bindables to find the primary

If `defaultProperty` is not specified, it defaults to `'value'` for backward compatibility.

### 🎫 Issues

- Closes #2343
- Closes #2340

## 👩‍💻 Reviewer Notes

Key files:

1. **`packages/template-compiler/src/interfaces-template-compiler.ts`** - Added `defaultProperty` to `IAttributeComponentDefinition`, removed `primary` from `IComponentBindablePropDefinition`

2. **`packages/runtime-html/src/compiler/template-compiler.ts`** - Updated `ResourceResolver.bindables()` to use `defaultProperty`

3. **`packages/runtime-html/src/resources/custom-attribute.ts`** - Added `defaultProperty` to `CustomAttributeDefinition`

4. **`packages/runtime-html/src/bindable.ts`** - Removed `primary` from `BindableDefinition`

Framework resources updated:
- `portal.ts` - `defaultProperty: 'target'`
- `repeat.ts` - `defaultProperty: 'items'`
- `load.ts` (router) - `defaultProperty: 'route'`
- `validation-errors-custom-attribute.ts` - `defaultProperty: 'errors'`
- `virtual-repeat.ts` - `defaultProperty: 'items'`

Error code 714 (`compiler_primary_already_existed`) has been removed.

Documentation updated:
- `docs/user-docs/templates/custom-attributes.md`
- `docs/user-docs/components/bindable-properties.md`
- `docs/user-docs/developer-guides/error-messages/*/aur0714.md` (marked as removed)

## 📑 Test Plan

All existing and updated tests still pass.

## 📦 Changeset

## ⏭ Next Steps

None.

### DIFF
--- a/docs/user-docs/TOC.md
+++ b/docs/user-docs/TOC.md
@@ -18,6 +18,13 @@
 * [Aurelia for New Developers](getting-started/aurelia-for-new-developers.md)
 * [Hello World Tutorial](getting-started/quick-start-guide/README.md)
 * [Intermediate Tutorial](getting-started/intermediate-tutorial.md)
+* [Extended Tutorial](getting-started/extended-tutorial/README.md)
+  * [Step 1: Project setup + app shell](getting-started/extended-tutorial/step-1-project-setup.md)
+  * [Step 2: Routing + nested layouts](getting-started/extended-tutorial/step-2-routing-and-layout.md)
+  * [Step 3: Overview page + filters + events](getting-started/extended-tutorial/step-3-projects-overview.md)
+  * [Step 4: Detail route + guards](getting-started/extended-tutorial/step-4-project-detail-and-guards.md)
+  * [Step 5: Router events + polish](getting-started/extended-tutorial/step-5-router-events-and-polish.md)
+  * [Step 6: Route data + auth roles](getting-started/extended-tutorial/step-6-route-data-and-roles.md)
 
 ### Coming from Another Framework?
 * [From React to Aurelia](getting-started/from-react.md)
--- a/docs/user-docs/advanced-scenarios/binding-and-templating-internals.md
+++ b/docs/user-docs/advanced-scenarios/binding-and-templating-internals.md
@@ -166,7 +166,7 @@ export class InterpolationInstruction {
 
 **Instruction Types:**
 - `PropertyBindingInstruction` - `.bind`, `.one-way`, `.two-way`
-- `ListenerBindingInstruction` - `.trigger`, `.delegate`, `.capture`
+- `ListenerBindingInstruction` - `.trigger`, `.capture`
 - `RefBindingInstruction` - `ref` attribute
 - `IteratorBindingInstruction` - `repeat.for`
 - `HydrateElementInstruction` - Custom elements
--- a/docs/user-docs/advanced-scenarios/building-phone-apps.md
+++ b/docs/user-docs/advanced-scenarios/building-phone-apps.md
@@ -127,7 +127,7 @@ The complete Capacitor guide covers:
   </header>
 
   <main class="mobile-content">
-    <router-view></router-view>
+    <au-viewport></au-viewport>
   </main>
 
   <nav class="mobile-nav">
--- a/docs/user-docs/advanced-scenarios/performance-optimization-techniques.md
+++ b/docs/user-docs/advanced-scenarios/performance-optimization-techniques.md
@@ -453,7 +453,7 @@ export class EventComponent {
     window.addEventListener('resize', this.resizeHandler);
   }
   
-  detached(): void {
+  detaching(): void {
     window.removeEventListener('resize', this.resizeHandler);
   }
   
@@ -481,7 +481,7 @@ export class SubscriptionComponent {
     );
   }
 
-  detached(): void {
+  detaching(): void {
     this.subscriptions.forEach(sub => sub.dispose());
     this.subscriptions = [];
   }
@@ -669,7 +669,7 @@ export class InfiniteScroll {
     this.scrollContainer.addEventListener('scroll', this.onScroll.bind(this));
   }
 
-  detached(): void {
+  detaching(): void {
     this.scrollContainer.removeEventListener('scroll', this.onScroll.bind(this));
   }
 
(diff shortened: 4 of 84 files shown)


bigopon:
Is it still automatically selecting the first bindable if no default(primary) is declared, or everything must be explicit? It feels like a bit heavy on the configuration side if so.
Cc @ekzobrain

ekzobrain:
> Is it still automatically selecting the first bindable if no default(primary) is declared, or everything must be explicit? It feels like a bit heavy on the configuration side if so. Cc @ekzobrain

I think it should not look at bindables at all. The only exception may be BC mode for migration as I described in linked issue.
It always uses the property name that is specified by `defaultProperty`, which has a default value of `value`. So in this case:

```ts
class CA {
  @bindable prop1;
  @bindable prop2;
  value;
}
```

a default property will still be 'value', but it will not be reactive as it has no `@bindable` decorator

fkleuver:
> Is it still automatically selecting the first bindable if no default(primary) is declared, or everything must be explicit? It feels like a bit heavy on the configuration side if so. Cc @ekzobrain

No, it no longer automatically selects the first bindable.

The behavior is:
1. If `defaultProperty` is specified -> that property is used as the primary
2. If `defaultProperty` is NOT specified -> it defaults to `value`
3. If no bindable exists with the matching name -> the compiler automatically creates a bindable with that name

This is actually less configuration-heavy for the common case:
- If your default property is named `value` -> zero configuration needed (works like before)
- Only if you want a different default property name -> add `defaultProperty: 'myProp'`

The old implicit algorithm (picking the first bindable) was arguably more confusing because bindable declaration order could affect behavior, and you had to understand the undocumented fallback chain. Now it's explicit: either use `value` (automatic) or explicitly declare which property you want.

ekzobrain:
> > > Is it still automatically selecting the first bindable if no default(primary) is declared, or everything must be explicit? It feels like a bit heavy on the configuration side if so. Cc @ekzobrain
> > 
> > 
> > I think it should not look at bindables at all. The only exception may be BC mode for migration as I described in linked issue. It always uses the property name that is specified by `defaultProperty`, which has a default value of `value`. So in this case:
> > ```ts
> > class CA {
> >   @bindable prop1;
> >   @bindable prop2;
> >   value;
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > a default property will still be 'value', but it will not be reactive as it has no `@bindable` decorator
> 
> You make a fair point in principle, but your scenario of a plain `value` property receiving values without being a bindable (if I understood you correctly) wouldn't work - the compiler needs that metadata. The auto-creation of the bindable definition is actually necessary for the binding system to function at all.
> 
> To be precise, even with non-reactive properties like `.one-time`, the compiler:
> 
>     1. Requires a bindable definition to know what property to target
> 
>     2. Uses `primaryBindable.name` to generate the `SetPropertyInstruction`
> 
>     3. Throws an error if you try to bind to a property that isn't a bindable
> 
> 
> So the flow is:
> 
>     1. Compiler needs bindable metadata to emit instructions -> requires `BindableDefinition`
> 
>     2. Runtime (`SetPropertyRenderer`) can do direct assignment to plain properties, but it needs the instruction from the compiler telling it which property.
> 
> 
> You can't bypass that infrastructure without breaking important invariants about how the binding system works, which would have unforeseen ripple effects.
> 
> If you want non-reactive behavior, the answer is "use `.one-time` binding mode", not "don't make it a bindable"

Does this account that default property may already be declared as bindable and have a two-way binding for example?
```ts
class ca {
  @bindable({ mode: BindingMode.twoWay }) value;

  binding() {
    setTimeout(() => this.value = 1, 100);
  }
}
```

```html
<div ca.bind="twoWayBoundProp">
```

Will `twoWayBoundProp` be correctly updated to `1` here?

PR #2345, opened 2026-01-08
fix: don't try to handle multi bundle clash

bigopon:
## 📖 Description

Sometimes ago I was trying to prepare for a scenario where multiple Aurelia applications can work & share objects together. When an object is observed in multiple applications (different bundles), the observation from different bundles need to work well together and thus, some markers on the global objects were employed. This is probably not necessary, as most of the time the observer is cached on the object, which is practical enough. I think it's better we leave it natural, and fix any issues that arise later rather than trying to predict/anticipate, which is likely inadequate/wrong.

cc @fkleuver @Sayan751 @ekzobrain

### DIFF
--- a/packages/runtime/src/array-observer.ts
+++ b/packages/runtime/src/array-observer.ts
@@ -17,7 +17,7 @@ import {
 } from './subscriber-collection';
 import { rtDef, rtDefineHiddenProp, rtDefineMetadata, rtGetMetadata } from './utilities';
 import { addCollectionBatch, batching } from './subscriber-batch';
-import { type IIndexable, isFunction } from '@aurelia/kernel';
+import { isFunction } from '@aurelia/kernel';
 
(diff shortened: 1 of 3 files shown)


fkleuver:
You're removing a shared global WeakMap mechanism that was designed to allow multiple Aurelia bundables to share observer instances. I'm concerned about how this will impact HMR which is something that isn't fully implemented and tested yet.

The existing HMR e2e tests in this repo only looks at component state preservation, bindable value retention and UI re-rendering. They do not test collection reactivity after HMR, so any bugs caused by this change wouldn't be caught by the current test setup.

I do not yet have proof that the code you're removing is truly necessary, and I am generally in favor of removing code that does not have a proven and justified existence, but I'd prefer to wait with this one until the SSR/HMR stack is done and properly tested. Then we can circle back to see if this can be safely removed. It's easier that way than the other way around.

PR #2349, opened 2026-01-11
fix(effect): consistent getter & expression observation behavior

### LINKED ISSUE #2175: IObservation.watch() and IObservation.watchExpression() behavior inconsistency
Inconsistency is with the first run. Will provide complete description later, this is a placeholder for the issue.

# 🐛 Bug Report

Repro link:

## 🌍 Your Environment

| Aurelia          |
| Language         |
| Browser          |
| Bundler          |
| Operating System |
| NPM/Node/Yarn    |

bigopon:
## 📖 Description

As pointed out in #2175 , `observation.watch`/`watchExpression` are behaving differently during the initial call when the callback returns undefined. This PR aligns the behaviors of the two. Also cleanup some interfaces in the observer locator so it doesn't expose more than necessary information for `.getExpressionObserver`.

### 🎫 Issues

Closes #2175 

cc @ekzobrain

### DIFF
--- a/packages/runtime/src/observation.ts
+++ b/packages/runtime/src/observation.ts
@@ -61,8 +61,8 @@ export class Observation implements IObservation {
     callback: (value: R, oldValue: R | undefined) => unknown,
     options?: IWatchOptions,
   ): IEffect;
-  public watch<T extends object, R>(
-    obj: T,
+  public watch<R>(
+    obj: object,
     expression: string,
     callback: (value: R, oldValue: R | undefined) => unknown,
     options?: IWatchOptions,


packages/runtime/src/expression-observer.ts
```diff
@@ -11,23 +11,10 @@ import { subscriberCollection } from './subscriber-collection';
 export interface ExpressionObserver extends IObserverLocatorBasedConnectable, ICollectionSubscriber { }
 export class ExpressionObserver implements IObserverLocatorBasedConnectable, ISubscriber {
```
ekzobrain: May be fix interfaces as well? Fo example `ExpressionObserver` implements `ISubscribable`, but does not declare it. I think that just aliging with `ComputedObserver` will be enough.

ekzobrain:
Also, now, as `watch` and `watchExpression` have almost identical code (the only difference is a different underlying observer used) - may be lets merge them into `watch` with an overload? Just as a `@watch` decorator.

bigopon:
@ekzobrain by separating them we have a clear signal whether expression parser is used in the application code, it's likely easier to be picked up by AOT, but maybe it doesn't matter. cc @fkleuver

bigopon:
Sometimes the argument passed is a variable, and won't it require much more work to analyse, and sometimes not possible to tell with certainty? Having said that, It wouldn't matter a lot, since the expression parser isn't that big, and applications tend to get more dynamic later.

bigopon:
@ekzobrain done, thanks, I kind of wanted that too but wasn't motivated enough until your comment 😛

fkleuver:
> Sometimes the argument passed is a variable, and won't it require much more work to analyse, and sometimes not possible to tell with certainty? Having said that, It wouldn't matter a lot, since the expression parser isn't that big, and applications tend to get more dynamic later.

It's true that it can't always be statically analyzed but AOT has explicit guards for up to which point it will try, and produce warnings when it can't reliably determine the nature of a call site. If you enable aggressive optimization, AOT will tell you e.g. "cannot analyze dynamic expression in this file at line X" and then the developer can make the choice: either just let it bail and conservatively include more of the runtime, or change their call site to be more statically analyzable.

Issue #2273, opened 2025-10-10
Handling child routes of parameterized routes better


Sayan751:
# 🐛 Bug Report

Copied from #2259:

Having multiple accessible routes breaks direct links to child routes in the following situation:

> route 1: `[ 'a', 'a/:id' ]`
> child-route 2 under component 'a' : `'x'`

Component 'x' can be navigated to by links on page 'a' (`load='x'`), but cannot be accessed directly unless an id is provided.
For example, the following path can be followed:
`/ -> /a -> /a/x`
or
`/ -> /a/2 -> /a/2/x`
But a direct link `load='/a/x'` will not work, as the base route assumes the `x` is the id from the first route.

That's an understandable (but undesirable) effect, but we also get an UnknownRouteError if the user refreshes the page.
This either needs clearly documenting when using child routes with a multiple route path component, or generate a warning/error if a child route is defined under these ambiguous situations.

Repro link:

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Aurelia          |
| Language         |
| Browser          |
| Bundler          |
| Operating System |
| NPM/Node/Yarn    |

Sayan751:
@jsobell Thanks for the reproduction! This is an artifact of eagerly choosing a route. I will try to find a fix for this.

PR #2355, opened 2026-01-13
feat(router): eager loading

Sayan751:
# Pull Request

## 📖 Description

This PR tries to solve the problem addressed in the issue #2273.

In a nutshell, the problem is as follows:
- The parent with routes `['parent', 'parent/:id']` configures a child `['child']`
- Works:  when users navigate from `/parent` to `/parent/child`
- Does not work: when users navigate directly to `/parent/child`

IMO the inherent lazy configuration of the routes contributed to this problem. Thus, this PR makes an attempt to eagerly load the routing configurations. In other words, with `RouterConfiguration.customize({ useEagerloading: true })`, the router builds the following routing table right when the application starts.

```
/parent -> ce-parent
/parent/:id -> ce-parent'
/parent/child -> [ce-parent, ce-child]
/parent/:id/child -> [ce-parent', ce-child]
```

> ce-parent' denotes the ce-parent endpoint with required `:id` parameter

The approach comes with trade-off:
- As the name suggests, the lazy loading of the components is not respected, as the configuration name suggests, as all components related to routing are loaded during bootstrap.
- This approach relies on the configured path, and hence usage of the empty paths is discouraged.

The difference between default configuration and eager loading can be seen below.

- [Default configuration](https://stackblitz.com/edit/au2-router-reproduction-of-2273?file=package.json,src%2Fmy-app.ts,src%2Fmain.ts)
- [Eager loading](https://stackblitz.com/edit/au2-router-eager-loading-2273?file=package.json,src%2Fmy-app.ts,src%2Fmain.ts)

## 👩‍💻 Reviewer Notes

- In case of eager loading, a single route recognizer is used for all route contexts as opposed to the default settings, where a separate route recognizer is used for every route context.
- Thus a single route recognizer has all the information.
- When registering child routes, the parent routes are prepended, and the combined path pattern is registered to the route recognizer.
- The same approach is applied when recognizing the routes. In this case, the child routes are matched relative to the parent route state.

## 📦 Changeset

### DIFF
--- a/packages/router/src/router.ts
+++ b/packages/router/src/router.ts
@@ -1,4 +1,4 @@
-import { IContainer, ILogger, DI, IDisposable, onResolve, Writable, onResolveAll, Registration, resolve, isObjectOrFunction, IModuleLoader } from '@aurelia/kernel';
+import { IContainer, ILogger, DI, IDisposable, onResolve, Writable, onResolveAll, Registration, resolve, isObjectOrFunction } from '@aurelia/kernel';
 import { CustomElement, CustomElementDefinition, IPlatform } from '@aurelia/runtime-html';
 
 import { createEagerInstructions, IRouteContext, RouteConfigContext, RouteContext } from './route-context';
@@ -102,7 +102,6 @@ export class Transition {
   }
 }
 
-type RouteConfigLookup = WeakMap<RouteConfig, RouteConfigContext>;
 type RouteConfigContextLookup = WeakMap<RouteConfigContext, IRouteContext>;
 type ViewportAgentLookup = Map<ViewportAgent | null, RouteConfigContextLookup>;
 
@@ -192,7 +191,6 @@ export class Router {
   /** @internal */ private readonly _logger: ILogger =  /*@__PURE__*/ resolve(ILogger).root.scopeTo('Router');
   /** @internal */ private readonly _events: IRouterEvents = resolve(IRouterEvents);
   /** @internal */ private readonly _locationMgr: ILocationManager = resolve(ILocationManager);
-  /** @internal */ private readonly _moduleLoader: IModuleLoader = resolve(IModuleLoader);
   public readonly options: Readonly<RouterOptions> = resolve(IRouterOptions);
 
   public constructor() {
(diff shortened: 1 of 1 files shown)


packages/route-recognizer/src/index.ts
```diff
@@ -58,18 +73,54 @@ export class RecognizedRoute<T> {
...
+
+  /** @internal */
+  public _importFrom(route: RecognizedRoute<T>): void {
+    const params: Record<string, string | undefined> = Object.create(null);
+    for (const key in this.params) {
+      params[key] = this.params[key];
+    }
+    for (const key in route.params) {
+      params[key] = route.params[key];
+    }
```
bigopon: did `{ ...this.params, ...route.params }` cause any issues?

packages/route-recognizer/src/index.ts
```diff
@@ -466,15 +617,56 @@ export class RouteRecognizer<T> {
     return endpoint;
   }
 
-  public recognize(path: string): RecognizedRoute<T> | null {
-    let result = this.cache.get(path);
-    if (result === void 0) {
-      this.cache.set(path, result = this.$recognize(path));
+  public recognize(path: string, relativeTo: RecognizedRoute<T>[] | null = null): RecognizedRoute<T>[] | null {
```
bigopon: this is more of an internal package, so maybe a change in the return should be ok. Probably we should mark the entire package as internal.

bigopon (commented):
I think the changes are quite nice, it enables really ambiguous url scenarios, which sometimes may be desirable. At least refreshing browser won't be an issue in those cases anymore.

For the router changes I think they are fine as long as the test work, and we should start marking every public method as internal and only leave `.load()` open, as we are in RC now. pretty much that's the only contract we guarantee that there will be no changes. Same for the route recogniser, we should probably mark it as an internal package too, and don't put a guaratee of no API change on it.

packages/__tests__/src/router/route-recognizer.spec.ts
```diff
@@ -3005,4 +3006,136 @@ describe('router/route-recognizer.spec.ts', function () {
       assert.throws(() => sut.add({ path: route2, handler: null }), `Cannot add ambiguous route. The pattern '${route2}' clashes with '${route1}'`);
     });
   }
+
+  describe('eager-loading - single recognizer', function () {
```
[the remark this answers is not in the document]
Sayan751: Exactly. I think it is easier to find relevant tests and add new tests when needed.

PR #2361, opened 2026-01-15
chore: add warning for ungarded recursive el

bigopon:
## 📖 Description

Sometimes it gets really confusing when we get a maximum call stack error from unguarded self recursive elements. Add a dev only warning to alert when such case happens.

### 🎫 Issues

To prevent confusing error like the mentioned Stackblitz here https://discourse.aurelia.io/t/au1-to-au2-differences-with-gloabl-registration/6556/6?u=bigopon

### DIFF
--- a/.changeset/cyan-masks-cough.md
+++ b/.changeset/cyan-masks-cough.md
@@ -1,0 +1,5 @@
+---
+"@aurelia/template-compiler": patch
+---
+
+add a warning when an unguarded self-recursive component is detected


packages/template-compiler/src/template-compiler.ts
```diff
@@ -152,6 +152,19 @@ export class TemplateCompiler implements ITemplateCompiler {
...
+    if (__DEV__) {
+      if (compiledDef.instructions.some(row => row.some(instr => {
+        if (instr.type === itHydrateElement) {
+          return (instr as HydrateElementInstruction).res === compiledDef.name
+            || ((instr as HydrateElementInstruction).res as ICompiledElementComponentDefinition)?.name === compiledDef.name;
+        }
+        return false;
+      }))) {
+        // eslint-disable-next-line no-console
+        console.warn(`[DEV:aurelia] Detected ungarded self-referencing component name "${compiledDef.name}" in compiled instructions. This may lead to infinite recursion at runtime.`);
```
fkleuver:
```suggestion
        console.warn(`[DEV:aurelia] Detected unguarded self-referencing component name "${compiledDef.name}" in compiled instructions. This may lead to infinite recursion at runtime.`);
```

PR #2369, opened 2026-01-24
refactor(router): path syntax and transition plan

Sayan751:
# Pull Request

## 📖 Description

Supersedes #2292  and #2279 .

It addresses the following issues.

### 1. Path syntax

This PR implements the path syntax changes proposed by @bigopon here: https://github.com/aurelia/aurelia/issues/2256#issuecomment-3360075505. Below is a short recap:

- `/path`: This represents an absolute path from the root of the application.
- `path`, `./path`: These represent a relative paths from the current routing context.
- `../path`, or `../../path`: These represent relative paths from the ancestor routing context; every `../` moves one level up in the routing context hierarchy.

### 2. Overriding `transitionPlan` from `Router#load`

This is reported in [Discord](https://discord.com/channels/448698263508615178/1243519563283435520/1427940294506319904). 

This PR ensures that an explicit `transitionPlan` provided like following via `Router#load` is enforced.

```ts
this.router.load('test', { transitionPlan: 'replace', queryParams: { message: this.newMessage } });
```

## 📦 Changeset

### DIFF
--- a/.changeset/slow-parts-argue.md
+++ b/.changeset/slow-parts-argue.md
@@ -1,0 +1,27 @@
+---
+"@aurelia/router": patch
+---
+
+refactor(router): path syntax and transition plan
+
+It addresses the following issues.
+
+### 1. Path syntax
+
+This PR implements the path syntax changes proposed by @bigopon here: https://github.com/aurelia/aurelia/issues/2256#issuecomment-3360075505. Below is a short recap:
+
+- `/path`: This represents an absolute path from the root of the application.
+- `path`, `./path`: These represent a relative paths from the current routing context.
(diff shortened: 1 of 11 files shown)


ekzobrain:
> @ekzobrain Although those paths might be considered valid, I think they need to be unsupported for now. Personally, as an application developer, I would find the usages of those paths very confusing. The docs now outline the supported path syntaxes. If there is a real-world need to support those edge cases, we can reconsider those later.

Merge this right now, it's ok. I agree that smth like './../a' is strange. But '..' is a really valid use case. Nevertheless i would like to suggest a bit different approach to path resolving. I think it wont be hard to implement it, while it will allow to use any valid relative URI path. I will open a separate issue for this after you finish current PRs

bigopon (approved):
Nice work @Sayan751 👍

PR #2370, opened 2026-01-24
feat(vite-plugin): inline css into entry bundle

Vheissu:
Adds a new CSS injection plugin that collects CSS assets from the build output, concatenates their sources, and injects them as a runtime style tag into entry chunks. Removes the separate CSS assets from the bundle to ensure inlined CSS is loaded with the bundle. Includes helper guards to identify CSS assets and entry chunks and runs in the post-build phase for proper ordering.


3cp (approved):

bigopon:
For this, we probably don't want to just do it, it should at least be in the form of a configurable property, no?

Vheissu:
@bigopon Agreed. I'm changing the code to be configurable, that's way better than my initial thought process. Also thanks @3cp

PR #2376, opened 2026-01-30
feat(au-compose): add flush mode binding

### LINKED ISSUE #2373: au-compose async recomposition
# 🙋 Feature Request

Consider the following example:
```ts
@customElement({
  name: 'test',
  template: '<au-compose component.bind="component" model.bind="model"/>'
})
class test {
  index = 0;  

  components = [
    { cmp: ViewModelThatAcceptsArray, data: []},
    { cmp: ViewModelThatAcceptsObject, data: {} }
  ]
  
  attached() {
    setTimeout(() => this.index = 1, 3000);
  }

  get component() {
    return this.components[this.index].cmp;
  }

  get model() {
    return this.components[this.index].data;
  }
}
```

Here `au-compose` initially renders `ViewModelThatAcceptsArray` component with `[]` data 
Then we change both component and data. But in fact `au-compose` updates them separately - in the order binding were defined in template. In our case first `component` will be updated and rendered by `au-compose`, but with OLD `[]` data.
Then, afterwards model binding will be propagated and then `ViewModelThatAcceptsObject` will get an updated model.

## 🤔 Expected Behavior

Have an ability to recompose only when both `component` and `model` are updated.

## 😯 Current Behavior

Only sync recompose mode is supported now.

## 💁 Possible Solution

Add an option to make `au-compose` recompose only when both `component` and `model` are updated. This may be achieved by adding a new bindable parameter (may be named `flushMode`/`recompose` = `sync/async`) and using async `propertiesChanged()` callback for that purpose. This will also improve compatibility with V1 as using this flag will allow to align behavior of V2 `au-compose` with V1 version.

bigopon:
## 📖 Description

For the template
```html
<au-compose component.bind="component" model.bind="model" flush-mode="async" />
```
and the component
```ts
class App {
  index = 0;

  components = [
    { cmp: { activate: (model) => models.push(model) }, data: [] },
    { cmp: { activate: (model) => models.push(model) }, data: {} }
  ];

  get component() {
    return this.components[this.index].cmp;
  }

  get model() {
    return this.components[this.index].data;
  }
}
```
Changing `index` should result in a coherent updates of the `<au-compose>` composition, but because of binding sync nature, the component.bind and model.bind are reacted in 2 different rounds causing `activate` on the vm to be called twice.

This PR adds a flush-mode "async" so that applications have more tool to control au-compose behavior.

## 🎫 Issues

Closes #2373

cc @ekzobrain


estaylorco:
@bigopon Does this solve the problem where the component doesn't refresh if only the model changes?

PR #2382, opened 2026-02-07
feat(ast): ability to track dependency on read in template during method call

### LINKED ISSUE #2381: feature request: ability to declare observation within a method call from template
## 🔦 Context

Currently any function call in template is considered pure, and only the parameters will be observed to trigger a re-evaluation when something changes. This is good in many cases, but it leaves other use cases to be desired.

An example would be where a function is supposed to be reused in both template and component code, and the function happens to be complex enough that declaring all the dependencies in the template wouldn't make sense, or feasible.

We probably can look into providing a way to make observing function call easier.

## 💁 Possible Solution

Maybe we can introduce a decorator to decorate a class method so that when ast evaluations happens, it will call the function with a proxy-wrapped object, and let the binding that currently observes automatically collect the dependencies.

## 💻 Examples

```ts
class MyComponent {

  @astTracked()
  myMethod() {
    return this.reads + this.a + this.bunch + this.of + this.properties
  }
}
```

cc @fkleuver @Sayan751 @ekzobrain @Vheissu

bigopon:
## 📖 Description

As described in #2381 , this PR adds a way to enable automatic tracking for property reads in a method call, as it's often an inconvenient part of using function call in template.

While we get automatic track on read with getter, it does not always make sense, or feasible to employ a bunch of getters everywhere. Sometimes a reusable method makes a class much better designed/understandable.

### 🎫 Issues

Close #2381

## Examples

Usages:
- `@computed`
- `@computed()`
- `@computed('prop', 'nested.prop')`
- `@computed(instance => instance.prop + instance.prop2)`
- `@computed({ deps: ['prop', 'nested.prop'] })`
- `@computed({ deps: ['prop', instance => instance.prop2] })`

Behavior:
- `deps` omitted (or `deps: null/undefined`) falls back to proxy-based tracking.
- `deps: []` explicitly disables tracking for the decorated method.
- `deps` with strings/getter functions enables explicit dependency tracking.
- Applying `@computed` again on the same method overrides prior metadata.

### DIFF
--- a/.changeset/bright-ravens-wait.md
+++ b/.changeset/bright-ravens-wait.md
@@ -4,18 +4,18 @@
 
 Extend `@computed` decorator to support method decoration for declaring/tracking dependencies when called from an observation context (e.g. a template binding or another computed observation). A normal function call will not trigger any observation.
 
+Note: method usage of `@computed` is experimental. Syntax and behavior may change before final release.
+
 Usages on methods:
-- `@computed` — proxy-based auto-tracking
-- `@computed('prop', 'nested.prop')` — explicit string dependencies
-- `@computed(instance => instance.prop + instance.prop2)` — getter function dependency
-- `@computed({ deps: ['prop', 'nested.prop'] })` — config object with string deps
-- `@computed({ deps: vm => vm.prop })` — config object with getter function dep
+- `@computed` - proxy-based auto-tracking
+- `@computed('prop', 'nested.prop')` - explicit string dependencies
+- `@computed(instance => instance.prop + instance.prop2)` - getter function dependency
+- `@computed({ deps: ['prop', 'nested.prop'] })` - config object with string deps
+- `@computed({ deps: vm => vm.prop })` - config object with getter function dep
 
 Usages on getters:
-- `@computed({ deps: ['prop1', 'prop2'] })` — explicit deps (uses ControlledComputedObserver)
-- `@computed({ flush: 'sync' })` — no deps, auto-tracking with sync flush
-- `@computed('prop1', 'prop2')` — shorthand for deps array
+- `@computed({ deps: ['prop1', 'prop2'] })` - explicit deps (uses ControlledComputedObserver)
+- `@computed({ flush: 'sync' })` - no deps, auto-tracking with sync flush
+- `@computed('prop1', 'prop2')` - shorthand for deps array
 
-Breaking: `deps` no longer allows mixing strings and functions in the same array.
 `getComputedObserver` now accepts an optional `ComputedPropertyInfo` parameter directly instead of reading from an internal WeakMap.
-
--- a/docs/user-docs/essentials/reactivity.md
+++ b/docs/user-docs/essentials/reactivity.md
@@ -171,6 +171,8 @@ Basides the above basic usages, the `computed` decorator also supports a few mor
 When you call a method from a template (not a getter), use `@computed` to control dependency tracking for that call.
 Observation only activates when the method is called from an observation context (e.g. a template binding or another computed observation). A normal function call will not trigger any observation.
 
+> Note: Using `@computed` on methods is currently experimental. The syntax and behavior may change before the final release.
+
 ```ts
 import { computed } from 'aurelia';
 
--- a/packages/runtime/src/computed-decorators.ts
+++ b/packages/runtime/src/computed-decorators.ts
@@ -120,7 +120,6 @@ export function computed<TThis extends object>(
       return;
     }
 
-    // Handle getters (existing logic)
     const getterOptions: ComputedPropertyInfo = isObject(targetOrOptionsOrDependency)
       ? targetOrOptionsOrDependency as ComputedPropertyInfo
       : {};


Sayan751 (approved):
Great addition, thanks @bigopon! A few follow-ups:
- Can you please also add some tests where `useProxy: false`?
- This needs to be documented.

ekzobrain:
I doubt a bit about name of the new deco `astTracked`. It seems too technical (internal) for me. Before proposing a better (IMHO) name I have a question: is this decorator supposed to change method behaviour when it is used inside a getter? Like this:

```ts

@astTracked()
getName(param) {
  return param + this.param2 + this.param3;
}

get name() {
  return this.getName(this.param);
}
```
Will `astTracked` usage influence `name` getter dependencies, and if yes - then how exactly?

packages/__tests__/src/3-runtime-html/ast-track-decorator.spec.ts
```diff
@@ -0,0 +1,381 @@
...
+
+    component.obj.prop1 = 'obj2';
+    await Promise.resolve();
+    assertText('obj1-value1');
+    assert.strictEqual(callCount, 1);
+
+    component.prop1 = 'value2';
+    await Promise.resolve();
+    assertText('obj2-value2');
+    assert.strictEqual(callCount, 2);
```
Sayan751: Nice 👍

packages/__tests__/src/3-runtime-html/ast-track-decorator.spec.ts
```diff
@@ -0,0 +1,381 @@
...
+        return 'ok';
+      }
+    });
+
+    assertText('ok');
+
+    component.nested.prop = 'value2';
+    await Promise.resolve();
+    assertText('ok');
+  });
```
Sayan751: Without a call count assertion, this test looks a bit incomplete. 😄

packages/__tests__/src/3-runtime-html/ast-track-decorator.spec.ts
```diff
@@ -0,0 +1,381 @@
...
+    assertText('value1');
+    assert.strictEqual(callCount, 1);
+
+    component.prop = 'value2';
+    await Promise.resolve();
+    assertText('value1');
+    assert.strictEqual(callCount, 1);
+  });
+
+  it('string dependency mode does not auto-track method reads', async function () {
```
Sayan751: Can we rename this test? To me, the name does not fully explain the objective of the test.

packages/__tests__/src/3-runtime-html/ast-track-decorator.spec.ts
```diff
@@ -0,0 +1,381 @@
...
+    assertText('u1-value1-value2');
+    assert.strictEqual(callCount, 1);
+
+    component.prop = 'next';
+    await Promise.resolve();
+    assertText('u2-next-value2');
+    assert.strictEqual(callCount, 2);
+  });
+
+  it('supports mixed deps in options object', async function () {
```
Sayan751: Nice 👍

packages/__tests__/src/3-runtime-html/ast-track-decorator.spec.ts
```diff
@@ -0,0 +1,381 @@
...
+    });
+
+    assertText('ok');
+
+    component.nested.prop = 'value2';
+    await Promise.resolve();
+    assertText('ok');
+  });
+
+  it('disables tracking when deps is an empty array', async function () {
```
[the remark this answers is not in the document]
bigopon: it's like an consistency enforcement, since we have `deps: []` in `@computed` with the meaning: disable tracking.

ekzobrain:
I feel a bit uncomfortable that new decorator works only for templates. If the same behavior (turning fuction to a parametrized getter) is not applicable for VM code (my example above), then this feels more like a plugin for certain narrow cases rather than a core feature

ekzobrain:
Ok, I see. I talk not about how it works internally, but how it would look from user perspective. As for me, this is redundant as a core feature, in it's current implementation. I also met cases that this feature is meant to solve, but that was just a few times per several large projects. If you look into an old 'addons' PR by Brendon - i think it would be the best place for it. It better fits into a 'swiss knife' toolkit. I think @fkleuver agrees with me.

bigopon:
Not having to manually pull the trigger is the point of this PR, though I do see point of redirecting the need to an existing infra. The issue is application has no way to hook into ast evaluation of a function, and function evaluation shouldn't be observing (reads automatically track) dependency by default. So I came up with the solution we see in this PR: a knob to let user enable automatic tracking.

In order for binding signaler to work, we need to be able to fire it at all appropriate levels. Imagine a list with items with `id`/`value` properties. Every where we edit the `id` or `value` of an item, we need to tell the signaler to fire so that bindings can know to update. it may work for a few items, but in a repeat, you have to have that signal on all the items, and then suddenly editing one item id becomes editing all items, from the perspective of reactivity.

packages/__tests__/src/3-runtime-html/ast-track-decorator.spec.ts
```diff
@@ -0,0 +1,381 @@
...
+    assertText('u1-value1');
+    assert.strictEqual(callCount, 1);
+
+    component.nested.prop = 'value2';
+    await Promise.resolve();
+    assertText('u2-value2');
+    assert.strictEqual(callCount, 2);
+  });
+
+  it('tracks configured function dependency via @astTrack', async function () {
```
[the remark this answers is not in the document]
bigopon: added 👍

packages/__tests__/src/3-runtime-html/computed-method.spec.ts
```diff
@@ -0,0 +1,381 @@
...
+        return 'ok';
+      }
+    });
+
+    assertText('ok');
+
+    component.nested.prop = 'value2';
+    await Promise.resolve();
+    assertText('ok');
+  });
```
bigopon: yeah it was off without 1, added, thanks @Sayan751 👍

bigopon:
@ekzobrain an example where signal isn't desirable
```html
<div repeat.for="item of items">
  <div>label: ${item.label}</div>
  <div>${getTotalValue() & signal : 'item-change'}</div>
</div>
```
```ts
class MyClass {
  getTotalValue(item) {
    return item.value * (1 + item.tax) * Math.max(0.5, 1 - item.discount);
  }
}
```
In order to have the binding `getTotalValue()` updated properly, everywhere `item.value`/`.tax`/`.discount` changes, you'll need to add the code `signaler.dispatch(...)`. The areas that have those changes may not be close, or relevant to each other at all.

in my app, the equivalent of `getTotalValue` accesses even more properties, and there will be a lot more signaler code needed if I'm to use that. Besides, manually triggering the signaler makes it feel like something is missing, when most of our observation is already automatic.

bigopon:
@Sayan751 @ekzobrain @fkleuver now the `@computed` behaves almost identical for both getter/method, added tests to verify its behavior when:
- method called directly from template
- method called from a getter being observed

Should be good to go now.

PR #2387, opened 2026-02-14
chore(validation): deprecated the email rule

### LINKED ISSUE #1919: Validation Package - Email validation
https://github.com/aurelia/aurelia/blob/2fca6ea3703f51a2b7b87d719bfa45c9bb35d6e1/packages/validation/src/rule-provider.ts#L311

Not sure if this is the intended behavior.

At the moment the behaviour is like this

`test` -> not valid
`test@` -> not valid
`test@test`  -> valid
`test@test.` -> not valid
`test@test.c` -> valid

Should `test@test` also not be valid?

If yes, then changing the regex to something like this:
```
^[a-zA-Z0-9.!#$%&'*+\/=?^_`{|}~-]+@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)*\.[a-zA-Z0-9]{2,}$
```

Should fix it.

Sayan751:
# Pull Request

## 📖 Description

The non-RFC-compliant email rule is now deprecated. It is recommended to apply a RFC compliant email address parser in a custom rule via `.satisfiesRule()` or `.satisfies()` instead.

### 🎫 Issues

Closes #1919

## 📦 Changeset

### DIFF
--- a/.changeset/beige-groups-push.md
+++ b/.changeset/beige-groups-push.md
@@ -1,0 +1,9 @@
+---
+"@aurelia/validation": patch
+---
+
+chore: removed email address validation rule
+
+This is a BREAKING CHANGE. The decission is tken because the regex/pattern-based email validation was anyway suboptimal. It is better to employ a proper email address parser and create a custom rule for that.
+
+Fixes #1919
--- a/.changeset/beige-groups-push.md
+++ b/.changeset/beige-groups-push.md
@@ -1,9 +1,0 @@
----
-"@aurelia/validation": patch
----
-
-chore: removed email address validation rule
-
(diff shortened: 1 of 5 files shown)


bigopon:
is it out of our reach to implement a proper email validation, regardless how big? it can still be an add-on export right in this package. If not then this PR is fine I guess.

bigopon:
Do we know whether other validation frameworks provide partial/complete email validation? if our regex is practical enough to make frontend development simple (not having to carry a big plugin to do an email validation), it wouldn't be too bad to keep, since the real validation should happen on the server anyway.

bigopon (approved):
nice work @Sayan751 👍

PR #2389, opened 2026-02-21
fix: better satisfies rule types

### LINKED ISSUE #2347: Pass generic types to `satisfies` parameter
# 🙋 Feature Request

`satisfies` condition delegate parameters don't have types

## 🤔 Expected Behavior

Delegate parameters should be typed

## 😯 Current Behavior

Delegate parameters have `any` type which upsets @typescript-eslint/no-unsafe-argument eslint rule

## 💁 Possible Solution

Update functions definition to 
```
public satisfies(condition: RuleCondition<TObject, TValue>)
```

Same for `rules` property
```
export interface IValidationRules<TObject extends IValidateable = IValidateable> {
  rules: PropertyRule<TObject>[];
```

and any other ones which don't specify generics

bigopon:
## 📖 Description

As per title, currently we don't pass the types through from validation rules.

### 🎫 Issues

Closes #2347 

cc @Sayan751 @MaximBalaganskiy

### DIFF
--- a/packages/validation/src/rule-provider.ts
+++ b/packages/validation/src/rule-provider.ts
@@ -529,11 +529,11 @@ export class ValidationRules<TObject extends IValidateable = IValidateable> impl
       rule = new PropertyRule<TObject>(this, this.messageProvider, new Property(expression, name));
       this.rules.push(rule);
     }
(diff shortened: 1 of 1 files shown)


Sayan751 (approved):
Nice work @bigopon 👍

PR #2391, opened 2026-03-04
fix(router): href handling via load and as-element

Sayan751:
# Pull Request

## 📖 Description

This issue is reported in [discord](https://discord.com/channels/448698263508615178/1477954486948462643).

This is how it looks like now after the change.
image

## 📦 Changeset

- [ ] Added a changeset via `npx changeset` (or not needed for this PR)


packages/router/src/resources/load.ts
bigopon: we probably don't want to do the code href & !href twice for controller & !controller. Also a bit: ternary should have positive case handled first 😃

PR #2396, opened 2026-03-21
fix(router): empty path handling for hash-based setup

### LINKED ISSUE #2393: Error: AUR3500 when navigating to root
# 🐛 Bug Report
When an app is hosted under a subfolder, e.g. http://localhost/mobile/, and router is configured to use #, changing the url to http://localhost/mobile/#/ from http://localhost/mobile/#/other or calling `window.history.back()` causes a AUR3500 error

Repro link: https://github.com/MaximBalaganskiy/router-issue

## 🤔 Expected Behavior

Navigation should happen without errors

## 🌍 Your Environment

| Aurelia          | 2.0.0-rc0

Sayan751:
# Pull Request

### 🎫 Issues

Closes #2393 

## 📦 Changeset

### DIFF
--- a/packages/__tests__/karma.conf.cjs
+++ b/packages/__tests__/karma.conf.cjs
@@ -11,6 +11,13 @@ const baseKarmaArgs = 'karma start karma.conf.cjs  --browsers=ChromeDebugging --
 const cliArgs = process.argv.slice(2).filter(arg => !baseKarmaArgs.includes(arg));
 const hasSingleRun = process.argv.slice(2).includes('--single-run');
 
+// Extract --grep option
+let grep = undefined;
+const grepIndex = cliArgs.indexOf('--grep');
+if (grepIndex !== -1 && grepIndex + 1 < cliArgs.length) {
+  grep = cliArgs.splice(grepIndex, 2)[1]; // Remove --grep and its value from cliArgs
+}
+
 const junitCircleCi = require('./z-scripts/circleci-junit-reporter.cjs');
 
 module.exports =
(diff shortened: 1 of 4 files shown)


packages/__tests__/karma.conf.cjs
```diff
@@ -171,6 +178,7 @@ module.exports =
         bail: config['bail'],
         ui: 'bdd',
         timeout: 5000,
+        ...(grep && { grep }),
```
bigopon, line 181: 👍

packages/__tests__/src/router/location-manager.spec.ts
```diff
@@ -546,6 +546,45 @@ describe('router/location-manager.spec.ts', function () {
         await au.stop(true);
       });
     }
+
+    // https://github.com/aurelia/aurelia/issues/2393
+    it('handles navigation to root in subfolder deployment with hash routing (issue #2393)', async function () {
```
bigopon: which part of the test is related to "subfolder deployment"?

bigopon (approved):
nice work @Sayan751 👍

packages/__tests__/src/router/location-manager.spec.ts
Sayan751: Thanks @bigopon for noticing that. I first wrote the test differently like the repro-repo in the linked issue. But as per my analysis, the root cause has nothing to do with subfolder deployment. Hence, I later simplified the test, but forgot to change the name, which is now done.

PR #2401, opened 2026-04-12
fix(router): handling encoded parameter values

### LINKED ISSUE #2398: Router strange behavior when child-router route parameter contains space
**tl;dr** 
Aurelia 2.0-rc.1 + router 

When a child router route parameter value contains a space, reloading page (hitting f5) causes the entire route to be appended at the end of current route. 
Subsequent refresh fails, because such route is invalid.

This does not happen at the top level routes, regardless of the spaces being present or not.

Repro live: https://migajek.github.io/au2-router-bug-space/
Repro code: https://github.com/migajek/au2-router-bug-space?tab=readme-ov-file

** more context**
The same behavior will occur if you open the link directly, for instance by clicking the middle button on the "Location 2 with space" - opening directly in new tab, causes the route in the URL bar to be already broken.

Sayan751:
# Pull Request

### 🎫 Issues

Fixes #2398 

## 📦 Changeset

### DIFF
--- a/.changeset/major-guests-invite.md
+++ b/.changeset/major-guests-invite.md
@@ -1,0 +1,5 @@
+---
+"@aurelia/route-recognizer": patch
+---
+
+fix(router): handling encoded parameter values. fixes #2398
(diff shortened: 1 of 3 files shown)


Sayan751:
@bigopon I am merging this one. If you notice something, I can open a subsequent PR later.

PR #2413, opened 2026-05-18
fix(router): correctly handle dom operation with containerless

### LINKED ISSUE #2403: <au-viewport containerless> throws "appendChild on Node" error since beta.24
# 🐛 Bug Report
 
`<au-viewport containerless>` throws "Failed to execute 'appendChild' on 'Node'" since beta.24
 
Repro link: https://stackblitz.com/edit/router-containerless-v2-0-0-beta-24?file=src%2Fmy-app.html (router-lite)
Repro link: https://stackblitz.com/edit/router-containerless-v2-0-0-beta-23?file=src%2Fmy-app.html (router-lite)
Repro link: https://stackblitz.com/edit/router-containerless-v2-0-0-rc-1?file=src%2Fmy-app.html (router)
 
## 🤔 Expected Behavior
 
`<au-viewport containerless>` renders routed content without the `<au-viewport>` wrapper element in the DOM, as it did in `v2.0.0-beta.23`.
 
## 😯 Current Behavior
 
The app crashes immediately on `v2.0.0-beta.24` and all later versions up to `rc.1` with:
 
```
[ERR Router] AUR3271: Transition T(id:1,trigger:'api',instructions:[VPI(c:'',viewport:default)]) failed with error: TypeError: viewportHost.append is not a function
```
 
Removing `containerless` fixes the crash, but leaves an unwanted `<au-viewport>` element in the DOM.
 
| Version | Status |
|---|---|
| `v2.0.0-beta.23` | Works |
| `v2.0.0-beta.24` through `rc.1` | Broken |
 
## 💁 Possible Solution
 
The regression was introduced by commit [`402c746`](https://github.com/aurelia/aurelia/commit/402c746) (#2128, @bigopon). The new DOM manipulation in `ComponentAgent._activate()` calls `viewportHost.append(...)`, but when the viewport is `containerless`, its host is a **Comment node** which does not support `append`.
 
```ts
// packages/router-lite/src/component-agent.ts
case MountTarget.location:
  viewportHost.append(controller.location!.$start!, controller.location!); // ← crashes when viewport is containerless

https://github.com/aurelia/aurelia/blob/083d25fc9735b8a292209c2b75067e9a1d4e408f/packages/router/src/component-agent.ts#L110
```
 
## 🔦 Context
 
Encountered this while upgrading an aurelia application from `v2.0.0-beta.23` to `v2.0.0-rc.1`. The `containerless` attribute on `<au-viewport>` is used to avoid polluting the DOM with an extra wrapper element.
 
## 💻 Code Sample
 
```html
<!-- my-app.html -->
<au-viewport containerless></au-viewport>
```

bigopon:
## 📖 Description

As reported in #2403 , when a containerless attribute is forced on `<au-viewport>`, the existing code path isn't handling insert before correctly for DOM. This PR fixes it

### 🎫 Issues

Resolves #2403 thanks @euglv


packages/router/src/component-agent.ts
```diff
@@ -95,7 +95,7 @@ export class ComponentAgent<T extends IRouteViewModel = IRouteViewModel> {
    */
   private _mountToViewport(): void {
     const controller = this._controller;
-    const viewportHost = this._ctx.vpa.hostController.host;
+    const viewportHost = this._getViewportMountParent();
```
Sayan751: Unused; can be removed.

packages/__tests__/src/router/router-containerless.spec.ts
```diff
@@ -62,4 +62,26 @@ describe('router/router-containerless.spec.ts', function () {
...
+      template: 'root <au-viewport containerless></au-viewport>'
+    })
+    class App {}
+
+    const { host } = await start({ appRoot: App });
+
+    assert.html.textContent(host, 'root foo');
+    assert.strictEqual(null, host.querySelector('au-viewport'));
+    assert.includes(host.innerHTML, '<foo>foo</foo>');
+  });
```
Sayan751, line 86: Can one more test be added for nested containerless viewports?

Sayan751 (commented):
Great work @bigopon!

packages/router/src/component-agent.ts
```diff
@@ -132,6 +132,50 @@ export class ComponentAgent<T extends IRouteViewModel = IRouteViewModel> {
     }
   }
 
+  /**
+   * Resolve the actual DOM parent that routed content should mount into.
+   * Containerless viewports are mounted at a render location, so their parent node
+   * is the insertion point rather than `hostController.host`.
+   */
+  private _getViewportMountParent(): Node {
```
Sayan751: Can be removed, it seems.

bigopon:
@Sayan751 resolved the comments, pls help review again. The previous version were about ensuring we are not performing any dom operation in case the controller host/location already connected to the target host. But it's probably not necessary. we can add that any time when some observable issue comes.

Sayan751 (approved):
Looks good to me. Probably add a `@internal` to the `_insertToViewport`?

PR #2414, opened 2026-05-18
fix(vite-plugin): less intrusive way to pair/handle html modules

### LINKED ISSUE #2063: Vite build watch cannot handle html imports

### LINKED ISSUE #2371: vite: Image assets referenced in html are not included as assets in vite build
# 🐛 Bug Report

It seems as if `` assets are not included in the vite build. I can get the images to show when using `vite` but not `vite build` - the build does not include the images as an asset.

Repro link: https://stackblitz.com/edit/au2-conventions-ox6fa5hr?file=src%2Fmy-app.html

## 🤔 Expected Behavior

I would expect that I can reference a relative image like `` and that this image would be included in the vite build output.

## 😯 Current Behavior

Various combinations of `` can make the image show in the dev server output, but the image is never included in the build output.

## 🌍 Your Environment

| Aurelia          | 2.0.0-rc.0
| Language         | typescript/html
| Browser          | chrome
| Bundler          | vite
| Operating System | windows
| NPM/Node/Yarn    |

### LINKED ISSUE #2411: vite-plugin: non-`production` Vite modes silently strip compiled HTML templates from lazy route chunks
## Summary

`@aurelia/vite-plugin@2.0.0-rc.1` (and current `master`) gates its HTML-import rewrite on a literal equality check against `'production'`:

https://github.com/aurelia/aurelia/blob/master/packages-tooling/vite-plugin/src/index.ts#L62-L67

```ts
transformHtmlImportSpecifier: (s) => {
  return $config.mode === 'production'
    ? s.replace(/\.html$/, '.$au.ts')
    : s;
},
```

For any Vite mode that isn't *literally* `'production'` — e.g. `prod`, `staging`, `production-eu`, or any custom mode name — the rewrite is silently skipped, the `.html` template import is never resolved into a compiled `.$au.ts` virtual module, and **the compiled template is stripped from the resulting lazy chunk**.

At runtime this surfaces as Aurelia's router being unable to resolve any lazy route component:

```
[ERR Router] AUR3271: Transition T(id:1,trigger:'api',instructions:[VPI(c:'')])
  failed with error: Error: AUR3174: Failed to resolve VR(viewport:'default',component:'unnamed-20')
```

The class is registered but with synthetic `unnamed-N` metadata because no CustomElement definition was attached. End-user impact: blank page across the entire app on `vite build --mode <not-production>`.

## Reproduction

1. Create a minimal Aurelia 2 app with at least one lazy route (`component: import('./route')`).
2. Configure `vite.config.ts` with `aurelia({ useDev: true })`.
3. Build with `vite build --mode prod` (or any non-`production` mode).
4. Inspect the route chunk in `dist/assets/<route>-route-*.js` — compare with the same build under `vite build` (default mode `production`).
5. The non-`production` build's chunk will be smaller and will not contain the template-derived literals (class names, text fragments) from the route's `.html` file.

## Real-world impact

We hit this in production on `liverty-music/frontend@v1.0.0`. We had named our prod build mode `prod` so Vite would load a `.env.prod` overlay (per Vite's standard `.env.<mode>` convention). The dev build (default `mode === 'production'`) worked correctly. The prod build silently shipped template-less route chunks, producing a blank page at `https://liverty-music.app/` until we identified the root cause and switched the architecture.

Full write-up of the diagnosis and our workaround (move all env config to runtime so `mode` stays `'production'` always): https://github.com/liverty-music/specification/tree/main/openspec/changes/archive/2026-05-16-adopt-runtime-config-for-frontend

## Suggested fix

The gate should be "is this a build (not a dev server)?" rather than "is the mode named `production`?". One of these would correctly handle every mode name:

```ts
// Option A — gate on command
$config.command === 'build'

// Option B — gate on Vite's resolved isProduction flag (Vite ≥ 5)
$config.isProduction
```

Both are one-line changes. We'd prefer the gate to fire for *any* `vite build` invocation regardless of `--mode` naming, because the HTML→virtual-`.$au.ts` rewrite is a build concern, not a config-mode concern. The Vite dev server (`vite` / `vite dev`) is the only context where the rewrite must be skipped (HMR uses the raw HTML path).

## Why this is hard to catch

- The build *succeeds* with no warning.
- The resulting chunk *loads* with no console error.
- The page reaches the router, which then fails to resolve the (chunk-defined) route class because its metadata is incomplete.
- The router's error references the synthetic `unnamed-N` name, not the original class name, so grepping the error message gives no hint that templates were stripped.
- A "smoke check passes env values are embedded in the bundle" spot-check does NOT catch this — env values come from Vite's `import.meta.env` substitution, which is independent of the HTML rewrite.

## Workaround (for other consumers hitting this)

If you can't change Vite mode, the canonical fix is to make your build env-agnostic and source per-environment values at runtime (e.g., `fetch('/config.json')` at bootstrap). That keeps `mode === 'production'` for every `vite build` invocation and removes the entire interaction surface. We documented our migration here in case it's useful to others: https://github.com/liverty-music/frontend/blob/main/docs/runtime-config.md

Happy to provide a minimal reproduction repo if it would help.

bigopon:
## 📖 Description

This PR reworks the Vite convention path for HTML views so Aurelia no longer compiles paired .html files by reading them directly from disk. Conventional TS resources now import a separate Aurelia view module via ?au-view, and the Vite plugin resolves that module as a thin proxy that re-exports the original .html module.

This keeps the separate Aurelia module identity needed for convention processing, while letting the underlying HTML file continue through Vite html pipeline.

Webpack behavior is intentionally unchanged. To introduce a query based rework, we need applications to change their webpack config, which makes the plugin usage unintuitive. We probably need to introduce a webpack plugin for that. Without AI webpack internal API/doc is quite a deterrent, but now maybe we can afford to go a bit further.

### Details

- Vite convention rewrites now target `*.html?au-view` instead of the previous synthetic compiled-module path.
- The Vite plugin resolves and loads `?au-view` as a proxy module that re-exports the original .html module.

### 🎫 Issues

Resolves #2411 
Resolves #2371
Resolves #2063


bigopon:
cc @3cp this is based on your idea, i'll do some tests with the per PR package and verify

packages-tooling/__tests__/src/plugin-conventions/preprocess-html-template.spec.ts
```diff
@@ -160,6 +160,32 @@ export function register(container) {
     assert.equal(result.code, expected);
   });
 
+  it('processes template with transformed html dependency specifier', function () {
+    const html = '<import from="./hello-world.html" /><template></template>';
+    const expected = `import { CustomElement } from '@aurelia/runtime-html';
+import * as d0 from "./hello-world.html?au-view";
+export const name = "foo-bar";
+export const template = "<template></template>";
```
3cp: This is different from my idea. What you did is using `.html?au-view` for a html module (for hello-world.html here), but to utilise standard vite toolchain on html file, we have to rewrite this line to
```js
export template from "./cleaned-up-foo-bar.html";
```

So that vite toolchain can process that "cleaned-up-foo-bar.html" (real name is probably should be original html file name like foo-bar.html) content.

Right now, vite toolchain has no way to process our "inner html content" in a js string.
```js
export const template = "<template>inner html content</template>";
```

3cp (changes requested):

packages-tooling/__tests__/src/plugin-conventions/preprocess-html-template.spec.ts
3cp: My idea was:
One au2 html source file -> two output resources after au2 vite plugin:
1. a piece of code, and if you call it foo-bar.html.js/ts, then you can simplify the checkers in the plugin, just need to skip `.html.js/ts` resources.
2. a piece of clean html stripped off all au2 meta tags, probably reuse the original source file name foo-bar.html.

Then vite standard toolchain can continue to process that clean html resource foo-bar.html.

bigopon:
@3cp thanks for the comment, I'm pretty sure that's my original understanding, not sure what went wrong in the middle.
Regardless, I've tweaked the implementation to follow that, and it's ready for review again. Pls help have a look.

I've also add a mocha test (with a bunch of mock), so we can have quicker feedback cycle, just in case, but I still used both the e2e and the node based test to verify.

bigopon:
After testing with my app, I realized it's not as simple as introducing a module in the middle. A normal compilation result would look like this:

```
// app.ts
import * as __def from './app.html.au.ts'

// app.html.au.ts
import template from './app.html';
```

The line `import template from './app.html'` is where it gets tricky:
- we cannot just transform the code returned to JS/TS, because it breaks production build
- we cannot leave it as html string, because in production build doesn't know what to do with a html module, it doesn't have a default export, or any code exported at all
- we cannot add `?raw` to the specifier, `./app.html?raw` automatically bypasses all internal html pipeline of Vite, which means no ``/`<video src>` etc...

It seems our best options is to still use:
- our `.$au.ts` (or `.html.au.ts`) & inline the template in the virtual file 
- add a plugin to handle processing of assets related rags (copy Vite internal code) to resolve #2371

and add some more code to fix #2411 & #2063

3cp:
I will have some test too.

packages-tooling/plugin-conventions/src/preprocess-html-template.ts
```diff
@@ -137,9 +139,16 @@ export function preprocessHtmlTemplate(
...
   statements.forEach(st => m.append(st));
-  m.append(`export const name = ${s(name)};
-export const template = ${s(options.transformHtml?.(html) ?? html)};
-export default template;
+  if (!inlineTemplate) {
+    m.append(`import * as __au2Template from ${s(templateSourceModuleSpecifier)};\n`);
+  }
+  m.append(`export const name = ${s(name)};\n`);
+  if (!inlineTemplate) {
+    m.append(`export const template = __au2Template.default;\n`);
```
3cp: If I understand correctly, this `inlineTemplate` didn't change the fact that `processHtmlTemplate` (and the higher level `preprocess`) is still an one-to-one generator.

You didn't export the clean `html` (the js var) string to the output of `processHtmlTemplate` `{code, map}`.

Should not we update the output shape to `{code, map, html?}`?

3cp: Since you didn't save the clean `html`. I assume the result of `__au2Template.default` is the original html content which contains all meta tags like `<import>`?

3cp:
I think part of issue of your `The line import template from './app.html' is where it gets tricky` is due to you didn't generate a clean `app.html` like I commented in your code, you were still reading the original `app.html` file.

The other problem is vite doesn't by default support importing html content in js code. The vite doc shows html as an entry point, and map them into routes, and that is seemly the only scenario it covered.

The html import can be done with a plugin. But again, once you use a plugin, it doesn't go through vite's html toolchain.
```js
import { defineConfig } from 'vite'
import { string } from 'vite-plugin-string'

export default defineConfig({
  plugins: [string({ include: '**/*.html' })],
})
```

Vite's dealing of html files seems very opinionated, it doesn't anticipate anyone to import html in js code. Now I don't think my idea would work :(

bigopon:
> Vite's dealing of html files seems very opinionated, it doesn't anticipate anyone to import html in js code.

And it's even more blocked by the fact they force a html processor in the end so you can never just have your own html -> code transformer in the pipeline. It's all or nothing. And with all the necessary requirements, plus the help of AI, i think we should take the nothing route, and replicate everything.

packages-tooling/plugin-conventions/src/preprocess-html-template.ts
bigopon: yes, but with an exception that our html loader will strip the meta tags. But that doesn't work as rollup doesn't understand html import by default, and thus will give back no code with `import _template from 'el.html'` in production build.

bigopon:
I'll leave this branch where it is, even though there's some extra work related to #2411, for future reference.

bigopon:
After some more digging, I see current vite plugin order is:

```ts
...normalPlugins,
isBuild && buildHtmlPlugin(config),
...postPlugins,
...buildPlugins.post
```

This means if we follow your idea @3cp , we can make our html plugin enforced `post`, and it'll run after the index html plugin from vite internal. The issue is at that point, it will no longer be just a HTML string exported as a JS module, since the vite html plugin treats all `.html$` files as its targets (a vite app probably assume multiple entries via multiple root .html files)

So this is actually likely a lost cause, at least until there's some changes in the core of vite. Maybe the change is to have Vite mark entry module, and the buildHtmlPlugin will only touch those modules, instead of anything and everything html like it is at the moment.

PR #2418, opened 2026-05-20
fix (router): Deep link to a child with empty parent route

### LINKED ISSUE #2406: Deep link to a child route fails when the parent route has `path: ""` (`AUR3401`)
# 🐛 Bug Report

When the root component (`@route` on `App`) registers a child route with `path: ""`
that mounts a layout component, and that layout component declares its own
non-empty child routes, navigation works correctly **only** if the user first
loads the application at `/`. Loading the application directly at any of the
nested routes (a deep link, e.g. `/shops`) throws:

```
Uncaught (in promise) UnknownRouteError: AUR3401: Neither the route 'shops'
matched any configured route at 'app' nor a fallback is configured for the
viewport 'default' - did you forget to add 'shops' to the routes list of the
route decorator of 'app'?
```

The router does not descend into the empty-path parent's route configuration to
resolve the remaining URL segment when the URL is matched fresh at startup
(deep link). After the empty-path parent has been mounted via a "/" load, the
exact same URL (`/shops`) resolves correctly.

Repro link: https://github.com/rmja/aurelia-router-repro

1. `npm install`
2. `npm start`
3. Open `http://localhost:5173/` — page renders, `shell-layout` redirects to
   `shops`, URL becomes `/shops`. Clicking the "Billing accounts" link also
   works. 
4. Hard-reload (or open a new tab) on `http://localhost:5173/shops` directly.
   The console throws `AUR3401` and nothing renders. 

The same happens for `/billing-accounts`.

## 🤔 Expected Behavior

Loading `/shops` from a cold start should resolve `App` → `ShellLayout`
(empty-path parent) → `Shops`, the same way it does once the user has visited
`/` first.

## 😯 Current Behavior

The router tries to match `shops` against the routes declared on `App` only,
sees no entry for `shops`, and throws `AUR3401`. It does not fall through into
the `path: ""` parent's child route table at the root recognition step.

## 💁 Possible Solution

In `dist/esm/index.dev.mjs`, around `createAndAppendNodes`:

```js
([rr, ...childrenRoutes] =
  ctx.routeConfigContext.recognize(path, false, node.pathFromRoot) ?? [null]);
const residue = rr?.residue ?? null;
// If the residue matches the whole path it means that empty route is configured,
// but the path in itself is not configured.
// Therefore the path matches the configured empty route and puts the whole path
// into residue.
if (rr == null || residue === path) {
    // ... try route-id ...
    return createFallbackNode(vi, node, log);
}
```

When the URL is `shops` and `App` has only an empty-path route, `recognize`
matches the empty path with `residue === "shops"`, so the condition
`residue === path` is true and the code falls through to `createFallbackNode`
instead of mounting the empty-path parent and forwarding `shops` as a child
instruction to its viewport.

This makes deep linking with a layout-style empty-path parent unusable unless
every nested route is also re-declared at the root (or a wildcard fallback is
added at the root).

## 🔦 Context

I have tried with `useEagerLoading: true` but that does not seem to be the way to go.

## 💻 Code Sample

A repro link is included above but the gist of the route configuration is:

`src/app.ts`:

```ts
const routes: Routeable[] = [
  {
    path: "",
    component: import("./shell-layout"),
  },
];

@route({ routes })
export class App {}
```

`src/shell-layout.ts`:

```ts
const routes: Routeable[] = [
  { path: "",                redirectTo: "shops" },
  { path: "shops",           component: import("./shops"),           title: "Shops" },
  { path: "billing-accounts",component: import("./billing-accounts"),title: "Billing accounts" },
];

@route({ routes })
export class ShellLayout implements IRouteViewModel {}
```

## 🌍 Your Environment

- `aurelia` `2.0.0-rc.1`
- `@aurelia/router` `2.0.0-rc.1`
- `@aurelia/vite-plugin` `2.0.0-rc.1`
- `vite` 7
- Browser: any (reproduces in Chromium)

I already do:)

Sayan751:
# Pull Request

### 🎫 Issues

Closes #2406

## 📦 Changeset

- [ ] Added a changeset via `npx changeset` (or not needed for this PR)

### DIFF
--- a/packages/__tests__/src/router/smoke-tests.spec.ts
+++ b/packages/__tests__/src/router/smoke-tests.spec.ts
@@ -7534,4 +7534,63 @@ describe('router/smoke-tests.spec.ts', function () {
       await au.stop(true);
     });
   }
+
+  /**
+   * Reproduction: App has a single empty-path route pointing to a ShellLayout component.
+   * ShellLayout owns the real navigation routes ('shops', 'billing-accounts').
+   * Hard-reloading at /shops (or /billing-accounts) used to throw AUR3401.
+   * Each row deep-links to `initialPath`, then navigates in-app to `navPath`.
+   */
+  for (const { initialPath, initialContent, navPath, navContent } of [
+    { initialPath: '/shops',            initialContent: 'shops',   navPath: 'billing-accounts', navContent: 'billing' },
+    { initialPath: '/billing-accounts', initialContent: 'billing', navPath: 'shops',            navContent: 'shops'   },
+  ]) {
+    it(`deep-link to '${initialPath}' then navigate to '${navPath}' - GH issue #2406`, async function () {
+      @customElement({ name: 'shops-page', template: 'shops' })
+      class ShopsPage { }
+
+      @customElement({ name: 'billing-page', template: 'billing' })
+      class BillingPage { }
+
+      @route({
+        routes: [
+          { path: '', redirectTo: 'shops' },
+          { path: 'shops', component: ShopsPage },
+          { path: 'billing-accounts', component: BillingPage },
+        ],
+      })
+      @customElement({ name: 'shell-layout', template: '<au-viewport></au-viewport>' })
+      class ShellLayout { }
+
+      @route({
+        routes: [
+          { path: '', component: ShellLayout },
+        ],
+      })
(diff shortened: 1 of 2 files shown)


bigopon (approved):
nice work @Sayan751 , and very nice bug report @rmja 👍

PR #2419, opened 2026-05-21
fix(vite-plugin): don't enforce development resolution for all packages when use dev is true

### LINKED ISSUE #2400: @aurelia/vite-plugin causes incorrect version of dependency to be included
# 🐛 Bug Report

When useDev is enabled in @aurelia/vite-plugin, dependency resolution conditions are globally changed in a way that can cause non-Aurelia packages to resolve to unintended builds. In my case, importing typesense causes axios to resolve to its Node build path instead of the browser build, which breaks browser behavior.

Repro link: https://github.com/ivanbacher/vite-aurelia-test

in `vite.config.ts` toggle `useDev: true` / `useDev: false`

## 🤔 Expected Behavior

Enabling useDev should only influence Aurelia package resolution (for dev exports), and should not alter resolution behavior of unrelated third-party dependencies.

With useDev enabled, importing typesense should still keep axios on the browser adapter/build in browser apps.

## 😯 Current Behavior

With useDev true:
 - Importing typesense causes axios to resolve through Node-oriented path internally (including Node adapter internals such as zlib-related code paths).

With useDev false:
- The issue disappears and browser axios resolution is used as expected.

## 🔦 Context

I am trying to use typesense in a browser Aurelia + Vite app while also using useDev true for Aurelia developer experience.

This issue makes dependency behavior inconsistent and surprising:

Same app and bundler
- Only difference is useDev true/false
- Unrelated package resolution changes as a side effect

## 🌍 Your Environment

Aurelia | @aurelia/vite-plugin 2.0.0-rc.1
Language | TypeScript
Browser | Browser environment (repro in dev server)
Bundler | Vite 7.3.1
Operating System | macOS
NPM/Node/Yarn | npm

bigopon:
## 📖 Description

Per #2400 , we need a different way to resolve development build of aurelia when `useDev` is true, this PR adds that.

### 🎫 Issues

Resolves #2400

### DIFF
--- a/packages-tooling/__tests__/src/vite-plugin/vite-plugin.spec.ts
+++ b/packages-tooling/__tests__/src/vite-plugin/vite-plugin.spec.ts
@@ -51,6 +51,51 @@ describe('vite-plugin', function () {
     } as unknown as import('vite').ResolvedConfig;
   }
 
+  function createConfig(mode: string) {
+    return {
+      mode,
+      resolve: { alias: [], dedupe: [], conditions: [] },
+    } as unknown as import('vite').UserConfig;
+  }
+
+  function createPluginContext() {
+    return {
+      meta: { watchMode: false },
+      async resolve(id: string) {
+        return { id: `/resolved/${id}` };
+      },
+    };
+  }
+
+  it('does not mutate global resolve conditions when useDev is enabled', function () {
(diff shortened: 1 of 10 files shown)


bigopon:
Thanks @ivanbacher

PR #2421, opened 2026-05-25
fix(runtime): prevent task queue false deadlock detection

fkleuver:
# Pull Request

## 📖 Description

This hardens the task queue so large finite workloads no longer trip the recursive deadlock guard just because many tasks are queued during a drain.

Automatic `queueTask` drains now process work in timed slices and continue through host timer turns until the queue is idle. Manual `runTasks()` calls remain synchronous and keep an internal guard for runaway synchronous flushes.

This also improves task queue error reporting. Multiple task failures now surface as `TaskQueueAggregateError`, which extends `AggregateError`, preserves the original errors, sets `cause` to the first failure, and includes a more useful task-queue-specific message.

Because automatic drains may now yield between chunks, queued runtime-html work now re-checks lifecycle state before mutating DOM or observation state. This prevents delayed layout writes and `show` updates from acting after teardown or unsubscribe.

### 🎫 Issues

No active github issues as far as I could tell, but there were some reports on Discord that prompted a deeper investigation.

## 👩‍💻 Reviewer Notes

The main contract change is that very large automatic drains may span more than one host turn. Code that needs full completion should await `tasksSettled()` rather than assuming a single microtask is enough for every possible queue size.

`configureTaskQueue({ flushBudget })` can be used to control automatic drain pacing. The manual deadlock guard or timer internals are not exposed until specific use cases require otherwise.

The runtime-html guards are paired with the queue change: once yielding is possible, user input, stop/detach, or unsubscribe can happen before a later queued chunk runs.

## 📦 Changeset

- [x] Added changeset: `.changeset/task-queue-hardening.md`

### DIFF
--- a/packages/__tests__/src/3-runtime-html/show.integration.spec.ts
+++ b/packages/__tests__/src/3-runtime-html/show.integration.spec.ts
@@ -1,5 +1,5 @@
 import { Aurelia, customElement, ICustomElementViewModel, ICustomElementController } from '@aurelia/runtime-html';
-import { runTasks, tasksSettled } from '@aurelia/runtime';
+import { configureTaskQueue, queueTask, runTasks, tasksSettled } from '@aurelia/runtime';
 import { assert, TestContext } from '@aurelia/testing';
 
 function createFixture() {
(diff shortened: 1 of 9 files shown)


packages/__tests__/src/3-runtime-html/show.integration.spec.ts
```diff
@@ -121,6 +121,58 @@ describe('3-runtime-html/show.integration.spec.ts', function () {
 
       await au.stop();
     });
+
+    it('does not apply a queued show update after detaching during a yielded flush', async function () {
```
bigopon, line 125: 👍 though why is only show being updated?

bigopon (commented):
Looks good to me, tested with an app and there doesn't seem anything out of ordinary. I'm concerned that `await Promise.resolve()` no longer guarantees things are done, this kind of destroys the deterministic nature we currently have.

packages/__tests__/src/3-runtime-html/show.integration.spec.ts
fkleuver: Sorry for the late reply. Only show is being updated because other queued DOM paths (content, interpolation) already have the controller-state guard. This was simply the only one that was missing. It would have had to be fixed regardless of this PR.

fkleuver:
> Looks good to me, tested with an app and there doesn't seem anything out of ordinary. I'm concerned that `await Promise.resolve()` no longer guarantees things are done, this kind of destroys the deterministic nature we currently have.

That guarantee only existed for queues with no async tasks in them, and the documented guarantee has always been `tasksSettled()`. So yes, there are now more situations where a quick `await Promise.resolve()` won't cut it anymore, which seems mainly relevant in simple tests, but there they still work for small queues (which seems to hold up for all existing tests we have). For larger tests which have more than 256 tasks you'd probably be using `tasksSettled()` anyway.

Issue #2422, opened 2026-05-27
Default text translations are inserted as HTML


Vheissu:
## Summary

`@aurelia/i18n` appears to parse default text translations as HTML. The docs say default `t="key"` output should escape markup unless `[html]` is explicitly used, but both default text and explicit `[text]` create live DOM nodes.

## Reproduction

```html
<div id="default-i18n" t="payload"></div>
<div id="explicit-text-i18n" t="[text]payload"></div>
<div id="html-i18n" t="[html]payload"></div>
```

```ts
const payload =
  'Translation text <img src="x" onerror="document.body.dataset.i18nXss = \'fired\'"> <strong>HTML node created</strong>';
```

## Expected

`t="payload"` and `t="[text]payload"` should render the string as text. The `<img>` and `<strong>` tags should be escaped.

Only `t="[html]payload"` should create HTML.

## Actual

Both text paths create a live `<img onerror>` node. The inline event can execute.

## Reproduction

Codepen: https://codepen.io/DigitalSea/pen/NPbwjbo

## Relevant code

`packages/i18n/src/t/translation-binding.ts`

`_preprocessAttributes()` defaults to `textContent`, but `_addContentToTemplate()` always uses:

```ts
parser.innerHTML = content;
```

## Proposed fix

Use separate paths for text and HTML content:

- `textContent`: append a text node
- `[text]`: append a text node
- `[html]`: parse as HTML
- decide whether `[prepend]` / `[append]` remain HTML-capable or need explicit HTML variants

Add regression tests proving that `t="payload"` and `t="[text]payload"` do not create DOM nodes from markup.

fkleuver:
Fixed by https://github.com/aurelia/aurelia/pull/2424

PR #2424, opened 2026-05-28
fix(i18n): ensure text content is escaped

Vheissu:
Default and [text] replacements use textContent so markup is not parsed as HTML. [html] still permits markup. Added tests to verify markup is escaped for default and [text] replacements.

Relates to issue #2422

### DIFF
--- a/packages/i18n/src/t/translation-binding.ts
+++ b/packages/i18n/src/t/translation-binding.ts
@@ -309,8 +309,17 @@ export class TranslationBinding implements IBinding {
 
     this._addContentToTemplate(template, content.prepend, marker);
 
-    // build content: prioritize [html], then textContent, and falls back to original content
-    if (!this._addContentToTemplate(template, content.innerHTML ?? content.textContent, marker)) {
+    // [html] body content parses markup; default/[text] body content stays text.
+    let hasContent = this._addContentToTemplate(template, content.innerHTML, marker);
+
+    if (!hasContent && content.textContent !== void 0 && content.textContent !== null) {
+      const textNode = this._platform.document.createTextNode(content.textContent);
+      Reflect.set(textNode, marker, true);
+      template.content.append(textNode);
+      hasContent = true;
+    }
+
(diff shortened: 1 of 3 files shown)


packages/i18n/src/t/translation-binding.ts
```diff
@@ -309,8 +309,17 @@ export class TranslationBinding implements IBinding {
...
-    if (!this._addContentToTemplate(template, content.innerHTML ?? content.textContent, marker)) {
+    // [html] body content parses markup; default/[text] body content stays text.
+    let hasContent = this._addContentToTemplate(template, content.innerHTML, marker);
+
+    if (!hasContent && content.textContent !== void 0 && content.textContent !== null) {
+      const textNode = this._platform.document.createTextNode(content.textContent);
+      Reflect.set(textNode, marker, true);
+      template.content.append(textNode);
+      hasContent = true;
+    }
```
bigopon: if we already add html to the node, and then later create text content out of the parsed html, wouldn't that be already too late?

packages/i18n/src/t/translation-binding.ts
```diff
@@ -307,15 +307,19 @@ export class TranslationBinding implements IBinding {
...
-      const textNode = this._platform.document.createTextNode(content.textContent);
-      Reflect.set(textNode, marker, true);
-      template.content.append(textNode);
+    this._addHtmlContentToTemplate(template, content.prepend, marker);
+
+    // Only explicit [html] content is parsed. Default and [text] translations stay literal,
+    // while prepend and append retain their documented HTML-capable behavior.
+    let hasContent = false;
+    if (content.innerHTML != null) {
+      hasContent = this._addHtmlContentToTemplate(template, content.innerHTML, marker);
```
bigopon, line 316: the naming of `innerHTML` tripped me, i had to open the code 😃

bigopon (approved):
nice work @Vheissu @fkleuver 👍

PR #2427, opened 2026-05-28
feat(router): add setTitle to route nodes

### LINKED ISSUE #2416: The title property on RouteNode is readonly
# 🐛 Bug Report

I am trying to set the page title in `loading()`:

```ts
@route({ routes })
export class ShopLayout implements IRouteViewModel {
  private readonly api = resolve(IAdminApi);

  async loading(params: { shopId: ShopId }, next: RouteNode) {
    const shop = await this.api.shops.getShop(params.shopId);
    next.title = shop.name;
  }
}
```

but it complains that `next.title` is readonly. It works just fine though - it specifically sets "the part of the title" that belongs to this route node.
The docs says that there should be a `next.setTitle()`
https://github.com/aurelia/aurelia/blob/24bb2d1fecc3148e5b88f627ae474ed11fac1fbe/docs/user-docs/developer-guides/migrating-to-aurelia-2/side-by-side-comparison.md?plain=1#L853
but it does not seem to exist.

## 🤔 Expected Behavior

I expect `RouteNode.title` to not be readonly.

## 😯 Current Behavior

It is currently readonly and I get the typescript error `Cannot assign to 'title' because it is a read-only property.`

## 🌍 Your Environment

| Aurelia          | `2.0.0-rc.1`
| Language         | `typescript`

Vheissu:
## Summary

- Add `RouteNode#setTitle()` for updating a route node title part during navigation lifecycle/router hooks while keeping `RouteNode.title` readonly.
- Cover dynamic route-node titles with a router test that asserts both `document.title` and `ICurrentRoute.title`.
- Align router docs around supported title paths: route config title parts, `setTitle`, navigation `title`, `buildTitle`, `updateTitle`, and direct `document.title` caveats.

Closes #2416.

### DIFF
--- a/docs/user-docs/router/api-reference.md
+++ b/docs/user-docs/router/api-reference.md
@@ -219,7 +219,7 @@ interface IRouterOptions {
   /** History interaction strategy. Default: 'push' */
   historyStrategy?: HistoryStrategy | ((instructions: ViewportInstructionTree) => HistoryStrategy);
 
-  /** Custom title builder function. Return null to skip title updates */
+  /** Custom final title builder function. Return null to skip document title updates */
   buildTitle?: ((transition: Transition) => string | null) | null;
 
   /** Generate navigation model for menus. Default: true */
@@ -252,7 +252,7 @@ interface INavigationOptions {
   /** Override router's history strategy for this navigation */
   historyStrategy?: HistoryStrategy | ((instructions: ViewportInstructionTree) => HistoryStrategy);
 
-  /** Page title for this navigation */
+  /** Final title for this navigation when no custom buildTitle is configured */
   title?: string | ((node: RouteNode) => string | null) | null;
 
   /** Separator between hierarchical titles. Default: ' | ' */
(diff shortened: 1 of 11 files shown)


docs/user-docs/router/api-reference.md
```diff
@@ -252,7 +252,7 @@ interface INavigationOptions {
   /** Override router's history strategy for this navigation */
   historyStrategy?: HistoryStrategy | ((instructions: ViewportInstructionTree) => HistoryStrategy);
 
-  /** Page title for this navigation */
+  /** Final title for this navigation when no custom buildTitle is configured */
```
Sayan751: Since the title can also be a function, The word "Final" sounds kind of odd.

Sayan751 (approved):
@Vheissu looks good to me. I would have simply removed the `readonly` modifier though, since the `get`/`setTitle` are now serving two completely different purposes. And also because, the get/set pattern is uncommon in AU2.

PR #2435, opened 2026-06-04
fix(runtime): ensure start/stop return promises

### LINKED ISSUE #2433: Incorrect typings for Aurelia.start/stop
# 🐛 Bug Report

Currently the call to `start()` and `stop()` on `Aurelia` returns `Promise<void> | void`. I believe that it previously returned `Promise<void>`. Looking at e.g. `start()` it seems that it always returns a promise, so why the explicit "or void"?

Image

Repro link:

## 🤔 Expected Behavior

`start` and `stop` should always return a promise. And if they already do that, then the typings should be fixed.

## 😯 Current Behavior

The typings indicate that `start` and `stop` can return void.

## 🌍 Your Environment

| Aurelia          | 2.0.0-rc.1
| Language         | typescript 6

Vheissu:
Ensure Aurelia.start() and Aurelia.stop() always return Promise<void> (even for synchronous paths) and reuse in‑flight promises so concurrent calls share the same promise. Added unit and type tests to verify the behaviour. Fixes #2433.

### DIFF
--- a/packages/runtime-html/src/aurelia.ts
+++ b/packages/runtime-html/src/aurelia.ts
@@ -118,7 +118,7 @@ export class Aurelia implements IDisposable {
 
   /** @internal */
   private _startPromise: Promise<void> | void = void 0;
-  public start(root: IAppRoot | undefined = this.next): void | Promise<void> {
+  public start(root: IAppRoot | undefined = this.next): Promise<void> {
     if (root == null) {
       throw createMappedError(ErrorNames.no_composition_root);
     }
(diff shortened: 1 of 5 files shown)


packages/runtime-html/src/aurelia.ts
```diff
@@ -141,11 +141,20 @@ export class Aurelia implements IDisposable {
         this._dispatchEvent(root, 'au-started', root.host);
       });
     });
+    return isPromise(startPromise)
+      ? this._startPromise = startPromise
+      : Promise.resolve();
```
bigopon, line 146: i dont think we can have this change without calling it a breaking change.

this code breaks
```ts
const startResult = au.start();
if (startResult instanceof Promise) {
  // breaks
}
```

If the intention is to fix concurrent in flight promise, then we only need the `this._startPromise = startPromise` part, no?

fkleuver:
Closing as not planned. #2433 was resolved as intentional API design, so making `start` and `stop` always return promises is not a change we plan to adopt.

PR #2439, opened 2026-07-28
fix(runtime): detach collection length and size observers

fkleuver:
# 🐛 Bug Fix

## 📖 Description

Collection length and size observers remained attached to their owning collection after their final subscriber unsubscribed, causing unnecessary collection-change work.

## 💁 Your Solution

Detach the observer from its owner on the zero-subscriber transition.

## 📑 Test Plan

Added lifecycle regression coverage for array length, map size, and set size observers.

## 📦 Changeset

- [x] Added a changeset

### DIFF
--- a/packages/runtime/src/collection-length-observer.ts
+++ b/packages/runtime/src/collection-length-observer.ts
@@ -136,6 +136,6 @@ function subscribe(this: CollectionLengthObserverImpl, subscriber: ISubscriber):
 
 function unsubscribe(this: CollectionLengthObserverImpl, subscriber: ISubscriber): void {
   if (this.subs.remove(subscriber) && this.subs.count === 0) {
-    this.owner.subscribe(this);
+    this.owner.unsubscribe(this);
   }
 }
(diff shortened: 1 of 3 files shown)


packages/runtime/src/collection-length-observer.ts
```diff
@@ -136,6 +136,6 @@ function subscribe(this: CollectionLengthObserverImpl, subscriber: ISubscriber):
 
 function unsubscribe(this: CollectionLengthObserverImpl, subscriber: ISubscriber): void {
   if (this.subs.remove(subscriber) && this.subs.count === 0) {
-    this.owner.subscribe(this);
+    this.owner.unsubscribe(this);
```
bigopon, line 139: well this is quite a big typo.

bigopon (approved):

packages/runtime/src/collection-length-observer.ts
fkleuver: yep

PR #2441, opened 2026-07-29
fix(router): preserve static nav and derive convention route ids

fkleuver:
# 🐛 Bug Fix

## 📖 Description

- **Static navigation:** Object-form route configurations ignored a component’s static `nav` value.
- **Convention route IDs:** Convention-only routes could expose an undefined or stale ID when their effective path was resolved later.

## 💁 Your Solution

- **Static navigation:** Preserve the static `nav` value when no explicit override is supplied.
- **Convention route IDs:** Derive the ID from the primary effective path, including child configuration and configuration-hook overrides.

## 📑 Test Plan

Added RouteConfig normalization coverage for both static navigation precedence and convention-derived route IDs.

## 📦 Changeset

- [x] Added a changeset

### DIFF
--- a/.changeset/fix-route-config-normalization.md
+++ b/.changeset/fix-route-config-normalization.md
@@ -1,0 +1,7 @@
+---
+"@aurelia/router": patch
+---
+
+Object-form route configurations now preserve a component's static `nav` value when no explicit override is supplied.
+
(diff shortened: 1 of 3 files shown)


Sayan751 (approved):
LGTM

Issue #2250, opened 2025-09-17
[router]: Inconsistent href generation


qraynaud:
# 🐛 Bug Report

I found out that when using a hierarchical router, the href generated are inconsistent at best. I tested with the following hierarchy:
* home
* sub
  * :id

Repro link: https://stackblitz.com/edit/au2-conventions-dtdfbyev

## 🤔 Expected Behavior

The links on the sub page should always be:
* `/sub/1`
* `/sub/2`
* `/sub/3`

## 😯 Current Behavior

The links looks like this [originally](https://au2conventionsdtdfbyev-nast--9000--96435430.local-corp.webcontainer.io/):
* `/1` (missing `/sub` prefix when using `href` and not `load`)
* `/sub/2`
* `/sub/3`

If you refresh the page on either sub page details (eg `/sub/1`), the links [become](https://au2conventionsdtdfbyev-nast--9000--96435430.local-corp.webcontainer.io/sub/2):
* `/1` (still missing `/sub`)
* `/sub/<X>/2` (where `<X>` is the parameter of the refreshed page and should not be there)
* `/sub/<X>/3` (where `<X>` is the parameter of the refreshed page and should not be there)

_Note:_ All links are always working properly and generating the correct URL once navigated to, the wrong URL is the one generated in the `href` tag and viewed by hovering the link. A big issue is that ctrl-clicking the link is failing since this does open a new tab with a bad URL.

## 💁 Possible Solution

I think the router is not correctly interpreting the correct context to generate the URL.

## 🔦 Context

I was trying to test aurelia 2 since I’m already a user of aurelia 1 and have a project in production with it. I wanted to assess if it was stable enough to migrate or create new project on with it. This bug with the router is a nogo for me though. Or maybe I’m doing something wrong?

ekzobrain:
Also faced mutiple issues with wrong href generation, just did not have time to post it yet. Will ad my cases here is they differ from the described above.

Sayan751:
@euglv I am not sure that I follow your last comment. I would suggest opening a new issue, preferably with a minimal reproduction. Below is what I have tried, and it seems to be working correctly; I am using `RouterConfiguration.customize({ useUrlFragmentHash: true })`.

![Image](https://github.com/user-attachments/assets/63a3531e-f006-46cc-91a5-b5e486ea4190)

PR #2443, opened 2026-07-29
fix(router): normalize navigation paths and base handling

fkleuver:
# 🐛 Bug Fix

## 📖 Description

#2369 established the router's current context-relative navigation contract:

- `/path` starts at the application route root;
- `path` and `./path` start at the current route context;
- each leading `../` selects one parent route context.

This PR fixes cases where the implementation violates that contract or where the materialized browser URL does not represent the destination selected by the router. It does not change the public navigation syntax or add another route resolution policy.

### Part 1: Correctness under the current contract

- **Application-root navigation:** Removing the configured base can erase the leading `/` before route-context selection. A rooted instruction issued from a nested component can consequently resolve in that nested context.
- **Parent traversal:** More leading `../` prefixes than available route parents can survive normalization and leak into the rendered browser URL.
- **Deployment bases:** Base matching treats `/app` as a prefix of `/apple`.
- **Hash-mode `load` URLs:** `load` links can omit a non-root application base even though `href`, intercepted navigation, and history-mode `load` retain it.

### Part 2: Relationship to the earlier work

- #2250 and #2254 established that rendered links, intercepted clicks, native activation, direct entry, and reload must represent the same destination.
- #2279 explored similar path behavior by adding nearest-ancestor recognition. #2369 superseded that work with explicit route-context prefixes. This PR does not resume the additional recognition path.
- #2391 made `load` materialize its generated target URL on native anchors, `as-element` anchors, and other elements. This PR completes its deployment-base handling in hash mode.
- #2256 also contains a broader proposal to separate URL navigation, route identity, and multi-viewport navigation at the public API level. That design and its compatibility implications remain separate from these fixes.

## 💁 Your Solution

- **Application-root navigation:** Record a scalar instruction's root intent before removing the deployment base, then use that retained intent only to select the application route context.
- **Parent traversal:** Consume every leading parent prefix while clamping route-context traversal at application root.
- **Readonly normalization:** Normalize into replacement instruction values rather than rewriting caller-supplied instruction arrays or viewport instruction objects.
- **Deployment bases:** Strip a base only at an exact path, query, or fragment boundary.
- **Hash-mode `load` URLs:** Materialize `load` targets through the location manager in both routing modes.

Route recognition remains in the existing authoritative pipeline. This PR does not add the ancestor-recognition layer from #2279 or introduce another route grammar.

## 📑 Test Plan

- Cover application-root and excess-parent navigation from a nested route under history and hash routing, through both `load` and `href`.
- Include a same-named nested route so an ancestor lookup cannot hide loss of the application-root marker.
- Cover frozen instruction arrays and viewport instruction objects at the prefix-normalization boundary.
- Cover exact, trailing-slash, query, fragment, and partial-segment base matching in browser and server location managers.
- Assert that non-root hash `load` targets retain the deployment base.
- Run the full framework build, full router Node/JSDOM suite, router lint, and router Playwright suite.

## 👩‍💻 Reviewer Notes

This PR deliberately implements the existing route-context semantics; it is not general URL-reference resolution. A future explicit URL-target API can resolve and recognize public URLs separately without adding a second recognizer here. This PR therefore does not add bare `..`, general dot-segment processing, URL-segment-relative handling of multi-segment routes, or independently relative entries within one multi-viewport instruction array.

Base stripping also still lacks provenance. With deployment base `/app`, the browser URL `/app/app/dashboard` can be stripped once by location reading and again by scalar instruction normalization. Solving that pre-existing collision requires provenance across browser events, SSR, and programmatic input.

## 📦 Changeset

- [x] Added `.changeset/calm-routes-agree.md`

### DIFF
--- a/.changeset/calm-routes-agree.md
+++ b/.changeset/calm-routes-agree.md
@@ -1,0 +1,9 @@
+---
+"@aurelia/router": patch
+---
+
+Fix application-root navigation under configured base paths. Root-prefixed string instructions now retain their application-root meaning when the base is removed, and a base such as `/app` is stripped only at a path, query, or fragment boundary—not from paths such as `/apple`.
+
+Consume excess leading `../` prefixes while clamping route-context traversal at the application root. This prevents unresolved parent syntax from leaking into browser URLs after intercepted navigation has already resolved at the root.
+
+Keep hash-mode `load` links under the configured application base, matching `href` links and the destination used by router navigation.
--- a/packages/__e2e__/6-router/src/router.basic.spec.ts
+++ b/packages/__e2e__/6-router/src/router.basic.spec.ts
@@ -169,10 +169,13 @@ test.describe('router.basic', () => {
         await page.goto(`${baseURL}/?useUrlFragmentHash=true`);
       }
 
+      // `href1` is a native href. The remaining values are produced by `load`
+      // on an as-element anchor, a native anchor, and a button. Every surface
+      // must materialize the same application-qualified URL in both modes.
       const href1 = `${baseURL}${useUrlFragmentHash ? '/#' : ''}/sub/1`;
-      const href2 = `${useUrlFragmentHash ? '/#' : baseURL}/sub/2`;
-      const href3 = `${useUrlFragmentHash ? '/#' : baseURL}/sub/3`;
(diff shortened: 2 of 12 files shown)


fkleuver:
cc @Sayan751

Sayan751 (approved):
LGTM. Thank you @fkleuver!

PR #2445, opened 2026-07-29
fix(runtime-html): include Z in default key mappings

fkleuver:
# 🐛 Bug Fix

## 📖 Description

The default keyboard mapping omitted `z`, `90`, and `122` because its generated A-Z range stopped at Y.

## 💁 Your Solution

Generate all 26 letter mappings.

## 📑 Test Plan

Added event-modifier integration coverage for `:z`, `:90`, and `:122`.

## 📦 Changeset

- [x] Added a changeset

### DIFF
--- a/packages/runtime-html/src/binding/listener-binding.ts
+++ b/packages/runtime-html/src/binding/listener-binding.ts
@@ -155,7 +155,7 @@ export const IKeyMapping = /*@__PURE__*/createInterface<IKeyMapping>('IKeyMappin
     space: 'Space',
     tab: 'tab',
     // by default, maps the key a-z and A-Z to their respective keycodes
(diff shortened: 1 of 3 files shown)


packages/runtime-html/src/binding/listener-binding.ts
```diff
@@ -155,7 +155,7 @@ export const IKeyMapping = /*@__PURE__*/createInterface<IKeyMapping>('IKeyMappin
     space: 'Space',
     tab: 'tab',
     // by default, maps the key a-z and A-Z to their respective keycodes
-    ...Array.from({ length: 25 }).reduce((acc: Record<string, string>, _, idx) => {
+    ...Array.from({ length: 26 }).reduce((acc: Record<string, string>, _, idx) => {
```
bigopon, line 158: oops

bigopon (approved):

PR #2448, opened 2026-07-30
fix(runtime-html): support reactive object destructuring in repeats

fkleuver:
# 🐛 Bug Fix

## 📖 Description

Object destructuring in repeat declarations produced empty locals or accidentally resolved values from parent scopes.

## 💁 Your Solution

Support shallow properties and aliases as reactive one-way locals. Reject unsupported or unsafe targets with AUR0177.

## 📑 Test Plan

Added parser and repeat coverage for reactivity, aliases, keyed replacement, lifecycle, SSR hydration, and invalid targets. Verified the repeat benchmarks.

## 📦 Changeset

- [x] Added a changeset


docs/user-docs/developer-guides/error-messages/0151-to-0179/README.md
```diff
@@ -34,5 +34,6 @@ These errors occur during expression parsing in templates and binding expression
 * [AUR0174](aur0174.md) - Arrow function default parameters not supported
 * [AUR0175](aur0175.md) - Arrow function destructuring not supported
 * [AUR0176](aur0176.md) - Rest parameter must be last
+* [AUR0177](aur0177.md) - Unsupported object binding pattern
```
bigopon, line 37: 👍

packages/__tests__/src/3-runtime-html/repeater.destructered-declaration.spec.ts
```diff
@@ -127,4 +135,290 @@ describe('3-runtime-html/repeater.destructered-declaration.spec.ts', function ()
       );
     }, { app: App, template: `<div repeat.for="[k,ps] of map">\${k}:<template repeat.for="p of ps"> \${p.name.slice(0,3)}-\${p.age}</div>` });
   }
+
+  describe('object binding patterns', function () {
+    it('projects shorthand and aliased locals before the repeated body binds', function () {
+      const { assertText, getAllBy } = createFixture(
+        `<div repeat.for="{ id: orderId, name } of orders"><span data-id.one-time="orderId">\${id}/\${orderId}/\${name}</span></div>`,
```
bigopon, line 147: I've disabled the lint rules, we could just use `'` or `"`, having to `\$` a lot reads quite difficult. But we have a mix of ` and " so dont worry about changing it, just fyi.

packages/__tests__/src/3-runtime-html/repeater.destructered-declaration.spec.ts
```diff
@@ -127,4 +135,290 @@ describe('3-runtime-html/repeater.destructered-declaration.spec.ts', function ()
...
+          public items = [
+            { $index: 'source-index', constructor: 'source-constructor' },
+          ];
+        },
+      );
+
+      assertText('source-index/source-constructor');
+    });
+
+    it('hydrates SSR-adopted rows with reactive object-pattern locals', async function () {
```
bigopon, line 180: 👍 the mix of one-time in the assertion confuses me why a bit, but nice test 👍

packages/__tests__/src/3-runtime-html/repeater.destructered-declaration.spec.ts
```diff
@@ -127,4 +135,290 @@ describe('3-runtime-html/repeater.destructered-declaration.spec.ts', function ()
...
+      assert.strictEqual(labelReads, 2, 'another selected property reuses the computed value');
+
+      // A newly inserted row must receive the same declaration binding as the
+      // views created during the repeat's initial activation.
+      component.orders.push(Object.assign(new Order('Katherine', 'Johnson'), { status: 'ready' }));
+      await tasksSettled();
+      assertText('Grace Lovelace — completeKatherine Johnson — ready');
+    });
+
+    it('reuses keyed views while reconnecting their locals to replacement objects', async function () {
```
bigopon, line 286: 👍

packages/__tests__/src/3-runtime-html/repeater.destructered-declaration.spec.ts
```diff
@@ -127,4 +135,290 @@ describe('3-runtime-html/repeater.destructered-declaration.spec.ts', function ()
...
+        'the replacement source is observed once',
+      );
+
+      oldOrders[0].name = 'stale';
+      replacementOrders[1].name = 'Ristretto';
+      await tasksSettled();
+      assertText('2:Green tea1:Ristretto');
+    });
+
+    it('keeps destructured locals one-way when a body binding assigns them', async function () {
```
bigopon, line 334: it feels to me either behavior could be allowed. May I know the thought direction of disabling it?

packages/__tests__/src/3-runtime-html/repeater.destructered-declaration.spec.ts
```diff
@@ -127,4 +135,290 @@ describe('3-runtime-html/repeater.destructered-declaration.spec.ts', function ()
...
+        },
+      );
+
+      assertText('Ada/Lovelace');
+      component.snapshots.length = 0;
+
+      batch(() => {
+        order.first = 'Grace';
+        order.last = 'Hopper';
+      });
```
bigopon, line 377: should the batch call be in the destucturing evaluation automatically?

packages/__tests__/src/3-runtime-html/repeater.destructered-declaration.spec.ts
```diff
@@ -127,4 +135,290 @@ describe('3-runtime-html/repeater.destructered-declaration.spec.ts', function ()
...
+      second.name = 'Green tea';
+      await au.start();
+      assertText('Green tea');
+      assert.strictEqual(secondObserver.subs.count, 1, 'the row after restart connects once');
+
+      await stop(true);
+      assert.strictEqual(secondObserver.subs.count, 0, 'disposal leaves no source subscription');
+    });
+
+    it('throws when a repeated item cannot be object-destructured', function () {
```
bigopon, line 505: 👍

packages/expression-parser/src/expression-parser.ts
```diff
@@ -1364,7 +1364,36 @@ function parseArrayLiteralExpression(expressionType: ExpressionType): ArrayBindi
...
+  ...Object.getOwnPropertyNames(Object.prototype),
+  '$index',
+  '$length',
+  '$odd',
+  '$even',
+  '$first',
+  '$middle',
+  '$last',
+  '$previous',
+  '__items__',
```
bigopon, line 1382: does it make sense to add `$item`? could be reserved so in the future we can support both destructuring & accessing the object.

packages/runtime-html/src/resources/template-controllers/repeat.ts
```diff
@@ -165,8 +173,24 @@ export class Repeat<C extends Collection = unknown[]> implements ICustomAttribut
 
     this._refreshCollectionObserver();
     const dec = forOf.declaration;
-    if(!(this._hasDestructuredLocal = dec.$kind === 'ArrayDestructuring' || dec.$kind === 'ObjectDestructuring')) {
-      this.local = astEvaluate(dec, this.$controller.scope, binding, null) as string;
+    switch (dec.$kind) {
+      case 'ArrayDestructuring':
+      case 'ObjectDestructuring':
+        this._declaration = { kind: 'destructuring', value: dec };
```
bigopon, line 182: doesn't object binding pattern cover this as well? I'm not familiar wiith the syntax so these looks similar to me.

packages/runtime-html/src/templating/controller.ts
```diff
@@ -1470,6 +1470,23 @@ function createWatchers(
...
 
+/**
+ * Adds a binding that must run before bindings already rendered for a controller.
+ *
+ * This is intentionally separate from `addBinding`: declaration-style bindings are
+ * uncommon, and the normal append path should not pay for an ordering option.
+ *
+ * @internal
+ */
+export function prependBinding(controller: IController, binding: IBinding): void {
```
bigopon: 😃 well if it lives long enough it's bound to lack something. I sometimes wanted prepend as well. But that is kind of a leak on the internal of controller. Maybe/not we can just make it a method on controller.

bigopon (commented):
Nice work @fkleuver , only a few comments. 👍

packages/__tests__/src/3-runtime-html/repeater.destructered-declaration.spec.ts
fkleuver: `.one-time` is a bind-order probe. It proves that destructured locals exist before the repeated body binds. A reactive binding could repair itself later and conceal that bug

packages/__tests__/src/3-runtime-html/repeater.destructered-declaration.spec.ts
fkleuver: I treated the destructured values as reactive local variables. This is similar to normal JS destructuring where changes to the source refresh the local, but assigning the local does not implicitly mutate the source object.
Write-through would require treating every destructured local as a two-way proxy, which is both more complex and potentially surprising (especially with aliases, getters and replaced items).

packages/__tests__/src/3-runtime-html/repeater.destructered-declaration.spec.ts
fkleuver: It already happens at the correct boundary. The current implementation batches the publication of all projected locals. The outer `batch()` in this test is a seperate thing to turn two source assignments into one transaction because we can't infer that arbitrary consecutive source assignments belong together

packages/expression-parser/src/expression-parser.ts
fkleuver: Yeah agree, good call

packages/runtime-html/src/templating/controller.ts
fkleuver: Yeah you're right, should just put it on the controller.

packages/runtime-html/src/resources/template-controllers/repeat.ts
fkleuver: No they're different contracts, but you're right that this is messy. This is actually AST tech debt. The full object-destructuring parser was removed in 2021 but the old `ObjectDestructuring` branch stayed reachable for manually constructed or serialized ASTs, but normal `{ ... } of ...` templates have used `ObjectBindingPattern` ever since.

I'll rename the new file to match what it actually uses. I'll revisit the AST problem later in a separate PR.

PR #2450, opened 2026-07-31
fix(plugin-conventions): detect static $au definitions

### LINKED ISSUE #2437: `plugin-conventions`: naive `.includes('$au')` check silently disables template pairing when a class body contains the substring `$au` (e.g. a `$authId` GraphQL variable)
### Summary

In `@aurelia/plugin-conventions`, `preprocessResource` decides how to attach the conventional HTML template to a view-model class. When it reaches the implicit-element branch it does a **naive text substring check** for `$au` on the class source to detect the `static $au = {}` resource convention:

```ts
const elementStatement = unit.contents.slice(implicitElement.pos, implicitElement.end);
if (elementStatement.includes('$au')) {
  const sf = createSourceFile('temp.ts', elementStatement, ScriptTarget.Latest, true);
  const result = transform(sf, [createAuResourceTransformer()]);
  ...
}
```
(`packages-tooling/plugin-conventions/src/preprocess-resource.ts`, still present on `master`.)

If **any** text inside the class body contains the substring `$au` — even inside a string/template literal that has nothing to do with a `static $au` property — the plugin takes the `$au` branch, runs `createAuResourceTransformer()` (which finds no `static $au` member and does nothing useful), and **never injects the `@customElement(...)` definition**. The class is emitted without any custom-element registration, i.e. it silently stops being a custom element. No error or warning is produced; the build succeeds.

This is very easy to hit with GraphQL, where variables are written `$name` — e.g. `$authId`, `$authCode`, `$author` all contain `$au`.

### Minimal reproduction

`my-element.js` (with a paired `my-element.html`):

```js
export class MyElement {
  foo () {
    // any string containing the substring "$au" is enough
    return 'query ($auId: String) { a(id: $auId) { b } }'
  }
}
```

Running `preprocessResource` on this file:

```js
import { preprocessResource } from '@aurelia/plugin-conventions'

const contents = `export class MyElement {\n  foo () {\n    return 'query ($auId: String) { a(id: $auId) { b } }'\n  }\n}\n`
const res = preprocessResource(
  { path: 'my-element.js', contents, filePair: 'my-element.html' },
  { hmr: false, enableConventions: true }
)
console.log(res.code.includes('customElement('))
```

- **Control** (same file with `$id` instead of `$auId`): logs `true` — template is paired, class becomes a custom element.
- **Repro** (`$auId`): logs `false` — no `@customElement`, class is not registered.

### Expected

The template should be paired with the class regardless of arbitrary `$au` substrings appearing inside method bodies / string literals. Detection of the `static $au` convention should be AST-aware (an actual static class member named `$au`), not a text `.includes('$au')` on the whole class statement.

### Actual

The class is silently emitted without a custom-element definition.

### Downstream impact (how it surfaced for us)

The component still imports fine, but at runtime the router's eager path generation (`resolveCustomElementDefinition` → `_resolveLazy`, used when generating `load`/`href` links for sibling routes) resolves the lazy `() => import()` module, finds no custom element in it, and throws:

```
AUR3175: [object Promise] does not appear to be a component or CustomElement recognizable by Aurelia; make sure to use the @customElement decorator for your class if not using conventions.
```

Because eager link generation resolves *sibling* route modules, a single affected component takes down navigation for the whole router section (the dev-server error overlay then blocks the entire page), even on pages that don't render the affected component. That makes the root cause quite hard to trace back to "a GraphQL variable happens to start with `$au`".

### Suggested fix

Replace the `elementStatement.includes('$au')` heuristic with an AST check for an actual `static $au` property declaration on the class (the plugin already has `isStatic$auProperty` / `isStaticPropertyDeclaration(node, '$au')` helpers that could drive this decision).

### Environment

- `@aurelia/plugin-conventions` `2.0.0-rc.1` (latest published; bug also present on `master`)
- Build: webpack + `@aurelia/webpack-loader`

fkleuver:
# 🐛 Bug Fix

## 📖 Description

`plugin-conventions` mistook any `$au` text inside a class (including GraphQL variables such as `$authId`) for static Aurelia resource metadata. This prevented paired HTML templates from being registered.

Fixes #2437.

## 💁 Your Solution

Detect actual static `$au` properties using the existing TypeScript AST instead of searching the class source text.

## 📑 Test Plan

Added regressions for `$au` text in strings and instance `$au` properties, while retaining coverage for genuine static `$au` definitions.

## 📦 Changeset

- [x] Added a changeset

### DIFF
--- a/.changeset/fix-static-au-detection.md
+++ b/.changeset/fix-static-au-detection.md
@@ -1,0 +1,5 @@
+---
+"@aurelia/plugin-conventions": patch
+---
+
+Fix conventional template pairing for classes containing `$au` in strings or other non-resource code. Only an actual static `$au` property now selects the static resource-definition path.
--- a/packages-tooling/__tests__/src/plugin-conventions/preprocess-resource.spec.ts
+++ b/packages-tooling/__tests__/src/plugin-conventions/preprocess-resource.spec.ts
@@ -446,6 +446,60 @@ export class FooBarBindingCommand {}
     assert.equal(result.code, expected);
   });
 
+  it('pairs the template when a class string contains $au text', function () {
+    const code = `
+export class FooBar {
+  query() {
+    return 'query ($authId: String) { user(id: $authId) { id } }';
+  }
+}
+`;
+    const expected = `import { customElement } from '@aurelia/runtime-html';
+import * as __au2ViewDef from './foo-bar.html';
+
+@customElement(__au2ViewDef)
+export class FooBar {
+  query() {
+    return 'query ($authId: String) { user(id: $authId) { id } }';
+  }
+}
+`;
+    const result = preprocessResource(
+      {
+        path: path.join('bar', 'foo-bar.js'),
+        contents: code,
+        filePair: 'foo-bar.html'
+      },
+      preprocessOptions({ hmr: false })
+    );
+    assert.equal(result.code, expected);
+  });
+
+  it('pairs the template when $au is an instance property', function () {
+    const code = `
+export class FooBar {
+  $au = 'instance state';
+}
+`;
+    const expected = `import { customElement } from '@aurelia/runtime-html';
+import * as __au2ViewDef from './foo-bar.html';
+
+@customElement(__au2ViewDef)
+export class FooBar {
+  $au = 'instance state';
+}
+`;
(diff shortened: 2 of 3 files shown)


bigopon (approved):
✅

PR #2452, opened 2026-08-21
refactor(convention): make ability to stub file apis provider to official

bigopon:
## 📖 Description

Even though we plan to let the language server take over all the convention work, it's still better to have the core utility of plugin convention to be flexible. This PR separates the plugin convention into 2 parts:

- core convention work which is filesystem host agnostic
- platform/host related export for as host providers implementation

So vite/parcel/webpack can plug in the core convention easier, also for usage in the browser.

cc @3cp @fkleuver

### DIFF
--- a/packages-tooling/plugin-conventions/src/options.ts
+++ b/packages-tooling/plugin-conventions/src/options.ts
@@ -18,6 +18,11 @@ export interface IFileUnit {
   readFile?(path: string): string;
 }
 
+export interface IFileUnitHost {
+  fileExists(unit: IFileUnit, path: string): boolean;
+  readFile(unit: IFileUnit, path: string): string;
+}
+
 export interface IOptionalPreprocessOptions {
   defaultShadowOptions?: { mode: 'open' | 'closed' };
   // More details in ./preprocess-html-template.ts
@@ -35,6 +40,7 @@ export interface IOptionalPreprocessOptions {
   // Whenn CSSModule is in use, stringModuleWrap is ignored.
   useCSSModule?: boolean;
   hmr?: boolean;
+  isDev?: boolean;
   enableConventions?: boolean;
   hmrModule?: string;
   /**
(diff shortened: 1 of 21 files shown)


packages-tooling/plugin-conventions/src/options.ts
```diff
@@ -18,6 +18,11 @@ export interface IFileUnit {
   readFile?(path: string): string;
 }
 
+export interface IFileUnitHost {
+  fileExists(unit: IFileUnit, path: string): boolean;
+  readFile(unit: IFileUnit, path: string): string;
+}
```
3cp, line 24: Since we try to abstract file apis, should these two be async functions? 

For example in browser env, FileReader api is async. That's the only api for local file reading in browser mode. I know those sync apis can be used for browser mode with simulated files, but they cannot work with real user local files with FileReader.

3cp:
Over all, LGTM. just added some thought on the abstraction.

packages-tooling/plugin-conventions/src/options.ts
3cp: Async api also enables reading remote files.

bigopon: Thanks for this, we actually cannot just turn it into async. jess transformer requires processor to be in sync only. At least from this https://github.com/jestjs/jest/blob/52db46e50404c993d3f10dd1481e861cc04ae22e/packages/jest-transform/src/types.ts#L123-L127

That means if we want to provide async api, it needs to be 2 different processor exports: 1 async and 1 sync, the sync processor will require sync file host and the async processor can work with both.

Maybe we can do this. Only jest needs the sync while the rest can be just fine with whatever, and browser based convention parsing can use the async one instead.

3cp: Let's stay on sync for now. Jest is almost out of picture. Vite and vitest work much easier.

Parcel and webpack are going out too. We might can drop the support of the three in few years.

3cp (approved):

packages-tooling/plugin-conventions/src/options.ts
bigopon: yeah, at least async can always call sync, and can adapt easily to a sync API. We can avoid a lot of unnecessary code too.

bigopon:
a reference in case we want to provide async process apis in the future https://github.com/aurelia/aurelia/tree/feat/async-process

PR #2453, opened 2026-08-22
feat(if): support else if

### LINKED ISSUE #1217: request(template): add else-if attribute
# 🙋 Feature Request

At the moment, many important pieces of the JS syntaxes are supported in our template, except `else if` in `if/else if/else` combo. With `link`, it seems easier to support this feature now. Consider adding this.

## 🤔 Expected Behavior

```html
<div if.bind="isRock">Rock</div>
<div else-if.bind="isPaper">Paper</div>
<div else>Scissor</div>
```
The above template should work similarly with the following JS block:
```tsx
if (isRock) {
  <div>Rock</div>
} else if (isPaper) {
  <div>Paper</div>
} else {
  <div>Scissor</div>
}
```

bigopon:
# ✨ New Feature

## 📖 Description

Adds `else if` chains by placing `else` and `if.bind` on the same element:

```html
<div if.bind="isA">A</div>
<div else if.bind="isB">B</div>
<div else if.bind="isC">C</div>
<div else>D</div>
```

Conditions are evaluated lazily, and the first matching branch owns the active view. Chains work with native elements, direct custom elements, and explicit `<template>` branches.

`else` and `if.bind` must be adjacent template controllers on the same element to form an `else if` branch. Other template-controller combinations keep their existing behavior.

### Implementation

- Each `else if` is represented as an `else` branch containing a lazy nested `if`, preserving the existing view lifecycle and cache behavior.
- The compiler records proven same-element adjacency on the template-controller instruction instead of reconstructing authoring syntax from nested view definitions.
- Serialized SSR definitions preserve this optional instruction metadata during hydration.
- Failed value-driven async branches finish teardown before a later branch can attach.

### 🎫 Issues

Closes #1217

## 📑 Test Plan

- Added compiler coverage for aliases, resolved and unresolved resources, and absent, separated, or reversed link targets.
- Added integration coverage for native, custom-element, containerless, and explicit-template branches; lazy and nested chains; caching; long jumps; async failure, supersession, and teardown; and SSR adoption.
- Added serialized-definition coverage for present, nested, and absent template-controller metadata.

## 📦 Changeset

The `plugin-conventions` release note for #2452 is kept in a separate changeset so each entry retains its own PR attribution.


ekzobrain:
Oh oh. I would prefer not to do it like this. I imagine this syntax may be very comfusing with nested conditional templates, even this flat one looks ambiguous. While it mimics js syntax, for the templates else-if looks much clearer imho.

bigopon:
> I imagine this syntax may be very comfusing with nested conditional templates, even this flat one looks ambiguous. While it mimics js syntax, for the templates else-if looks much clearer imho.

Can you pls help give some example showing where the confusion will be, probably in a more standard template, rather than my test template above. I think we can discuss between `else-if` vs `else if` for a bit more before making any final conclusion.

bigopon:
If we are to implement `else-if`, we will likely need to document the difference between the following:
```html
<div if>
<div else if>
```
vs
```html
<div if>
<div else-if>
```
I find it a bit weird that we have 2 ways to expresss the same intention, any suggestion @ekzobrain?

.changeset/gold-aliens-own.md
```diff
@@ -0,0 +1,7 @@
+---
+"@aurelia/runtime-html": minor
+"@aurelia/plugin-conventions": minor
```
fkleuver: Is this intentional or did plugin-conventions accidentally slip in?

docs/user-docs/getting-to-know-aurelia/template-controllers.md
```diff
@@ -166,6 +188,8 @@ Use this hook whenever two controllers must share state (think `switch` / `case`
 - **Container strategy matters** – Setting `containerStrategy: 'new'` ensures each rendered view gets a fresh child container (see `PromiseTemplateController`); the default `'reuse'` is faster but shares services.
 - **Lifecycle must be balanced** – Always deactivate and dispose views you created. Built-ins call `.dispose()` in `detaching` to release DOM and avoid leaks.
 
+**Async swaps are phased** â€“ When one branch returns a promise from `detaching()` and the next returns a promise from `attaching()`, Aurelia processes the swap in stages. The old branch leaves first, then the new branch starts attaching, and the new DOM can be present before the async attach promise resolves.
```
fkleuver: There's `â€“` mojibake here

packages/__tests__/src/3-runtime-html/if.integration.spec.ts
```diff
@@ -241,6 +256,1016 @@ describe(`3-runtime-html/if.integration.spec.ts`, function () {
...
+
+      component.step = 2;
+      await tasksSettled();
+      await tasksSettled();
+      assertText('c');
+      assert.deepStrictEqual(evalCounts, { a: 2, b: 1, c: 1 });
+
+      component.step = 3;
+      await tasksSettled();
+      await tasksSettled();
```
fkleuver: Why do we need two `tasksSettled()` calls here?

packages/runtime-html/src/resources/template-controllers/if.ts
```diff
@@ -184,6 +188,86 @@ export class If implements ICustomAttributeViewModel {
...
+    : tcInstruction.res.name === 'if';
+};
+
+const hasChainedIf = (factory: IViewFactory): boolean => {
+  const rows = factory.def.instructions;
+  if (rows.length !== 1) {
+    return false;
+  }
+  const row = rows[0];
+  return row.length === 1 && isIfInstruction(row[0]);
```
fkleuver: This will cause a nested `if` to be mistaken for a syntactic `else if` because it only checks whether the `else` factory contains one instruction row with one `if`. This would produce the same shape:
```html
<div if.bind="outer">A</div>
<div else><span if.bind="inner">B</span></div>
<div else>C</div>
```

The third `else` should throw AUR0810 because a plain `else` terminates the chain. Instead, it is now accepted and `C` is wired into the nested `if`.

We may need compiler-owned provenance saying "these `else` and `if` attributes came from the same element". I don't think instruction-count inference alone can establish that

packages/__tests__/src/3-runtime-html/if.integration.spec.ts
bigopon: there' 2 activation cycles happening. 1 is for the first view deactivation, and other is the next view activation. each of them has their own async lifecycle, and is not connected to the other somehow.

packages/runtime-html/src/resources/template-controllers/if.ts
bigopon: resolved, thanks for the test case 👍

docs/user-docs/getting-to-know-aurelia/template-controllers.md
bigopon: resolved, thanks

.changeset/gold-aliens-own.md
bigopon: intentional, i forgot about this in another PR

bigopon:
One thing probably should be worth noting in our doc is that
```html
<div if>
<div else if>
<div else>
```
is effectively similar to JS
```ts
if {}
else {
  if {}
  else {}
}
```
So the template inside the last else is not a sibling of the first ib, but a child of it.

fkleuver:
> Oh oh. I would prefer not to do it like this. I imagine this syntax may be very comfusing with nested conditional templates, even this flat one looks ambiguous. While it mimics js syntax, for the templates else-if looks much clearer imho.

I understand the readability concern. `else-if` is visually clearer when viewed as one standalone directive, and Vue provides a reasonable precedent for that spelling.

The original #1217 request also used `else-if`, so history does not settle the spelling by itself. Looking through Aurelia’s own history, though, I still lean toward `else if`. `else` and `if.bind` on the same element have been treated as composed template controllers for a long time; generated integration tests have covered that form since 2018. Before this PR, a single composed fallback could already work:

```html
<div if.bind="A"></div>
<div else if.bind="B"></div>
```

What was incomplete was continuation of the chain. A following `else` (either another `else if` or the final fallback) could not link to the inner `If` and threw `AUR0810`. The `AUR0810` documentation even described the complete form as valid for a time. We removed that example after the runtime mismatch was reported.

This PR completes that existing composition:

```html
<div if.bind="A"></div>
<div else if.bind="B"></div>
<div else if.bind="C"></div>
<div else></div>
```

The nested-template concern is handled structurally. The compiler marks the relationship only when `else` and `if` are adjacent template controllers on the same element. An `if` nested inside an ordinary `else` branch is not mistaken for an `else if`. The integration tests cover that distinction.

We could still introduce `else-if` as compiler sugar, but we would then need to support two spellings or deliberately reject a form Aurelia has already compiled and partially supported.

For that reason, my preference is to make `else if` the canonical form.

PR #2456, opened 2026-08-22
fix(runtime): normalize computed getter dependencies

fkleuver:
# 🐛 Bug Fix

## 📖 Description

Getter callback dependencies supplied through `@computed({ deps })` were accepted by the public API but passed to an array-only observer, causing observation to fail. Calling `@computed()` on a getter also lost automatic dependency tracking.

## 💁 Your Solution

Normalize getter dependencies when the decorator is applied. Callback dependencies now use the existing function-observer path. Omitted dependencies retain proxy-based automatic tracking.

## 📑 Test Plan

Added regression coverage for callback dependency forms, automatic tracking, dynamic recollection and teardown, synchronous flushing, and deep returned graphs.

## 📦 Changeset

- [x] Added a changeset

### DIFF
--- a/packages/runtime/src/observer-locator.ts
+++ b/packages/runtime/src/observer-locator.ts
@@ -80,6 +80,7 @@ export class ObserverLocator {
 
   public getObserver(obj: unknown, key: PropertyKey): IObserver;
   public getObserver<T, R>(obj: T, key: ComputedGetterFn<T, R>): IObserver<R>;
+  public getObserver(obj: unknown, key: PropertyKey | ComputedGetterFn): IObserver;
   public getObserver(obj: unknown, key: PropertyKey | ComputedGetterFn): IObserver {
     if (obj == null) {
       throw createMappedError(ErrorNames.observing_null_undefined, key);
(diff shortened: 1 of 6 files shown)


packages/runtime/src/observer-locator.ts
```diff
@@ -80,6 +80,7 @@ export class ObserverLocator {
 
   public getObserver(obj: unknown, key: PropertyKey): IObserver;
   public getObserver<T, R>(obj: T, key: ComputedGetterFn<T, R>): IObserver<R>;
+  public getObserver(obj: unknown, key: PropertyKey | ComputedGetterFn): IObserver;
```
bigopon, line 83: it's intentional that i omitted this, so we should have only 1 or another, or is this always required?

bigopon (approved):
👍

packages/runtime/src/observer-locator.ts
fkleuver: This allows users (and our tests) to also pass things through dynamically. As long as it's the last overload, the specific ones are still automatically preferred.

PR #2457, opened 2026-08-23
fix(runtime): correct observable class decorator typing

fkleuver:
# 🐛 Bug Fix

## 📖 Description

Object-form `@observable` configuration works as a class decorator at runtime, but its typing caused TypeScript to reject valid class usage with TS1270.

## 💁 Your Solution

Give object-form configuration distinct class and field decorator signatures. Runtime behavior remains unchanged.

## 📑 Test Plan

Added regression coverage for object-configured class and field decorators, including `name`, `callback`, and `set`.

## 📦 Changeset

- [x] Added a changeset

### DIFF
--- a/packages/__tests__/src/3-runtime-html/decorator-observable.spec.ts
+++ b/packages/__tests__/src/3-runtime-html/decorator-observable.spec.ts
@@ -132,6 +132,50 @@ describe('3-runtime-html/decorator-observable.spec.ts', function () {
     assert.strictEqual(callCount, 2);
   });
 
+  it('supports object configuration as class and field decorators', function () {
+    const changes: [string, number, unknown][] = [];
+
+    @observable({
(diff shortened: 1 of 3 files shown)


packages/__tests__/src/3-runtime-html/decorator-observable.spec.ts
```diff
@@ -132,6 +132,50 @@ describe('3-runtime-html/decorator-observable.spec.ts', function () {
     assert.strictEqual(callCount, 2);
   });
 
+  it('supports object configuration as class and field decorators', function () {
```
bigopon, line 135: we should use the api test instead of a real test, or they can be supplement of each other. it's probably at the end of the observable file, or in test-api.ts

bigopon (approved):

PR #2458, opened 2026-08-23
fix(runtime-html): roll back partially bound interpolations

fkleuver:
# 🐛 Bug Fix

## 📖 Description

When a multi-part interpolation failed during binding, completed parts remained subscribed because the aggregate never reached its bound state.

## 💁 Your Solution

Roll back completed interpolation parts when a later part or the initial target update fails, then rethrow the original error.

## 📑 Test Plan

Added unit coverage for both failure paths and an integration regression proving that a failed conditional view releases its subscriptions and can reactivate normally.

## 📦 Changeset

- [x] Added a changeset

### DIFF
--- a/packages/runtime-html/src/binding/interpolation-binding.ts
+++ b/packages/runtime-html/src/binding/interpolation-binding.ts
@@ -131,10 +131,19 @@ export class InterpolationBinding implements IBinding, ISubscriber, ICollectionS
     const partBindings = this.partBindings;
     const ii = partBindings.length;
     let i = 0;
-    for (; ii > i; ++i) {
-      partBindings[i].bind(scope);
+    try {
+      for (; ii > i; ++i) {
+        partBindings[i].bind(scope);
+      }
+      this.updateTarget();
+    } catch (error) {
+      this._scope = void 0;
+      // `i` tracks completed binds, including every part after a target failure; unwind them in reverse activation order.
+      while (i > 0) {
+        partBindings[--i].unbind();
+      }
+      throw error;
     }
-    this.updateTarget();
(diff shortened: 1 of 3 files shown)


packages/runtime-html/src/binding/interpolation-binding.ts
```diff
@@ -131,10 +131,19 @@ export class InterpolationBinding implements IBinding, ISubscriber, ICollectionS
...
+      for (; ii > i; ++i) {
+        partBindings[i].bind(scope);
+      }
+      this.updateTarget();
+    } catch (error) {
+      this._scope = void 0;
+      // `i` tracks completed binds, including every part after a target failure; unwind them in reverse activation order.
+      while (i > 0) {
+        partBindings[--i].unbind();
+      }
```
bigopon, line 144: we don't handle error/rollback subscription anywhere, is it right that we do it here? Normally if the binding can keep rolling forward, its subscriptions are automatically clean. If the binding cannot, then the whole app breaks anyway.

fkleuver: I think you make a fair point. The partial interpolation state is observable in isolation, and older implementations left it reachable by normal cleanup, but we do not currently seem to have a realistic user scenario where the retained subscriptions cause a concrete problem.

fkleuver:
After further review, I'm closing this. We do not generally provide transactional recovery from binding errors. Adding interpolation-specific rollback would introduce a new cleanup policy without addressing partial work in the throwing part or equivalent failure states in other binding types.
The implementation and regression coverage therefore add more complexity than the impact justifies.

PR #2460, opened 2026-08-23
fix(runtime-html): enforce structural if/else adjacency

fkleuver:
# 🐛 Bug Fix

## 📖 Description

`else` is documented and generally understood as the alternative branch of the immediately preceding sibling `if`: two branches occupying one structural position.

Currently, Aurelia does not establish that relationship from template structure. `Else.link` selects the last controller registered under the parent. Because ordinary DOM nodes are absent from that controller list, this template is accepted:

```html
<div if.bind="condition">if</div>
<p>unrelated content</p>
<div else>else</div>
```

The `else` branch is then rendered at the `if` location, before the `<p>`. Registering a custom element or attribute on the intervening node can make otherwise identical markup throw AUR0810 instead. A nested `if` inside an intervening wrapper can even capture the outer `else`.

Consequently, the meaning of `else` currently depends on incidental implementation details:

- Whether an intervening node creates an Aurelia controller
- Which resources happen to be registered
- How the compiler lowers or removes intervening content
- The order in which controllers are linked

This is neither a stable feature nor a useful compatibility guarantee. Supporting non-adjacent branches would require deliberately specifying where the alternative branch renders and how it interacts with intervening bindings, controllers, projections, lifecycles, and animations. It should not emerge accidentally from a controller-array lookup.

## 💁 Your Solution

Associate `else` with the immediately preceding sibling `if` in both the authored template and the live render structure. Formatting whitespace and HTML comments remain allowed; any other sibling content ends the pair and produces AUR0810.

Checking the runtime DOM alone is insufficient because compilation can erase significant source structure:

```html
<div if.bind="condition">if</div>
${message}
<div else>else</div>
```

```html
<div if.bind="condition">if</div>
<let message.bind="value"></let>
<div else>else</div>
```

Interpolations become binding targets, `<let>` is removed, local-template declarations are extracted, and projection templates may be unwrapped. By the time `Else.link` runs, a marker-only implementation can no longer distinguish these cases from genuine adjacency.

The compiler therefore records generic source-node provenance on template-controller instructions:

- The source node that produced the instruction
- That node’s previous significant sibling
- One root-wide identity space covering nested templates and projections

The renderer carries the source identity to the render location. `Else.link` then requires both:

1. The preceding live render marker belongs to an `if` at the same DOM boundary.
2. That `if` is the exact previous significant sibling recorded by the compiler.

In language-design terms, the relationship is defined by the source template, not by whichever nodes happen to survive lowering. Lowering must preserve the semantics rather than become the semantics.

The provenance is resource-neutral and additive to the instruction contract; the template compiler does not contain an `if`/`else` special case. Framework-generated and serialized instructions preserve it. Legacy or hand-authored instruction sets with no provenance on either side retain marker-based compatibility, while partially populated provenance fails closed.

## ⚠️ Compatibility

This intentionally rejects templates that previously relied on intervening content between `if` and `else`. Although such templates happened to work in some cases, the behavior was registration-sensitive, could associate with the wrong conditional, and did not have well-defined branch-placement semantics.

Applications relying on it should move the intervening content outside the pair or express the alternative with an explicit `if.bind` condition.

Making this correction before GA gives Aurelia a compositional rule we can document and preserve. If non-adjacent conditional branches are ever supported, they should be introduced as an explicit feature with intentionally designed semantics.

## 📑 Test Plan

Added regression coverage for:

- Whitespace and comments
- Text, interpolations, `<let>`, and local-template declarations
- Registered and unregistered elements and attributes
- Nested conditional capture
- Repeat views and collection mutation
- Same-slot, cross-slot, and unwrapped projection content
- JIT SSR/hydration and serialized `needsCompile: false` definitions
- Complete, legacy, and partially populated provenance

## 📦 Changeset

- [x] Added a changeset

cc @bigopon - this should be either merged (or intentionally rejected) before #2453

### DIFF
--- a/.changeset/tidy-branches-link.md
+++ b/.changeset/tidy-branches-link.md
@@ -1,0 +1,8 @@
+---
+"@aurelia/runtime-html": patch
+"@aurelia/template-compiler": patch
+---
+
+`else` now enforces its documented relationship with an `if` on the immediately preceding sibling element. Formatting whitespace and HTML comments may remain between them. Text, interpolations, `<let>` bindings, local-template declarations, or elements now end the pair, preventing an unrelated or nested `if` from capturing the branch.
+
+Templates that previously relied on an intervening plain element should move it outside the pair or use an explicit `if.bind` condition.
--- a/docs/user-docs/developer-guides/error-messages/runtime-html/aur0810.md
+++ b/docs/user-docs/developer-guides/error-messages/runtime-html/aur0810.md
@@ -6,18 +6,18 @@
 
 ## Description
 
-This error occurs during template linking when Aurelia encounters an element with an `else` custom attribute, but the immediately preceding element sibling in the DOM (after template compilation) does not have an `if.bind` attribute or another `else` attribute. The `else` attribute must directly follow an `if` or another `else` to form a conditional chain.
+This error occurs during template linking when Aurelia encounters an element with an `else` custom attribute that does not immediately follow an element with `if.bind`. Formatting whitespace and ordinary HTML comments are ignored, but non-whitespace text, interpolations, and intervening elements—including `<let>` or `<template>`—end the conditional pair.
 
 ## Cause
 
-1.  **Incorrect Placement:** The element with the `else` attribute is not placed immediately after the element with the corresponding `if.bind` or `else` attribute in the HTML template. There might be other elements, comments (that aren't removed during compilation), or significant whitespace between them.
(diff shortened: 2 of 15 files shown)


fkleuver:
After consideration and discussion with @bigopon , I'm withdrawing this concern. Non-adjacent `if` / `else` branches were already outside the documented contract, so this PR does not need to define or correct the behavior. We should just avoid adding coverage that makes the incidental behavior contractual.
Will turn this into an aurelia-ls lint error instead.

Issue #2315, opened 2025-12-18
Add an async hook for template loading (official replacement for AU1 loader.loadText override)


fscck:
First of all, I would like to thank you for this great framework. We have been using Aurelia 1 for quite some time and we are thrilled with its clarity on the one hand and its great adaptability on the other.

We are currently transforming our application to Aurelia 2. We have to make quite a few adjustments, but that's okay with a version jump.

# 🙋 Feature Request
In AU1 we overrode loader.loadText to lazily intercept the original HTML of a view at first access, send it to our server for transformation, then let Aurelia compile and cache the result.

## 🤔 Expected Behavior
In AU2, there should also be a simple way to adapt the existing template asynchronously.

## 😯 Current Behavior
The documented template compiler hooks are synchronous (compiling) and run before any async lifecycle like loading. There is no public async loader hook to react at the moment the template source is loaded.

AU2 migration docs emphasize bundler ?raw imports, dynamic composition, and runtime definitions, but none expose an async template‑source hook.

## 💁 Possible Solution

1. Introduce an interface ITemplateSourceResolver with resolveTemplateSource(...): Promise<string>, invoked before compilation for a component’s template.

Framework calls it once on first access; returned HTML becomes the input for the (synchronous) compilation; compiled fragment caching remains effective.

2. Alternatively, add an async compiler hook compilingAsync(...): Promise<string | HTMLElement> that runs prior to compilation and can await network calls.

Benefits: Enables lazy, targeted loading and transformation of templates (server‑side or otherwise) without preloading all pages and without complex router/host wrappers. Keeps Aurelia’s compiled‑view caching intact.

## 🔦 Context
AU1 Code:

```
aurelia.loader.loadText = async (url) => {
	if (url.indexOf("-dynamic.html") > -1) {
		const content: any = await oldTextLoader(url).catch(() => {
			/* error handling*/
		});
		let name: string = url;
		name = name.replace(".html", "");

		const request: any = {
			PageName: name,
			Content: content,
			InteractiveScriptParameter: {
				RunTimeMode: "Interactive"
			}
		};
		return WebSocketClient.serviceCall("Layout.GetLayout", request, "layout").then..
```
`

bigopon:
[...]

Issue #2142, opened 2025-04-06
au-compose deactivate hook not firing on component or model change


davidsk:
# 🐛 Bug Report

When binding a component and / or model to the au-compose element, the deactivate hook on the currently bound component is not being called.

Repro link: https://stackblitz.com/~/github.com/davidsk/aurelia-compose-deactivate

## 🤔 Expected Behavior

When either the bound component or model is changed, the deactivate hook should be called.

## 😯 Current Behavior

When either the bound component or model is changed, the deactivate hook is not being called.

## 💁 Possible Solution

🤷‍♂️

## 🔦 Context

I am evaluating the options Aurelia provides for building list/detail views in a LoB application.  Often times the lists in these scenarios will be hierarchical and the node representing many different types of entity and therefore the component backing the detail will need to be swapped out. 

## 💻 Code Sample

https://github.com/davidsk/aurelia-compose-deactivate

## 🌍 Your Environment

| Software         | Version(s) |
| ---------------- | ---------- |
| Aurelia          | beta23
| Language         | Typescript 5.7.2
| Browser          | Edge (Latest)
| Bundler          | Vite 6.0.3
| Operating System | Windows 10
| NPM/Node/Yarn    | NPM 9.6.5

bigopon:
thanks @davidsk , au-compose only has `activate`. Shouldn't you be able to use `detached`/`unbinding` lifecycle for this?

davidsk:
I suppose so, I am just going off of the [documentation](https://docs.aurelia.io/components/component-lifecycles#others). 

```
For <au-compose>, there are extra lifecycle hooks that are activate/deactivate. Refers to [dynamic composition doc](https://docs.aurelia.io/getting-to-know-aurelia/dynamic-composition) for more details
```

bigopon:
Ahh I see, that's why I wasn't able to find it in the au-compose doc. It probably should be removed. We will leave it as is for now, unless someone asks for this feature.

davidsk:
> Shouldn't you be able to use detached/unbinding lifecycle for this?

I added these hooks and unless I'm doing something wrong, detached isn't called when swapping out the component and neither hook is called when swapping out the model.

> It probably should be removed. We will leave it as is for now

I don't follow, are you suggesting leaving the documentation as it is?

davidsk:
> we'll need a reproduction of that, preferably on Stackblitz
[...]

PR #2461, opened 2026-08-24
fix(runtime-html): preserve async lifecycle correctness during overlap

fkleuver:
## 📖 Description

This PR aligns asynchronous lifecycle coordination with the framework feature that owns each operation. Synchronous operations continue inline and retain their `void` completion paths.

### Resulting behavior

- Repeat runs one reconciliation at a time, composes queued collection changes against the current row generation, and waits for accepted row lifecycle work before continuing or tearing down.
- `<au-compose>` owns structural generations through loading, activation, retirement, deactivation, and disposal. `composing` represents structural work. Components own Promises returned by model-only `activate(model)` calls.
- A stop requested during application startup receives its own completion Promise and runs after activation. Joined stop callers preserve disposal requests across root replacement and reentrant `au-started` handlers.
- Initial If and Switch branch failures reach `start()`. Value-driven updates keep their established cleanup and recovery behavior.
- Lifecycle hooks and AppTasks preserve the original failure value. Aurelia observes started task Promises and reports the first known failure promptly.

### Documentation

The lifecycle guides describe hook order, DOM timing, application boundaries, and practical async patterns. The Repeat and dynamic-composition guides document their queue and teardown behavior. The runtime-html error reference includes the active AUR0655 and AUR0809 codes.

### 🎫 Relevant history

<details>
<summary>Issues and PRs</summary>

- **Partial activation and deactivation:** #1760, #1762
- **Repeated composition during asynchronous attachment:** #1299
- **Repeat reconciliation precedents:** #1119 / #1511, #1425 / #1429, #1720 / #1737
- **`<au-compose>` behavior:** #2142, #2376, #2428
- **Value-driven branch recovery:** #2322
- **Async template loading and hydration:** #1126, #1813, #2315, #2459

The Repeat links document earlier forms of the same reconciliation invariant.

</details>

PR #2459 extends asynchronous work into template resolution and hydration. When rebased, it can be evaluated against these owner-level guarantees while retaining its own feature and integration review.

## 📑 Test Plan

- Added deterministic integration coverage for Repeat reconciliation, structural composition, branch activation, and application start/stop overlap.
- Repeat coverage includes keyed and unkeyed arrays, duplicate keys, Map, Set, numeric ranges, destructured locals, SSR-adopted rows, teardown, restart, and failure paths.
- Package builds, lint, focused lifecycle tests, runtime-html, Router, Router Direct, and Chrome suites pass locally.

## 📦 Changeset

- [x] Added a changeset


packages/runtime-html/src/resources/template-controllers/repeat.ts
```diff
@@ -529,90 +760,108 @@ export class Repeat<C extends Collection = unknown[]> implements ICustomAttribut
...
-        : Promise.all(promises)) as unknown as Promise<void>;
+    // A synchronous lifecycle hook may mutate the observed collection. Teardown
+    // owns the row set accepted at entry, not whatever length is visible later.
+    const length = views.length;
+    for (let i = 0; i < length; ++i) {
+      const view = views[i];
+      // Repeat retains these rows for a possible owner restart. Releasing them
+      // would let Controller return an ordinary row to the factory cache while
+      // it is still referenced by this owner.
+      void view.deactivate(initiator, $controller);
```
bigopon, line 788: why did we remove the tracking from deactivation? the comment doesn't clearly explain how/why so I'm quite confused.

packages/runtime-html/src/resources/template-controllers/repeat.ts
```diff
@@ -703,6 +955,35 @@ let maxLen = 16;
...
 let tailIndices = new Int32Array(maxLen);
 
+/**
+ * Compose two observer maps that describe consecutive collection mutations.
+ * The result maps the latest collection directly back to the collection owned
+ * by the in-flight reconciliation.
+ *
+ * @internal
+ */
+function composeIndexMaps(previous: IndexMap, current: IndexMap): IndexMap {
```
bigopon, line 969: 👍 , we can also have a bindable for automatic composing index maps , maybe can help with avoiding batch for array mutation.

packages/runtime-html/src/resources/template-controllers/switch.ts
```diff
@@ -89,11 +90,20 @@ export class Switch implements ICustomAttributeViewModel {
...
+    (this as Writable<Switch>).promise = void 0;
+    const deactivate = (): void => {
+      // The ancestor initiator tracks descendant async teardown. Switch only
+      // needs to keep its own queued case work ahead of this final deactivation.
+      void this.view.deactivate(initiator, this.$controller);
+    };
+    if (!isPromise(pending)) {
+      return deactivate();
+    }
+    return pending.then(deactivate);
```
bigopon, line 107: is there some easy to understand catch phrase when and when not to bother with waiting?

packages/runtime-html/src/app-root.ts
```diff
@@ -174,12 +174,25 @@ export class AppRoot<
...
+    try {
+      for (let i = 0; i < appTasks.length; ++i) {
+        const task = appTasks[i];
+        if (task.slot === slot) {
+          results.push(task.run());
+        }
       }
-      return results;
-    }, [] as (void | Promise<void>)[]));
+    } catch (error) {
```
bigopon, line 185: 👍

bigopon (commented):
nice work, only a few minor comments.

packages/runtime-html/src/resources/template-controllers/repeat.ts
fkleuver: Because the old repeat Promise[] aggregation apparently didn't meaningfully track ordinary row hooks anyway. The repeater only needs local tracking when it actually owns the operation. If an ancestor owns the teardown, it has no continuation after deactivating the rows. I'll add a clearer comment

packages/runtime-html/src/resources/template-controllers/switch.ts
fkleuver: Wait at the boundary that owns the operation (who has to do the work after the promise settles). Usually this is the change handler

packages/runtime-html/src/resources/template-controllers/repeat.ts
fkleuver: Yeah a microtask could wrap the change handler but I'm not sure what would be a good place to put that api

PR #2464, opened 2026-08-25
fix(runtime-html): close view lifecycle ownership gaps

fkleuver:
# 🐛 Bug Fix

## 📖 Description

Three lifecycle cleanup paths could leave view ownership incomplete:

- An ancestor-driven child lifecycle failure rejected both the initiator Promise and an unobserved descendant Controller Promise.
- `if` with `cache: false` created fresh branches but did not dispose views it stopped retaining.
- `release()` lost its state when a synthetic view entered deactivation, preventing the existing `ViewFactory` cache-or-dispose logic from running.

## 💁 Your Solution

- Observe descendant-local lifecycle Promises when an ancestor owns the public operation. The initiator still receives the original failure.
- Dispose uncached `if` and `else` views after successful deactivation, including dormant views retained before caching was disabled. Failed branch cleanup also finishes before a replacement can activate.
- Preserve the released state through deactivation until `unbind()` returns the view to its configured factory cache or disposes it.

The released-state change adds no allocations to synchronous lifecycle paths. Its measured bundle impact is 28 minified bytes.

### 🔗 Related Work

- Follows #2461, which established owner-local async lifecycle coordination and exposed these remaining cleanup gaps.
- Precedes #2453, which will build its `else if` lifecycle behavior on the corrected base `If`, Controller, and synthetic-view ownership semantics.

## 📑 Test Plan

- Added Node and Chrome coverage for lifecycle error identity and duplicate unhandled rejections.
- Added `if` lifecycle coverage for synchronous and asynchronous cleanup, rapid supersession, cache changes, restart, nesting, and activation failure.
- Added real Repeat/ViewFactory coverage for synchronous and asynchronous recycling, pending activation, disabled caching, and cache overflow.

## 📦 Changeset

- [x] Added a changeset

### DIFF
--- a/packages/runtime-html/src/resources/template-controllers/if.ts
+++ b/packages/runtime-html/src/resources/template-controllers/if.ts
@@ -149,22 +149,14 @@ export class If implements ICustomAttributeViewModel {
     // ends If's ownership instead, so every owned slot is disposed directly.
     const ifView = this.ifView;
     const elseView = this.elseView;
-    const currentView = this.view;
-    const ownsView = view !== void 0
-      && (currentView === view || ifView === view || elseView === view);
-    this.ifView = this.elseView = void 0;
-    if (currentView === view || currentView === ifView || currentView === elseView) {
-      this.view = void 0;
+    view?.dispose();
+    if (ifView !== view) {
+      ifView?.dispose();
     }
-    if (ownsView) {
-      view.dispose();
+    if (elseView !== view) {
+      elseView?.dispose();
     }
-    if (ifView !== void 0 && ifView !== view) {
-      ifView.dispose();
-    }
-    if (elseView !== void 0 && elseView !== view && elseView !== ifView) {
-      elseView.dispose();
-    }
+    this.ifView = this.elseView = this.view = void 0;
   }
 
   /**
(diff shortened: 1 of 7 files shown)


packages/runtime-html/src/resources/template-controllers/if.ts
```diff
@@ -121,21 +122,55 @@ export class If implements ICustomAttributeViewModel {
...
+    // ends If's ownership instead, so every owned slot is disposed directly.
+    const ifView = this.ifView;
+    const elseView = this.elseView;
+    const currentView = this.view;
+    const ownsView = view !== void 0
+      && (currentView === view || ifView === view || elseView === view);
+    this.ifView = this.elseView = void 0;
+    if (currentView === view || currentView === ifView || currentView === elseView) {
+      this.view = void 0;
+    }
```
bigopon: why do we need this check? shouldn't we dispose everything inside an error branch?

fkleuver: Yeah you're right, this was doing something for a hypothetical edge case that couldn't be reproduced. Simplified it

bigopon (approved):
nice work @fkleuver 👍

PR #2469, opened 2026-09-05
fix(runtime-html): preserve array observation in content interpolation

fkleuver:
# 🐛 Bug Fix

## 📖 Description

Text interpolations could stop updating when an expression reevaluated to the same array, such as after refreshing a parent object.

## 💁 Your Solution

Keep array results observed and refresh their rendered text, including edits made before a queued update runs.

## 📑 Test Plan

- Parent-object refresh that retains the same array.
- Array edits queued before or after the refresh.
- Array replacement and transitions to/from `null`.
- Unbinding with pending updates, cached-view rebinding, and teardown cleanup.

## 📦 Changeset

- [x] Added a changeset

### DIFF
--- a/.changeset/fresh-array-content.md
+++ b/.changeset/fresh-array-content.md
@@ -1,0 +1,5 @@
+---
+"@aurelia/runtime-html": patch
+---
+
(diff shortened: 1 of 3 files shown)


bigopon (approved):
nice find 👍

PR #2471, opened 2026-09-05
Fix expression parsing and optional parent-scope access

fkleuver:
# 🐛 Bug Fix

## 📖 Description

Fixes three issues in existing binding syntax:

- Array destructuring depended on whitespace before `of`. Ordinary array literals could also be mistaken for declarations.
- Optional `$parent` access could fail when a view had no parent scope.
- Constructor expressions such as `entries.push(new Entry)` consumed tokens belonging to the surrounding expression.

## 💁 Your Solution

- Use the iterator parsing context to recognize array declarations.
- Carry optional-parent guards through parsing, evaluation and serialization.
- Remove the extra token advance after constructor expressions.

## 📑 Test Plan

- Restored iterator test cases and added repeat/interpolation regressions.
- Added optional-parent integration tests for composition and nested scopes.
- Verified strict evaluation, AST round-trips and template type-checking.
- Added constructor integration tests and checks for precedence and malformed input.

## 📦 Changeset

- [x] Added changesets for all three fixes.


packages/runtime/src/ast.eval.ts
```diff
@@ -146,7 +161,13 @@ export const {
         return currentScope ? currentScope.bindingContext : void 0;
       }
       case ekAccessScope: {
-        const obj = getContext(s, ast.name, ast.ancestor) as IBindingContext;
+        const obj = getContext(s, ast.name, ast.ancestor);
+        if (obj == null) {
+          if (e?.strict && !isOptionalAncestorMissing(s, ast.ancestor, ast.optionalAncestor)) {
```
bigopon: the name is quite confusing, why care missing if it's optional is what popped up for me.

packages/expression-parser/src/expression-parser.ts
```diff
@@ -1329,7 +1332,7 @@ function parseCoverParenthesizedExpressionAndArrowParameterList(expressionType:
  * ,
  * Elision ,
  */
-function parseArrayLiteralExpression(expressionType: ExpressionType): ArrayBindingPattern | ArrayLiteralExpression {
```
bigopon, line 1332: 👍

packages/expression-parser/src/expression-parser.ts
```diff
@@ -1341,7 +1344,7 @@ function parseArrayLiteralExpression(expressionType: ExpressionType): ArrayBindi
         break;
       }
     } else {
-      elements.push(parse(Precedence.Assign, expressionType === etIsIterator ? etNone : expressionType) as IsAssign);
```
bigopon, line 1344: why is this change?

.changeset/constructor-expression-boundaries.md
```diff
@@ -0,0 +1,5 @@
+---
+"@aurelia/expression-parser": patch
+---
+
+Constructor expressions without parentheses now work inside larger binding expressions, such as `entries.push(new Entry)` and `[new Map, label]`. The parser also reports malformed text following these expressions instead of silently consuming it.
```
bigopon, line 5: 👍

bigopon (commented):
looks good to me 👍

packages/runtime/src/ast.eval.ts
fkleuver: I'll rename to `isMissingScopeGuarded`, that would express the intention better. We only reach this check after scope lookup has failed in strict mode. It determines whether `?.` permits returning `undefined` rather than throwing.

The parser stores `$parent` paths as an ancestor count, so we retain where the optional access occurred. With a parent but no grandparent, `$parent?.$parent.name` still has an unguarded final access (`$parent.$parent?.name` permits the missing grandparent).

packages/expression-parser/src/expression-parser.ts
fkleuver: Iterator arrays now go directly to `parseArrayDestructuring`, so `parseArrayLiteralExpression` cannot receive `IsIterator` anymore. That makes the conversion to `etNone` unreachable. The iterable after `of` is already parsed separately in `IsChainable` mode.

bigopon (approved):

PR #2472, opened 2026-09-06
Fix keyed method receivers and observation

fkleuver:
# 🐛 Bug Fix

## 📖 Description

Bracket-notation calls such as `model[action](item)` lost their `this` receiver and bypassed the observation behavior of equivalent dot calls.

Keyed event-handler references, such as `click.trigger="model[action]"`, also lost their receiver.

## 💁 Your Solution

- Resolve the receiver once and share one inline path for method invocation and observation.
- Preserve receiver binding for keyed handler references.

## 📑 Test Plan

- Added binding regressions for dynamic calls and handler references, including evaluation order and optional calls.
- Covered `@computed` dependencies and array updates through replacement and teardown.

## 📦 Changeset

- [x] Added a changeset.


.changeset/keyed-method-receivers.md
```diff
@@ -0,0 +1,7 @@
+---
+"@aurelia/runtime": patch
+---
+
+Method calls through bracket notation, such as `model[action](item)`, now preserve their receiver. Computed-method dependencies and array-method observation follow the same behavior as dot calls.
```
bigopon, line 5: computed method + array method means keyed access fn call right?

bigopon (approved):
👍

.changeset/keyed-method-receivers.md
fkleuver: Yes, specifically their keyed-call forms:

- `model[action](item)` where the selected method uses `@computed`.
- `items[method](...)` where the selected array method is something like `join` or `filter`.

"Computed method" means `@computed`-decorated method.

PR #2474, opened 2026-09-06
fix(runtime-html): reconcile initial select option models

fkleuver:
# Pull Request

## 📖 Description

Fixes initial selection for static options using `model.bind`. The select previously synchronized before its option models and matcher were ready, leaving preselected values unselected.

Adds a select-local reconciliation after the initial bindings run. This also handles cached-view rebinding. Ordinary two-way updates retain their existing immediate path.

## 📑 Test Plan

Added integration coverage for:

- Direct bindings and values supplied through `<let>` and conditional views.
- Object models, matchers, and one-time/to-view bindings.
- Cached-view reuse and unbind/rebind before the queue drains.
- Model updates, two-way writeback, and native selection defaults.

## 📦 Changeset

- [x] Added a changeset


packages/runtime-html/src/observation/select-value-observer.ts
```diff
@@ -124,6 +128,18 @@ export class SelectValueObserver implements INodeObserver {
...
+    // so schedule from setValue; active two-way updates keep their immediate path.
+    if (!this._observing && !this._isQueued) {
+      this._isQueued = true;
+      queueTask(() => {
+        if (this._isQueued) {
+          this._isQueued = false;
+          this.syncOptions();
+        }
+      });
+    }
```
bigopon, line 142: we already have the mutation observer based syncing, I think if there are fix to be done, it should be long to there isntead of here

fkleuver: That wouldn't fix this particular bug. The existing mutation observer handles DOM changes. Here, the options already exist (binding later assigns `option.model` and `select.matcher`), which are JS properties and produce no DOM mutation notification.

Also, `to-view` and `one-time` bindings never start that mutation observer.

We're already reusing `syncOptions()`. What this adds is the missing initialization layer. Moving the fix into the mutation callback would leave these cases untouched.

PR #2476, opened 2026-09-09
 fix(runtime): observe nullable deep computed dependencies

fkleuver:
# 🐛 Bug Fix

## 📖 Description

Computed getters with explicit deep dependencies could throw when a dependency was `null` or `undefined`, including during initial binding or when clearing a loaded model.

## 💁 Your Solution

Skip deep traversal for absent values. Keep observing the dependency so an assigned object and its nested changes are tracked.

## 📑 Test Plan

Added integration coverage for:

- Loading, clearing, and replacing nullable models in both flush modes.
- Nested updates, subscription cleanup, and rebinding after teardown.

## 📦 Changeset

- [x] Added a changeset


bigopon (approved):
👍

PR #2477, opened 2026-09-09
fix: preserve optional-chain continuation and grouping

fkleuver:
# 🐛 Bug Fix

## 📖 Description

Optional chains such as `record?.format().label` could throw when `record` was absent. Computed keys could also run despite a short-circuit, and expression round-trips could lose optionality or grouping.

## 💁 Your Solution

Preserve short-circuiting through subsequent accesses and calls. Retain a small grouping marker so `record?.details.label` stays distinct from `(record?.details).label`, including through serialization and template type checking.

Grouped method calls retain their receivers and existing dependency tracking.

## 📑 Test Plan

- Added regressions for loading/replacement, skipped keys and arguments, and grouped calls.
- Verified parser/unparser and validation round-trips, plus TS/JS template checks.

## 📦 Changeset

- [x] Added a changeset


packages/runtime/src/ast.eval.ts
```diff
@@ -90,6 +90,13 @@ export const {
   const ekCustom = 'Custom';
   const getContext = Scope.getContext;
 
+  // Only receiver/callee recursion may carry this value. Arguments and other expressions evaluate
+  // independently; a grouped or public result converts it back to undefined, including on reentry.
+  const shortCircuit = {};
+  function skipChain(ast: { parenthesized?: true }, chain: boolean | undefined): unknown {
```
bigopon, line 96: We are returning skipChain returned value to the caller, any modification on this object affects all the subsequent evaluation (check property exists, identity comparison etc...). Is this analysis correct?

fkleuver: No, not correct. Normal evaluation returns `undefined`, not the shared marker. The marker's properties are never read, and modifying them would not change its identify or affect subsequent identity comparisons

